PR #165, opened 2014-11-17
Throttle implementation

bbockelm:
Latest attempt at merging the throttles code.  Brushed off the spiderwebs, fixed some atomics issues; everything looks good right now.

Works fine with an OSS; throttles don't apply for AIO or SendFile.  Not sure what I'd need to do for either of those.


bbockelm:
Alright, documentation updated!

PR #180, opened 2014-12-02
New XrdFileCache option to invoke script on new file attach

alja:
Add 'newfilescript' option to invoke a command when new file is opened from proxy file cache server.

This was requested by Doug Benjamin.


abh3:
Hi Alja,

Yes, truncate() should propagate into the cache (well, only for writable caches anyway).

Andy

alja:
The feature can be implemented using OFS notification. There is no need to make change in  caching proxy.

PR #186, opened 2014-12-15
XrdSys: Delete channel object before returning

esindril:
Is there any particular reason why the channel object is not deleted in this case?
According to valgrind this is a leak ...

Thanks,
Elvin


abh3:
OK, I did the research on this and simply deleting the object in the proposed context will likely corrupt storage in certain circumstances. The problem is that pollErr1 is set for various reasons -- not all of which leave the channel object clean. So, I will need to go through each instance and make sure that the channel is left in a deletable state before this can be merged.

PR #197, opened 2015-02-04
free() dnarray only when necessary

artemharutyunyan:
xrootdfs attempts to free() an uninitialized pointer when redirector returns an empty list of data nodes. This patch is supposed to fix that.


ljanyst:
Hi Artem,

   thanks! I will merge it. Line 338 is sufficient the rest is no-op.

Cheers,
   Lukasz

PR #209, opened 2015-03-05
ProxyFileCache purge :: fix order of deletion, optimize sorting 

alja:
With this implementation file with the oldest close time are deleted.
Bug was reported by Anna Olson.


alja:
@abh3 
I googled a bit about solaris multimap insert.  The first hit https://zeromq.jira.com/browse/LIBZMQ-368 says std::pair first entry should be const.

I have changed that, but could not test it on solaris

PR #252, opened 2015-07-07
packaged XrdSysFAttr.hh in the private headers

sponce:


abh3:
Hi Sebastien, Why does this need to be in the public headers? You are supposed to use XrdSysXattr.hh. The XrdSysFattr.hh was always meant to be private and it would be unworkable to freeze all future development on that code by making this header file public.

bbockelm:
Hi @abh3 -

Maybe I'm reading the commit wrong, but isn't the entire change to list `XrdSysFAttr.hh` in the _private_ header list?

Brian

abh3:
Anything that gets added to the XrdHeaders.cmake file is installed as a public header. So, the change makes the file public which it should not be as it's the actual implementation. XrdSysXAttr.hh is the virtual one and is supposed to be public, which it is.

sponce:
??? So if I had a header to XROOTD_PRIVATE_HEADERS (note the private), it becomes public ???
My intention was indeed that it's distributed in the private header package, so that I can compile against it without needing a checkout of xroot.
Now you are saying that I'm supposed to use XrdSysXattr.hh. That would be perfect, but how do I access XrdSysFAttr::Xat ? Any hidden way to get it ? I do not see anything in XrdSysXattr

abh3:
Ah, I messed the placement of the addition and thought it was added to the public header section. My apologies about that. You already have all the needed access. Simply go to the XrdSysXAttr.hh file, scroll to the bottom, read te section that tells you how to access the native implementation. It's really very simple.

sponce:
Thanks to the help of Andy, I managed to use the new API without the need for exporting the header. So forget about this pull request

PR #268, opened 2015-07-23
add MANIFEST.in for python bindings

otron:
* Requested by @esindril (see [original PR to xrootd-python](https://github.com/xrootd/xrootd-python/pull/12).
* Adds a MANIFEST.in file for the python bindings.
* Is not by itself sufficient to generate a python source distribution:
  * The `setup.py` file generated via `setup.py.in` and `cmake` looks for XrootD libraries and resources within the XrootD source directory -- assuming they can be found wherever they were found on the system on which the `setup.py` file was generated. This is not portable.

### DIFF
--- a/bindings/python/MANIFEST.in
+++ b/bindings/python/MANIFEST.in
@@ -1,0 +1,6 @@
+include README.rst
+recursive-include tests *
+recursive-include examples *
+recursive-include docs *
+recursive-include src *
+recursive-include libs *


ljanyst:
It is portable, because `rpath` is stripped from the binaries and the linker only looks for sonames in whatever search paths are configured on the system.

ljanyst:
If you really need to generate a standalone python source package, then probably having another `setup.standalone.py' is the way to go.

esindril:
Hi @otron ,

Lukasz's suggestion is probably the simplest and cleanest. I've created a branch in my github checkout of xrootd and added a new setup_pypi.py file to properly generate the sdist package. The current one from the XRootD package is not able to do this as the setup.py MANIFEST assumes all the necessary python files are in the same directory which is not the case. 

Therefore, check it out - it should properly build the sdist pkg. If you are satisfied and all works as you expect it, I can merge it into the master. The downside bit is that it pollutes the xrootd source directory when you create the sdis pkg, but this is a one off when creating the package to put it in PyPI. The alternative would be to copy all into the build directory ... also not nice.

The branch I created is: https://github.com/esindril/xrootd/tree/enhance_python_sdist

Cheers,
Elvin

otron:
Looks good! I am working on the tox config file that'll automate the upload process as well as getting installation to work on OSX.

otron:
Hello again, @esindril.
As it is now, the version number is generated/retrieved upon installing the python package using a script that isn't portable. I'd like to change it so that the version number is generated/retrieved as a part of the package uploading process via the tox script we discussed last week.

The tox script requires the addition of a single config file in the xrootd project directory. Executing it will generate and upload the python package for you, allowing you to easily integrate this with your existing release process. 

I'll submit a PR to your branch with the required changes when it's ready.

otron:
@esindril: I've created a pull request to your branch with the changes I found necessary to make it work.
[Link to the PR](https://github.com/esindril/xrootd/pull/1)

esindril:
Thank a lot @otron. I'll have a look over it and include it in the next release.

esindril:
Fixed by commit: https://github.com/xrootd/xrootd/commit/98aebca7bfaeecd4c049ec341f66af90b49cb6f7

PR #272, opened 2015-07-31
Fix selinux policy for xrootd due to new SELinux change https://fedor…

wrabcak:
…aproject.org/wiki/Changes/SELinuxPolicyStoreMigration


simonmichal:
thanks for pushing it upstream

PR #285, opened 2015-09-11
Do not forward the 'tried=' CGI within the proxy interface.

bbockelm:
This may cause the proxy to respond with 'file not found', causing
upstream redirectors to believe the file is not present behind
the proxy at all - even if it's just 'not found' because the only
source was excluded.

### DIFF
--- a/src/XrdPss/XrdPss.cc
+++ b/src/XrdPss/XrdPss.cc
@@ -973,9 +973,26 @@ const char *XrdPssSys::P2CGI(int &cgilen, char *cbuff, int cblen,
    if (!Cgi1) {cgilen = strlen(Cgi2); return Cgi2;}
    if (!Cgi2) return Cgi1;
 
-// Compose the two cgi elements together
-//
-   cgilen = snprintf(cbuff, cblen, "%s&%s", Cgi1, Cgi2);
+// Strip out any 'tried='
+// Then, Compose the two cgi elements together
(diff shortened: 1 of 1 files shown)


abh3:
Thanks Brian.

bbockelm:
Actually - be ready to revert this one.  After I sent the PR, Andrea reported it didn't seem to work (I assume he'll be able to provide us more information on Monday).

alja:
@bbockelm 
I'm concerned this will effect hadoop healing proxy. In this case we must used 'tried=' because we don't want that file is served from the cluster where files were corrupt.

bbockelm:
For what it's worth, '?tried=' can (and often is) ignored.  You can't depend on it functioning.

That said, I'm not sure this fixed things for Andrea anyway.

alja:
I hope '?tried=' will be reliable when servers will be updated to new version.

Anyway, healing proxy server need to pass the tried argument. Maybe there is some other way to fix the issue reported in the first comment.

@abh3 
Brian and I both agree to revert this PR.

abh3:
Hi Alja,

Thank you for reminding me of this. Brian, should be back out your patch for tried= or leave it in. We will soon be cutting 4.3.0 and we need to know ASAP.

Andy

bbockelm:
I think we should definitely revert it - solving this via DPM patch appears to cover the root of the issue.

PR #286, opened 2015-09-14
Implement blacklisting in a FileCache decision plugin.

bbockelm:
The current release does very little with the decision plugin architecture - we ship source code for a trivial plugin (which always returns true) but we don't appear to build it or ship it (of course, why would we?).

This pull request adds a new plugin, "blacklist decision", which can be enabled with the following configuration line:

```
pfc.decisionlib libXrdBlacklistDecision.so /etc/xrootd/blacklist-cache.cfg
```

where /etc/xrootd/blacklist-cache.cfg is a list of globs.  For example:

```
/bar/baz/foo
/foo/*/some_file
/foo/*/.*
```

This is shipped as part of the xrootd-server-libs RPM.

@alja @osschar


osschar:
+1

PR #294, opened 2015-10-08
Revert "Do not forward the 'tried=' CGI within the proxy interface."

abh3:
Reverts xrootd/xrootd#285


abh3:
OK, we know the issue was actually a DPM thing, so the patch is not needed.

Issue #313, opened 2015-12-02
segmentation fault in XrdSysLinuxSemaphore::Wait (often seen with gcc 5.1.1 + xrootd 4.2.3)


smithdh:
This is probably the same issue #310 but I opened another issue just incase it is different.
This crash is frequently (possibly always) seen when running one of the ROOT test which use XrdCl, on fedora 22:

{code}
#0  0x00007f02a41e3cf9 in ?? () from /lib64/libgcc_s.so.1
#1  0x00007f02a41e502b in ?? () from /lib64/libgcc_s.so.1
#2  0x00007f02a41e544d in _Unwind_Resume () from /lib64/libgcc_s.so.1
#3  0x00007f0299c5f406 in Wait (this=<optimized out>)
    at /home/dhsmith/root-master-build/XROOTD-prefix/src/XROOTD/src/./XrdSys/XrdSysLinuxSemaphore.hh:175
#4  Get (this=0x20a45f8)
    at /home/dhsmith/root-master-build/XROOTD-prefix/src/XROOTD/src/./XrdCl/XrdClSyncQueue.hh:67
#5  XrdCl::JobManager::RunJobs (this=0x20a45e0)
    at /home/dhsmith/root-master-build/XROOTD-prefix/src/XROOTD/src/XrdCl/XrdClJobManager.cc:146
#6  0x0000000000800000 in ?? ()
#7  0x0000000000000000 in ?? ()
{code}

I can reproduce it (at least every time I try it) in this way: Compile xrootd 4.2.3 on fedora 22, with the standard cmake system. i.e. this compiles it with -O2 optimisation. Running this example against XrdCl gives a seg fault:

{code}
#include "XrdCl/XrdClFile.hh"
using namespace XrdCl;
int main() {
   File *f = new File();
   f->Open("xroot://nosuchhost.example.edu//path", OpenFlags::Read,  Access::None, (uint16_t)0);
   delete f;
  return 0;
}

c++ -g <exmaple.cpp> -lXrdCl -lpthread

#0  0x00007ffff7051cf9 in ?? () from /lib64/libgcc_s.so.1
#1  0x00007ffff705302b in ?? () from /lib64/libgcc_s.so.1
#2  0x00007ffff705344d in _Unwind_Resume () from /lib64/libgcc_s.so.1
#3  0x00007ffff7bb2c06 in Wait (this=<optimized out>)
[...]

PR #321, opened 2015-12-23
Order pthread cleanup statements

### LINKED ISSUE #310: segmentation fault in xrdfs
This issue was reported to the redhat bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1278968

See the bugzilla report for details.

ellert:
Change the order of some statements when registering pthread clean-up functions.
Fixes #310 and #313.

### DIFF
--- a/src/XrdSys/XrdSysLinuxSemaphore.hh
+++ b/src/XrdSys/XrdSysLinuxSemaphore.hh
@@ -155,13 +155,13 @@ namespace XrdSys
             {
               int r = 0;
 
-              pthread_cleanup_push( Cleanup, pValue );
               pthread_setcanceltype( PTHREAD_CANCEL_ASYNCHRONOUS, 0 );
+              pthread_cleanup_push( Cleanup, pValue );
 
               r = syscall( SYS_futex, pValue, FUTEX_WAIT, newVal, 0, 0, 0 );
 
-              pthread_setcanceltype( PTHREAD_CANCEL_DEFERRED, 0 );
               pthread_cleanup_pop( 0 );
+              pthread_setcanceltype( PTHREAD_CANCEL_DEFERRED, 0 );
 
               if( r == 0 )               // we've been woken up
                 break;


ljanyst:
This is incorrect but very interesting. If this indeed helps then wrapping the syscall in brackets should help as well.

```
{
   r = syscall( SYS_futex, pValue, FUTEX_WAIT, newVal, 0, 0, 0 );
}
```

I can't reproduce the problem on Debian with gcc-5.3, so this may be a compiler quirk on RedHat. It wouldn't be the first time when xroot is affected by a broken compiler.

ljanyst:
This is causing a data race because you may get a `SIGCANCEL` from another thread before before you install the cancellation handler. While it's true that the consequences of this data race are not grave, they may "only" induce a performance hit in the worst case, it's probably not a good idea to allow that. This code tries to provide a remedy for an issue that went unnoticed in glibc for years and hit EOS pretty hard.

Look at how these functions are implemented. `pthread_cancel_push` and `pthread_cancel_pop` are essentially:
```
do {
  _pthread_cleanup_class __clframe (routine, arg)
  // your code goes here
} while(0);
```
where `_pthread_cleanup_class` is essentially the same as `XrdSysMutexHelper`.
https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/nptl/pthread.h;h=fd0894efd210933791e4f89cb8af709b545dcb93;hb=HEAD#l548

As for `pthread_setcanceltype`, it essentially does:
```
cancelhandling |= CANCELTYPE_BITMASK;
```
and checks if the thread should be cancelled immediately.
https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_setcanceltype.c;h=442211cd8110ca4db2cce1142a61d9268082ffb5;hb=HEAD

I am sure there is a better solution, even if it has to be a different combination of these.

simonmichal:
I think Mattias already pushed it as a patch to EPEL so it's better to be in sync.

As you mentioned in #313 the original semaphore bug seems to be fixed in glibc 2.21 (https://sourceware.org/bugzilla/show_bug.cgi?id=12674) which is already available in f22, so I think in 4.3.0 we can go back to the posix semaphores for f22 and f23, and for older platforms reorder the two statements back in order to avoid the race condition.

ljanyst:
Well, that's up to you. IMO, being in sync for the sake of being in sync when you know the code is buggy is probably not the best idea.

PR #323, opened 2015-12-24
Doxygen fixes

ellert:


ljanyst:
Make sure this goes well with ROOT. They do `-Wall -Werror`.

PR #334, opened 2016-02-13
Don't throw exceptions in destructors

ellert:
Throwing exceptions in a destructor causes warnings with gcc 6, and since -Werror is used this becomes an error:

XrdSys/XrdSysPthread.hh: In destructor 'XrdSysSemaphore::~XrdSysSemaphore()':
XrdSys/XrdSysPthread.hh:356:31: error: throw will always call terminate() [-Werror=terminate]
XrdSys/XrdSysPthread.hh:356:31: note: in C++11 destructors default to noexcept

There are two destructors in the code that does this. This pull request comments out these lines of code.


abh3:
Thanks for the update. We found the same problem a couple of weeks ago and haven't pushed the fix into the repo yet. Our solution was to replace the throw with an abort which is equivalent to what we really wanted.

PR #336, opened 2016-02-17
Fix strict aliasing issues

ellert:
```
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc: In static member function 'static int XrdSysDNS::getAddrName(const char*, int, char**, char**, char**)':
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc:187:51: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
       int hn = getHostName((struct sockaddr &)ip[i], names, 1, errtxt);
                                                   ^
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc: In static member function 'static int XrdSysDNS::Host2Dest(const char*, sockaddr&, char**)':
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc:522:47: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
    if (!getHostAddr(hbuff, (struct sockaddr &)InetAddr, errtxt)) return 0;
                                               ^~~~~~~~
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc: In static member function 'static int XrdSysDNS::Host2IP(const char*, unsigned int*)':
/builddir/build/BUILD/xrootd-4.2.3/src/XrdSys/XrdSysDNS.cc:550:47: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
    if (!getHostAddr(hname, (struct sockaddr &)InetAddr)) return 0;
                                               ^~~~~~~~
```

### DIFF
--- a/src/XrdSys/XrdSysDNS.cc
+++ b/src/XrdSys/XrdSysDNS.cc
@@ -184,7 +184,8 @@ int XrdSysDNS::getAddrName(const char *InetName,
 
       // The name
       char *names[1] = {0};
-      int hn = getHostName((struct sockaddr &)ip[i], names, 1, errtxt);
+      struct sockaddr *ipaddr = (struct sockaddr *)&ip[i];
+      int hn = getHostName(*ipaddr, names, 1, errtxt);
       if (hn)
          Name[i] = strdup(names[0]);
       else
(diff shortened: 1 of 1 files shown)


abh3:
Thank you Mattias for the fix. We thought we would do this using unions but your solution is just as nice :-)

PR #352, opened 2016-04-18
pfc-V2: Set RAM usage required parameter

alja:
Two changes in thos PR:
* pfc.ram a required parameter
* specifiy prefetch level as pfc.prefetch <n>
* update readme


alja:
## example of configuration file
http://uaf-2.t2.ucsd.edu/~alja/entireFile.cfg

PR #358, opened 2016-04-21
Use cmake to determine the existence of openssl funcs. Ignore the openssl constants

ffurano:
This version of the openssl method detection works for me on slc5 and rawhide.


abh3:
Before I pull it, does it work in SLC6, Fedora, and Ubuntu?

xrootd-dev:
Hi Andy,

  I tried it for SLC5, SLC6, latest Ubuntu, fedora rawhide and it compiled fine.

Fabrizio

PR #360, opened 2016-04-27
pfc-V2: Fix bug in selecting file to prefetch

alja:


alja:
This change protects from spurious wakeups in GetNextFilePrefetch(). I replaced 'if' with 'while' to re-evalue nonenmpty vector state.

PR #406, opened 2016-08-19
Update name of ceph development package in spec file.

brunocanning:
Since the 'hammer' release of ceph, the package "ceph-devel" no longer exists, having been replaced by the meta package "ceph-devel-compat". This change updates the spec file accordingly.

### DIFF
--- a/packaging/rhel/xrootd.spec.in
+++ b/packaging/rhel/xrootd.spec.in
@@ -62,7 +62,7 @@ BuildRequires: cppunit-devel
 %endif
 
 %if %{?_with_ceph:1}%{!?_with_ceph:0}
-BuildRequires: ceph-devel
+BuildRequires: ceph-devel-compat
 %endif
 
 BuildRequires:	doxygen


simonmichal:
From what I've seen the ceph-devel-compat has been there at least since Firefly release ;-) It does obsolete the ceph-devel package in the spec file so IMHO everything is OK.

brunocanning:
Hi Simon,

I can't see "ceph-devel-compat" in our local mirrors here at RAL for firefly or giant or "ceph-devel" for hammer, infernalis or jewel. We are mirroring from download.ceph.com. I think this change will be OK, XrootD and Ceph have only been usable together since the more recent releases of both.

Best wishes,
Bruno

simonmichal:
Hi Bruno,

Hmm that's interesting, I actually checked what's in epel7 (http://koji.fedoraproject.org/koji/buildinfo?buildID=738434) and there's ceph-0.80.7-0.8.el7 (so firefly) and it contains ceph-devel-compat ;-)

In download.ceph.com the ceph-devel-compat indeed appears only in hammer.

Changing the build requires to ceph-devel-compat would mean that we wont be able to build xrootd-ceph with giant any more, right? (giant does have radosstriper and ceph-devel)

On the other hand, all the releases that have devel-compat, have a respective obsolete entry in the spec file for the devel package, which means that the dependencies will be handled correctly.

Just considering the pros and cons ;-)

Cheers,
Michal

brunocanning:
Hi Michal,

I meant to get in touch about this. I notice you've put a more elegant solution into the spec file than the one I proposed here. I tested it in December when I build XRootD 4.5 with ceph it it works just fine, so closing this PR is a good thing. Should I delete the branch?

Best wishes,
Bruno

simonmichal:
Hi Bruno,

Yes, please go ahead and delete the branch. Sorry I forgot to notify you ;-)

Thanks, 
Michal

Issue #411, opened 2016-09-02
override dual-stack: false on the server side


zvada:
Hello!

We're experiencing issues with older clients vs redirector IPv6 enabled where client host is dual-stack. Relevant piece of behavior is from the log as follows:

`[2016-09-02 19:40:10.529888 +0200][Debug  ][XRootDTransport   ] [cmsxrootd.fnal.gov:1094 #0.0] Sending out kXR_login request, username: zvada, cgi: ?xrd.cc=ch&xrd.tz=1&xrd.appname=xrdcp&xrd.info=, dual-stack: false`
...
...

```
[2016-09-02 19:41:00.221320 +0200][Debug  ][File              ] [0x2000000@root://cmsxrootd.fnal.gov:1094//store/mc/RunIISpring16reHLT80/DYJetsToLL_M-50_TuneCUETP8M1_13TeV-amcatnloFXFX-pythia8/AODSIM/PUSpring16RAWAODSIM_reHLT_80X_mcRun2_asymptotic_v14-v1/40000/003A70D5-E55C-E611-9B14-0025905C53A8.root] Error while opening at cms-xrd-transit.cern.ch:1094: [ERROR] Server responded with an error: [3011] No servers are available to read the file.
[0B/0B][100%][==================================================][0B/s]  
Run: [ERROR] Server responded with an error: [3011] No servers are available to read the file.
$ xrdcp -version
v4.0.4
```

Forcing XRD_NETWORKSTACK=IPv4 solve the issue on client side, though:

```
$ XRD_NETWORKSTACK=IPv4 xrdcp -f root://cmsxrootd.fnal.gov//store/mc/RunIISpring16reHLT80/DYJetsToLL_M-50_TuneCUETP8M1_13TeV-amcatnloFXFX-pythia8/AODSIM/PUSpring16RAWAODSIM_reHLT_80X_mcRun2_asymptotic_v14-v1/40000/003A70D5-E55C-E611-9B14-0025905C53A8.root /dev/null 
[80MB/3.495GB][  2%][=>                                                ][8.889MB/s] 
$ xrdcp -version
v4.0.4
```

Suggestion is provide on the server side a configuration parameter to force dual-stack to true - we'd assume that IPv4 is always present. Clients >=v4.3.x handle dual-stack situation well, but for users running anything below that would be preferable be cautions and act on the server side with config tweak if possible.

Thanks,
Marian

abh3:
Hi Marian,

[...]

PR #420, opened 2016-09-27
Allow the server to assume a IPv6-only client is misconfiged

bbockelm:
This adds a new option to the `xrd.network`, `assumev4`, that allows the server to assume a IPv6-only client is actually also able to speak IPv4.

It's an unpleasant approach - but allows us to gracefully deal with older clients that incorrectly set `dual-stack=false`.

Solves #411.


abh3:
Hi Brian,

I though we were going to figure out at what releasethis problem was 
solved and make a more intelligent decision here. I hate such options 
because once added they get rarely removed. So, what release was this 
fixed?

Andy

PR #426, opened 2016-10-23
PSS: Add url prefix functionality to the PSS plugin

esindril:
The administrator can specify the following configuration directive
for a PSS plugin to proxy the incoming connections from clients to
another proxy server:

pss.urlprefix root://other_proxy_host:other_proxy_port//

Note that the url prefix needs to be a valid endpoint for an XRootD
service and must end with double slashes.


esindril:
This is not needed.

PR #429, opened 2016-10-31
File caching proxy V2 and new pss async interface

alja:
This pull request contains two different sets of changes:

* Implementation of pss async interface (from pssasync branch)
* Next version of proxy file cache code


abh3:
V2 of Caching Proxy also includes async proxy implementation.

PR #444, opened 2016-12-19
Make Xrdhttp secure by default, by rejecting proxy cert in the absence of a proper SecXtractor plugin

ffurano:
Fix some minor memory leaks
Change the secxtractor interface to use the new lib and initialize it properly

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -101,6 +101,53 @@ XrdSysError XrdHttpProtocol::eDest = 0; // Error message handler
 XrdSecService *XrdHttpProtocol::CIA = 0; // Authentication Server
 
 
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
 /******************************************************************************/
 /*            P r o t o c o l   M a n a g e m e n t   S t a c k s             */
 /******************************************************************************/
@@ -298,6 +345,8 @@ int XrdHttpProtocol::GetVOMSData(XrdLink *lp) {
   // No external plugin, hence we fill our XrdSec with what we can do here
   peer_cert = SSL_get_peer_certificate(ssl);
   TRACEI(DEBUG, " SSL_get_peer_certificate returned :" << peer_cert);
+  ERR_print_errors(sslbio_err);
+  
   if (peer_cert && peer_cert->name) {
     
     // Add the original DN to the moninfo. Not sure if it makes sense to parametrize this or not.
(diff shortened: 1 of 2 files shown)


bbockelm:
A few issues with this one:
- Breaks backward compatibility with existing plugins.
- No versioning bump when breaking the backward compatibility.
- Now requires a client certificate for access, which is new behavior.  We currently allow unauthenticated clients to download, but allow additional authorization based on the ACC.
- Lots of strange whitespace.
- Changes in timeouts - can you add comments as to why?
- Lots of old code commented out.  Can probably just delete that and rely on git for history.

ffurano:
Hi Brian,

 thanks for having a look.

On 12/20/2016 03:44 AM, Brian Bockelman wrote:
> A few issues with this one:
> 
>   * Breaks backward compatibility with existing plugins.

 indeed, we can add a  deprecation in the spec, when we decide how to release it.
 Or would you prefer a runtime check ?

>   * No versioning bump when breaking the backward compatibility.

 Uhm, I had increased the version number on the plugin. Which version number are you referring to ?

>   * Now requires a client certificate for access, which is new behavior. We currently allow unauthenticated clients to download,
>     but allow additional authorization based on the ACC.


 Do you mean unauthenticated but still through https ? Ok, should be the option SSL_VERIFY_FAIL_IF_NO_PEER_CERT,
I'll have a try.


>   * Lots of strange whitespace.
>   * Changes in timeouts - can you add comments as to why?

 It was one second, less than a roundtrip in some cases. Do you see any problem ? I did with one second.

>   * Lots of old code commented out. Can probably just delete that and rely on git for history.


 Ok!
 I'm doing these points except the versioning thing, we need one more exchange for that imo.

Fabrizio

xrootd-dev:
Ok, I've submitted a little pull request containing:

 - now xrdhttp allows unauthenticated https clients
   (you also need to update xrdhttpvoms if you use it)
 - remove blank lines and comments

 About versioning we should decide what to do after the Christmas break. So far
the xrdhttpvoms was not released officially in any repo, but I agree that we should
prevent future version-based troubles.
 We should decide mainly how to release it (xrootd tree or alone in epel or wlcg)
and how to defend from version alchemies.

 Please let me know how it goes on your side.
 Fabrizio

bbockelm:
For the versioning - I didn't see a version bump in this PR?  Did it go in earlier?

@abh3 will have to provide input here with regards to the best way to handle the version incompatibilities, but we need to generally keep ABI compatibility I believe.

ffurano:
Now I see, you were referring to the xrootd version. I don't know what to say about that, so I'd leave it
to Andy/Michal. So far for me it's stuff into the master.

I agree with keeping the ABI compat for public interfaces. Not for private ones though. I'd like
to discuss how to release the voms extractor. In the absence of other decisions, I'll try
to release it to wlcg in January, after CERN reopens.

Fabrizio

PR #448, opened 2017-01-13
[RFC] GCC 7 error cleanup

davidlt:
This PR is marked as RFC thus should not be merged. Its purpose is to discuss potential fixes to errors while compiling with GCC 7 [should be in Stage4 this week].

Check inline comments for explanation and discussions.

Signed-off-by: David Abdurachmanov <David.Abdurachmanov@cern.ch>

### DIFF
--- a/src/XrdApps/XrdMpxStats.cc
+++ b/src/XrdApps/XrdMpxStats.cc
@@ -473,6 +473,7 @@ int main(int argc, char *argv[])
                  else {Say.Emsg(":", "Invalid format - ", optarg); Usage(1);}
                  break;
        case 'h': Usage(0);
+                 break;
        case 'p': if (!(Port = atoi(optarg)))
                     {Say.Emsg(":", "Invalid port number - ", optarg); Usage(1);}
                  break;
--- a/src/XrdCl/XrdClZipArchiveReader.cc
+++ b/src/XrdCl/XrdClZipArchiveReader.cc
@@ -243,7 +243,7 @@ class ZipArchiveReaderImpl
     {
       delete pBuffer;
       ClearRecords();
-      pArchive.Close();
+      XRootDStatus st = pArchive.Close();
     }
 
     File                           pArchive;
--- a/src/XrdClient/XrdClient.cc
+++ b/src/XrdClient/XrdClient.cc
@@ -1372,7 +1372,7 @@ bool XrdClient::OpenFileWhenRedirected(char *newfhandle, bool &wasopen)
 	Info(XrdClientDebug::kHIDEBUG,
 	     "OpenFileWhenRedirected", "Stripping off the 'delete' option." );
 
-	options &= !kXR_delete;
+	options &= ~kXR_delete;
 	options |= kXR_open_updt;
     }
 
@@ -1380,7 +1380,7 @@ bool XrdClient::OpenFileWhenRedirected(char *newfhandle, bool &wasopen)
 	Info(XrdClientDebug::kHIDEBUG,
 	     "OpenFileWhenRedirected", "Stripping off the 'new' option." );
 
-	options &= !kXR_new;
+	options &= ~kXR_new;
 	options |= kXR_open_updt;
     }
 
--- a/src/XrdCms/XrdCmsManTree.cc
+++ b/src/XrdCms/XrdCmsManTree.cc
@@ -79,7 +79,7 @@ int XrdCmsManTree::Connect(int nID, XrdCmsNode *nP)
 {
    static CmsDiscRequest discRequest = {{0, kYR_disc, 0, 0}};
    XrdSysMutexHelper Monitor(myMutex);
-   char mybuff[8];
+   char mybuff[16];
    int i;
 
 // Rule 0: If we aborted tell the client to just stop doing this
--- a/src/XrdCms/XrdCmsReq.cc
+++ b/src/XrdCms/XrdCmsReq.cc
@@ -134,17 +134,17 @@ void XrdCmsReq::Reply_Error(int ecode, const char *emsg, int elen)
 // Translate the error name
 //
    switch(ecode)
-         {case ENOENT:       eval = kYR_ENOENT;
-          case EPERM:        eval = kYR_EPERM;
-          case EACCES:       eval = kYR_EACCES;
-          case EIO:          eval = kYR_EIO;
-          case ENOMEM:       eval = kYR_ENOMEM;
-          case ENOSPC:       eval = kYR_ENOSPC;
-          case ENAMETOOLONG: eval = kYR_ENAMETOOLONG;
-          case ENETUNREACH:  eval = kYR_ENETUNREACH;
-          case ENOTBLK:      eval = kYR_ENOTBLK;
-          case EISDIR:       eval = kYR_EISDIR;
-          default:           eval = kYR_EINVAL;
+         {case ENOENT:       eval = kYR_ENOENT; break;
+          case EPERM:        eval = kYR_EPERM; break;
+          case EACCES:       eval = kYR_EACCES; break;
+          case EIO:          eval = kYR_EIO; break;
+          case ENOMEM:       eval = kYR_ENOMEM; break;
+          case ENOSPC:       eval = kYR_ENOSPC; break;
+          case ENAMETOOLONG: eval = kYR_ENAMETOOLONG; break;
+          case ENETUNREACH:  eval = kYR_ENETUNREACH; break;
+          case ENOTBLK:      eval = kYR_ENOTBLK; break;
+          case EISDIR:       eval = kYR_EISDIR; break;
+          default:           eval = kYR_EINVAL; break;
          };
 
 // Make sure that elen includes a null byte
--- a/src/XrdCns/XrdCnsConfig.cc
+++ b/src/XrdCns/XrdCnsConfig.cc
@@ -150,6 +150,7 @@ int XrdCnsConfig::Configure(int argc, char **argv, char *argt)
                     else NoGo = NAPath("'-a'", Spec.argval);
                  break;
        case 'B': Opts |= optNoCns;
+                 [[gnu::fallthrough]];
        case 'b': bPath = Spec.argval;
                  break;
        case 'c': cPath = Spec.argval;
--- a/src/XrdCns/XrdCnsLogRec.hh
+++ b/src/XrdCns/XrdCnsLogRec.hh
@@ -58,8 +58,8 @@ struct Ctl
 
 struct Arg
       {char Type;                   // Event code
-       char Mode[3];                // Mode (create, inv, mkdir)
-       char SorT[12];               // Size (closew, inv) | TOD (eol)
+       char Mode[6];                // Mode (create, inv, mkdir)
+       char SorT[24];               // Size (closew, inv) | TOD (eol)
        char Mount;                  // Mount Index
        char Space;                  // Space Index
        char Rsvd[6];                // Reserved (blank filled)
--- a/src/XrdCrypto/XrdCryptosslX509.cc
+++ b/src/XrdCrypto/XrdCryptosslX509.cc
@@ -507,8 +507,8 @@ const char *XrdCryptosslX509::IssuerHash(int alg)
       if (issueroldhash.length() <= 0) {
          // Make sure we have a certificate
          if (cert) {
-            char chash[15] = {0};
-            snprintf(chash,15,"%08lx.0",X509_NAME_hash_old(cert->cert_info->issuer));
+            char chash[30] = {0};
+            snprintf(chash, sizeof(chash), "%08lx.0", X509_NAME_hash_old(cert->cert_info->issuer));
             issueroldhash = chash;
          } else {
             DEBUG("WARNING: no certificate available - cannot extract issuer hash (md5)");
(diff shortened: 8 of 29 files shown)


src/XrdApps/XrdMpxStats.cc
```diff
@@ -473,6 +473,7 @@ int main(int argc, char *argv[])
                  else {Say.Emsg(":", "Invalid format - ", optarg); Usage(1);}
                  break;
        case 'h': Usage(0);
+                 break;
```
davidlt: `Usage(0)` will call `exit(0)`, but adding `break` is an extra failsafe that shouldn't hurt.

src/XrdCl/XrdClZipArchiveReader.cc
```diff
@@ -243,7 +243,7 @@ class ZipArchiveReaderImpl
     {
       delete pBuffer;
       ClearRecords();
-      pArchive.Close();
+      XRootDStatus st = pArchive.Close();
```
davidlt: The declaration of `Close()` has an attribute to check return value, but I could't figure out what to do (we are in destructor). Just saving return value tricks the compiler to avoid the warning, but probably is not the right approach.

src/XrdClient/XrdClient.cc
```diff
@@ -1372,15 +1372,15 @@ bool XrdClient::OpenFileWhenRedirected(char *newfhandle, bool &wasopen)
 	Info(XrdClientDebug::kHIDEBUG,
 	     "OpenFileWhenRedirected", "Stripping off the 'delete' option." );
 
-	options &= !kXR_delete;
```
davidlt: This does not seem right at the first look. `kXR_delete` is a positive integer (enum), which would be translated to `true` here and negating that would result in `false` (i.e. 0).

src/XrdClient/XrdClient.cc
```diff
@@ -1372,15 +1372,15 @@ bool XrdClient::OpenFileWhenRedirected(char *newfhandle, bool &wasopen)
...
+	options &= ~kXR_delete;
 	options |= kXR_open_updt;
     }
 
     if (fOpenPars.options & kXR_new) {
 	Info(XrdClientDebug::kHIDEBUG,
 	     "OpenFileWhenRedirected", "Stripping off the 'new' option." );
 
-	options &= !kXR_new;
```
davidlt: Same here.

src/XrdCms/XrdCmsManTree.cc
```diff
@@ -79,7 +79,7 @@ int XrdCmsManTree::Connect(int nID, XrdCmsNode *nP)
 {
    static CmsDiscRequest discRequest = {{0, kYR_disc, 0, 0}};
    XrdSysMutexHelper Monitor(myMutex);
-   char mybuff[8];
+   char mybuff[16];
```
davidlt: Compiler now studies string format to understand what size buffer is required (min and max bytes). According to compiler this buffer (an many other) are too small to fit potential values. Note that in most (or all) cases `snprintf` is used thus we don't have buffer overflow, but string can be truncated.

src/XrdCms/XrdCmsReq.cc
```diff
@@ -134,17 +134,17 @@ void XrdCmsReq::Reply_Error(int ecode, const char *emsg, int elen)
...
-          case EPERM:        eval = kYR_EPERM;
-          case EACCES:       eval = kYR_EACCES;
-          case EIO:          eval = kYR_EIO;
-          case ENOMEM:       eval = kYR_ENOMEM;
-          case ENOSPC:       eval = kYR_ENOSPC;
-          case ENAMETOOLONG: eval = kYR_ENAMETOOLONG;
-          case ENETUNREACH:  eval = kYR_ENETUNREACH;
-          case ENOTBLK:      eval = kYR_ENOTBLK;
-          case EISDIR:       eval = kYR_EISDIR;
-          default:           eval = kYR_EINVAL;
```
davidlt: In this case whatever `ecode` is we set `eval` to `kYR_EINVAL`. This looks wrong, thus probably we forgot to add `break`s otherwise the whole thing can be reduced to a single line: `eval = kYR_EINVAL;`.

src/XrdCns/XrdCnsConfig.cc
```diff
@@ -150,6 +150,7 @@ int XrdCnsConfig::Configure(int argc, char **argv, char *argt)
                     else NoGo = NAPath("'-a'", Spec.argval);
                  break;
        case 'B': Opts |= optNoCns;
+                 [[gnu::fallthrough]];
```
davidlt: Fallthroughs must be explicitly marked. Note, that `[[fallthrough]]` attribute was added to C++17 and GNU makes it available in C++11 and C++14 via `[[gnu::fallthrough]]`. Note that Clang had for years `[[gnu::fallthrough]]`. One can also use GNU style attribute. They also support simple C++ comments with specific patterns. Read more here: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html (Search for `-Wimplicit-fallthrough`).

src/XrdCns/XrdCnsLogRec.hh
```diff
@@ -58,8 +58,8 @@ struct Ctl
 
 struct Arg
       {char Type;                   // Event code
-       char Mode[3];                // Mode (create, inv, mkdir)
-       char SorT[12];               // Size (closew, inv) | TOD (eol)
```
davidlt: In these two cases we always lost the last char, because `snprintf` needs to write terminating NULL byte. Thus we need to increase buffers in order not to loose the last digit.

src/XrdCrypto/XrdCryptosslX509.cc
```diff
@@ -507,8 +507,8 @@ const char *XrdCryptosslX509::IssuerHash(int alg)
       if (issueroldhash.length() <= 0) {
          // Make sure we have a certificate
          if (cert) {
-            char chash[15] = {0};
-            snprintf(chash,15,"%08lx.0",X509_NAME_hash_old(cert->cert_info->issuer));
+            char chash[30] = {0};
+            snprintf(chash, sizeof(chash), "%08lx.0", X509_NAME_hash_old(cert->cert_info->issuer));
```
davidlt: I don't fully understand this string format. Why do you need _width_ set to 8 and also ".0" at the end. Just looks a bit awkward. Does your hash really end with ".0"?

davidlt:
Especially we need to review all fallthroughs and double check that increasing buffer sizes do not cause any regressions for whatever reason.

abh3:
Michal and I looked over the changes. As you suggested, these cannot be pulled for a number of reasons but the intended changes are essentially correct. We have worked over the changes to address compiler portability  and backward compatibility issues but git head should now compile cleanly except for XrdHttpReq.cc (another developer is working on that as the solution is rather messy). Could you give your compiler a go and see if we missed anything?

davidlt:
Will try to compile master with GCC 7 and report back in 1-2 days.

davidlt:
You missed the following:
```
src/XrdClient/XrdClient.cc: In member function 'virtual bool XrdClient::OpenFileWhenRedirected(char*, bool&)':
src/XrdClient/XrdClient.cc:1375:14: error: enum constant in boolean context [-Werror=int-in-bool-context]
  options &= !kXR_delete;
              ^~~~~~~~~~
src/XrdClient/XrdClient.cc:1383:14: error: enum constant in boolean context [-Werror=int-in-bool-context]
  options &= !kXR_new;
              ^~~~~~~
```
You are using boolean operation where you needed a bitwise operation.

```
src/XrdCl/XrdClZipArchiveReader.cc: In destructor 'XrdCl::ZipArchiveReaderImpl::~ZipArchiveReaderImpl()':
src/XrdCl/XrdClZipArchiveReader.cc:246:23: error: ignoring return value of 'XrdCl::XRootDStatus XrdCl::File::Close(uint16_t)', declared with attribute warn_unused_result [-Werror=unused-result]
       pArchive.Close();
                       ^
In file included from src/XrdCl/XrdClZipArchiveReader.cc:27:0:
src/XrdCl/XrdClFile.hh:121:20: note: declared here
       XRootDStatus Close( uint16_t timeout = 0 ) XRD_WARN_UNUSED_RESULT;
                    ^~~~~
```

I am not fully sure how to properly handle this inside `~ZipArchiveReaderImpl()`. I know, that only saving return value (but not using it) already removed the error, but is not a right solution probably. Maybe dtor should be marked as `noexcept(false)` and should throw an exception/terminate.

Will check further. Now it compiles up to 71%.

davidlt:
One more:
```
src/XrdHttp/XrdHttpReq.cc: In member function 'int XrdHttpReq::PostProcessHTTPReq(bool)':
src/XrdHttp/XrdHttpReq.cc:1585:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
             }
             ^
src/XrdHttp/XrdHttpReq.cc:1587:11: note: here
           default: //read or readv
           ^~~~~~~
```

src/XrdCrypto/XrdCryptosslX509.cc
bbockelm: Yup, that's the correct format.  I believe it's to allow for collisions (so you can have a CA with `.0`, `.1`, `.2`, ....

davidlt:
Seems that's all to compile with GCC 7.

src/XrdCl/XrdClZipArchiveReader.cc
bbockelm: There is an attribute for this situation added in C++17; is there a pre-existing `gnu:` attribute that we could use?

abh3:
Yes, we did so on purpose. The old client is no longer supported and we no 
longer fix it. Starting with 4.6 it will also no longer be compiled so you 
wouldn'tsee that message. At 5.0 it will be completely removed.

Andy

abh3:
Ah, recall we have another developer working on that and he has come up 
with a fix but hasn't put it in the repo yet.

Andy

abh3:
Actually, the fix we will be using is to issue an error message is the 
status is not clean. That will be posted tomorrow.

Andy

src/XrdCl/XrdClZipArchiveReader.cc
davidlt: Looks like to void this official way would be to add a cast to void, i.e., ` (void) pArchive.Close();` but GCC currently doesn't detect that (bug #66425), Clang does. According to  C++17 standard we have `nodiscard` which is unused-result and it says that compile should only warn/error if cast to void is missing.

davidlt:
For example Fedora 26 should come with GCC 7.1 and mass rebuild is expected next month.
xrootd is still at 4.5 for Fedora 26 thus it still will need a fix downstream.

Same for CMSSW (at least for now).

Btw, is there ETA for 4.6?

david

abh3:
Yes, we wanted to release it this week but some CMS sites that were 
testing some mods to improve performance let that slip through the cracks. 
So, they are testing it now (a delay of a few days). Plus, now that you 
brought the gcc 7 problems to our attention we wanted to fix those as 
well (oh yes, thank you for posting the fixes -- that helped a lot!).. 
So, if all goes well, next week.

Andy

src/XrdCrypto/XrdCryptosslX509.cc
gganis: Hi, the '.0' is because this information is mostly used to load the CA files which by convention are '<hash>.0' (for a reason which is likely to be the one that Brian explained). I agree that this could have been done differently.

davidlt:
Hi,

Quick question. If master will become 4.6 why old client is still not disabled?

Cheers,
david

simonmichal:
Hi David,

I will disable old client before I cut the release, just hadn't had time yet to do it ;-)

Cheers,
Michal

xrootd-dev:
Simply because we haven't gotten to it yet. That will be the case when we 
cut the official elease.

Andy

davidlt:
Fedora are already testing preparing for massive rebuild with GCC 7: http://pkgs.fedoraproject.org/cgit/rpms/gcc.git/log/

If everything goes fine there should be so nightly Fedora Rawhide builds for testing if you like, i.e. not need to build your own toolchain.

abh3:
We reworked this pull request to solve some backward comparability issues but essentially all of the recommendation have been applied.

ellert:
@abh3: Have you pushed the reworked changes to github? If possible I would like to backport the upstream changes to the Fedora package.

davidlt:
@ellert all changes are in GitHub, that's what I use for CMSSW + GCC 7.

ellert:
@davidlt: I know your pull request is on github. However, the last comment by @abh3 says "We reworked this pull request ..." and I was asking about this reworked version.

davidlt:
@ellert everything is already committed, but in pieces (thus multiple commits). This PR was just for notes of what is wrong while compiling with GCC 7. Thus you need to fish out those particular commits (should be easy to identify based on commit message).

simonmichal:
@ellert I am going to release 4.6.0, which includes all those changes, early this week. I'm not sure if it's worth doing a backport as you will be pushing 4.6.0 to EPEL/Fedora soon.

ellert:
I see it now. I got confused by the output of "git log". The dates are not in order, and after I saw entries from November I stopped looking further down.

ellert:
After extracting the changes from github I still see 2 issues. I have created pull request #457 to address them.

PR #457, opened 2017-02-06
Gcc7 issues

ellert:
Here are two issues I found while extracting the gcc 7 fixes from git.
1) The compilation fails where logical not is used instead if bitwise not in XrdClient.cc
2) In XrdSecProtocolpwd.cc the if statements added to replace the fall though doesn't actually fall through.

### DIFF
--- a/src/XrdClient/XrdClient.cc
+++ b/src/XrdClient/XrdClient.cc
@@ -1372,7 +1372,7 @@ bool XrdClient::OpenFileWhenRedirected(char *newfhandle, bool &wasopen)
 	Info(XrdClientDebug::kHIDEBUG,
 	     "OpenFileWhenRedirected", "Stripping off the 'delete' option." );
 
-	options &= !kXR_delete;
+	options &= ~kXR_delete;
 	options |= kXR_open_updt;
     }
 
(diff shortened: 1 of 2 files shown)


davidlt:
XrdClient will be deprecated. They are not fixing on purpose. They will remove/disable it from default build before cutting 4.6 release.

ellert:
There is other software, such as gfal and parts of root, that use the old client library. So removing it completely would be very disruptive. Disabling it in the default build while having a configuration option to enable it is OK, but removing the code from the sources or on purpose make it un-compilable I think would be a bad idea.
Backward compatible changes like the complete removal of a library usually happens for major releases, not minor.

simonmichal:
Indeed, when we were planning with Andy to remove the old client completely we didn't realized that the old client has some public header files, etc. So yes, Mattias is right.

@ellert you're also right about the fall through, good catch!!! Thanks!

ljanyst:
Just to add my two cents here. We announced the intention to drop the old client some six years ago and officially deprecated it on Jun 5th, 2014. That's a lot of time for people to move on. AFAIR, the parts of ROOT that depend on the old client are redundant, not used for years now and can be disabled. There was the same resistance to dropping the proprietary makefiles and autotools builds. Sometimes you just need to make a cut.

   Lukasz

ljanyst:
I also vaguely recall working with Alejandro to port GFAL to the new client circa beginning of 2014, so probably that's not a problem either.

simonmichal:
Thanks for your feedback! :-)

ljanyst:
No problem. Make sure you double check though. Have a special look at PROOF.

davidlt:
Have you modified XrdClient to include various `deprecated` attributes and `#warning` to be verbose as possible about it? Could you also share a list of deprecated headers, which I could use to cross-check who might be using them? Mainly I want to check CMSSW.

PR #461, opened 2017-02-10
Fix CRL verification bug

ellert:
Hi!

After uploading xrootd 4.6.0 to Fedora and EPEL I got some feedback that authentication failed with many sites, see:

https://bodhi.fedoraproject.org/updates/FEDORA-2017-0eca8dbb12

I could reproduce the problem, and this is my proposed fix.

This issue was not caused by the changes to support openssl 1.1.0, but was introduced with the improvements to the CRL handling that was part of the 4.6.0 release.

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -3946,12 +3946,15 @@ XrdCryptoX509Crl *XrdSecProtocolgsi::LoadCRL(XrdCryptoX509 *xca, const char *sub
       return crl;
    }
 
+   String caroot(subjhash);
+   if (caroot.endswith(".0"))
+     caroot.erase(caroot.length() - 2, 2);
+
    // Get the CA hash
-   String cahash(subjhash);
    int hashalg = 0;
-   if (strcmp(subjhash, xca->SubjectHash())) hashalg = 1;
-   // Drop the extension (".0")
-   String caroot(cahash, 0, cahash.find(".0")-1);
+   if (strcmp(caroot.c_str(), xca->SubjectHash())) hashalg = 1;
+
+   String cahash = caroot + ".0";
 
    // The dir
    String crlext = XrdSecProtocolgsi::DefCRLext;


gganis:
Thanks for reporting.
This part of the code was not touched recently. The recent patch introduced better checking which have probably triggered the problem.
What are the configuration options run by the servers showing the problem? And the CA,hashes involved?
Just to understand why we did not see it.

ellert:
I am not quite sure what the server configuration is, I debugged this from the client side.

The file that I found which I had access to and that triggered the problem was:
root://fax.mwt2.org:1094//pnfs/uchicago.edu/atlasdatadisk/rucio/data16_13TeV/eb/1a/DAOD_TOPQ1.10314911._000041.pool.root.1

The same file has replicas at other servers and those where not affected. If I used --debug 3 option to xrdcp I found this for the server that did not work:

[2017-02-10 17:32:27.470539 +0100][Debug  ][XRootDTransport   ] [fax.mwt2.org:1094 #0.0] kXR_protocol successful (type: manager [], protocol version 289)
[2017-02-10 17:32:27.597157 +0100][Debug  ][XRootDTransport   ] [fax.mwt2.org:1094 #0.0] Authentication is required: &P=gsi,v:10200,c:ssl,ca:d690e530

Whereas a server where the download worked says:

[2017-02-10 17:36:49.514501 +0100][Debug  ][XRootDTransport   ] [sedoor1.bfg.uni-freiburg.de:1094 #0.0] kXR_protocol successful (type: server [], protocol version 310)
[2017-02-10 17:36:49.549587 +0100][Debug  ][XRootDTransport   ] [sedoor1.bfg.uni-freiburg.de:1094 #0.0] Authentication is required: &P=gsi,v:10300,c:ssl,ca:7ecb2657.0|dd4b34ea.0

So the servers report different protocol versions (289 vs. 310) and the format for the ca option is different - the older server reports the only the md5 hash while the newer version reports both md5 and sha1, but more importantly the older version reports the hash only, while the newer version appends ".0" to the hashes (ca:d690e530 vs. ca:7ecb2657.0|dd4b34ea.0).

gganis:
Ok, I see. That seems a really old server version, before 2011 at least.
I am ok with your patch, which fixes a bug. 
I will merge, thanks.

Issue #463, opened 2017-02-13
xrootd 4.6.0 from epel-testing  is crashing on DPM instances


andrea-manzi:
I have already contacted by mail the devs and Mattias as EPEL packager and  i'm now opening also a ticket about it.

We are seeing Xrootd crashing on our DPM testbeds after the upgrade to Xrootd 4.6.0 from EPEL-testing

the backtrace:

```
(gdb) bt
#0  XrdCryptogsiX509Chain::Verify (this=0x661e90, errcode=@0x7fffffff959c, vopt=<value optimized out>) at /usr/src/debug/xrootd-4.6.0/src/XrdCrypto/XrdCryptogsiX509Chain.cc:109
#1  0x00007ffff6bc8c5a in XrdSecProtocolgsi::VerifyCA (opt=<value optimized out>, cca=0x661e90, CF=<value optimized out>) at /usr/src/debug/xrootd-4.6.0/src/XrdSecgsi/XrdSecProtocolgsi.cc:4252
#2  0x00007ffff6bcd169 in XrdSecProtocolgsi::GetCA (cahash=0x660fb0 "5168735f.0", cf=0x7ffff699b520, hs=0x0) at /usr/src/debug/xrootd-4.6.0/src/XrdSecgsi/XrdSecProtocolgsi.cc:4399
#3  0x00007ffff6bcdd6a in XrdSecProtocolgsi::GetSrvCertEnt (pfeRef=..., cf=0x7ffff699b520, timestamp=<value optimized out>, certcalist=...) at /usr/src/debug/xrootd-4.6.0/src/XrdSecgsi/XrdSecProtocolgsi.cc:5378
#4  0x00007ffff6bd4972 in XrdSecProtocolgsi::Init (opt=..., erp=0x7fffffffb530) at /usr/src/debug/xrootd-4.6.0/src/XrdSecgsi/XrdSecProtocolgsi.cc:653
#5  0x00007ffff6bd796a in XrdSecProtocolgsiInit (mode=<value optimized out>, parms=<value optimized out>, erp=0x7fffffffb530) at /usr/src/debug/xrootd-4.6.0/src/XrdSecgsi/XrdSecProtocolgsi.cc:2674
#6  0x00007ffff6de7d99 in XrdSecPManager::ldPO (this=0x7ffff6fefd60, eMsg=0x7fffffffb530, pmode=115 's', pid=<value optimized out>,
[...]

PR #466, opened 2017-02-15
Fix for CA chain verification segfault (issue #463)

ellert:
This is a proposed fix for issue #463.

### DIFF
--- a/src/XrdCrypto/XrdCryptogsiX509Chain.cc
+++ b/src/XrdCrypto/XrdCryptogsiX509Chain.cc
@@ -117,7 +117,8 @@ bool XrdCryptogsiX509Chain::Verify(EX509ChainErr &errcode, x509ChainVerifyOpt_t 
    }
    //
    // Check the end-point entity certificate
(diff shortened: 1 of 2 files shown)


andrea-manzi:
Hi Mattias, 
i just downloaded the rpms from Koji and the server is now starting without problems:-)
i will wait for the rpms from EPEL testing to do more tests and to confirm that everything is fine
thanks!
Andrea

ellert:
Here the XrdCryptogsiX509Chain::Verify is used to verify a chain of CA certificates, something I don't think it was originally intended to do.

In this case there should not be an EEC certificate in the chain, so the result for a verified CA certificate chain is that XrdCryptogsiX509Chain::Verify returns false (i.e. the verification fails as an EEC certificate chain) and the error returned is kNoEEC.

I.e. there is a failure reported if the chain verifies as an EEC chain (and hence is not a CA chain) or if the error returned is something else than kNoEEC.

gganis:
Hi, I have uploaded a patch to fix this problem, accounting for the case of not self-signed CAs .
I had previously managed to reproduce the problem on my machine and this patch fixes it.

Issue #465, opened 2017-02-14
xrootd 4.6.0 xrdcp treats missing CRL as an error


ellert:
This issue was reported as a comment in the bodhi update system in Fedora:

https://bodhi.fedoraproject.org/updates/FEDORA-2017-0eca8dbb12

When I first investigated this I found the problem with the incompatibility with older servers that I reported in #461 which showed a similar problem, but the original reporter came back and reported that his problem still was there after updating to the version fixing the older server incompatibility.

I tried to issue the exact same command that was used by the reporter, and it succeeded without problems for me. So I asked for more information.

The detailed logs that were provided showed that on the users system the CRL files for the CA signing the server's certificate ware not present. And if I move these CRL files away on my system I can reproduce the problem.

An absent CRL file should not cause an error. The CRL file being absent means "do not check". If the CRL file is there it must be valid (i.e. not expired) and those certificates that are revoked in it must not be validated.

The CRL checking code in 4.6.0 treats a missing CRL as an expired CRL, which is not correct. This is a regression form the 4.5.0 version.

The logs from the original reporter attempting the same xrdcp command with 4.5.0 and 4.6.0 installed on the same system are attached.

[xrdcp-4.5.0.txt](https://github.com/xrootd/xrootd/files/773755/xrdcp-4.5.0.txt)
[xrdcp-4.6.0.txt](https://github.com/xrootd/xrootd/files/773754/xrdcp-4.6.0.txt)

gganis:
This should be now fixed in the HEAD of master and stable-4.6.x .
Thanks for reporting.

ellert:
This issue has reappeared.

gganis:
Should be fixed now. Please confirm. Sorry for the inconvenience.

ellert:
The case when the CRL is absent works now. However, a warning appears in the terminal which I think should not be there:
170306 21:21:08 2183 secgsi_GetCA: CRL entry for '70d35895.0:1' needs refreshing: clean the related entry cache first (0x7fd3b002d700)

ellert:
Is anyone looking into removing the confusing warning?

gganis:
Sorry, I am not able to reproduce it.
What configuration produces this (valid/invalid/missing/expired CRL)?

ellert:
With the CRLs (70d35895.r0 and d690e530.r0) present in /etc/grid-security/certificates:

[...]

PR #467, opened 2017-02-15
Absent CRL is not an error (issue #465)

ellert:
This is a proposed fix for issue #465.

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -4008,7 +4008,7 @@ XrdCryptoX509Crl *XrdSecProtocolgsi::LoadCRL(XrdCryptoX509 *xca, const char *sub
    // If not required, we are done
    if (CRLCheck < 2 || (dwld == 0)) {
       // Done
-      return crl;
+      return crl ? crl : (XrdCryptoX509Crl*)(-1);
    }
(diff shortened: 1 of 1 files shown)


gganis:
I see the problem, but I believe the fix should be in the way the null return is interpreted.
I am preparing a patch. Let you know when I am done.

gganis:
I have just uploaded a patch for #465 fixing the interpretation of the return code.
So I am closing this PR without merging it.

ellert:
CRLCheck = 1 is supposed to implement the standard behaviour, i.e. "use if present". This means that a missing CRL should nor be considers an error. But it also means that a present but expired CRL should trigger an error. The code still treats a missing and an expired CRL the same. Now it accept both instead of rejecting both as it did before, but it should accept a missing CRL while rejecting an expired one.

gganis:
Expired CRL contain valid information though not updated. As documented (http://xrootd.org/doc/dev45/sec_config.htm#_Toc464225463) CRLCheck=1 is expected to trigger an error is the loaded CRL is invalid for other reasons (for example if the signature verification fails), 
but accepts expired but otherwise valid CRLs, as CRLCheck=2 does.

gganis:
Forgot to mention the a patch uploaded to day to the HEAD of stable-4.6.x and master makes sure that the documented behaviour is honoured.

PR #469, opened 2017-02-17
A valid legacy proxy should not trigger an error msg

ellert:
This addresses a comment received on the xrootd update for EPEL 7:
https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2017-9b2cd39ee3

"xrdcp from xrootd-client-4.6.0-3.el7.x86_64 downloads file but with ERROR message"


gganis:
I have applied a slightly modified patch.
Thanks for providing it.

PR #471, opened 2017-02-22
Rename InitCTX() to Init() and make it not pure virtual, as requested by Brian

ffurano:
Please review and let me know.
When this phase is finished we will push the secxtractor library to epel-testing

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1385,7 +1385,7 @@ int XrdHttpProtocol::InitSecurity() {
 
   }
 
-  if (secxtractor) secxtractor->InitCTX(sslctx, XrdHttpTrace->What);
+  if (secxtractor) secxtractor->Init(sslctx, XrdHttpTrace->What);
 
   ERR_print_errors(sslbio_err);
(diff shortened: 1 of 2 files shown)


ellert:
The signature in 4.5.0 was

    // Initializes an ssl ctx
    virtual int Init(SSL_CTX *, int) = 0;

So if you wanted to make it backward compatible I don't think this was right. Brian's (@xrootd-dev) comment read:

> - Rename InitCtx back to the original Init
> - Make InitSSL / FreeSSL not pure-virtual (put in the trivial inline implementation in the base class).

i.e. the not pure virtual comment referred to InitSSL and FreeSSL, not Init.

ellert:
In update 4.6.0-5 in EPEL testing I used the following patch in line with my previous comment:
http://pkgs.fedoraproject.org/cgit/rpms/xrootd.git/plain/xrootd-backward-compat.patch

ffurano:
Ok, thank you Mattias, so I was missing the second little point. I am making a pull request.

ffurano:
Actually I added the commit to this pull request.

ellert:
Shouldn't Init() still be pure virtual though? It was in 4.5.0.

PR #473, opened 2017-03-02
Ignore unnecessary parameter.

bbockelm:
Without explicitly ignoring the argument value, we trigger an assert to make sure all arguments have been processed - even if it is otherwise ignored.

Assert failure looked like:
```
170301 22:44:17 2410886 XrdFileCache_Config: pfc.nramread is deprecated, please use pfc.ram instead. Ignoring this directive.
xrootd: /home/cse496/bbockelm/projects/xrootd/src/XrdFileCache/XrdFileCacheConfiguration.cc:413: bool XrdFileCache::Cache::ConfigParameters(std::string, XrdOucStream&, XrdFileCache::TmpConfiguration&): Assertion `config.GetWord() == 0 && "Cache::ConfigParameters() lost argument"' failed.
```

triggered by the presence of `pfc.nramread 4`.


osschar:
Sigh, created #474 ...

PR #474, opened 2017-03-02
Consume parameter given to ignored pfc.nramread configuration directive.

osschar:


abh3:
This is (apparently) a duplicate of what Brian's pull request happens to be. According to the author of this pull request, we should accept Brian's pull request and close this one out. This shall be done.

PR #480, opened 2017-03-13
Don't write error when CRL is absent

ellert:
This addresses the bogus error message mentioned in #465.


gganis:
Merged. Thanks!

PR #486, opened 2017-03-24
Fix ability to read back a file written with O_RDWR flags + improvements in the logs

glpatcern:
The bug manifested itself only 2 times when the xrootd server attempted to read back a file written out of order (to compute the checksum).
The 'improvements' are actually about dropping way too many messages in the absence of a proper mechanism to classify them as debug level. @sponce, what do you think?


sponce:
I think this is all good !

PR #487, opened 2017-03-27
XrdApp: Add XrdClProxyPlugin implementation

esindril:
I've added as a .so library since it's not a shared library but a module which is loaded dynamically.

### DIFF
--- a/packaging/rhel/xrootd.spec.in
+++ b/packaging/rhel/xrootd.spec.in
@@ -654,6 +654,7 @@ fi
 %files libs
 %defattr(-,root,root,-)
 %{_libdir}/libXrdAppUtils.so.1*
+%{_libdir}/libXrdClProxyPlugin-4.so
 %{_libdir}/libXrdCks*-4.so
 %{_libdir}/libXrdCrypto.so.1*
 %{_libdir}/libXrdCryptoLite.so.1*
--- a/src/XrdApps.cmake
+++ b/src/XrdApps.cmake
@@ -2,6 +2,11 @@
 include( XRootDCommon )
 
 #-------------------------------------------------------------------------------
+# Modules
+#-------------------------------------------------------------------------------
+set( LIB_XRDCL_PROXY_PREFIX XrdClProxyPrefix-${PLUGIN_VERSION} )
+
+#-------------------------------------------------------------------------------
 # Shared library version
 #-------------------------------------------------------------------------------
 set( XRD_APP_UTILS_VERSION   1.0.0 )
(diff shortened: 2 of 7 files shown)


abh3:
Hi Elvin,

Almost there. Plugins generally cannot exist as pure ".so" files. We require that they be versioned. Take a look in XrdServer.cmake or (simpler) XrdSec.cmake to see how versioned plugin libraries are created. In the end you'll have something like libXrdClProxyPlugin-4.so where the "-4" is the version number. OK, this is what we do for the server. Michal, other than the security plugins which are auto-handled, do we do the same for other client-side plugins?

Andy

simonmichal:
Hi Andy,

The version check is done properly for the XrdClProxyPlugin:

When the plugin version doesn't match it is rejected:
> [simonm@idefix XrdCl]$ strings -a $XRD_PLUGIN | grep "@V"
> @V:XrdClGetPlugIn v3.9.9
> [simonm@idefix XrdCl]$ strings -a libXrdCl.so | grep "@V"
> @V:client v4.9.9
> [simonm@idefix XrdCl]$ ./xrdcp -f -d 1 root://esvm000//tmp/file1.dat /tmp/dump
> [2017-03-28 15:14:19.234129 +0200][Debug  ][Utility           ] Unable to process user config file: [ERROR] OS Error: No such file or directory
> [2017-03-28 15:14:19.234302 +0200][Info   ][Utility           ] Env: Importing from shell XRD_PLUGIN=/home/simonm/test/xrootd-3.9.9/build/src/libXrdClProxyPlugin.so as PlugIn
> [2017-03-28 15:14:19.234335 +0200][Debug  ][PlugInMgr         ] Initializing plug-in manager...
> [2017-03-28 15:14:19.234387 +0200][Debug  ][PlugInMgr         ] Loading default plug-in from /home/simonm/test/xrootd-3.9.9/build/src/libXrdClProxyPlugin.so...
> [2017-03-28 15:14:19.234611 +0200][Debug  ][PlugInMgr         ] Error while loading /home/simonm/test/xrootd-3.9.9/build/src/libXrdClProxyPlugin.so: Unable to load client plugin /home/simonm/test/xrootd-3.9.9/build/src/libXrdClProxyPlugin.so
> [2017-03-28 15:14:19.234625 +0200][Debug  ][PlugInMgr         ] Failed to load default plug-in from /home/simonm/test/xrootd-3.9.9/build/src/libXrdClProxyPlugin.so
> [3.015kB/3.015kB][100%][==================================================][3.015kB/s]  

If the plugin version does match it is loaded:
> [simonm@idefix XrdCl]$ strings -a $XRD_PLUGIN | grep "@V"
> @V:XrdClGetPlugIn v4.9.9
> [simonm@idefix XrdCl]$ strings -a libXrdCl.so | grep "@V"
> @V:client v4.9.9
> [simonm@idefix XrdCl]$ ./xrdcp -f -d 1 root://esvm000//tmp/file1.dat /tmp/dump
> [2017-03-28 15:15:05.885847 +0200][Debug  ][Utility           ] Unable to process user config file: [ERROR] OS Error: No such file or directory
> [2017-03-28 15:15:05.885951 +0200][Info   ][Utility           ] Env: Importing from shell XRD_PLUGIN=/home/simonm/test/xrootd-4.9.9/build/src/libXrdClProxyPlugin.so as PlugIn
> [2017-03-28 15:15:05.885971 +0200][Debug  ][PlugInMgr         ] Initializing plug-in manager...
> [2017-03-28 15:15:05.885992 +0200][Debug  ][PlugInMgr         ] Loading default plug-in from /home/simonm/test/xrootd-4.9.9/build/src/libXrdClProxyPlugin.so...
> [3.015kB/3.015kB][100%][==================================================][3.015kB/s] 

To be honest the XrdClProxyPlugin is the first plugin that will be included with xrootd core, so there is no convention yet. Hence, it's up to you whether we should follow the "-4" naming convention, or not ;-)

Cheers,
Michal

abh3:
Hi Michal,

There were two reasons that we adopted the “-n” convention. One was so that we could supports more than one version of a plugin at the same time (i.e. filenames would not conflict and could exist in parallel); the plugin manager automatically loads a compatible version, and b) to make it harder for people to link against the plugin (strictly forbidden by EPEL). As for the latter, I wanted to actual create non-“so” files for plugins since the dlsym doesn’t care what the file name is (e.g. <libname>.pi) but apparently cmake couldn’t handle that kind of thing, sigh. So, for consistency’s sake with server-side plugins it would probably make sense to add the “-n” suffix. Anyway, give it a try and see what happened with the “pinloader”.

Andy

PR #493, opened 2017-04-03
Fix memory leaks in SecEntity structure for XrdHttp.

bbockelm:
This fixes two problems:
1.  The `SecEntity` cleanup code in `XrdHttp` did not clean up two fields of the data structure.
2.  Whenever a security extractor is used (which should be always!), the host data was extracted twice - and not cleaned up in between - leaking the original data.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -291,8 +291,6 @@ XrdProtocol *XrdHttpProtocol::Match(XrdLink *lp) {
 int XrdHttpProtocol::GetVOMSData(XrdLink *lp) {
   TRACEI(DEBUG, " Extracting auth info.");
 
-  SecEntity.host = GetClientIPStr();
-
   X509 *peer_cert;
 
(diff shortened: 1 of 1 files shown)


bbockelm:
Please backport to 4.6.x series too - we're observing these memory leaks in the wild!

abh3:
Hi Brian,

Given that 4.6.0 less than ideal (putting it mildly) there doesn’t seem much reason to back port to it as we would prefer people not run it in the first place. How about 4.6.1?

Andy

bbockelm:
Yup, that's that's what I intended to mean by suggesting a backport to `4.6.x` series.  I just didn't want to wait for `4.7.0` for this fix.

abh3:
Fabrizio merged this fix into his pull request, so no need to merge this one.

abh3:
Actually, talking about 4.6.1; we are thinking of just releasing 4.7.0 as we have accumulated a lot of new stuff (mostly systemd related stuff). We might not but just wanted o give you a heads up in case you have thoughts about it.

Andy

PR #495, opened 2017-04-03
Fill the XrdSecEntity struct according to the last discussions,

ffurano:
in particular about the short name

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -303,27 +303,50 @@ int XrdHttpProtocol::GetVOMSData(XrdLink *lp) {
   if (peer_cert) {
 
     // Add the original DN to the moninfo. Not sure if it makes sense to parametrize this or not.
+    if (SecEntity.moninfo) free(SecEntity.moninfo);
     SecEntity.moninfo = X509_NAME_oneline(X509_get_subject_name(peer_cert), NULL, 0);
 
-    // Here we have the user DN, we try to translate it using the XrdSec functions and the gridmap
+    // Here we have the user DN, and try to extract an useful user name from it
     if (SecEntity.name) free(SecEntity.name);
+    SecEntity.name = 0;
+    // To set the name we pick the first CN of the certificate subject
+    // and hope that it makes some sense, it usually does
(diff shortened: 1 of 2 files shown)


bbockelm:
This appears to fix two out of three memory leaks also fixed by #493.  To make backporting easier, can you separate the bugfixes and new features to separate commits?  Or maybe just make sure #493 is merged first?

Thanks!

xrootd-dev:
Just to make sure I don’t make a mess. I first pull 493 and then afterwards 495 and no conflicts will arise, right?

Andy

bbockelm:
I suspect 495 won't be mergeable after 493 goes in.

But Fabrizio can then easily fix the merge - and you have an easier time backporting!

Brian

gganis:
Hi Fabrizio,

I am not 100% sure, but for XrdSecEntity.name in the case no username was specified, I thought we said to take the last 8 digits of the first '/CN=', without spaces, i.e., from '.../CN=Fabrizio Furano/...' we should have got 'ioFurano' . 
If that is the case, then I am not sure the code does that: the buffer is 64 in length and the space is not removed.
Do I remember wrongly?

Gerri

abh3:
Hi All,

Gerri remembers correctly. The 8 character name is the one that gets put into the loginid for "kXR_login”. The entity structure name is, in fact, the full distinguished name. Unless someone tells me otherwise in the next couple of hours, I will merge in 495 (which should include Brian’s fix) and then Fabrizion will have to generate a new pull request with the login-name fix.

Andy

xrootd-dev:
Ciao Gerri,

 indeed, I forgot to shorten the name. Good that you don't share the other doubts I had.
 Thank you for the review, I hope I'll be able to do this today-ish

f

xrootd-dev:
Hi again,

 I am on this small thing, and re-re-re-read the notes. They
were mentioning only the link name for this 8-chars thing. Not
SecEntity.name

 In fact I believe it would make no sense to mangle the username, would it ?

f

abh3:
Hi Fabrizio,

Obviously you figured it out beacause you had a pull request for this. 
Indeed, we were only talking about the link name not the qualified name.

Andy

ffurano:
Hi Andy,

 yes, a quick call to Gerri and we agreed on what you say. I hope it's good now,
in any case please let me know.

Fabrizio

PR #501, opened 2017-04-10
Support Python3

alexpearce:
This PR adds Python 3 support to the Python bindings.

The changes are relatively minor, but I haven't been able to run the Python tests. Are there instructions for doing so? There's an `env.py` file that it looks like I should source somehow, and I think I should have a server running on `localhost`, but I don't know where to start. Any help would be appreciated!

### DIFF
--- a/bindings/python/libs/client/file.py
+++ b/bindings/python/libs/client/file.py
@@ -21,6 +21,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 #-------------------------------------------------------------------------------
+from __future__ import absolute_import, division, print_function
 
 from pyxrootd import client
 from XRootD.client.responses import XRootDStatus, StatInfo, VectorReadInfo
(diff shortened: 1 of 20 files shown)


bindings/python/libs/client/file.py
```diff
@@ -42,7 +43,7 @@ def __exit__(self, type, value, traceback):
   def __iter__(self):
     return self
 
-  def next(self):
+  def __next__(self):
```
chrisburr: In order to maintain compatibility with python 2 you need to add `next = __next__` to the class.

alexpearce: Good spot, thanks! Fixed in 5164c6c.

alexpearce:
Sorry to bother, but is there a chance this could be merged?

simonmichal:
Hi Alex,

It seems to me that you have a typo in  bindings/python/libs/client/copyprocess.py : 40

you have:
> if 'success' in results:

should be:
> if 'status' in results:

Otherwise it looks good :-)
Anyway sorry for the delay!

Cheers,
Michal

alexpearce:
Thanks for the review! I've fixed the typo.

PR #503, opened 2017-04-12
Address two memory leaks

jthiltges:
- bpxy in XrdSecProtocolgsi::Authenticate
- SSL state after calling X509_verify_cert in XrdCryptosslX509VerifyChain


gganis:
Merged.
Ported also to stable-4.6.x to appear in 4.6.1 .
Thanks a lot!

Issue #502, opened 2017-04-11
Add continuous integration builds


alexpearce:
It's always nice for developers to have some immediate and obvious feedback when
making pull requests. I think a really nice way of giving them this is to have some form of continuous integration that runs for every new PR, like [using Travis CI](https://travis-ci.org/getting_started).

For xrootd, this could mean building the project (i.e. running `cmake ...` and then `make/make install`) and then running the tests (those under `tests` and those that are bindings specific). The CI configuration can be written like a matrix for different platforms and parameters, so one could easily set up CI for all combinations of linux and macOS, and Python 2.7 and 3.5, for example.

If the xrootd developers are willing to [flick the switch](https://travis-ci.org/getting_started) to enable Travis, I'm happy to work on the configuration file in a PR.

simonmichal:
Hi Alex,

Thanks for the offer! :-)

Actually, we do have a CI:
https://st-ci.cern.ch/

We use a Jenkins instance at CERN. Currently, there is also some ongoing work to mirror our github repo in cern gitlab and use gitlab runner for continuous integration.

That said, I would be more than happy to use Travis CI and have everything in one place!

But before we go ahead could you clarify few things:

1) Can we provide our own build nodes? Or does Travis provide their own build nodes? If we can only relay on Travis build nodes, how many do we get?
 
2) Are you willing to maintain the Travis CI for XRootD?

3) For what platforms can we build? Currently we are interested in epel6, epel7, latest 3 fedora releases, fedora-rawhide, and OSX, can we cover that with Travis?

4) At least the release builds should be done with mock, as they go to epel/fedora, can we do that with Travis CI?

5) Can we run unit/functional tests in docker containers?

Cheers,
Michal

alexpearce:
> Actually, we do have a CI:
> https://st-ci.cern.ch/

Aha! I did not know about this. That's already quite useful. It looks like it doesn't test GitHub PRs, but still 👍 

For your questions:

> Can we provide our own build nodes? Or does Travis provide their own build nodes? If we can only relay on Travis build nodes, how many do we get?

Travis provide the nodes. You're allowed to run whatever you like, but within a time limit of around an hour, I think.

> Are you willing to maintain the Travis CI for XRootD?

Yes.

> For what platforms can we build? Currently we are interested in epel6, epel7, latest 3 fedora releases, fedora-rawhide, and OSX, can we cover that with Travis?

The available environments are defined here:

https://docs.travis-ci.com/user/ci-environment/

[...]

PR #504, opened 2017-04-14
WIP: Add Travis CI configuration

alexpearce:
Given the positive reaction in #502, here's a PR with a Travis CI configuration file.

This PR is WIP while I figure out how to get as much as possible building on Travis (different compilers, Python versions, platforms, …). So, lots of little tests commits, then I'll rebase once things look good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 addons:
   apt:
     packages:
+      - libxml2-dev
       - libcppunit-dev
 language: cpp
 compiler:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: false
+addons:
+  apt:
+    packages:
+      - libcppunit-dev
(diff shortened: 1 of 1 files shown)


alexpearce:
Ah. This PR was far from complete 😅

simonmichal:
Ah. My bad, I was thinking it's just a stub to remove the warring that builds fail ;-)

Ho do we fix it?

alexpearce:
Right now, I'm [having trouble even building the package](https://travis-ci.org/xrootd/xrootd/jobs/222125844#L509), so this is the first thing to address.

Once that's done, I need to set up the builds on the multiple platforms and for the various Python versions XRootD supports. That step should be easier.

ljanyst:
You either revert the merge or reset the head to a past commit and force push. The second option is probably cleaner.

simonmichal:
Thanks for the tip :)

PR #512, opened 2017-05-05
xrootdfs: remove default fuse argument "allow_other" as it is impossible to unset

fizmat:
Motivation:
This prevented mounting if user_allow_other is not set in fuse.conf
So a user with no root access could not use xrootdfs and could not change this argument without recompiling.
Change:
Don't pass allow_other to fuse by default.
Result:
Now xrootdfs works if user_allow_other is disabled, and allow_other argument can be passed when mounting if it is needed.

### DIFF
--- a/src/XrdFfs/README
+++ b/src/XrdFfs/README
@@ -49,6 +49,7 @@ XROOTDFS_OFSFWD: if the xrootd cluster uses ofs.forward (or ofs.forward 3way)
     and expect the operations to be forwarded to the data nodes by the 
     redirector. Otherwise one can define XROOTDFS_OFSFWD to '0'. XrootdFS will 
     then go to individual data node for mv/rm/rmdir/trunc.
+XROOTDFS_NO_ALLOW_OTHER: do not pass option allow_other to fuse.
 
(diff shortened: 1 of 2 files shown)


wyang007:
Hi, sorry I lose track of your original request. 

I think completely remove "allow_other" may not work well as most users won't know this. So I wonder if you can do something like 
if (getenv("I_dont_allow_other"))
      cmdline_opts[2] = strdup("fsname=xrootdfs,...")    // no allow_other
else
      cmdline_opts[2] = strdup("fsname=xrootdfs, allow_other, ...")

You may want to have the environment name look like "XRD_FFS_XYZ". Let me know when you have another pull request

Many thanks!
Wei    |     yangw@slac.stanford.edu

fizmat:
Yes, it's a good idea to protect existing use-cases that may break with the change. The env variable seems to be the simplest place to place this extra option.
I added it and put it in the readme.

Do you think the defaults should ever change, with a long enough deprecation warning maybe?

wyang007:
Thanks! 
I think it is probably better to leave "allow_other" in the default simple because most users don't know what it is.

PR #515, opened 2017-05-11
xrootdfs: implement xrootdfs_create

fizmat:
Motivation:
In general, a create operation is nice to have.
Especially useful for a dcache xrootd door.
When FUSE has to use mknod, it results in an empty file being created, closed, then opened again for writing.
In dcache this does not work and needs an explicit check to allow overwriting the empty file.
Just appending to the empty file is not possible in dcache, and FUSE does not set the DELETE flag, as it does not expect it to be needed to append to an empty file.
Implementing a create operation makes copying onto dcache faster, as it does not require creating the file twice.

Change:
Implement xrootdfs_create that works as xrootdfs_mknod, except:
 - it opens the file over a numbered link, instead of unnumbered, using the same link numbering process as xrootdfs_open
 - it creates the wcache for the file
 - it does not close the file, and returns the file descriptor

Result:
Copying files on xrootdfs uses the new CREATE function.
The current release of dcache (without the workaround) xrootd door can be mounted over xrootdfs and used.
This should still work with xrootd clusters with/without cns and with/without sss, but was not tested. Is there an existing testing infrastructure or methodology for this?

### DIFF
--- a/src/XrdFfs/XrdFfsXrootdfs.cc
+++ b/src/XrdFfs/XrdFfsXrootdfs.cc
@@ -354,22 +354,54 @@ static int xrootdfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler,
     }
 }
 
-static int xrootdfs_mknod(const char *path, mode_t mode, dev_t rdev)
+static int xrootdfs_do_create(const char *path, const char *url, int oflags, bool use_link_id, int *fd)
+/* Actually implement creating a file as generally as possible.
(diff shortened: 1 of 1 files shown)


wyang007:
Thanks!

xrootdfs_mknod() may not be needed in the future (note we have "#define FUSE_USE_VERSION 26" at the top). So I may merge _create() and _do_create() and eliminate _mknod(). Low priority...

PR #521, opened 2017-05-21
Make generated RPMs and version numbers monotonically increasing.

bbockelm:
Working with RPMs and pre-releases is currently a pain - the output
of `makesrpm.sh` is not generally usable.  This is because the RPM
has distinct versioning schemes (4.6.1234 vs 20161215) depending
whether you are on an exact tag or not.  This means that upgrades
from a pre-release version to an official release is generally
impossible (at least, one wouldn't give a tester a pre-release out of
fear of messing up their system).

This introduces a new scheme for pre-releases:
```
$MAJ.$MIN.$BUG+1-pre
```

where maj / min / bug versioning is based on the most recent tag.

For example, if the latest release is 4.6.1, then the pre-release
would be 4.6.2.pre.  If `makesrpm.sh` is run, then the RPM NVR would
be

```
4.6.2.pre-0.164.git.gbddc68e
```

where `164` is the number of commits since the 4.6.1 tag and `gbddc68e`
is the current git short hash.

NOTE: when master is pointing at a new release series, we should then
tag something along the lines of the next release.  For example, once
master is pointing at 4.7.x, we should make a tag of 4.6.99.

With this scheme, testers can cleanly upgrade from a pre-release to
a final release, reducing the likelihood they need to trash their sytem
setup in order to do testing.

### DIFF
--- a/genversion.sh
+++ b/genversion.sh
@@ -4,7 +4,7 @@
 # Process the git decoration expansion and try to derive version number
 #-------------------------------------------------------------------------------
 EXP1='^v[12][0-9][0-9][0-9][01][0-9][0-3][0-9]-[0-2][0-9][0-5][0-9]$'
-EXP2='^v[0-9]+\.[0-9]+\.[0-9]+$'
+EXP2='^v[0-9]+\.[0-9]+\.[0-9]+(-pre)?$'
 EXP3='^v[0-9]+\.[0-9]+\.[0-9]+\-rc.*$'
 
 #-------------------------------------------------------------------------------
@@ -19,8 +19,24 @@ function getNumericVersion()
   fi
   VERSION=${VERSION/v/}
   VERSION=${VERSION//./ }
+  VERSION=${VERSION//-/ }
   VERSION=($VERSION)
   printf "%d%02d%02d\n" ${VERSION[0]} ${VERSION[1]} ${VERSION[2]}
+}
+
+#-------------------------------------------------------------------------------
+# Extract pre-release version number from git references
+#-------------------------------------------------------------------------------
+function getPrereleaseVersion()
+{
+  VERSION=`git describe --abbrev=0`
+  VERSION=${VERSION/v/}
+  VERSION=${VERSION//./ }
+  VERSION=${VERSION//-/ }
+  VERSION=($VERSION)
+
+  echo "v${VERSION[0]}.${VERSION[1]}.$((${VERSION[2]}+1))-pre"
+  return 0
 }
 
 #-------------------------------------------------------------------------------
@@ -171,10 +187,7 @@ else
       if test ${?} -eq 0; then
         VERSION="`git describe --tags --abbrev=0 --exact-match`"
       else
-        LOGINFO="`git log -1 --format='%ai %h'`"
-	if test ${?} -eq 0; then
-          VERSION="`getVersionFromLog $LOGINFO`"
-        fi
+        VERSION="`getPrereleaseVersion`"
       fi
     fi
     cd $CURRENTDIR
--- a/packaging/makesrpm.sh
+++ b/packaging/makesrpm.sh
@@ -6,6 +6,7 @@
 
 RCEXP='^[0-9]+\.[0-9]+\.[0-9]+\-rc.*$'
 CERNEXP='^[0-9]+\.[0-9]+\.[0-9]+\-[0-9]+\.CERN.*$'
+PREEXP='^[0-9]+\.[0-9]+\.[0-9]+-pre.*$'
 
 #-------------------------------------------------------------------------------
 # Find a program
@@ -138,6 +139,17 @@ fi
 #-------------------------------------------------------------------------------
 if test x`echo $VERSION | egrep $CERNEXP` != x; then
   RELEASE=`echo $VERSION | sed 's/.*-//'` 
+  VERSION=`echo $VERSION | sed 's/-.*\.CERN//'`
+fi
+
+#-------------------------------------------------------------------------------
+# Deal with pre-releases
+#-------------------------------------------------------------------------------
+if test x`echo $VERSION | egrep $PREEXP` != x; then
+  DESCRIBE=`git describe --tags`
+  DESCRIBE=${DESCRIBE//-/ }
+  DESCRIBE=($DESCRIBE)
+  RELEASE="0.${DESCRIBE[-2]}.git.${DESCRIBE[-1]}"
   VERSION=`echo $VERSION | sed 's/-.*\.CERN//'`
 fi
 


ljanyst:
Just to provide a bit of historical perspective. It was our intention to make upgrades from "pre-releases" as painful as possible. The rationale was the huge support workload for these "illegitimate" releases.

bbockelm:
Yup, I know.  That's why I do far less testing than I used to.  It also highly discourages other sites - I wouldn't dare give them an Xrootd test RPM. I always hand-edit the version first.  To much energy is required.

Since no one uses the illegitimate releases - instead, renaming them so they look real - doesn't that increase the workload?  Isn't it better to make them easy to use but also easy to identify?

ljanyst:
Back then sites and users were using half-tested non-release code en masse. It's true that it was much harder to distinguish, because we were using timestamps. I could sometimes get 50 or more emails a day, so even saying "sorry, not supported" took time and was a pain. Also, the stability of the global ecosystem as a whole was far worse. Your perspective is perfectly valid, but it's a trade off.

osschar:
I usuallly cut-n-paste from this:
`
ver=4.6.0
rel=0.1.beta

ver_re="s/(Version:\\s*).*/\${1}$ver/o;"
rel_re="s/(Release:\\s*).*/\${1}$rel%{?dist}%{?_with_cpp11:.cpp11}%{?_with_clang:.clang}/o;"

cd ~/xrootd/

# checkout appropriate branch

cd packaging
rm -f xrootd-*.rpm
./makesrpm.sh
rm -rf ~/rpmbuild/
rpm -i xrootd-*.src.rpm

perl -pi -e $ver_re ~/rpmbuild/SPECS/xrootd.spec
perl -pi -e $rel_re ~/rpmbuild/SPECS/xrootd.spec

grep Version: ~/rpmbuild/SPECS/xrootd.spec
grep Release: ~/rpmbuild/SPECS/xrootd.spec

# Make sure the above are ok

rpmbuild -bb ~/rpmbuild/SPECS/xrootd.spec

ssh matevz@xrd-cache-1.t2.ucsd.edu rm -rf server/RPMS/rhel7/*
rsync -ae ssh ~/rpmbuild/RPMS/ matevz@xrd-cache-1.t2.ucsd.edu:server/RPMS/rhel7
`

bbockelm:
Indeed - the idea behind this branch is you could simplify your process to:

```
pushd packaging; ./makesrpm.sh; popd
rpm -Uhv packaging/xrootd-*.src.rpm
rpmbuild -ba ~/rpmbuild/SPECS/xrootd.spec
```

ljanyst:
I am not saying it's not useful. I am just saying that I trusted some people more than I trusted others and it seems to have been a good enough discriminant. Anyways, the distribution channels for XRootD changed, so perhaps it's worth rethinking. Not my call anymore, just trying to be helpful.

ljanyst:
Since there is no definite linear history in git, you'll run into problems once you start having local patches.

ljanyst:
Furthermore, how would you distinguish branches other than by hash, which is pretty much meaningless.

bbockelm:
Indeed - since this linearizes things, it really only works for test builds from `master`.  It doesn't help much building off a private branch (but RPM doesn't really allow for this?).

ljanyst:
If you have your version of the script in the branch, does it matter which branch you're on? You could, in principle, have: `4.6.2.pre-0.164.git.abcdefgh` and `4.6.2.pre-0.164.git.ijklmnop` on two different branches. It's quite confusing.

simonmichal:
Hi Brian,

If I run makerpm.sh with your PR I'm getting the following version:

> 4.6.1.pre-0.165.git.g7184a9f

which is not in line with what you described, right? I would expect to have:

> 4.6.2.pre-0.165.git.7184a9fc

The $BUG should have been bumped, and short hash should point to the last commit, right?


Anyway, let's assume it works correctly for now, so if I create the pre-release version I'll get correctly:
> 4.6.2.pre-0.165.git.7184a9fc
And let's assume the next release will be 4.6.2-1.

Lets do:
> rpmdev-vercmp 4.6.2-1 4.6.2.pre-0.165.git.7184a9fc
> 4.6.2-1 < 4.6.2.pre-0.165.git.7184a9fc
This means yum will think the pre-release is a higher version than the actual release.

Similarly if we would have a release candidate: 4.6.2-0.rc1:
> rpmdev-vercmp 4.6.2-0.rc1 4.6.2.pre-0.165.git.7184a9fc
> 4.6.2-0.rc1 < 4.6.2.pre-0.165.git.7184a9fc
Again the same problem.

So it needs some rethinking ;-)

We could try something like:
4.6.1-1.pr521.165.git.7184a9fc
Where '4.6.1-1' is the latest release, 'pr521' is the branch name, 165 is the number of commits since last release, and '7184a9fc' is the short hash of the last commit.
Just an idea.

Cheers,
Michal

xrootd-dev:
Sent from my iPhone

> On May 26, 2017, at 3:02 PM, simonmichal <notifications@github.com> wrote:
> 
> Hi Brian,
> 
> If I run makerpm.sh with your PR I'm getting the following version:
> 
> 4.6.1.pre-0.165.git.g7184a9f
> 
> which is not in line with what you described, right? I would expect to have:
> 
> 4.6.2.pre-0.165.git.7184a9fc
> 

Yeah - that's because 4.6.x isn't being merged into master.

So, the nearest tag to the master branch is the 4.6.0 pre-release.

We really ought to be merging through to master.

> The $BUG should have been bumped, and short hash should point to the last commit, right?
> 
> Anyway, let's assume it works correctly for now, so if I create the pre-release version I'll get correctly:
> 
> 4.6.2.pre-0.165.git.7184a9fc
> And let's assume the next release will be 4.6.2-1.
> 
> Lets do:
> 
> rpmdev-vercmp 4.6.2-1 4.6.2.pre-0.165.git.7184a9fc
> 4.6.2-1 < 4.6.2.pre-0.165.git.7184a9fc
> This means yum will think the pre-release is a higher version than the actual release.
> 
> Similarly if we would have a release candidate: 4.6.2-0.rc1:
> 
> rpmdev-vercmp 4.6.2-0.rc1 4.6.2.pre-0.165.git.7184a9fc
> 4.6.2-0.rc1 < 4.6.2.pre-0.165.git.7184a9fc
> Again the same problem.
> 
> So it needs some rethinking ;-)
> 

Damn.  I thought RPM understood that.

In that case, yes, we need to move the pre-release tag to the release string.

> We could try something like:
> 4.6.1-1.pr521.165.git.7184a9fc
> Where '4.6.1-1' is the latest release, 'pr521' is the branch name, 165 is the number of commits since last release, and '7184a9fc' is the short hash of the last commit.
> Just an idea.
> 

As long as the string starts with 0 (indicating a pre-release), OSG would be happy.

simonmichal:
Hi Brian,

It would be nice if the RPM naming convention would make sense for all brunches ;-)

We could also have an agreement that every now and then we tag a pre-release
(or have the RCs much earlier), this depends what would you like to test and where 
to install the pre-releases?

On the other hand, if you have a test setup(s), which you would like to update to latest
master on weekly basis, maybe a new naming convention would make more sense.

Let me know your thoughts!

Cheers,
Michal

bbockelm:
Unfortunately, I don't think there's any generic way to have `rpm` and `git` both be happy: you won't be able to create a linear sequence of numbers and handle arbitrary branches!

My goal here is to be able to always produce a pre-release RPM for testers from master and the latest stable series.  I would also like that RPM to have meaning to upstream - i.e., it should be easily traceable back to a particular commit.

Is there a document describing the branching and tagging strategy for Xrootd?  (Particularly, it seems quite strange to have the stable branches not merged back into master.)  Given a document like that, I can probably come up with a script that meets my above requirements.

ljanyst:
Why is that strange? You have main developments in master and small bugfix releases with cherry-picked patches in branches. That's the strategy used by most of the projects I have seen.

bbockelm:
:) Maybe we work on a mutually-exclusive set of projects?  Most of the ones I work on merge release branches through master as a mechanism to make sure patches in the stable series make it into the master branch.

Regardless - I don't care too much provided we can have a written policy for branching and tagging that I can code against to provide the two items above.

ljanyst:
Hah! Maybe :)

simonmichal:
@bbockelm I have to say that I don't like the idea that the rpm naming convention will work reliably only for master and stable branches, and will give random results for devel branches. If we cannot do something that works reliably for every one better not do it.

I suppose we could simply add a parameter to the makesrpm.sh script, that when present is used as the the Version in xrootd.spec. This way the dirty scripting would be moved to the build system that chooses to use this feature. Also, this would allow for including the build number in the version, which might be handy sometimes.

I do agree that a policy that every release tagged in a stable brunch gets merged back to master, would be useful.

abh3:
The reason it works is that we never apply patches to a stable branch. All 
patches must come from maste; hence by implication they will always be 
carried forward. We've never had a problem with this approach.

bbockelm:
@simonmichal - we have something that works poorly for master, stable, and devel branches.  What's the loss if we can "only" fix two of the three?

I'm OK if we move the logic to the build system, provided that we also include the SHA1 hash in the RPM release.

simonmichal:
@bbockelm : My bad, I thought you are using some random git hash for the devel branches, but you are actually using the right one prefixed with a 'g' as in git describe. 

Two questions:

1. Why do you need the SHA1 hash in the RPM release?

2. Why would we need the 4.6.99 ?
> NOTE: when master is pointing at a new release series, we should then
> tag something along the lines of the next release. For example, once
> master is pointing at 4.7.x, we should make a tag of 4.6.99.


Regarding the branching and tagging, we have a stable branch per 'minor' release (e.g. stable-4.6.x), both RC and release tags go to the stable branches. And I think that's it. 
(I will be merging the stable branches back to master starting with the next release, it seems to me like a reasonable thing to do).

In general I'm not against your idea, I think we just need to iron out the details ;-)

Would the following schema work for you:
xrootd-4.6.1-1.master.91.c41d4c13

This is very intuitive: it's 4.6.1-1 + 91 commits from branch master, the last one of them being c41d4c13.

I would skip both the 'g' and 'git' as we are using only git anyway. I think we can leave out the 'pre' interfix, as it carries no additional info.
(Also, it would be awesome,  if we could have a switch that allows to replace the 91 with a value chosen by the user.)

4.6.1-1 < 4.6.1-1.master.91.gc41d4c1 && 4.6.1-2 > 4.6.1-1.master.91.c41d4c13
which is what we want, right?

Let me know your thoughts!

bbockelm:
Toward your first question - I want the SHA1 available so I can easily go from output of `rpm -q` from a tester who is reporting an issue to a specific version of the code.

Something like `4.6.1-1.master.91.gc41d4c1` gets closer; it works at a technical level, but runs afoul of RPM packaging standards - basically, once an official release is done, the release number (`1.master.91.gc41d4c1`) is meant to be utilized by the distributions.  For example, `xrootd` in Fedora (https://koji.fedoraproject.org/koji/buildinfo?buildID=874518) had 8 rebuilds; this got the release number up to 9.  So, since distros start the release number at `1` for official releases, it's become common to see pre-releases start with `0`.

What about `4.6.1.1-0.master.91.gc41d4c1`?  That meets the criteria of RPM versioning and keeps the release number prefixed with 0.

We still have the problem of determining the "last release".  `git describe` gets confused because we don't merge stable branches through master.

Perhaps we really want "next release", not "last release"?  That is, if I build from `stable-4.6.x`, I want a snapshot build to be prefixed `4.6.1.1-0` or `4.6.2-0` while if I build from `master`, I'd want `4.6.99` (`.99` is used as it assumes it is before `4.7.0` but reasonably after any bugfix release of `4.6.x` series) or `4.7.0-0`.

I see three ways of doing this:
1.  Have Jenkins produce RPMs as build artifacts.  Keep the "next release version" information for each branch in the Jenkins build.  This builds on your idea of allowing `./makesrpm.sh` take the base version as a CLI argument.  So, the knowledge is stored outside the repository: people would have to go to Jenkins to get the RPMs or make reasonable guesses when they are building.
2.  Tag `master` immediately after forking a stable branch (such as `4.7.x.branch`?  Something...). So, the release knowledge would be stored inside git -- but it's a bit opaque.  This is the _closest_ option to what is done today.
3.  Actually store a "version number" somewhere in the repository.  This makes the next target very visible - but runs the chance of the tag and contents of the repository running out of sync.  *However*, this is something Travis-CI can catch!

There's no clear "winner" of the three to me.  Do you see an obvious route?  Maybe a fourth option?

xrootd-dev:
If all you want it to be able to get to the exact version of a release and still want to keep RPM standards then may I suggest that the hash be included in the “xrootd version string” but omitted from the the RPM version string. All other sequence numbers would be the same. The xrootd string is always displayed in the log file and you can query it at any time.

simonmichal:
@bbockelm : in general I would be happy with 4.6.1.1-0.master.91.gc41d4c1, I suppose we could even use something like 4.6.1.post-0.master.91.gc41d4c1, right?

Let me experiment a bit with margin stable branches back to master, so I'm sure that there are no side effects ;-) I think it is a policy we could adopt.

ljanyst:
IMHO, the clear winner is keeping things as they are. :)

It seems to me that you'll only be complicating everybody's life unnecessarily just so that some people could use one rpm command instead of two in a testing environment. Unless, of course, it is not the end goal of all this.

Merging things back to master only makes sense if you do significant developments in stable branches, which is not the case for XRootD. Besides, your linear numbering will be even more confusing if you do that. Go check that out for a couple of stable releases. :)

ljanyst:
You're trying to impose a linear ordering in a system that is based on a topological ordering. I think, you'll suffer more than you'll gain.

ljanyst:
How about a solution where you just manually tag pre-releases more often?

abh3:
Thank you Lukasz. I agree with you completely.

bbockelm:
Well "tagging pre-releases more often" sounds a lot like "having the build system produce usable RPMs".  It is one way forward I suggest and I think can be made workable.

It may be the simplest route *for development* is to keep things the way they are.  However, that's the worst route *for testing*.  I believe the lack of usable, accessible pre-release packages for testers has significantly decreased the participation in testing, which in turn has been a contributor to issues like we saw with 4.6.0.

ljanyst:
That's a fair point, but I think that having, say, bi-weekly upgradable testing snapshots is a better solution than having a naming scheme that kind of works, but not quite, even in the inteded case and goes against the basic paradigms of the underlying system (git).

bbockelm:
Indeed - that would be a welcome solution (especially if the source RPMs were one of the Jenkins build artifacts).

@simonmichal - would you be able to do something along these lines in Jenkins?

ljanyst:
Michał is on vacation for the next two weeks.

BTW. It would be interesting to see how many people install release candidates.

xrootd-dev:
Hi,

 I do, and also make my own builds.

 And I agree that the current naming scheme is a major annoyance, for many reasons that you
already mentioned.

 To keep it simple I would propose to add the HHMM of the build invokation
to the date in the rpm filenames. We do this for
all our projects since 6 years and it works fine without needing scripts or other hacks.

 Keeping also the version number in front of the datetime
would be a major plus, as right now it is stripped (why !??!)

Cheers
Fabrizio

simonmichal:
@ljanyst : I finally had a moment to check and you're right, merging the stable back to master messes up the history

So I would say we add a switch to the ./makesrpm.sh script so the default versioning schema can be replaced by whatever the user provides + I will add a bi-weekly build to jenkins that generates RPMs for testers, with a versioning schema like: xrootd-4.6.2-0.experimental.91.c41d4c13 .

If there are no major objections I will close this one when I'm back and go ahead with this simple plan ;-)

ljanyst:
Apologies for bringing this up again, but I just read a discussion on how glibc handles similar problems and this issue popped back into my mind. It's just for the reference.

In particular, Brian has said:

> :) Maybe we work on a mutually-exclusive set of projects? Most of the ones I work on merge release branches through master as a mechanism to make sure patches in the stable series make it into the master branch.

From https://lwn.net/Articles/736429/

> Like many projects, glibc fixes bugs first in its repository trunk — the branch that will eventually become 2.27 at the moment. The project does not make point releases, but that does not mean it leaves users of its previous releases entirely out in the cold. A branch in the repository is created for each release and, at the discretion of the manager for that particular release, important fixes are backported and added to that "maintenance branch" post-release. So distributors have a stream of nicely backported fixes for each release; what they don't have are point releases that contain a well-defined set of those fixes.

Coincidently, XRootD has a very similar approach ;)

PR #523, opened 2017-05-30
Fixes for XrdXrootdJob and XrdCnsSsi

vp1981:
* src/XrdXrootd/XrdXrootdJob.cc, src/XrdCns/XrdCnsSsi.cc: gcc 7.1+
    complains about verClient() [XrdXrootdJob] (-Werror=array-bounds) and
    Updt() [XrdCnsSsi] (-Werror=nonnull), make it "happy".

Signed-off-by: Vladimir Lomov <lomov.vl@gmail.com>

### DIFF
--- a/src/XrdCns/XrdCnsSsi.cc
+++ b/src/XrdCns/XrdCnsSsi.cc
@@ -60,7 +60,8 @@ struct XrdCnsSsiFRec
 {
 char Info[XrdCnsLogRec::FixDLen];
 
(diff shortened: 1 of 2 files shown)


vp1981:
This is my first pull request, so there is possibility that I did something wrong.

vp1981:
As for the "fix": I use 'nullptr' but seems that it appears in C++11. I compile xrootd only for ROOT and it requires c++11 (and even c++14) so I think it is worth to begin to use C++11.

PR #544, opened 2017-07-19
Make XrdHTTP able to forward HTTP requests to an external, optional plugin (conceptually similar to CGI). Various stability fixes and improvements.

ffurano:
Make XrdHTTP able to forward HTTP requests to an external, optional plugin (conceptually similar to CGI). 

Various stability fixes and improvements.


simonmichal:
@ffurano : Could you please remove the changes to the '.gitignore' file from the commit, I suppose they were not intentional.

Thanks,
Michal

ffurano:
Should be done

PR #555, opened 2017-08-09
Use correct byte count when increasing byte count on a link.

osschar:


abh3:
Of course, myBytes is now useless. I wll remove it. Anyway, what was wrong with adding zero to bytes sent :-)

PR #571, opened 2017-08-22
XrdBuffer: Decrement total buffer count when freeing buffers

jthiltges:
If the total buffer count (totbuf) is incorrectly high, the request profile (bufprof) will also be disproportionately high: it's scaled by the buffer count. The buffer pool is not reshaped, with too many buffers being held.

This patch decrements the total buffer count as buffers are deleted, matching the behavior of XrdBuffXL.

### DIFF
--- a/src/Xrd/XrdBuffer.cc
+++ b/src/Xrd/XrdBuffer.cc
@@ -290,6 +290,7 @@ while(1)
                     delete bp;
                     bucket[i].numbuf--; numfreed++;
                     memhave -= memslot; totalo  -= memslot;
+                    totbuf--;
                    } else {bucket[i].numbuf = 0; break;}
            Reshaper.UnLock();
            memslot = memslot>>1;


bbockelm:
To be clear, we believe this patch fixes memory hoarding observed when `XrdHttp` is used.  Over time, we've seen the xrootd process hoard 40-60GB memory (linearly increases as a function of load) when HTTP is used heavily until systemd kills it.

jthiltges:
Related to this PR, I'm puzzled by the logic to reshape the XrdBuffer pool.

It scales the buckets of memory buffers proportional to the number of requests: If 50% of the requests are for buffers of size X, it adjusts the buffer count so that 50% of the buffers are size X.

However, the adjustment process doesn't consider the amount of memory being used. Assume the application is only using 1MB buffers: 100% of requests are for 1MB buffers. The target buffer count is 100% of size 1MB, which matches the current state. Nothing is freed, regardless of how much memory is allocated. Is my reading of XrdBuffManager::Reshape() correct?

PR #586, opened 2017-09-25
Make HTTP external handlers more flexible

### LINKED ISSUE #585: Improve XrdHttpExtHandler flexibility
I started looking into using the `XrdHttpExtHandler`, but immediately started to hit some problems.

1.  The only mechanism for handling a request is from path-based matching.  It's not possible to filter based on the verb (GET vs POST) or one of the headers.
    - Proposal: Either change the `MatchesPath` function to include the HTTP verb, or allow the `ProcessReq` to return a special value indicating the handler did not actually want to handle the request in the end.  The latter, of course, has a higher CPU overhead.
2.  I need more information from the authentication session than what is in the `XrdHttpExtReq` object.  For example - there's no VOMS information here (i.e., the contents of `vorg`).  As I started to add the information I need for making authorization decisions, I find myself quickly replicating the `XrdSecEntity` structure.
    - Proposal: Add a `const XrdSecEntity *` to the `XrdHttpExtReq`
3.  The `XrdHttpExtHandler` headers are not public, meaning I can't actually link against them yet.
    - Proposal: Once the above binary-compat-breaking changes are made, make the header public and require the standard version-checking scheme.

Thoughts?

bbockelm:
This improves the HTTP external handler environment and interface to have richer semantics.

- Pass down the `XrdOucEnv` object when the object is initialized; necessary to initialize other parts of Xrootd.
- Allow one to access the `XrdSecEntity` object associated with a request.
- Allow filtering based on HTTP verb in `MatchesPath`.
- Make this a public plug-in interface and declare ABI compatibility.

@ffurano - this is meant to fix #585

### DIFF
--- a/src/XrdHeaders.cmake
+++ b/src/XrdHeaders.cmake
@@ -104,7 +104,7 @@ set( XROOTD_PUBLIC_HEADERS
   XrdXrootd/XrdXrootdMonData.hh
   XrdXrootd/XrdXrootdBridge.hh
   XrdHttp/XrdHttpSecXtractor.hh
-
+  XrdHttp/XrdHttpExtHandler.hh
 )
 
 set( XROOTD_PRIVATE_HEADERS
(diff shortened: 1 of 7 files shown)


bbockelm:
@ffurano - would you have time to review this one?

PR #588, opened 2017-09-26
Allow XrdSfsGetDefaultFileSystem to be called multiple times

bbockelm:
When requesting the default filesystem, a statically-initialized singleton object is returned - but the `Configure` method of that singleton is invoked for each call to `XrdSfsGetDefaultFileSystem`.

Unfortunately, there are several pieces of the object's state that leak through after each `Configure` call (in the environment or in the `XrdOfs` object member variables).  This effectively limits us to a single invocation of `XrdSfsGetDefaultFileSystem`.

This PR resets the leaky pieces of state, allowing a valid OFS object to be returned each time.

### DIFF
--- a/src/XrdOfs/XrdOfsConfig.cc
+++ b/src/XrdOfs/XrdOfsConfig.cc
@@ -134,6 +134,11 @@ int XrdOfs::Configure(XrdSysError &Eroute, XrdOucEnv *EnvInfo) {
 //
    Eroute.Say("++++++ File system initialization started.");
 
+// Each time XrdSfsGetDefaultFileSystem is invoked, it'll reuse the same XrdOfs
+// object but call the Configure method again.  If poscLog is not reset, then
+// a stale value may be reused.
+   poscLog = NULL;
+
 // Establish the network interface that the caller must provide
 //
    if (!EnvInfo || !(myIF = (XrdNetIF *)EnvInfo->GetPtr("XrdNetIF*")))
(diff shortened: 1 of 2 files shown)


abh3:
I guess the real question is whether you want reinitialization every time XrdSfsGetDefaultFileSystem() is called. I would think the better solution is to make sure it gets initialized exactly once. That way we won't need to worry about future additions that might leak values. Plus changing the configuration mid-stream may cause other unintended results, So, I'd rather go with the single instance solution.

bbockelm:
I don't have any use cases that would necessitate re-`Configure`ing the OFS object.  Returning the same one each time, configured once, is fine by me.

If that's your preference, I can do a follow-up PR.

xrootd-dev:
Hi Brian,

Yes, please do. That should have been done in the first place. Who knew 
someone would call it nultiple times. Please make it thread-safe. Thanks!

Andy

abh3:
That would be great but the changes to XrdOfsConfig.cc are still in this merge. Could you remove those before I commit this?

bbockelm:
Actually, I'd like to keep the changes to `XrdOfsConfig`.  It's an improvement to the object's behavior, even if we don't strictly need it.  Any reason why we shouldn't?

abh3:
OK, but we have some backward compatibility issues.
if (redirect_val && strcmp(redirect_val, "0")) i  = isManager;
which replaced
if (getenv("XRDREDIRECT")) i  = isManager;
is not backward compatible and doesn't even make sense. Originally, we only cared that the envar was merely set and now you require it to be non-intuitively set to 0 (eh? not one?).

It's even worse because the same rule isn't applied to other getenv() statements that do the same. But, let's not do that as well, ok?

I fully accept that poscLog should be set to NULL, that's actually a bug.

So, can we edit this back to just set poscLog to null?

bbockelm:
@abh3 - I think you read the `strcmp` backward?  It is treating `0` as equivalent to `XRDREDIRECT` being unset.

Later on in the configuration code, for server roles, the existing code sets `XRDREDIRECT=0`.  Hence, on any follow-up configuration, the OFS tries to become a redirector.  This prevents multiple OFS objects from existing (or the same OFS object being configured twice).

xrootd-dev:
Hi Brian,

I wish you had explained hat earlier. I will look at this again. Now, we 
only configure once. Is there a reason to have more than one instance of 
the default OFS? Even if we wanted that  it would require a lot of work in 
getting rid of all the global statics. So, practically, you can't have 
more than one instance. Does that change anything?

Andy

PR #589, opened 2017-09-26
Allow multiple external handlers to be loaded by XrdHttp.

bbockelm:
This allows multiple external handlers to be loaded by `XrdHttp`.  The first loaded handler that matches a given request will handle it.

By allowing multiple plugins, we can have one plugin handle authorization requests that come in via HTTP and a second implement alternate HTTP verbs (see #586, for example).

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -93,7 +93,8 @@ SSL_CTX *XrdHttpProtocol::sslctx = 0;
 BIO *XrdHttpProtocol::sslbio_err = 0;
 XrdCryptoFactory *XrdHttpProtocol::myCryptoFactory = 0;
 XrdHttpSecXtractor *XrdHttpProtocol::secxtractor = 0;
-std::vector<XrdHttpExtHandler *> XrdHttpProtocol::exthandler;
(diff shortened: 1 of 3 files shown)


ffurano:
Hey Brian, I have pushed stuff into your branch, I see it in your branch but I can't see it in this pull request pointing to your branch. Maybe github needs time to sync these exotic practices ?
Could you please tell me what do you see on your side?

ffurano:
Ah, it appeared now, apparently github refreshes every 30 mins?

bbockelm:
@ffurano - I find that you may have to hit refresh on the PR issue to get it to show up in the webpage.  I've never seen a case where new commits didn't show up after a refresh -- there's always the possibility of backend issues.

Anyhow, glad all is solved.

ffurano:
OK, I have merged it and I'll run my stress tests again using the rpms. Since I have no use case for multiple handlers, please let me know if it works for your use cases.
Now the syntax for the directive is (I like giving names to plugins)

 *   Purpose:  To parse the directive: exthandler <name> <path> <initparm>
 * 
 *             <name>      a unique name (max 16chars) to be given to this
 *                         instance, e.g 'myhandler1'
 *             <path>      the path of the plugin to be loaded
 *             <initparm>  a string parameter (e.g. a config file) that is
 *                         passed to the initialization of the plugin

Thank you very much!

PR #597, opened 2017-10-04
Implementation of the option http.header2cgi

ffurano:


ffurano:
After merging the pull request, the docs should be added an additional paragraph containing something like the following:

-------------
6.3.7 header2cgi
  http.header2cgi <headerkey> <CGIkey>

Function:
When a request comes, pick the value associated to the HTTP header <headerkey> (if present) and append it to the CGI information of the path that xrootd will process internally, in the form "CGIkey=<value>"

Parameters:
headerkey: the header key that is looked for
CGIkey: the CGI key that will be appended to the internal CGI information

Defaults:
none

Notes:
1) <headerkey> and <CGIkey> can be equal

Example:
http.header2cgi Authorization authz

PR #611, opened 2017-11-01
Correct external handling logic.

bbockelm:
The test to see if there is an external plugin was inverted, meaning the wrong case in the conditional was always called.  This fixes the logic, avoiding a `SIGSEGV` I encountered while testing the improved `XrdHttp` code.

Additionally, this reverses the order of evaluation -- checking the (cheap, common) pointer value first, then invoking a function.

@ffurano


ffurano:
Hi Brian, I just merged, as it makes sense to me, even if I was not experiencing this. Do you regularly use multiple plugins ? Thank you!

bbockelm:
@ffurano - I was hitting a `SIGSEGV` with no plugins present.

PR #612, opened 2017-11-01
[pfc] Bugfix in reading of last access time for file purging.

osschar:
Access history in cache meta-data is fixed size now, make sure we do not read beyond the end.

The file purge checking code was reading beyond the end and could lead to random behavior or crashes once the number of times a file was accessed grew beyond the internal size limit (20).

Please merge this also into stable-4.7.x.


abh3:
I hope you understand that 4.7.1 has already been released. So, this will have to wait until 4.7.2 or 4.8.0, whichever comes first.

osschar:
Unfortunately, I know.

PR #629, opened 2017-11-22
[XrdCl] Make xrd.secuid take effect only if euid == 0

gbitzes:


bbockelm:
Is that the right approach?  Shouldn't you instead check for the correct capability?

After all, I can construct situations where user `xrootd` is allowed to do the UID-switch and situations where `root` is not.

abh3:
I think that Brian is correct here. Having euid 0 doesn't necessarily mean much in certain OS's. For instance, in BSD I can deny certain capabilities even for root.

gbitzes:
I agree this is not the best approach: It's a workaround for the fact "xrd.secuid" is passed on to the server, even though it's only needed locally on the client. This is a workaround to stop disk servers from getting confused.

I had a chat with Michal, he thinks it should be easy to add internal CGI parameters not passed on to the server, acting strictly as directives for the client.. This would solve the issue we're seeing on EOS.

Let's postpone this pull request for now, as the above would be a much cleaner solution.

simonmichal:
@gbitzes : could you please test 2be1931d8a058391be411859350ced50db583506 and let me know if it solves your problem?

gbitzes:
@simonmichal works great, thanks a lot. I also verified xrdcl.secuid doesn't end up on the server side by checking the logs, everything seems OK.

PR #636, opened 2017-12-20
Support HTTP chunked transfer encoding

bbockelm:
The chunked transfer encoding allows the response to be broken up into multiple send calls (potentially spread over a large period of time).

This is useful whenever one does not know the complete size of the response at the outset -- or if the response is generated over a period of time.

An example use case of this is an external handler that implements the WebDAV COPY verb: one might want to provide continuous feedback (say, every 30 seconds) as the third-party copy progress so the client doesn't timeout and give up.


bbockelm:
This PR fixes #636.

bbockelm:
@ffurano - any thoughts on this one?

ffurano:
I like the fact that this is demanded to an external plugin, if present, avoiding touching the internal logic
which is pretty touchy. Thank you very much Brian, I think we are going to use this.

PR #638, opened 2017-12-21
Allow parsing of unknown HTTP verbs.

bbockelm:
If a HTTP verb (such as COPY) is encountered in an request, the
current code will try to parse subsequent headers as a HTTP status
line (mostly encountering garbage, but sometimes actually succeeding!).

We should instead note that a valid -- but unknown -- verb was encountered
and parse the remaining headers as headers.  This is useful because
external handlers might actually understand verbs (again, COPY) that the
built-in handlers do not.


bbockelm:
@ffurano - any update here?

PR #639, opened 2017-12-21
Another attempt at getting HTTP PUT flags right.  Fix #637

### LINKED ISSUE #637: Correct combination of flags for "create with overwrite" for HTTP PUT?
Is it possible to "create with overwrite"?  This is needed for the HTTP PUT implementation, but doesn't seem possible.

We started with this:
```
xrdreq.open.options = htons(kXR_mkpath | kXR_open_updt | kXR_new );
```

this successfully creates the file, but doesn't allow it to be overwritten.  So, it was then changed to:

```
xrdreq.open.options = htons(kXR_mkpath | kXR_open_wrto );
```

This overwrites the file, but doesn't allow one to be created.

A bit of `grep`ing through the code, it seems like this does work

```
xrdreq.open.options = htons(kXR_mkpath | kXR_open_wrto | kXR_delete);
```

Any thoughts or opinions on this?  It would imply that HTTP PUT requires delete permission (which I recall is distinct from being able to create files?)

bbockelm:
Originally, HTTP PUT could create but not overwrite.  While fixing
this, the flags were changed -- but it could overwrite and not create!

With this change, I have tested that both overwriting and creating
work.

We likely need additional follow-up because, with this change, a
user cannot perform HTTP PUT with the "create" authorization alone.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -1090,7 +1090,7 @@ int XrdHttpReq::ProcessHTTPReq() {
         l = resourceplusopaque.length() + 1;
         xrdreq.open.dlen = htonl(l);
         xrdreq.open.mode = htons(kXR_ur | kXR_uw | kXR_gw | kXR_gr | kXR_or);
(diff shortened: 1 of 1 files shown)


bbockelm:
Fixes #637

PR #654, opened 2018-02-05
Include crypt.h on Linux

ellert:
With the update of glibc to version 2.27 crypt.h is not included recursively by some other include statement any longer but must be included explicitly.


gganis:
Thanks!
I have just merged it.

PR #657, opened 2018-02-09
Fix fedora rawhide build, part2.

simonmichal:
A couple of fixes for Fedora rawhide, GCC 8

### DIFF
--- a/src/XrdCms/XrdCmsRRQ.hh
+++ b/src/XrdCms/XrdCmsRRQ.hh
@@ -117,8 +117,20 @@ void *Respond();
 
 struct Info
       {
-        Info(): Add2Q(0), PBack(0), Resp(0), Multi(0), luFast(0), luSlow(0),
-                rdFast(0), rdSlow(0) {}
+        Info()
+	{Reset();}
+	
+	void Reset()
+	{Add2Q  = 0;
+	 PBack  = 0;
+	 Resp   = 0;
+	 Multi  = 0;
+	 luFast = 0;
+	 luSlow = 0;
+	 rdFast = 0;
+	 rdSlow = 0;
+	}
+
        long long Add2Q;    // Number added to queue
        long long PBack;    // Number that we could piggy-back
        long long Resp;     // Number of reponses for a waiting request
--- a/src/XrdDig/XrdDigAuth.cc
+++ b/src/XrdDig/XrdDigAuth.cc
@@ -233,10 +233,6 @@ bool XrdDigAuth::Parse(XrdOucStream &aFile, int lNum)
    int  k, n, bLeft = sizeof(buff);
    bool aOK = false, tfVal;
 
-// Clear the arec the fastest way possible
-//
-   memset(aEnt.eP, 0, sizeof(XrdDigAuthEnt));
-
 // Get the record type tokens first
 //
    while((var = aFile.GetToken()) && *var)
--- a/src/XrdSec/XrdSecEntity.hh
+++ b/src/XrdSec/XrdSecEntity.hh
@@ -69,13 +69,29 @@ const    char   *tident;                  // Trace identifier always preset
                                           // it in your XrdSfsFileSystem::Disc()
                                           // implementation, as needed.
          XrdSecEntity(const char *pName = "")
-                     : name(0), host(0), vorg(0), role(0), grps(0),
-                       endorsements(0), moninfo(0), creds(0), credslen(0),
-                       rsvd(0), addrInfo(0), tident(""), sessvar(0)
-                     {strncpy(prot, pName, XrdSecPROTOIDSIZE-1);
+                     {Reset();
+	              strncpy(prot, pName, XrdSecPROTOIDSIZE-1);
                       prot[XrdSecPROTOIDSIZE-1] = '\0';
                      }
         ~XrdSecEntity() {}
+
+	void Reset()
+	{
+          memset( prot, 0, XrdSecPROTOIDSIZE );
+	  name = 0;
+	  host = 0;
+	  vorg = 0;
+	  role = 0;
+	  grps = 0;
+	  endorsements = 0;
+	  moninfo = 0;
+	  creds = 0;
+	  credslen = 0;
+	  rsvd = 0;
+	  addrInfo = 0;
+	  tident = 0;
+	  sessvar = 0;
+	}
 };
 
 #define XrdSecClientName XrdSecEntity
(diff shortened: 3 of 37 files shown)


src/XrdCms/XrdCmsRRQ.hh
```diff
@@ -117,8 +117,20 @@ void *Respond();
...
+	{Add2Q  = 0;
+	 PBack  = 0;
+	 Resp   = 0;
+	 Multi  = 0;
+	 luFast = 0;
+	 luSlow = 0;
+	 rdFast = 0;
+	 rdSlow = 0;
+	}
+
```
abh3: What is the compiler complaining about here that makes you change all inline initialization to a function call? I am concerned that this structure may defeat compiler optimization.

src/XrdDig/XrdDigAuth.cc
```diff
@@ -233,10 +233,6 @@ bool XrdDigAuth::Parse(XrdOucStream &aFile, int lNum)
    int  k, n, bLeft = sizeof(buff);
    bool aOK = false, tfVal;
 
-// Clear the arec the fastest way possible
-//
-   memset(aEnt.eP, 0, sizeof(XrdDigAuthEnt));
-
```
abh3: Uh, this doesn't look kosher. You just eliminated initializing aEnt so I doubt his will work.

src/XrdSec/XrdSecEntity.hh
```diff
@@ -69,13 +69,29 @@ const    char   *tident;                  // Trace identifier always preset
                                           // it in your XrdSfsFileSystem::Disc()
                                           // implementation, as needed.
          XrdSecEntity(const char *pName = "")
-                     : name(0), host(0), vorg(0), role(0), grps(0),
-                       endorsements(0), moninfo(0), creds(0), credslen(0),
-                       rsvd(0), addrInfo(0), tident(""), sessvar(0)
-                     {strncpy(prot, pName, XrdSecPROTOIDSIZE-1);
+                     {Reset();
+	              strncpy(prot, pName, XrdSecPROTOIDSIZE-1);
                       prot[XrdSecPROTOIDSIZE-1] = '\0';
```
abh3: How does this square with ABI compatability? The in-lining differs.

src/XrdSsi/XrdSsiResponder.cc
```diff
@@ -207,7 +207,7 @@ void XrdSsiResponder::ReleaseRequestBuffer()
   
 XrdSsiResponder::Status XrdSsiResponder::SetMetadata(const char *buff, int blen)
 {
-   XrdSsiMutexMon(spMutex);
+   XrdSsiMutexMon lck(spMutex);
 
 // If we don't have a request or the args are invalid, return an error.
 //
```
abh3: You may get merge conflicts here if you didn't use the latest githead.

src/XrdCms/XrdCmsRRQ.hh
simonmichal: The reason I add the 'Reset' method is here:

https://github.com/xrootd/xrootd/commit/60da79db805d122a7190535b8dead0f306e5a2f9#diff-bffded318632dc8e33cbdb6c4dcc1d34L134

In XrdCmsRRQ.cc the compiler was complaining that an object with a non-trivial constructor is being initialized with memset. As the constructor does basically the same as 'Reset' I though that implementing the constructor in terms of 'Reset' method would simplify the code. IMHO the compiler will be smart enough to first inline 'Reset' and then the constructor. That said, I don't have any strong feelings on this one, I can revert to the old constructor implementation if you wish (as long as the 'Reset' stays).

src/XrdDig/XrdDigAuth.cc
simonmichal: Again, the compiler was complaining that an object with a non-trivial constructor is being initialized with memset. 

Now the fix rational:

aEnt.eP is a newly created instance of 'XrdDigAuthEnt':
https://github.com/xrootd/xrootd/blob/fedora-rawhide-fix/src/XrdDig/XrdDigAuth.cc#L227

note that I modified the default constructor of 'XrdDigAuthEnt' :
https://github.com/xrootd/xrootd/commit/60da79db805d122a7190535b8dead0f306e5a2f9#diff-a726e008bf733c6a23ce19da88801122R56

now, it initializes all the members to 0, and clears member arrays, so it is equivalent to the memset I removed.

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -541,7 +541,7 @@ int XrdSecProtocolkrb5::Authenticate(XrdSecCredentials *cred,
 int XrdSecProtocolkrb5::Init(XrdOucErrInfo *erp, char *KP, char *kfn)
 {
    krb_rc rc;
-   char buff[1024];
+   char buff[2048];
```
gganis: Out of curiosity, why 2048 ?

src/XrdSec/XrdSecEntity.hh
simonmichal: The code compiled against the old version of constructor will have the old version of the constructor inlined, there's no ABI compatibility problem. The new and the old version of the constructor are functionally equivalent so it is OK.

src/XrdSsi/XrdSsiResponder.cc
simonmichal: I did rebase against the latest git HEAD :-) so no conflicts :-)

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
simonmichal: The problem was that in here:

https://github.com/xrootd/xrootd/blob/fedora-rawhide-fix/src/XrdSeckrb5/XrdSecProtocolkrb5.cc#L585

"Unable to start sequence on the keytab file " gets written to buff, plus 'krb_kt_name', which is up to 1024 (at least that's what the compiler thinks), so the compiler was complaining that 'buff' might be too short. I just picked the next round number ;-), it could have been shorter.

gganis:
The changes in the parts I am concerned more directly look fine, as far as I can see.
I have seen that are a number of hardcoded 1024 becoming hardcoded 2048 ... perhaps we could define a generic buffer length parameter and used that consistently?

src/XrdDig/XrdDigAuth.cc
abh3: OK, got it.

src/XrdSec/XrdSecEntity.hh
abh3: Well, that's only true if the Reset() is actually inlined, isn't it?But I guess we are not providing regressed compatibility (i.e. new code running with old libraries).

simonmichal:
@gganis : in general I agree that having a constance would be better, but for now we lack a common header with global constance, also re-factoring hardcoded buffer sizes in this PR will decrease readability. So I would decouple this kind of re-factoring from this PR ;-)

Issue #659, opened 2018-03-01
XrdPosix: rename does not automatically replaces existing file


gganis:
This issue was discussed by email last summer but I realized that I never filled an issue in here.

According to the rename man page:
If newpath already exists, it will be atomically replaced, ...

However, this is not honored by XrdPosix::rename: if the target file exists the call always fails.

In the above-mentioned email exchange the reason for this was tracked back to the difficulty to accomplish a 'true' rename behavior when dealing with non-POSIX file systems.
As a possible 'partial' fix it was proposed to allow a 'true' rename as long as the existing target is neither remote nor symlinked.
Could this be implemented?

Gerri

abh3:
This has been addressed by commit 23adaf.

PR #660, opened 2018-03-01
XrdOss: implement proposed solution for better rename POSIX compliancy

gganis:
Implements enhancement #659

### DIFF
--- a/src/XrdOss/XrdOssRename.cc
+++ b/src/XrdOss/XrdOssRename.cc
@@ -56,7 +56,7 @@
 extern XrdSysError OssEroute;
 
 extern XrdOucTrace OssTrace;
-  
+
 /******************************************************************************/
 /*                                R e n a m e                                 */
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


src/XrdOss/XrdOssRename.cc
```diff
@@ -110,12 +110,11 @@ int XrdOssSys::Rename(const char *oldname, const char *newname,
...
-// Make sure that the target file does not exist
+// If on a remote FS or symlink, make sure that the target path does not exist
 //
-#if 0
-   retc2 = lstat(local_path_New, &statbuff);
-   if (!retc2) return -EEXIST;
-#endif
+   if (!(retc2 = lstat(local_path_New, &statbuff)))
+      { if (remotefs || (statbuff.st_mode & S_IFMT) == S_IFLNK) return -EEXIST;
+      }
```
abh3: What this means is that any symlinked FS will not allow a rename on top of an existing link. Is that what you really want? That excludes any virtualized filesystem even if local.

gganis: The symlink part comes from your original comment. In my case I am mostly interested to local file, no symlinks. However, I see that the man page of rename says that

if newpath refers to a symbolic link, the link will be overwritten

So I suggest that we only leave the check on 'remotefs'. What do you think?

abh3:
Well, unfortuantely, in a symlinked FS that is not a simple operation. The 
symlink refers to a data file and the data file has an xattr that records 
its symlink name. All of that has to change to successfully overwrite the 
symlink. Merely removing the symlink test would really screw things up for 
many people. I guess that's why we forbid renaming on top of an existing 
file. I don't have a problem of restricting destructive renames to simple 
file systems as long as that's OK with you.

Andy

gganis:
>  I don't have a problem of restricting destructive renames to simple
file systems as long as that's OK with you.
Tha's ok with me. Does it mean that the PR can stay as it is?
Gerri

abh3:
Yes, I'll merge it.

Andy

PR #676, opened 2018-03-24
Always have OpenSSL read/write data through the XrdLink object.

### LINKED ISSUE #675: XrdHttp does not handle XrdLink counters correctly for HTTPS
When reading / writing data to a HTTPS socket, `XrdHttpProtocol.cc` will perform operations on the socket directly.

This is not ideal for two reasons:
1.  The `isIdle` counter on the `XrdLink` object is not reset to zero, meaning that the TCP connection will eventually be abruptly disconnected.  I have hit this in production mid-transfer, particularly in setups with aggressive session reuse.
2.  The link bytes counter is not incremented, meaning the server will not have correct summary statistics.

bbockelm:
Previously, we just handed OpenSSL the underlying file descriptor.  This worked fine, protocol-wise.  However, this didn't interact well with the Xrootd framework itself as the idle counter was not update, nor were proper statistics recorded.

With this commit, we create a custom BIO object for OpenSSL to use; this object simply forwards the requests to an underlying `XrdLink`.

Fixes #675

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -545,6 +545,7 @@ BIO *XrdHttpProtocol::CreateBIO(XrdLink *lp)
 
   BIO *ret = BIO_new(m_bio_method);
 
+  ret->shutdown = 0;
   ret->ptr = lp;
   ret->init = 1;
   return ret;
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -105,7 +105,8 @@ XrdScheduler *XrdHttpProtocol::Sched = 0; // System scheduler
 XrdBuffManager *XrdHttpProtocol::BPool = 0; // Buffer manager
 XrdSysError XrdHttpProtocol::eDest = 0; // Error message handler
 XrdSecService *XrdHttpProtocol::CIA = 0; // Authentication Server
-
+int XrdHttpProtocol::m_bio_type = 0; // BIO type identifier for our custom BIO.
+BIO_METHOD *XrdHttpProtocol::m_bio_method = NULL; // BIO method constructor.
 
 /******************************************************************************/
 /*            P r o t o c o l   M a n a g e m e n t   S t a c k s             */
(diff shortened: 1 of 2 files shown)


bbockelm:
This has been running on a production host at UCSD for a few days - no problems.

abh3:
I don't think Fabrizzio is complaining about that. My understanding is 
that the merge fails.

Andy

ffurano:
Hi Brian,

 little issue #1 ... for some reason the github instructions for merging give an error. They want pull -r instead of the suggested pull. Don't know if it's one more git issue, and probably don't want to

 bigger issue #2 ... I fetched your branch, and something does not seem right to me. It's true that a sequence of requests (e.g. davix) will almost always work, yet a banal OPTIONS done with telnet will not get a response. The server may claim that a response has been sent but the client never receives it. I'm 100% sure that this used to work, and it was one of my basic tests.

bbockelm:
@ffurano - yup, I eventually hit the same issue at UCSD (just a matter of time).  It was an inadvertent default that caused a deadlock -- the BIO defaulted to shutting down the XrdLink, which triggered a deadlock.

See updated PR.

ffurano:
Yup, seems to work now, thank you.

ffurano:
Do we know what is the timeout for XrdLink ?

I let a connection idle, how much time will it last?

abh3:
It's configurable. The deflt is to not time out the links. So, it varies 
from server to server.

Andy

ffurano:
Indeed, that idle connection is still working. Can't leave it up much longer by now.

For the records, how do I give a timeout to an XrdLink ?

xrootd-dev:
It's the xrd.timeout directive. See "idle" here:

http://xrootd.org/doc/dev48/xrd_config.htm#_Toc496911326

ffurano:
Thank you Brian, I can't find anything wrong with naive tests, so I merged it.
A tougher set of tests will be run by the DPM nightlies, however I don't expect surprises.

simonmichal:
@bbockelm & @ffurano :
this PR broke our builds for fedora rawhide, could you please fix it?
https://gitlab.cern.ch/dss/xrootd/-/jobs/1202416

Thanks,
Michal

bbockelm:
@ffurano - could you tackle this?  I've got a 1.5 weeks of travel + proposals coming up... I'll really struggle to look in deeply here.

PR #679, opened 2018-03-29
Third-party-copy plugin for XrdHttp

bbockelm:
This PR integrates the external xrootd-tpc plugin (see https://github.com/bbockelm/xrootd-tpc) into the main Xrootd source itself.

The plugin will be built if HTTP is enabled and, when loaded, will handle the COPY WebDAV verb via a callout to `libcurl`.  Multistreaming is supported, as well various extensions used by the WLCG community (`Copy-Header` support to add specific headers to the TPC).

To enable, one can add the following to the configuration file:
```
http.exthandler xrdtpc libXrdHttpTPC.so
```


bbockelm:
If it's not clear, this is NOT meant for 4.8.2.  It might be worthwhile to merge relatively soon afterward.

ffurano:
Thank you very much Brian. We will surely use it a lot

simonmichal:
Hi guys,
I have reverted it, as I stated in the revert comment:

The PR does not follow XRootD coding conventions and brakes the build on vanilla RHEL6 platform in presence of libcurl-devel package. There should be a check in the CMake infrastructures that ensures the XrdTpc package will be only built if a proper version of libcurl-devel is present (or XrdTpc should built with version of libcurl-devel provided in raw RHEL6 environment). Moreover the new component should be packaged properly (if necessary a _with_tpchttp macro should be tested in the spec file).

@abh3 can provide additional detail.

PR #706, opened 2018-05-16
cryptossl/RSA : fix ExportPrivate and ExportPublic signatures

gganis:
This was broken by a partial change back in Dec 2017
(commit 0421bc639e1f1e581e2f597264f445ffa3380660)
Should restore proper behaviour when XrdSecGSIDELEGPROXY is set to 2.

NB: this breaks backward compatibility of some ExportPublic and ExportPrivate methods .


gganis:
Wrong PR. Removing.

PR #710, opened 2018-05-19
Allow XRootD client to accept subjectAltNames.

bbockelm:
With this, if the CN doesn't follow the expected matching rules, then the client will iterate through the listed `subjectAltNames` to determine whether the certificate matches the current host.

This includes support for CNs and SANs with wildcards.

One thing I noticed when testing this branch is XRootD is internally still doing a reverse DNS lookup on the IP address it connects to when filling out the `XrdSecEntity` structure.  This is widely considered insecure despite the fact it was commonly done at the grid level until a few years ago.  I'd strongly suggest we remove the reverse DNS lookup on the client side once there's appropriate SAN support.

### DIFF
--- a/src/XrdCrypto/XrdCryptoX509.cc
+++ b/src/XrdCrypto/XrdCryptoX509.cc
@@ -252,3 +252,77 @@ int XrdCryptoX509::DumpExtensions(bool)
    ABSTRACTMETHOD("XrdCryptoX509::DumpExtensions");
    return -1;
 }
+
+/**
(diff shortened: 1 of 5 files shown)


PerilousApricot:
FWIW - this would be super nice for our (Vanderbilt) site, especially since we're having to pay per-certificate. If you have a large site, requiring individual certs for each host gets to be quite expensive

gganis:
Looks good to me. I'll merge it.

PR #730, opened 2018-06-07
XrdClStream.cc: typo

jmuf:


jmuf:
sorry for this, but am seeing this message way too often on my screen to ignore..

bbockelm:
Ha!  I was complaining about the same thing just this week... it's been irritating me for so long that it's now endearing.

PR #731, opened 2018-06-07
Use DNS lookups to expand non-FQDNs

### LINKED ISSUE #725: [XrdSecgsi] regression in 47eb688dc2131c30edd57eba525f95bb2ea842ba
While testing the recent changes in XrdSecgsi I noticed that a regression has been first introduced in 47eb688dc2131c30edd57eba525f95bb2ea842ba (and **haven't been fixed** by 1533382e11418c24916c93392973ac1c3fcab552).

Consider XRootD with CERN Grid host certificate running on host.cern.ch. When the client tries to authenticate after issuing a command like:
```
 xrdfs host ls /tmp
```
(note hostname was given without _.cern.ch_ domain)
it fails and following error can be seen in the logs:

```
secgsi_getCredentials: server certificate CN 'host.cern.ch' does not match the expected format(s): '[*/]host[/*]' (default); exceptions are controlled by the env XrdSecGSISRVNAMES CF: 0x7fffeee1c360
secgsi_ErrF: Secgsi: ErrParseBuffer: server certificate CN 'host.cern.ch' does not match the expected format(s): '[*/]host[/*]' (default); exceptions are controlled by the env XrdSecGSISRVNAMES: kXGS_cert
```

@bbockelm : Before the commit in question the _host_ hostname was properly resolved into _host.cern.ch_ , currently we do DNS lookups only for IP addresses. Could you explain why the lookup has been removed for plain hostnames?

bbockelm:
Fixes #725.  @simonmichal - please test.

If the user provides a non-FQDN (such as `xrootd` instead of `xrootd.cern.ch`), this will detect the situation and use `getaddrinfo` to expand to the canonical name.

Note that the canonical name may be somewhat unexpected as it resolves CNAMEs all the way down to an A record.  For example, if you set your search domain to `unl.edu`, then the canonical name for `www` is not `www.unl.edu` but rather `unlcms-rs.autosplit.its-sys.unl.edu`.

This also allows the environment variable `XrdSecGSITrustDNS` to be set to 0; if it is, all DNS lookups are skipped.

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -35,6 +35,8 @@
 #include <sys/param.h>
 #include <pwd.h>
 #include <sys/types.h>
+#include <sys/socket.h>
+#include <netdb.h>
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <dirent.h>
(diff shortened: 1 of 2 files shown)


abh3:
I really would prefer that NetAddr class be used to resolve the missing domain instead of recoding all of that in-line here. It just becomes a maintenance nightmare. So,c ould you reduce the dozen or so lines with one call to NetAddrInfo?

bbockelm:
I don’t think anything in NetAddr does precisely this.  Should I add a new static method?

abh3:
The ref to netaddr describes the socket and hname is obtained from that object subject to the DNR option. So, it’s intimately tied to the passed netaddr object. So, simply calling the Name() function on that object will accomplish what you want. The only reason we pass in hname is for backward compatibility reasons and people should be using the endPoint reference but we didn’t have time to change it in every security protocol (so we didn’t). You might want to consider doing that instead and not using hname at all. Anyway, if you call Name() on endpoint it will do a getaddrinfo() call with the same options you have to get a canonical name for the endpoint.

bbockelm:
Hi Andy,

I simplified things a bit so we don't rely on DNS at all for the majority of cases (anything that looks like a FQDN) but still handles the case reported by @simonmichal.

The problem with using the trick of appending a `.` is that it still relies on DNS - hence exposing the security issue we were trying to avoid in the first place.

With this, I believe things are secure by default except in detectable backward compatibility scenarios.

Brian

simonmichal:
I can confirm that it does solve the problem in #725.

bbockelm:
@gganis - what's the next step here?

gganis:
Hi Brian,
As I wrote privately, although less relevant now, I think we should provide the possibility to control this also server side. We can do it with an env, but usually we do it with switches. 
I have sent you a patch doing that and slightly renaming the env.
If you do not want to apply the complete patch, can you at least rename the env (you seemed to agree upon)?

Gerri

bbockelm:
Hi Gerri,

Ah yes, I forgot about that patch.  Will apply it momentarily.

Brian

bbockelm:
@gganis - patch applied.

gganis:
@bbockelm Thanks, merged!

PR #732, opened 2018-06-07
Refactor user name extraction from certificate CN

mpatrascoiu:
When using `sec.potocol gsi` with `-authzfun:libXrdSecgsiAUTHZVO.so` and `-authzfunparms with empty vo2usr` , the behavior is to extract the user name from the certificate's /CN field. 

This commit refactors the name extraction to only the first CN encountered.
A text limitation to N characters is respected, where N = length of entity.vorg.

Previously, it would grab all the text after the CN field, stopping either at the end or when N characters were met.

### DIFF
--- a/src/XrdSecgsi/XrdSecgsiAuthzFunVO.cc
+++ b/src/XrdSecgsi/XrdSecgsiAuthzFunVO.cc
@@ -143,7 +143,7 @@ int XrdSecgsiAuthzFun(XrdSecEntity &entity)
    static XrdSysMutex Mutex;
    const char *vtxt = "", *etxt = 0;
    char vbuff[(g_maxvolen+1)*2];
-   int i, n;
+   int idx = 0, n;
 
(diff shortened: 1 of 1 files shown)


bbockelm:
I'm not sure you want the first one.  Maybe the last one?  For example, here's my CERN DN:

`/DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=bbockelm/CN=659869/CN=Brian Paul Bockelman`

With this patch, the username would be `659869`, no?

mpatrascoiu:
Hello Brian,

With the change, the username would be bbockelm.

A complete solution would be to add another argument to the `-authzfunparms` to decide between the first CN, last CN or all of them but I'm not sure that would be something which is needed.

xrootd-dev:
Another concern is that some sites may expect the old behavior and changing this would break things.

ffurano:
Hi,
 for the records, I have no preference for the solutions, they are both equally bad to me (not that I see a good one elsewhere)

Issue #713, opened 2018-05-23
XrdHttp should have appropriate HTTP status code for permission denied


bbockelm:
In #691, @olifre noted the `XrdHttp` plugin always returns a 404 (File Not Found) in the case of an error.

This is particularly confusing when there's a permission denied, as that would more naturally map to a "403 Unauthorized User".

Let's implement the translation between permission denied and 403 status code.

(There are, of course, many other translations we could make... let's start with something concrete and move forward).

abh3:
Didn't this get fixed on some recent pull request?

ffurano:
Hi Andy,

 I think so, but did not have time to verify... sorry. I'll verify today :-)

f


Il 21/08/18 06:43, Andrew Hanushevsky ha scritto:

ffurano:
Hi Andy,

 yes AFAIK :-)

f

abh3:
OK, so I am closing this as solved.

ffurano:
yup

PR #735, opened 2018-06-08
XrdHttp: Xrd error code to HTTP status translation

mpatrascoiu:
This change introduces a new method which translates from an Xrd error code into an appropriate HTTP 
status code.

It introduces 2 new private variables in XrdHttpReq: `httpStatusCode` & `httpStatusText`.
`httpStatusCode`: HTTP Status compliant code
`httpStatusText`: the Xrd error text, if present, or a minimal description otherwise
Default values are 500 and "Unrecognized error".

During the request post-processing, the branches which end in an error will send back the status code and text mentioned above.

This pull request addresses issue #713

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1426,9 +1426,12 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc, const char *hea
     ss << desc;
   } else {
     if (code == 200) ss << "OK";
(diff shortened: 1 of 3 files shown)


mpatrascoiu:
Hello Brian,

Good catch. I've did a small commit where I'm also passing the `httpStatusText.length()` to the `SendSimpleResp` function.

ffurano:
Hi,
I'm starting my little doublechecks now. I agree that this is a nice little improvement
Thank you Mihai
Fabrizio

PR #749, opened 2018-06-19
Restructure GSI proxy delegation options

gganis:
This patch re-organizes how proxy delegation is handled and controlled in GSI, and it adds the possibility to save the proxy in Entity.creds .

On the server side, the two switches controlling delegation have been 'cleaned': the first switch enables or disables delegation, the second determines where the delegated proxy will be saved. The new meanings are:
              -dlgpxy:0              no delegated proxy [default]
                           1              ask the client to sign a delegated proxy request

              -exppxy:none       delegated proxy available in memory (via a server calle to getCredentials)
                           :=creds    delegated proxy available in Entity.creds
                           :<file_template>  delegated proxy available in the indicated file which can include the
                                           following customization tags: <host>, <vo>, <group>, <user>, <rtag>; e.g.
                                           /tmp/x509up_u<user>_<rtag> .

(rtag is a 6 hex chars random string).

On the client, the env XrdSecGSISIGNPROXY is used to enable or deny proxy signature; default is 1, that is enabled. The env XrdSecGSIDELEGPROXY is used to determine the time of delegated proxy: 1 means standard delegated proxy, i.e. a proxy signed by the initial proxy (proxy request created by teh server); 2 means forwarding of the initial proxy.

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -967,12 +967,14 @@ char *XrdSecProtocolgsi::Init(gsiOptions opt, XrdOucErrInfo *erp)
          DefBits = opt.bits;
       //
       // Delegate proxy options
-      if (opt.dlgpxy == 1)
-         PxyReqOpts |= kOptsDlgPxy;
-      if (opt.dlgpxy == 2)
-         PxyReqOpts |= kOptsFwdPxy;
-      if (opt.sigpxy > 0 || opt.dlgpxy == 1)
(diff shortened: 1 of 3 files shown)


bbockelm:
Can we disable delegation by default?

I'm somewhat uncomfortable with the client handing over my proxy to an external server (especially as there are some interesting scenarios to think through about malicious servers joining a federation...).  I'd rather start slow and work through some of these cases.

gganis:
@bbockelm, @abh3 The last patch removes one env client side and disables delegation by default.

PR #762, opened 2018-07-10
Macaroons plugin for Xrootd

bbockelm:
This PR implements support for generated and utilizing authorization tokens with the Macaroon format.


bbockelm:
@simonmichal - we need to have libmacaroons installed on Travis-CI in order to test the feature.  How would you like to go about it?  Do you want me to just tweak the `.travis.yml`?

simonmichal:
@bbockelm : I would say that the macaroon plug-in should only be built if libmacaroons is present. This will shield us from failures on platforms where libmacaroons is not available. Of course then the spec file needs to be tweaked so the plug-in gets only packaged for EPEL-7 and Fedora platforms (my understanding is that there is no libmacaroons for EPEL-6). 

The idea is that our build script will install the dependencies with yum-builddep / dnf builddep, and cmake will only build the plug-in if respective dependencies are in place.

BTW Could you rewrite the history so there are fewer but more meaningful commits in the PR, ideally one called something like _Macaroons plugin for Xrootd_ ;-)

bbockelm:
>  I would say that the macaroon plug-in should only be built if libmacaroons is present. 

Sure, no problem.  Will start working on this.

> Could you rewrite the history so there are fewer but more meaningful commits in the PR

Hm -- I'm the sort of person who really likes to keep history.  What if I rewrote the commit messages to be prefixed with `[XrdMacaroons]`?

simonmichal:
@bbockelm : `[XrdMacaroons]` prefix is good enough for me :-) thanks!

bbockelm:
@simonmichal - OK, that was about my limit of abilities in `git`, but I managed to rewrite all the bits of history.

Also fixed the builds as requested.

Probably ready to be looked at again.

simonmichal:
Thanks a lot :-) for me it looks good :-)

bbockelm:
What's next in moving this one forward?

PR #778, opened 2018-07-25
Prevent double-parsing of parameters for ofs.ckslib.

### LINKED ISSUE #777: ofs.ckslib consumes two lines of configuration
It appears that when `ofs.ckslib * path` is present in the configuration file, then `GetRest` is invoked twice:

- Once in `XrdCksConfig::ParseLib`
- Once in `XrdOfsConfigPI::RepLib`

This has the side-effect of causing the `ofs.ckslib` line to consume the contents of the next line, making the following line ignored by the configuration parser.

To reproduce, set

```
ofs.ckslib * /var/lib/xrootd-hdfs/lib/libXrdHdfs.so
testline
```

and look for the following at startup:
```
=====> ofs.ckslib * /var/lib/xrootd-hdfs/lib/libXrdHdfs.so  testline
```

bbockelm:
Fixes #777

### DIFF
--- a/src/XrdOfs/XrdOfsConfigPI.cc
+++ b/src/XrdOfs/XrdOfsConfigPI.cc
@@ -280,7 +280,7 @@ bool XrdOfsConfigPI::Parse(TheLib what)
                               if (CksConfig->ParseLib(*Config, libType))
                                  return false;
                               if (libType) cksLcl = libType == 1;
-                              RepLib(theCksLib, CksConfig->ManLib(), nullParms);
+                              RepLib(theCksLib, CksConfig->ManLib(), nullParms, false);
                               return true;
                              }
                           Eroute->Emsg("Config", "Checksum version error!");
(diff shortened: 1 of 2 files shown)


abh3:
OK, I am confused. Could you explain exactly what the problem was and how this fixes it?

bbockelm:
Hi @abh3 -

The description of the problem is in ticket #777, no?

Without this, the code path that calls `ofs.ckslib` invokes `Config->GetRest` twice, meaning that `ofs.ckslib` is consuming two lines of configuration.

For example, the following config snippet will fail to parse:

```
ofs.ckslib * /var/lib/xrootd-hdfs/lib/libXrdHdfs.so
if exec xrootd
fi
```

because the handling of `ofs.ckslib` will also consume the tokens `if exec xrootd`, meaning that Xrootd complains of an unmatched `fi`.

Brian

abh3:
Thanks, the example is just what was needed for the history behind this patch.

PR #783, opened 2018-07-30
[XrdTpc] Fix order of chaining when parsing xrootd.fslib.

bbockelm:
Add comments to prevent future confusion over how this line is parsed.

Prior to this, a `xrootd.fslib` line that specifies a plugin chain:

```
xrootd.fslib /usr/lib64/libXrdMultiuser.so default
```

would not actually create the base default plugin.

### DIFF
--- a/src/XrdTpc/XrdTpcConfigure.cc
+++ b/src/XrdTpc/XrdTpcConfigure.cc
@@ -69,14 +69,24 @@ bool TPCHandler::ConfigureFSLib(XrdOucStream &Config, std::string &path1, bool &
         path2 = val;
     }
     if (!(val = Config.GetWord()) || !strcmp("default", val)) {
-        if (path2 == "libXrdThrottle.so") {
+        // There is not a second path specified or we requested the default path.
+        // Configuration of the form "xrootd.fslib /some/path.so"
+        // or                        "xrootd.fslib /some/path.so default"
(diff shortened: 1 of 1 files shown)


bbockelm:
@simonmichal - I think Fabrizio might be on vacation.  Would you be able to review this PR?

ffurano:
Hi, I am around, yet a bit spotty.
This PR seems to be a fix to XrdTPC to make it able to load correctly a chain of Sfs plugins. The fix seems small and clean in a much larger environment. I may understand what this is for, however I miss a properly malicious way to do a meaningful test involving the other libs that are mentioned.
I would accept it just because it makes sense to me.

abh3:
Hi Fabrizio,

Not to worry. It got merged as it, as you said, was localized and seemed 
clean.

Andy

bbockelm:
Hi,

Yup - this addresses a bug that we discovered when testing the `multiuser` plugin at Florida.  To reproduce, add:

```
http.exthandler xrdtpc libXrdHttpTPC.so
xrootd.fslib libXrdMultiuser.so default
```

I think this would also reproduce:

```
http.exthandler xrdtpc libXrdHttpTPC.so
xrootd.fslib libXrdThrottle.so default
```

(i.e., chain with the throttle plugin but do not use the `xrootd.fslib throttle default` abbreviation)

Regardless, thanks for the merge!

Brian

PR #789, opened 2018-08-01
Replace 'dl' with ${CMAKE_DL_LIBS} to make it work on systems without libdl

yurivict:
… libdl


abh3:
I think we have a problem here (and you're going to hate it). The project has to support relatively old version of all sorts of things, including cmake. From my reading, $(CMAKE_DL_LIBS) was introduced in version 3.x and we need this to function in version 2.x (well, 2.8 or higher). So, I think additional work will be needed on this patch to satisfy this requirement. Otherwise, it's a great idea.

yurivict:
I added the cmake version check. When the version is <3.0 it will be like before this patch.

PR #807, opened 2018-08-28
File operation workflows

kjamrog:


simonmichal:
Depends if you compile with clang or gcc ;-P
( I known you would like it ;-)

PR #811, opened 2018-08-29
Fix callback of protocol plugin after bridge delay

### LINKED ISSUE #810: XrdHttp loses requests under modest concurrency
After applying a workaround for issue #809, I notice that `ab` will quite consistently fail with only few number of repeated requests:

```
$ ab -k -s 30 -n 400 -c 100 http://host.example.com/some/path/hello_world
This is ApacheBench, Version 2.3 <$Revision: 1430300 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking hcc-briantest7.unl.edu (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
apr_pollset_poll: The timeout specified has expired (70007)
Total of 398 requests completed
```

This is a fairly vanilla setup - should be reproducible with the HTTP module on top of a POSIX filesystem.

I haven't been able to diagnose it precisely, but it seems to only occur when this line pops into the log:

```
180828 16:07:34 880  ofs_Stall: Stall 3: File hello_world is being staged; estimated time to completion 3 seconds for /some/path/hello_world
```

I admit, I don't understand why the file would ever be considered as being staged with the default OFS plugin.  However, the behavior is very much as if there's a callback not occurring.

*NOTE* this issue is based on an investigation into user complaints about the service.  It may be a synthetic benchmark but it shows something actually observed in the wild.

bbockelm:
Fix #810

### DIFF
--- a/src/XrdOfs/XrdOfsHandle.cc
+++ b/src/XrdOfs/XrdOfsHandle.cc
@@ -503,17 +503,9 @@ do{xP = XrdOfsHanXpr::Get(); hP = xP->Handle;
   
 int XrdOfsHandle::WaitLock(void)
 {
-   int ntry = LockTries;
-
 // Try to obtain a lock within the retry parameters
 //
-   while(ntry--)
-        {if (hMutex.CondLock()) return 1;
-         if (ntry) XrdSysTimer::Wait(LockWait);
-        }
-
-// Indicate we could not get a lock
-//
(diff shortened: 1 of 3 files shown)


abh3:
Thank you Brian for both patches, especially the one to update some old code.

PR #812, opened 2018-08-29
Obey the `Connection` request header.

### LINKED ISSUE #809: XrdHttp does not handle Connection header correctly
XrdHttp should have better handling of the `Connection` header.  Without it, HTTP 1.0 clients may timeout when trying to download files.

For example, using the venerable `ab` ("ApacheBench") utility with the following invocation:
```
ab -n1 http://host.example.com/foo
```
it will hang and fail the connection because it sets `Connection: close` in the request and waits for the TCP connection to close before considering the request complete (note: if you are trying to reproduce, it will only do this for a `20x` response).

My reading of things is that the response behavior should be:
   * If the request asks to keep the connection alive, keep it alive.
   * If the request asks to close the connection, close it.
   * If the request doesn't provide a `Connection` header, it should be kept alive as we respond with `HTTP 1.1` (the default for that version).

bbockelm:
Additionally, if the client is making a HTTP/1.0 request, make sure
to disable keepalive by default.

Fix #809 

@ffurano

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -31,14 +31,14 @@ int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header
 {
   if (!prot) return -1;
   
-  return prot->SendSimpleResp(code, desc, header_to_add, body, bodylen);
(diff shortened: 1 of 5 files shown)


ffurano:
Thank you Brian, this is a nice one too, makes sense.
I have merged it, now this will enter my heavy tests.

Issue #813, opened 2018-08-29
XrdHttp loses requests under modest concurrency - v2


bbockelm:
After fixing the issue in #810, I can confirm our production servers can now run the `ab` benchmark without dropping responses to requests or significant spikes in latency.

However, I still see failures in the LIGO workflow with the same symptom - the response looks like it isn't completed until after the client gives up.

Working on a simpler reproducer, but I wanted to note that there are issues that remain.

bbockelm:
Here's something that triggers a similar issue:
```
ab -k -H 'Range: bytes=1-2097152' -c 1 -s 10 -n 1 https://host.example.com:1094//path/to/bigfile
```

   * Works fine if you start at byte 1
   * Works fine without SSL.

I'm thinking maybe it's a SSL framing issue?

bbockelm:
Hm - actually, I suspect the `ab` thing above is actually something screwy with the benchmark tool.

I'm working on some larger tests, including copying a public dataset into the server so I can more easily share the setup with others.

xrootd-dev:
Hi Brian,

Also do suspect SSL you may have to do a BIO flush if SSL decides that is 
wants more data -- just a thought. It will be ineteresting if that was 
actually the case.

Andy

bbockelm:
Yeah, that was my first suspicion too.  However, OpenSSL man pages swear up and down that an explicit flush isn't needed (and I don't see any buffering).  I even tried adding printf statements .. no luck!

However, I was able to make progress by manually adding `Connection: close` to the headers, which triggered a new error message.  It appears the server is sending too little data and, when `keep-alive` is enabled, the client is timing out waiting for a response.

I found that, under load, `XrdHttpReq` is send too little data.  For example, given this request:
```
180830 22:41:28 16434 sysXrdHttp:  rc:125 got hdr line: GET //user/ligo//frames/O2/hoft_C02_clean/H1/H-H1_CLEANED_HOFT_C02-11822/H-H1_CLEANED_HOFT_C02-1182257395-4096.gwf HTTP/1.1
180830 22:41:28 16434 sysXrdHttp:  Parsing first line: GET //user/ligo//frames/O2/hoft_C02_clean/H1/H-H1_CLEANED_HOFT_C02-11822/H-H1_CLEANED_HOFT_C02-1182257395-4096.gwf HTTP/1.1
180830 22:41:28 16434 sysXrdHttp:  rc:34 got hdr line: Host: server.example.com:1094
[...]

PR #814, opened 2018-08-31
[XrdHttp] Reset filesize when the XrdHttpReq object is used.

bbockelm:
The `filesize` member was not being reset between requests.  If it is non-zero, the `XrdHttpReq` object uses the size when parsing header to potentially reduce the total read size (preventing it from going past the EOF).

Unfortunately, since the `filesize` was set to the size of the *previous* file transfer, this could cause the total response size to be less than the request, resulting in a shorter-than-expected response to the client.


bbockelm:
Fix #813

PR #824, opened 2018-09-12
Update xrdfs.1 - document 'BackUpExists' stat flag

jmuf:
trivial: "BackUpExists" (i.e file is stored on archival media) is implemented but not in the manpage.


murrayc3:
This half a joke.

Vlado read the following paper from Symantec Corporation:

http://ccbjournal.com/articles/11650/archiving-e-discovery-backup-recovery

The paper says backup is for recovery while archive is for discovery.  Do we have an archived bit? :-)

PR #826, opened 2018-09-17
Pfc purge v2

osschar:
Purge algorithm can now also consider space used by cached files and age of files. The main purpose is to allow several xcache instances to share the same set of disks.

### DIFF
--- a/docs/PreReleaseNotes.txt
+++ b/docs/PreReleaseNotes.txt
@@ -19,6 +19,7 @@ Prerelease Notes
   * **[XrdApps]** Implement xrdqstats command to display summary monitoring.
   * **[XrdSsi]** Provide summary monitoring information to report stream.
   * **[TPC]** Allow number of streams to use to be passed to the server.
+  * **[Proxy]** Implement new options in pfc.diskusage for better control of purging.
 
 + **Major bug fixes**
   * **[SSI]** Do not leak memory when a fatal error occurs. Fixes #775
@@ -33,5 +34,5 @@ Prerelease Notes
   * **[Proxy]** Add method to cache to get local file path of cached file.
   * **[All]** Place protocol definition under a modified BSD license.
   * **[Proxy]** Avoid auth failure due to URL cgi directives.
-
+  * **[Proxy]** Fix wrong logic in cache purge algorithm.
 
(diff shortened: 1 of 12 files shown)


bbockelm:
Aren't you worried by the fact that GCC seems to think that `GetSizeInBytes()` can be replaced by the constant `0`?  Is there something the compiler is seeing which we aren't?

osschar:
No, not worried :) Found this:
  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61294
And then, the call to Info::GetSizeInBytes() (inline) can return 0 ... leading to optimizer producing a different path (and memset call) for == 0 case. And this brings up the warning that tells you you might have transposed the parameters (as people usually set memory to 0, not to 255 as we do here).

bbockelm:
Ah, clever GCC.

Can you add a link to the GCC bug explaining why you don't use memset here?

(Alternately, is there an attribute we can set to suppress warnings?)

osschar:
I could put something like:
`
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Werror"
  memset(xxx, 255, GetSizeInBytes());                       /* no diagnostic for this one */
#pragma GCC diagnostic pop
`
or add comment like:
`
   // The following should be: memset(m_store.m_buff_synced, 255, GetSizeInBytes());
   // but GCC produces an overzealous 'possible argument transpose warning' and
   // xrootd build uses warnings->errors escalation.
   // See also: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61294
`

bbockelm:
I like both!  For the comment, also include the compiler version that hit the issue.  That way, we can also remove the workaround at a later date.

Brian

simonmichal:
@osschar : commit 87d162be15e5f41ad23bc85fa6b86e7d9f64e058 broke the OSX build:
https://gitlab.cern.ch/dss/xrootd/-/jobs/2164942
(there's no _posix_fallocate_ on OSX platform)

Could you fix it?

osschar:
Thanks for the heads up! I'm on to it now.

osschar:
@simonmichal the fix is in #827, apparently mozilla and KDE had the same problem :)

PR #827, opened 2018-09-19
[Pfc] Fix compilation on OSX

osschar:
Add definition of posix_fallocate equivalent for OSX.

### DIFF
--- a/src/XrdFileCache.cmake
+++ b/src/XrdFileCache.cmake
@@ -27,8 +27,7 @@ add_library(
   XrdFileCache/XrdFileCacheIO.cc            XrdFileCache/XrdFileCacheIO.hh
   XrdFileCache/XrdFileCacheIOEntireFile.cc  XrdFileCache/XrdFileCacheIOEntireFile.hh
   XrdFileCache/XrdFileCacheIOFileBlock.cc   XrdFileCache/XrdFileCacheIOFileBlock.hh
-  XrdFileCache/XrdFileCacheDecision.hh
-  XrdFileCache/posix_fallocate_osx.hh)
+  XrdFileCache/XrdFileCacheDecision.hh)
 
 target_link_libraries(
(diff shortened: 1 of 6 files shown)


abh3:
Please don't define functions in header files. Create a cc file for the actual code. This avoids having the same function defined twice should some other package need it. Also, this properly belongs in XrdSys. Have you checked the author's license terms?

osschar:
It's defined as static ... so it is sort of inline. I wanted to avoid modifying the code more than absolkutely necessary as I suspected you will want it to go elsewhere ...

I'm all for putting this in XrdSys. So, pray, tell me where to put it and I shall modify the code.

I checked the license and complied with it (it said one can remove their licensing terms and replace them with lgpl 2.1 or higher -- I used lgpl 3).

abh3:
Pretty easy. Put the header with an "extern" definition to he function which is defined in the corresponding "cc" file. So, here we have (I shortened the names to make them more manageable):

XrdFileCache/posix_fallocate_osx.hh -> XrdSys/XrdSysFallocate.hh
If the have something like this

#ifndef __APPLE__
#include <fcntl.h>
#else 
........ the whole thing sans the function body with the following....
extern int posix_fallocate(int fd, off_t offset, off_t len);
#endif 
----------------
Then place the actual function body in 
XrdSys/XrdSysFallocate.cc

Make sure to if/def it so it only compiles on macOS.

Then in add the header and cc file entry in XrdUtils.cmake in the obvious place.

osschar:
Thanks, on to it! Do you really want me to put extern for the function declaration?

osschar:
Hmmh ... I checked if it is sorted and figured it is not so I put it at the end. Nope ... XrdUtils.cmake, XrdSys section still doesn't look sorted to me (but I might be sort-blind). Do you want me to sort the whole XrdSys section?

PR #835, opened 2018-10-04
Support multiple IO objects working with the same file

osschar:
This is needed for n2n mapped files and for forwarding-mode
proxies.

Access stats are still collected on per file basis. If several IOs all
read from the same file at overlapping times, these will all be
reported as a single access. This can be improved if needed.

When a block fetch fails on certain IO, other IOs will reissue the
block request. The offending IO will get the error code result
associated with its request. When prefetching, the bad IO will be
removed from the prefetch IO map so it doesn't cause trouble again.

### DIFF
--- a/src/XrdFileCache/XrdFileCache.cc
+++ b/src/XrdFileCache/XrdFileCache.cc
@@ -86,8 +86,11 @@ XrdOucCache2 *XrdOucGetCache2(XrdSysLogger *logger,
    pthread_t tid1;
    XrdSysThread::Run(&tid1, ProcessWriteTaskThread, (void*)(&factory), 0, "XrdFileCache WriteTasks ");
 
-   pthread_t tid2;
-   XrdSysThread::Run(&tid2, PrefetchThread, (void*)(&factory), 0, "XrdFileCache Prefetch ");
+   if (factory.RefConfiguration().m_prefetch_max_blocks > 0)
+   {
+      pthread_t tid2;
+      XrdSysThread::Run(&tid2, PrefetchThread, (void*)(&factory), 0, "XrdFileCache Prefetch ");
+   }
 
    pthread_t tid;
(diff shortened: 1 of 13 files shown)


simonmichal:
@osschar : this PR broke our builds for all Fedora and Debian platforms ;-)

Could you please have a look?

https://gitlab.cern.ch/dss/xrootd/pipelines/532653
https://gitlab.cern.ch/dss/xrootd/-/jobs/2275458

```
[ 49%] Building CXX object src/CMakeFiles/xrdpfc_print.dir/XrdFileCache/XrdFileCacheInfo.cc.o
In file included from /root/rpmbuild/BUILD/xrootd/xrootd/src/XrdFileCache/XrdFileCacheInfo.cc:35:
/root/rpmbuild/BUILD/xrootd/xrootd/src/XrdFileCache/XrdFileCacheTrace.hh:29:15: error: 'const char* {anonymous}::TraceInt2Str(int)' defined but not used [-Werror=unused-function]
   const char* TraceInt2Str(int lvl)
               ^~~~~~~~~~~~
make[2]: *** [src/CMakeFiles/xrdpfc_print.dir/build.make:76: src/CMakeFiles/xrdpfc_print.dir/XrdFileCache/XrdFileCacheInfo.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:328: src/CMakeFiles/xrdpfc_print.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
```

abh3:
I'm on it. It's going to be not so straightforward to solve but solvable.

xrootd-dev:
Yeah, the fix is not great but workable. Apparently, newer version of gcc 
detect unused variables in global and anonymous namespaces. his isn't a 
problem in older versions of gcc.

Andy

PR #844, opened 2018-10-18
Always fall back to reverse DNS if trusted

### LINKED ISSUE #841: xrdcopy ignores subject alternative names from the x509 host certificate
Hello,

we recently have changed our layout for our dCache storage element, which also affected the xrootd capabilities. As you may know, [dCache](https://www.dcache.org/) has so-called "doors", that implement the support for different transfer protocols, including [xrootd](https://www.dcache.org/manuals/Book-3.2/config/cf-xrootd-fhs.shtml). That xrootd door may use a gsi plugin to enforce grid security standards for xroot transfers. With our changes however, we ran into a new issue, we avoided before by coincidence: The xrootd transfers fail with an "auth error", because the (primary) host name of the dCache xrootd door does not match the (primary) subject name in the host's certificate.

> Auth protocol handler for gsi refuses to give us more credentials Secgsi: ErrParseBuffer: server certificate CN 'f01-081-126.gridka.de' does not match the expected format(s): '[*/]dcachesrm-kit.gridka.de[/*]' (default); exceptions are controlled by the env XrdSecGSISRVNAMES: kXGS_cert
--- snippet from [xfer.log](https://github.com/xrootd/xrootd/files/2483307/xfer.log)

In my opinion, this is a bug in the xrdcopy tool, since the name we used in the transfer URL is listed among the subject alternative names inside the host certificate.
```
$ openssl x509 -in /etc/grid-security/dcache/hostcert.pem -noout -text | grep -A 1 "Subject Alternative"
            X509v3 Subject Alternative Name:
                DNS:f01-080-118.gridka.de, DNS:dcachedcap-kit.gridka.de, DNS:dcachegridftp-kit.gridka.de, DNS:dcachesrm-kit.gridka.de, DNS:dcachewebdav-kit.gridka.de, DNS:dcachexrootd-kit.gridka.de, DNS:f01-080-118-v4.gridka.de, DNS:f01-080-118-v6.gridka.de
```

The error may be overruled by exporting `XrdSecGSISRVNAMES="f01-080-118.gridka.de"`. Though that is not what we can demand from all our customers around the world ("Please ignore the security issues and simply trust our servers." 😉 ).

At first I blamed the plugin in dCache for this, but one of the developers from dCache.org told me, that they do not interfer with that. That is, they provide the host certificate as is to the client and it should extract the information that it is interested in - maybe you can find evidence for that in the [attached log file](https://github.com/xrootd/xrootd/files/2483307/xfer.log) of the transfer.

Lastly I want to mention that we're currently working with version 4.8.4 of xrdcopy and removing the gsi plugin from the dCache door will make the error go away, too (which means, that xroot transfers can happen without it).

I was unable to find similar complaints among the issues here on GitHub or the mail archive at xrootd.org. In case you need more information, I'd gladly provide it to you.

Best regards,
Xavier Mol.

bbockelm:
Whenever reverse DNS is trusted, this enables the "historical" configuration where the IP address of the connected socket is considered for reverse resolution (and the remote server's DN is compared against the resulting hostname).

IMHO, this is significantly more dangerous than "traditional' reliance on reverse DNS for non-Xrootd cases.  Any attacker that is performing a MITM is able to force the client to accept any valid host certificate.

So, if the attacker controlling DNS could force `foo.example.com` to resolve to attacker-controlled IP address `1.2.3.4` with reverse resolution of `badguy.io`, the client would accept the connection as long as `badguy.io` has a certificate from a valid authority.

Fix #841

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -308,6 +308,7 @@ XrdSecProtocolgsi::XrdSecProtocolgsi(int opts, const char *hname,
    // usage of DNS (processed on XrdSecProtocolgsiInit).
    // Default is to fallback to DNS lookups in limited
    // cases for backward compatibility.
+   expectedHost = NULL;
    if (TrustDNS) {
       if (!hname || !XrdNetAddrInfo::isHostName(hname)) {
          Entity.host = strdup(endPoint.Name(""));
@@ -337,6 +338,7 @@ XrdSecProtocolgsi::XrdSecProtocolgsi(int opts, const char *hname,
       // We have been told via environment variable to not trust DNS; use the exact
       // hostname provided by the user.
       Entity.host = strdup(hname);
+      expectedHost = strdup(hname);
    }
    epAddr = endPoint;
    Entity.addrInfo = &epAddr;
@@ -1031,6 +1033,7 @@ void XrdSecProtocolgsi::Delete()
    SafeDelete(sessionKsig);   // RSA key to sign
    SafeDelete(sessionKver);   // RSA key to verify
    SafeDelete(proxyChain);    // Chain with delegated proxies
+   SafeDelete(expectedHost);
 
    delete this;
 }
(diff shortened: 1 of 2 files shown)


abh3:
Well, the answer is yes but the unfortunate problem is we don't have a good way to get around it in an expeditious way. Fortunately, where this is needed for x509 based XRootD TPC is for source selection and that is inherently much less dangerous since if there is any kind of spoofing the checksum should catch it in the end. It also helps that XRootD TPC only allows pull requests which minimizes (though does not eliminate) the severity of the exposure.

PR #845, opened 2018-10-22
Allow user to control OpenSSL's verify depth

bbockelm:
Additionally, this increases the default depth.  We have observed cases "in the wild" where the default value is too low.


bbockelm:
Hold on to this one... I don't think this is the underlying problem here...

abh3:
From previous comments, it would appear that this fix is no longer needed. Is that true?

bbockelm:
Correct, I think this was a red herring.  Underlying problem was different; I'd prefer to not expose another configuration variable!

Issue #832, opened 2018-09-27
Feature request: glob functionality for python bindings


benkrikler:
Is it reasonable to consider adding glob functionality to the python bindings?  I'd be happy to try and implement this myself, but I wanted to check how it would be received and if there were suggested approaches.   I imagine this being a new function within the FileSystem module.  Given a path containing wildcards (potentially more than one), a list of files that match this (as per fnmatch) should be returned.

Just some context on why I'd like to do this:  I've previously added such functionality into the rootpy project ([here](https://github.com/rootpy/rootpy/blob/master/rootpy/utils/ext_glob.py)), but I now wanted to get such functionality without requiring ROOT to be installed rootpy.

simonmichal:
Hi Ben,

As I said in https://github.com/xrootd/xrootd-python/issues/19 thanks for taking initiative and sorry for the late response! 

Could you prepare a PR based on https://gitlab.cern.ch/fast-hep/public/fast-curator/blob/master/fast_curator/xrootd_glob.py, I'll try to do a review once it's done ;-)

Is this a full globbing functionality or a subset?

Michal

simonmichal:
Since #854 has been merged I'm closing it.

PR #854, opened 2018-11-05
Add a first attempt at a glob file

benkrikler:
This PR adds a `glob` function to extend the standard Python glob module to work with xrootd.  It addresses issue: #832 .  I've mainly pulled in code I wrote for  package I've been developing: https://gitlab.cern.ch/fast-hep/public/fast-curator/blob/master/fast_curator/xrootd_glob.py, which itself was an edited version of the code I added [to rootpy](https://github.com/rootpy/rootpy/blob/master/rootpy/utils/ext_glob.py).  

This something of a work-in-progress, I'm conscious that I should 1) add unit tests, and 2) add doc-strings.  I need to look more into what xrootd does for testing already, and wanted to push this up early on to get feedback.


benkrikler:
Do we think this could be reviewed and merged in?

simonmichal:
@benkrikler : sorry for the delay and thanks for submitting the PR!

In general it is a very nice enhancement for our Python bindings.

However, I noticed while testing your PR that if the path contains a component that does not exist following exception is being raised:
```
Traceback (most recent call last):
  File "test.py", line 6, in <module>
    print glob.glob( 'root://slc7-test//tmp/dir3/file*' )
  File "/home/simonm/xrootd/build/bindings/python/build/lib.linux-x86_64-2.7/XRootD/glob.py", line 59, in glob
    return xrootd_glob(pathname)
  File "/home/simonm/xrootd/build/bindings/python/build/lib.linux-x86_64-2.7/XRootD/glob.py", line 80, in xrootd_glob
    for entry in dirlist.dirlist:
AttributeError: 'NoneType' object has no attribute 'dirlist'
```

same happen when I try to glob elements of directory that has not been exported on the XRootD server, e.g. for a server where only `/tmp` is exported try:

`glob.glob( 'root://hostname//*' )`

Could you improve error handling?

Thanks, 
Michal

benkrikler:
Thanks for taking a look. You're right, and in fact I've hit those issues in our own version of this, but had forgotten to fix them.  I'll try to solve this as soon as possible, and try adding some unit tests for the code in the process.

Coming back to this with fresher eyes, I'm wondering if I've located the `glob.py` module in the right place.  Should it really be below `src/` ?  Looking at the setup.py file I guess this should really be imported via the `pyxrootd` package rather than the `XrootD` one?

simonmichal:
@benkrikler : thanks for bringing this up, I would actually say it should go to bindings/python/libs/client

simonmichal:
@benkrikler : we are planning another feature release soon, if you could update this PR (as we discussed) it would be included in 4.10.0 ;-)

benkrikler:
Thanks for the reminder!  I'll push to update this quickly so we can get it in by then.

benkrikler:
I think I've put in all I wanted to (moved under `lib/client`, better unit testing, check the status of the File query).  The only thing that might be useful is a docstring or two, I think.

benkrikler:
Hi @simonmichal. Do you think this is ok to go in?  Would you like me to make any more changes?

simonmichal:
@benkrikler : sorry for the delay, I was out of office for few days, it looks all good :-) again thanks a lot :-)

PR #863, opened 2018-11-20
[POSIX] Avoid SEGV due to race condition between open() and close().

bbockelm:
This should protect against a SEGV observed under load for file caches.

This is a backport of ae1fcd1142a7570dd97473edb54555d63b5377f0, originally authored by @abh3.

Note: I'm not sure if there will be further releases in the 4.8.x series, but I wanted to make sure we at least had a PR we can point sites at if they had questions about this issue.

### DIFF
--- a/src/XrdPosix/XrdPosixCallBack.hh
+++ b/src/XrdPosix/XrdPosixCallBack.hh
@@ -59,7 +59,7 @@ virtual     ~XrdPosixCallBack() {}
 };
 
 //-----------------------------------------------------------------------------
-//! @brief An abstract class to define a callback for open file requests.
+//! @brief An abstract class to define a callback for file I/O requests.
 //!
 //! This abstract class defines the callback interface for Fsync(), Pread(),
 //! Pwrite(), and VRead(). Async I/O is not supported for Read(), Readv(),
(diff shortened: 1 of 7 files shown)


abh3:
I see what you are trying to do. My concern is the ref/unref calls for each I/O request. Do you see any way of avoiding those?

bbockelm:
@abh3 - this is your code; I'm not sure I know the answer to your question.

osschar:
And you'll also need the other set of changes that went into XrdFileCache, I think I posted both commits into your original bug report.

abh3:
Ah, I got confused because this pull request came from fix_posix_race branch on bbockelm/xrootd. This patch is already in git head and 4.9.0-RC1, Why is this pull request here as it seems to duplicate patch ae1fcd1 with a cryptic comment that it is a back port? Anyway, we can fix any performance problems by switching to use std::atomics but this is best left to when we drop support for RH6. Can we skip this?

bbockelm:
@abh3 - right.  This is a PR against the 4.8.x branch, posted for people who need a stable XCache between now and when they can switch to the 4.9.x series.  It's not meant to be merged, just here so folks can find a backport.

simonmichal:
Since 4.9.0 is out I'm closing this one.

PR #867, opened 2018-12-05
Append signed Diffie-Hellman paramaters to the unsigned

wyang007:
The server appends signed Diffie-Hellman parameters to the unsigned so that the client can verify. If the verification fails, x509 proxy delegation will be turned off. This change is backward compatible to the older clients.

### DIFF
--- a/src/XrdSecgsi/XrdSecProtocolgsi.cc
+++ b/src/XrdSecgsi/XrdSecProtocolgsi.cc
@@ -49,6 +49,7 @@
 #include "XrdOuc/XrdOucPinLoader.hh"
 #include "XrdOuc/XrdOucStream.hh"
 #include "XrdOuc/XrdOucEnv.hh"
+#include "XrdOuc/XrdOuca2x.hh"
 
 #include "XrdSut/XrdSutAux.hh"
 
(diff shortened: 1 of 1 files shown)


abh3:
Could you un-comment the message

std::cerr <<"secgsi: proxy delegation forbidden when server does not provide signed DH parameter!\n"

that indicates delegation was turned off. Without that message we will be chasing delegation problems for days. With the message we will immediately know why i didn't work. A small price to pay for a message.

abh3:
Ah, are you telling me this is a server-side issue? That is, the message 
only appears at the server? If so, I think this is a bit more complicated 
then. If it's a client-side issue then the check with turnoff is correct.

On Wed, 5 Dec 2018, Wei Yang wrote:

> I put it there and comment it out because I feel the message is quite 
misleading: In a TPC case, if the source does not send DH parameters (but 
the destination does), this message will come out, even though the client 
never intended to delegate the proxy to the source.

wyang007:
It is at the client side. XrdSecProtocolgsi::ClientDoCert() is called by client only.

--
Wei Yang  |  yangw@slac.stanford.edu<mailto:yangw@slac.stanford.edu>  |  650-926-3338(O)

xrootd-dev:
OK, then do what I did when checking if DNS was used and if delegation is 
enabled, urn it off with a message. Otherwise, stay silent.

wyang007:
Andy, I switch to use x2b and b2x, and test it. Please review and merge.

wyang007:
replaced by PR #869

PR #869, opened 2018-12-07
support signing of server DH public parameters

gganis:
Revised version of PR #867


wyang007:
Hi Gerri,
I will try this out to see how it works. One concern by just looking at the code is that in line 1777, you directly added the signed DH pub pars to the bucket. In my previous test, the bck->buffer sometimes had \0 in the middle, and the receiving end would get this wrong. So I used bytes2hex() (or later Andy's XrdOucac2::b2x()) to convert it to hex string before we send it over the wire. Is this still a concern?

wyang007:
Forgot another thing: we are asked to turn off delegation if the DH pub pars exchange is not encrypted. Can we also add this in 869?

PR #876, opened 2018-12-12
Return correct content-range header

PerilousApricot:
According to the HTTP spec, the response from a "Range" request should have a
"content-range" header in the form of "range_begin-range_end/doc_length". But
xrootd currently returns the length of the range intead of the range of the whole
file. Pass in the length of the file and additionally change the printf string since the
file size is stored in a long long


ffurano:
Absolutely agree, that was actually a bug

PR #877, opened 2018-12-12
more on memory leaks

wyang007:
restore the original using of X509_NAME_dup(X509_get_subject_name(xpi))
change the way X509_EXTENSION (stack) is freed
included Gerri Ganis's patch to clean up proxyChain.


wyang007:
Tested on both centos 7 (openssl 1.0.2k) and rhel6 (openssl 1.0.1e)

PR #879, opened 2018-12-13
xrdhttp: expose query string in external req handler

apeters1971:
I integrated XrdHttp into EOS/MGM, but I missed access to the query CGI in the external handler. Would be great to have this added to 4.9, otherwise this external handler is really sort of useless. Otherwise performance looks great!

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -70,7 +70,7 @@ verb(req->requestverb), headers(req->allheaders) {
   resource = req->resource.c_str();
   int envlen = 0;
   
-  query = req->opaque ? req->opaque->Env(envlen) : "";
+  headers["xrd-http-query"] = req->opaque?req->opaque->Env(envlen):"";
   
   // These fields usually identify the client that connected
 
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -68,7 +68,7 @@ XrdHttpExtReq::XrdHttpExtReq(XrdHttpReq *req, XrdHttpProtocol *pr): prot(pr),
 verb(req->requestverb), headers(req->allheaders) {
   // Here we fill the request summary with all the fields we can
   resource = req->resource.c_str();
-  int envlen=0;
+  int envlen = 0;
   
   query = req->opaque ? req->opaque->Env(envlen) : "";
   
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -70,7 +70,7 @@ verb(req->requestverb), headers(req->allheaders) {
   resource = req->resource.c_str();
   int envlen=0;
   
-  query = req->opaque?req->opaque->Env(envlen):"";
+  query = req->opaque ? req->opaque->Env(envlen) : "";
   
   // These fields usually identify the client that connected
 
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -25,6 +25,7 @@
 #include "XrdHttpExtHandler.hh"
 #include "XrdHttpReq.hh"
 #include "XrdHttpProtocol.hh"
+#include "XrdOuc/XrdOucEnv.hh"
 
 /// Sends a basic response. If the length is < 0 then it is calculated internally
 int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long int bodylen)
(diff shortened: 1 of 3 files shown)


src/XrdHttp/XrdHttpExtHandler.cc
```diff
@@ -67,6 +68,9 @@ XrdHttpExtReq::XrdHttpExtReq(XrdHttpReq *req, XrdHttpProtocol *pr): prot(pr),
 verb(req->requestverb), headers(req->allheaders) {
   // Here we fill the request summary with all the fields we can
   resource = req->resource.c_str();
+  int envlen=0;
+  
+  query = req->opaque?req->opaque->Env(envlen):"";
```
bbockelm:
```suggestion
  query = req->opaque ? req->opaque->Env(envlen) : "";
```

src/XrdHttp/XrdHttpExtHandler.cc
```diff
@@ -67,6 +68,9 @@ XrdHttpExtReq::XrdHttpExtReq(XrdHttpReq *req, XrdHttpProtocol *pr): prot(pr),
 verb(req->requestverb), headers(req->allheaders) {
   // Here we fill the request summary with all the fields we can
   resource = req->resource.c_str();
+  int envlen=0;
```
bbockelm:
```suggestion
  int envlen = 0;
```

bbockelm:
Forgot to mention - once I recompile plugins against the new header, things appear to work as expected.

apeters1971:
Hi Brian,
an option would be to add the query information to the resource variable. This would not break the ABI., but might break the behaviour of existing plug-ins.

bbockelm:
I'm actually quite happy with breaking the ABI -- it certainly seems to be the right thing to do.

I just want to make it a clean failure as opposed to a segfault.

Do you know the versioning system well enough to do this?  I don't but I might have some time later today.

xrootd-dev:
Hi Brian,

Take a look at XrdVersionPlugin.hh. You may also want to put the plugin 
into the "strict" list which requires it to be versioned. Also, it's not 
clear which plugin(s) you are talking about. In any case, each such plugin 
also needs to have the XrdVERSIONINFO macro defined.

bbockelm:
Hm - it is indeed listed as required there, but it doesn't seem to be functioning right.  Delving...

bbockelm:
@apeters1971 - can you try this patch:

```
--- a/src/XrdVersionPlugin.hh
+++ b/src/XrdVersionPlugin.hh
@@ -96,7 +96,7 @@
         XrdVERSIONPLUGIN_Rule(DoNotChk,  4,  0, XrdgetProtocol                )\
         XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdgetProtocolPort            )\
         XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdHttpGetSecXtractor         )\
-        XrdVERSIONPLUGIN_Rule(Required,  4,  8, XrdHttpGetExtHandler          )\
+        XrdVERSIONPLUGIN_Rule(Required,  4,  9, XrdHttpGetExtHandler          )\
         XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdSysLogPInit                )\
         XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdOssGetStorageSystem        )\
         XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdOssStatInfoInit            )\
```

ffurano:
Hi,
I am not happy with breaking the ABI, as the external handler is used in DPM prod sites. Since XrdHTTP+handler is the DPM core, they would all break.

apeters1971:
That works for me.

apeters1971:
There are only two other ways, to get the query exposed without breaking the ABI. 
1) add the query to the resource parameter
2) add the query as a fake header entry

ffurano:
Very good point Andreas.
The query can safely be added as a fake, nicely named header item in the map XrdHttpExtReq::headers

apeters1971:
Maybe the way to go is to add a fake header for 4.X while we add the query as a parameter for 5

ffurano:
I don't have strong opinions on this. Maybe you prefer to change one line of code less when passing to 5.0 ?

simonmichal:
I would refactor it for 5.0 so it uses the pimpl idiom ;-)  this way in post-5.0 you will be able to add and remove as many things as you want without breaking ABI.

bbockelm:
@ffurano - I don't understand the opposition to breaking the ABI.  In #585, I proposed making this API public in order to some sort of ABI stability for users (such as the DPM plugins) and you wrote:

> I am definitely not a philosopher of ABI compatibility, and I have put those headers with the private ones exactly to have more freedom, until the interface stabilizes, which can take quite some time.

In that case, we broke the ABI for the 4.8.0 release to have more convenient access to the `XrdSecEntity`.  So, everyone last year had to recompile their plugins to upgrade to the 4.8.0 release (just as we would have them doing here).

If we can no longer do fixes to clear oversights the API, why don't we mark it as public?

ffurano:
Hi Brian. I stay not a philosopher. The difference is that right now that
stuff reached production sites, a practical issue, so I wrote "I don't like it".
I also think that the solution proposed by Andreas solves the problem very well.

apeters1971:
The modifications are now ABI compatible I just expose a special header 'xrd-http-query' and added the version plugin patch from Brian.

ffurano:
Merged. I will also trigger an immediate round of DPM tests to give a nice pre-Christmas word

PR #883, opened 2018-12-18
OAuth2-based Macaroon acquisition

bbockelm:
This PR implements the OAuth 2.0 client credentials workflow for acquiring a macaroon.  This provides a more standards-based approach to token acquisition compared to the existing method developed by @paulmillar, which has served us well but is admittedly a bit more ad-hoc.  With this PR, both the old method and new method ought to work.

This approach was originally implemented in StoRM by @andreaceccanti.  I wanted to have a second server implementation prior to doing the client side changes for FTS.

This is NOT meant for 4.9.0.

### DIFF
--- a/src/XrdMacaroons/XrdMacaroonsHandler.cc
+++ b/src/XrdMacaroons/XrdMacaroonsHandler.cc
@@ -17,6 +17,7 @@
 
 using namespace Macaroons;
 
+
 static
 ssize_t determine_validity(const std::string& input)
 {
@@ -95,6 +96,7 @@ Handler::GenerateID(const XrdSecEntity &entity, const std::string &activities,
     return result;
 }
 
+
 std::string
 Handler::GenerateActivities(const XrdHttpExtReq & req) const
 {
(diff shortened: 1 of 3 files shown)


src/XrdMacaroons/XrdMacaroonsHandler.cc
```diff
@@ -109,19 +145,167 @@ Handler::GenerateActivities(const XrdHttpExtReq & req) const
...
-    return !strcmp(verb, "POST");
+    return !strcmp(verb, "POST") || !strncmp(path, "/.well-known/", 13) ||
+           !strncmp(path, "/.oauth2/", 9);
+}
+
+
+int Handler::ProcessOAuthConfig(XrdHttpExtReq &req) {
+    if (req.verb != "GET")
+    {
+        return req.SendSimpleResp(400, NULL, NULL, "Only GET is valid for oauth config.", 0);
```
andreaceccanti: 405 method not supported would probably be more appropriate here?

src/XrdMacaroons/XrdMacaroonsHandler.cc
```diff
@@ -109,19 +145,167 @@ Handler::GenerateActivities(const XrdHttpExtReq & req) const
...
+    json_object_put(response_obj);
+    return retval;
+}
+
+
+int Handler::ProcessTokenRequest(XrdHttpExtReq &req)
+{
+    if (req.verb != "POST")
+    {
+        return req.SendSimpleResp(400, NULL, NULL, "Only POST is valid for token request.", 0);
```
andreaceccanti: same  as above, 405 probably more approriate?

src/XrdMacaroons/XrdMacaroonsHandler.cc
```diff
@@ -109,19 +145,167 @@ Handler::GenerateActivities(const XrdHttpExtReq & req) const
...
+        //std::cout << "Found key " << key << ", value " << value << std::endl;
+        if (key == "grant_type")
+        {
+            found_grant_type = true;
+            if (value != "client_credentials")
+            {
+                return req.SendSimpleResp(400, NULL, NULL, "Invalid grant type specified.", 0);
+            }
+        }
+        else if (key == "expire_in")
```
andreaceccanti: this "expire_in" grant parameter (which I guess is used to limit the access token lifetime) is something where we want to be aligned, or FTS will have to juggle. StoRM uses the 'lifetime' parameter. No problem to change on our side, but I suggest to have a profile document that defines these details somewhere and align implementations to such profile.

src/XrdMacaroons/XrdMacaroonsHandler.cc
```diff
@@ -314,11 +505,17 @@ int Handler::ProcessReq(XrdHttpExtReq &req)
...
     }
-    json_object_object_add(response_obj, "macaroon", macaroon_obj);
+    json_object_object_add(response_obj, oauth_response ? "access_token" : "macaroon", macaroon_obj);
+
+    json_object *expire_in_obj = json_object_new_int64(validity);
+    if (!expire_in_obj)
+    {
+        return req.SendSimpleResp(500, NULL, NULL, "Unable to create a new JSON validity object.", 0);
+    }
+    json_object_object_add(response_obj, "expire_in", expire_in_obj);
```
andreaceccanti: this should be "expires_in" according to the OAuth standard

src/XrdMacaroons/XrdMacaroonsHandler.cc
bbockelm: Yes - this was a bit arbitrary on my part.  I didn't know of an "obvious" candidate here.

Any reason that `expires_in` versus `expire_in` versus `lifetime` might be preferred?  Shall we start flipping coins?

PR #885, opened 2018-12-19
Enable use of an unique IV in enc/dec cipher operations

gganis:
Enable use of a unique IV randomly generated instead of a zeroed one.


abh3:
Looks fine to me.

PR #888, opened 2018-12-30
Report the correct protocol name to the bridge.

bbockelm:
Tell the bridge whether the connection came in using HTTPS or HTTP; this way, monitoring packets can differentiate the two protocol types.


apeters1971:
Ah, I will check ... I sort of 'missed' that. I thought this works only, if the client sends his identity but obviously it does exactly what I want not matter if the client authenticates or not ...

PR #893, opened 2019-01-10
XrdHttp: allow to use StartSimpleResp and SendData without breaking ABI

apeters1971:
This is really sort of ugly, but there is no better way without breaking the ABI and if one cannot return data in portions, I cannot return files larger then few MB. I use two non existing HTTP codes to do this (0 = StartResponse 1=SendData) . If you have a better idea, let me know.

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -31,7 +31,20 @@
 int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long int bodylen)
 {
   if (!prot) return -1;
-  
+
+  // @FIXME
+  // - need this to circumvent missing API calls and keep ABI compatibility
+  // - when large files are returned we cannot return them in a single buffer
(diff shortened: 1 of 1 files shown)


bbockelm:
Do the clients not accept chunked responses, or do they not accept chunked responses where the size isn’t known?  Can you provide the correct Content-Length header and work around it?

What’s the CERNbox client based on ?

ffurano:
I would modify StartChunkedResp so that it avoids adding unconditionally the header line
  ss << "Transfer-Encoding: chunked";

and then simply use it by passing a Content-Length header. IMO it should work, what do you think?

apeters1971:
If you put content-length this includes also the chunked encoding tags \r\n\ + size headers ... that does not work. If we remove Transfer-Encodign: chunked in StartChunkedResps, the sender function still encodes chunks, which breaks the specification.

bbockelm:
Gotcha, thanks!

One last question - for future reference, what's the CERNbox client?  Seems strange that a HTTP client doesn't support chunked encoding.

ffurano:
Well, I think that every other workaround we imagined is not less ugly than this one
The hope is to turn this into a proper addition to the API for Xrootd 5

PR #896, opened 2019-01-14
XrdHttp: add the used protocol (https/http) as a special header i

apeters1971:
There is only a 'getter' function added to the header file and I am not sure, if this breaks ABI compatibility of the shared library. This could be postponed to version 5. It is important to know, if a call comes via a 'secure' connection or not in the plug-in.

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -84,6 +84,7 @@ verb(req->requestverb), headers(req->allheaders) {
   int envlen = 0;
   
   headers["xrd-http-query"] = req->opaque?req->opaque->Env(envlen):"";
+  headers["xrd-http-prot"] = prot->isHTTPS()?"https":"http";
   
   // These fields usually identify the client that connected
 
--- a/src/XrdHttp/XrdHttpProtocol.hh
+++ b/src/XrdHttp/XrdHttpProtocol.hh
@@ -124,6 +124,9 @@ public:
   /// Authentication area
   XrdSecEntity SecEntity;
 
+  /// called via https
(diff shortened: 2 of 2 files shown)


bbockelm:
@apeters1971 - after https://github.com/xrootd/xrootd/pull/888, I think this is correctly set in the `XrdSecEntity` object, no?

abh3:
Well, yes and no. The actual protocol is not recorded in SecEntity as SecEntity is generic and doesn't care what the communications protocol is. However, starting in R5 you can ask the SecEntity whether or not  the communications channel is secure (i.e. encrypted). So, from that standpoint if the answer is "yes" then in an HTTP context you can infer that HTTPS is being used (in an XRootD context you can infer that xroots  is being used). Does, that make this change unnecessary then?

bbockelm:
I'm a bit confused - is the protocol reported via monitoring different than the `prot` field of the `XrdSecEntity` structure?

abh3:
Yes, the “prot” field in the SecEntity structure is the *authentication* protocol used to fill out the structure. The communications protocol reported in monitoring comes from other information.

apeters1971:
Ah no, then please keep the merge request.

bbockelm:
Yup, please do!  This has been on my TODO list as I want to block the use of tokens over HTTP.

abh3:
Well, I don’t understand why this won’t work for you. If you see tokens then you ask whether they were sent using TLS. If so, accept them; if not, reject them. I don’t see any value in knowing specifically the HTTPS was used as it’s no different then asking is TLS was used. So, Andreas and Brian, tell me what’s missing here?

Andy

bbockelm:
Right - that's what I said.  This approach would work for me.

PR #901, opened 2019-01-31
add missing newlines to end of cfg files (SOFTWARE-3485)

edquist:


bbockelm:
For what it's worth - we noticed this missing EOL as our tests concatenate configuration directives to the configuration file.  This resulted in an invalid configuration line.

(That said, the OSG tests need to be updated regardless because configuration isn't parsed after the continue statement...)

PR #905, opened 2019-02-14
Fix 'deamon' typos

matyasselmeci:


abh3:
Well, I merged it (thanks) but it may start a firestorm as the urban dictionary (not always to be trusted) has a definition for deamon that could work as well! In fact, for tekkies it seems either spelling is acceptable.

PR #906, opened 2019-02-14
Handle clients that request multiple checksums.

bbockelm:
This takes the first checksum type that Xrootd recognizes and only calculates that.

It improves RFC compliance by allowing Xrootd to properly parse multiple checksum requests in a single `Want-Digest` header.

Before,

```
Want-Digest: md5,adler32
```

would result in Xrootd trying to calculate a checksum named `md5,adler32`.  After, given `md5` is a known checksum type, it will request a checksum for `md5`.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -54,6 +54,7 @@
 #include <functional> 
 #include <cctype>
 #include <locale>
+#include <string>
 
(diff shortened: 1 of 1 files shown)


bbockelm:
@ffurano @abh3 - please take a look at this one.  It's a simple bugfix affecting some of the DOMA tests and an excellent candidate for 4.9.1.

Issue #909, opened 2019-02-18
RFE: downgrade "secgsi_InitProxy: cannot access private key file:" to Debug


jmuf:
When trying several possible authentication methods, only one of them prints something:
```
/tmp$ xrdcp  root://eosproject-i00:1094//proc /tmp/foo
190218 14:42:24 17779 secgsi_InitProxy: cannot access private key file: /eos/user/A/ABC/.globus/userkey.pem
```
This has confused some of our users in the past - if the message precedes some other error, they assume that the two are linked.

I would suggest to downgrade this message to "Debug" level, perhaps unless the user has explicitly specified GSI to be be used (via "XrdSecPROTOCOL=gsi"). This would be consistent with how the other auth methods are being tried - these fail silently (i.e only visible at "Debug" level):
```
[2019-02-18 14:42:12.136860 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Authentication is required: &P=krb5,xrootd/eosproject.cern.ch@CERN.CH&P=gsi,v:10300,c:ssl,ca:5168735f.0|4339b4bc.0&P=unix&P=sss,0.13:/etc/eos.keytab
[2019-02-18 14:42:12.136871 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Sending authentication data
[2019-02-18 14:42:12.138679 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Trying to authenticate using krb5
[2019-02-18 14:42:12.139140 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Cannot get credentials for protocol krb5: Seckrb5: No or invalid credentials; No credentials cache found (p=xrootd/eosproject.cern.ch@CERN.CH).
[2019-02-18 14:42:12.145168 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Trying to authenticate using gsi
190218 14:42:12 17762 secgsi_InitProxy: cannot access private key file: /eos/user/A/ABC/.globus/userkey.pem
[2019-02-18 14:42:12.160033 +0100][Debug  ][XRootDTransport   ] [eosproject-i00:1094 #0.0] Cannot get credentials for protocol gsi: Secgsi: ErrParseBuffer: error getting user proxies: kXGS_init
[...]

PR #923, opened 2019-03-06
Debug fixes for secgsi

gganis:
This PR contains mostly fixes to improve debugging (including the change requested by issue #909) and a fix affecting the option to save the delegated proxy to a file (which is not used by TPC, though).

It would probably make sense to port this to stable-4.9.x and include it in 4.9.1 .


gganis:
Basically, without the change in commit 7fe4dce there is not way to save the proxy to a file. This is a bug, not affecting TPC because you use the memory save option in there.
The rest of the changes improve the debug dumps and cleanup the exchanged buffers from buckets not needed.

abh3:
OK, then I have no problems with the patch. I will merge it for 4.9.1.

Andy

PR #931, opened 2019-03-15
XrdHTTP: Support chunked encoding for PUT

### LINKED ISSUE #915: XrdHTTP PUT does not complain if content-length undefined
Consider the following "conversation":
```
    => PUT /somewhere/test_webdav_access_9 HTTP/1.1
    => User-Agent: PycURL/7.19.7
    => Host: xrootd006.example.com:1094
    => Accept: */*
    => Transfer-Encoding: chunked
    => Expect: 100-continue
    => 
    <= HTTP/1.1 100 Unknown
    <= Connection: Keep-Alive
    <= Content-Length: 0
    => 79
    => This is a test file generated by the nagios probe check-webdav from package nagios-plugins-webdav. It is safe to delete.
    => 
    => 0
    => 
    <= HTTP/1.1 200 OK
    <= Connection: Keep-Alive
    <= Content-Length: 3
    <= 
    <= :-)
```
The end result is an empty file `/somewhere/test_webdav_access_9` on the storage. 

According to RFC 2616 in (4.4 Message Length):
```
   If a request contains a message-body and a Content-Length is not given,
   the server SHOULD respond with 400 (bad request) if it cannot
   determine the length of the message, or with 411 (length required) if
   it wishes to insist on receiving a valid Content-Length.

   All HTTP/1.1 applications that receive entities MUST accept the
   "chunked" transfer-coding (section 3.6), thus allowing this mechanism
   to be used for messages when the message length cannot be determined
   in advance.
```
So for HTTP 1.1 compatibility, XRootD must accept the content and determine the size on the fly. 

This test was done with XRootD 4.9.0.

bbockelm:
This adds support for clients who would like to PUT files via chunked encoding.

Fixes #915

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -927,17 +927,17 @@ int XrdHttpReq::ProcessHTTPReq() {
   
   
   // Verify if we have an external handler for this request
+  if (reqstate == 0) {
+    XrdHttpExtHandler *exthandler = prot->FindMatchingExtHandler(*this);
+    if (exthandler) {
+      XrdHttpExtReq xreq(this, prot);
+      int r = exthandler->ProcessReq(xreq);
+      reset();
+      if (!r) return 1; // All went fine, response sent
+      if (r < 0) return -1; // There was a hard error... close the connection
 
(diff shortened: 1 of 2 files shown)


ffurano:
This is a complex one, looks sane to me so far

PR #947, opened 2019-03-27
[XCache] Fix memory leak

osschar:
Fix memory leak - block was not freed when writing to disk failed.


osschar:
Please don't merge this yet ... I might have spotted another issue!

xrootd-dev:
Good that you told me before my procrastinaton wore off :-)

osschar:
OK, this is ready for merge now!

PR #970, opened 2019-04-18
Include provided opaque information in the redirection

bbockelm:
If opaque information is given by the user as part of the URL (i.e., `GET /foo?bar=1`), then include this opaque information in the redirection string.

Previously, opaque information was dropped and only the resource was used.  This is problematic in cases where the opaque information contains important authorization details (think: signed URLs).

Additionally, the TPC handler is updated to perform correctly in this situation.

With this PR, if `http.header2cgi Authorization authz` is set in the configuration, the token will be moved from the headers to the opaque information _and_ now included in the redirect.  Thus, clients which drop the `Authorization:` header upon redirect (this includes the latest versions of `curl`!) will now work by default with an Xrootd cluster.

Fixes behavior observed in the WLCG DOMA testing.

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -84,6 +84,7 @@ verb(req->requestverb), headers(req->allheaders) {
   int envlen = 0;
   
   headers["xrd-http-query"] = req->opaque?req->opaque->Env(envlen):"";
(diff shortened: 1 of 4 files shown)


ffurano:
Here I merged it, as the usual test runs went fine. I just have some concerns over how many people may remember to add "http.header2cgi Authorization authz" to their configuration. At the same time making this the default may be excessive.

Issue #976, opened 2019-05-07
[XrdSecgsi] Client is reusing cached proxy although xrd.gsiusrpxy cgi points to a new one


simonmichal:
Here's a simple reproducer:
```
  FileSystem fs1( std::string( "root://user1@hostname//tmp?xrd.gsiusrpxy=/tmp/user1proxy" ) );
  DirectoryList *rsp1 = 0;
  XRootDStatus st1 = fs1.DirList( "/tmp", DirListFlags::None, rsp1 );
  std::cout << st1.ToString() << std::endl;

  FileSystem fs2( std::string( "root://user2@hostname//tmp?xrd.gsiusrpxy=/tmp/user2proxy" ) );
  DirectoryList *rsp2 = 0;
  XRootDStatus st2 = fs2.DirList( "/tmp", DirListFlags::None, rsp2 );
  std::cout << st2.ToString() << std::endl;
```
In this scenario, a new connection is created for `user1` and is authenticated with `/tmp/user1proxy` credential. Afterwards, a new connection is being created for `user2` and is being authenticated with the cached credential (again `/tmp/user1proxy`).

This is easy to spot if one uses expired proxy as the second credential, in this case the second
dirls should fail, but it never does.

chaen:
I confirm the fix works

simonmichal:
@chaen: Perfect! Thank you very much for testing!

PR #980, opened 2019-05-08
[XrdSecgsi] Make proxy cache path aware.

simonmichal:
This aims at fixing #976.

@gganis : I think we can simply use `UsrProxy` as the tag used to query proxy cache. Alternatively, we could hash the proxy location and create a tag like `Proxy: + hash` (alternative version in the comments). Please let me know which one suits us better!

Michal


gganis:
@simonmichal Looks good. I agree that using UsrProxy should do the job. And is only used there, so there is no risk of using different tags. I think you can go ahead and test it in the release candidate.
Thanks!

simonmichal:
Thanks, I cleaned up the PR and am merging it!

Issue #663, opened 2018-03-02
PSS with disk caching enabled produces bad files on the local cache when attempting writes


mpatrascoiu:
Hello,

I understand that writing is not supported in the case of a disk-caching Proxy.

However, when attempting to write a file to the cluster through the proxy (with PFC enabled), the operation will not succeed (operation not supported) but the file is written to the local disk cache and appears to be in bad state.

Future reads for the same file will be provided from the cache, serving the corrupted file.

This can also cause confusion because although the write operation is not supported, later on, the cache can still provide that file (even if in a bad state).

Additional info
===========
Setup: client --> proxy (with PFC) --> cluster
I am copying the _hello.txt_ to the proxy and then reading it into _hello_copied.txt_.

Checksums:
bea8252ff4e80f41719ea13cdf007273  hello.txt
36df9540a5ef4996a9737657e4a8929c  hello_copied.txt

File contents in hexa (ran through the xxd command): 
hello.txt
0000000: 4865 6c6c 6f2c 2057 6f72 6c64 210a       Hello, World!.
hello_copied.txt
0000000: 0000 0000 0000 0000 0000 0000 0000       ..............

Below are the logs from the MGM, proxy and client, as well as the proxy config.

[proxy_write-client.log](https://github.com/xrootd/xrootd/files/1775615/proxy_write-client.log)
[proxy_write-mgm.log](https://github.com/xrootd/xrootd/files/1775616/proxy_write-mgm.log)
[proxy_write-proxy.log](https://github.com/xrootd/xrootd/files/1775617/proxy_write-proxy.log)
[xrd.cf.pss.txt](https://github.com/xrootd/xrootd/files/1775619/xrd.cf.pss.txt)

alja:
@mpatrascoiu  Thanks for the detail report. I will look into it.

alja:
@mpatrascoiu I can reproduce the problem. 
File written in the origin is empty, but it is not corrupted.

It would good if writing option would be visible at the open stage, so the file would be rejected before creating cache IO object.
Thread 4 "xrootd" hit Breakpoint 1, XrdFileCache::Cache::Attach (this=0x65e880, io=0x7fffe8002880, Options=4)
    at /home/alja/xrd/master/xrootd/src/XrdFileCache/XrdFileCache.cc:162
162        if (Cache::GetInstance().Decide(io))
(gdb) bt
[...]

PR #987, opened 2019-05-13
Deny write access to a file in initial Prepare() call.

osschar:
This addresses #663.

Note that Write() calls in IO object were already failing with errno = ENOTSUP.

Write access into cache is not supported. If one really wants to open files with write access using the cache as the entry point, they should setup static redirect on write to a suitable location / service.


osschar:
@abh can you please review this change.

PR #1001, opened 2019-06-14
Fix digest names returned to clients.

bbockelm:
Digest name issues reported by @olifre.

@simonmichal - Should be included in 4.10, if possible.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -67,6 +67,32 @@
 #define TRACELINK prot->Link
 
 
+static std::string convert_digest_rfc_name(const std::string &rfc_name_multiple)
+{
+  std::stringstream rfc_name_multiple_ss;
+  rfc_name_multiple_ss << rfc_name_multiple;
+  for (std::string rfc_name; std::getline(rfc_name_multiple_ss, rfc_name, ','); ) {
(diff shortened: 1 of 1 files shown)


simonmichal:
@bbockelm : sure, however Fabritzio is on DPM workshop so we will have to wait until Monday with the next RC

olifre:
@simonmichal @bbockelm Thanks! 
Once it is in an RC, I can also roll it out to UNI-BONN endpoint and see how it fares in interoperability with dCache. I'll be on holidays next week, so no promises if I'll get to it already next week (but latest in two weeks from now).

olifre:
I can confirm with 4.10.0-rc3, which contains this PR, the observed issues are fixed (which showed up mainly in XRootD talking with dCache). 
I'm still unsure whether #1002 would not have been the "better" (or rather, cleaner) implementation - @bbockelm , what's your take on this?

bbockelm:
@olifre - #1002 is probably cleaner code, but this gets us the same place.  The bigger picture is the algorithm selection code needs to be rewritten anyway: we currently take the first supported algorithm on the list while we really should respect the client-provided priorities.

olifre:
> he bigger picture is the algorithm selection code needs to be rewritten anyway: we currently take the first supported algorithm on the list while we really should respect the client-provided priorities.

Ok, I see - thanks for the explanation!

PR #1002, opened 2019-06-14
Fix two problems with RFC-3230 handling in xrootd:

paulmillar:
First, the Digest response header copies the supplied Want-Digest
request header value.  A request with:

    Want-Digest: MD5,ADLER32

would yield a response like:

    Digest: MD5,ADLER32=<md5-value>

where the correct response would be:

    Digest: MD5=<md5-value>

Second, the method that calculates whether the response should be
encoded in hexadecimal or base64 expects an algorithm name, but is given
the Want-Digest request header value.  If the client makes a request with:

    Want-Digest: MD5, ADLER32

then the response is incorrectly encoded using hexadecimal.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -67,7 +67,7 @@
 #define TRACELINK prot->Link
 
 
-static XrdOucString convert_digest_name(const std::string &rfc_name_multiple)
+static std::string convert_digest_name(const std::string &rfc_name_multiple)
 {
   std::stringstream rfc_name_multiple_ss;
   rfc_name_multiple_ss << rfc_name_multiple;
@@ -989,12 +989,13 @@ int XrdHttpReq::ProcessHTTPReq() {
         const char *opaque = strchr(resourceplusopaque.c_str(), '?');
         // Note that doChksum requires that the memory stays alive until the callback is invoked.
         m_resource_with_digest = resourceplusopaque;
+        m_selected_digest = convert_digest_name(m_req_digest);
         if (!opaque) {
           m_resource_with_digest += "?cks.type=";
-          m_resource_with_digest += convert_digest_name(m_req_digest);
+          m_resource_with_digest += m_selected_digest.c_str();
         } else {
           m_resource_with_digest += "&cks.type=";
-          m_resource_with_digest += convert_digest_name(m_req_digest);
+          m_resource_with_digest += m_selected_digest.c_str();
         }
         if (prot->doChksum(m_resource_with_digest) < 0) {
           // In this case, the Want-Digest header was set and PostProcess gave the go-ahead to do a checksum.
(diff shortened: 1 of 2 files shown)


ffurano:
Hi Paul,

 Thank you very much for raising this issue and proposing a solution, what you say is correct, and the handling of the digest names was not perfect.

 I see that your proposal conflicts with the previous PR, and I have the impression that that same PR fixes the issues that you mention here. What do you think ?

https://github.com/xrootd/xrootd/commit/943a470c317bc8a1740fd19f0cd29d207d4d5677

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1736,7 +1739,7 @@ XrdHttpReq::PostProcessChecksum(std::string &digest_header) {
     }
 
     digest_header = "Digest: ";
-    digest_header += m_req_digest;
+    digest_header += m_selected_digest;
```
paulmillar: digest_header += convert_digest_rfc_name(m_req_digest);

paulmillar:
I see you've already pulled Brian's pull request: #1001.

In essence, Brian's patch and mine address the same issue.  This explains the merge-conflict.

As it happens, I developed my patch independently from Brian -- I only learn of his pull-request when I posted my pull-request to the email discussion about this issue.

Since you asked my opinion, here it is!

For me, there are some stylistic issues with Brian's patch.

- The patch violates the DRY principle (Don't Repeat Yourself), since the new method (`convert_digest_rfc_name`) is just the composition of `convert_digest_name` and `convert_xrootd_to_rfc_name` with a minor optimisation.

- The patch results in the input string being parsed multiple times, including the if-else-if chain using the (relatively) expensive strcasecmp (once to discover which algorithm to query xrootd; once again to discover which RFC-3230 algorithm name to include in response).  This is, perhaps, not terrible, but it's unnecessary.

- It "smells" like premature optimisation:  given `convert_digest_rfc_name` is just a composition, why not just apply `convert_digest_name` and `convert_xrootd_to_rfc_name` to make this obvious?  Given the method will be delayed by calculating the checksum value, the additional overhead will likely be insignificant.

- It makes it harder to fix the checksum priority ordering (the "p" values where the client says which algorithm it prefers), since the p-values would now need to be parsed in two functions (`convert_digest_rfc_name` and `convert_digest_name`).  This is a consequence of DRY violation.

- It also fails to close a potential cause of bugs in the future, since it continues to assume the returned checksum is the first algorithm the client requested.  A future change could modify which checksum is returned (e.g., prefer returning a cached checksum, if one of the requested checksums is already known).  As the returned checksum now comes from (re-)parsing the request header, updating the code to support checksum caching would require the author to be aware how the returned checksum name is calculated and modify it.

That all said, I believe his patch works.

ffurano:
Well, I see your points, my merging of Brian's PR first was just due to the ordering, I had not seen yours yet.

ioggstream:
@paulmillar we'll probably  remove MD5 from the new digest specs https://github.com/httpwg/http-extensions/issues/867 do you have any feedback?

paulmillar:
@ioggstream  Thanks for getting in touch. Sorry, I've only just become aware of httpwg/http-extensions#867.   As a quick comment: I think it depends on the context.

Our particular use-case involves data integrity verification, not against malicious changes but from accidental changes (often due to hardware malfunctioning).  The choice of ADLER32 largely stems from historical use (within the particle physics community).  In that sense, the ability to "fake" data is (currently) of lesser importance than the speed of the algorithm.

We are using RFC 3230 to provide end-to-end data integrity verification when transferring data between different sites throughout the world in a process called HTTP Third-Party Copy: where a client initiates the transfer, but the data is transferred directly between sites using a regular HTTP GET/PUT request.  This is currently an experimental protocol, with the bulk of data currently transferred using a protocol called GridFTP.

ioggstream:
@paulmillar the issue about ADLER32 is here https://github.com/httpwg/http-extensions/issues/828 feel free to contribute, so we can tune the proposal. 

The CRC32c algorithm is still available though, so in case you may pick that one.

PR #1007, opened 2019-06-19
Fix macaroon authorization config

bbockelm:
This PR fixes the macaroon configuration when using macros.  Without it, if the `all.sitename` line contained a macro substitution, then it would not be expanded in some circumstances, leading to failed authorization.


bbockelm:
@ffurano @simonmichal - Can you please look at this?

I've now had two sites bitten by the problem...

ffurano:
Hi Brian,
not that I am practical with the code of XrdMacaroons... yet we can have a look at how it goes if it works fine for you
Thanks
Fabriizo

PR #1012, opened 2019-06-25
[XrdHttp] Fix one-byte overrun in Tobase64()

jthiltges:
The allocated target buffer for base64 output is sometimes too short, leading to an overrun. (And I suspect it's causing memory corruption with eventual misbehavior and segfaults.)

A hex string may be the same length as the base64 representation. To allow for a terminating null on the base64 string, we need to allocate an additional byte.
```
string abcd
   hex 61626364
base64 YWJjZA==
```

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -1754,7 +1754,7 @@ XrdHttpReq::PostProcessChecksum(std::string &digest_header) {
         free(digest_binary_value);
         return -1;
       }
-      char *digest_base64_value = (char *)malloc(digest_length);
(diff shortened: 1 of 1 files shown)


jthiltges:
Hi @ffurano. We're working on fixing some xrootd stability and behavior issues at Nebraska, and an ASAN build uncovered this buffer overrun. It would be great to see it in the upcoming v4.10 release if there's time.

Thanks!

zvada:
Ideally, this would be good to get into `v4.10.0 RC5` that we can proceed with tests in larger pool of hosts. Or, directly to the release unless there are any plans for RC5. Do you think we can we make it, @simonmichal and @ffurano ?

PR #1014, opened 2019-07-01
[XCache] merge development branch osschar/pfc-dirstats with per directory stats collection

osschar:


osschar:
The failing build is clang due to an error introduced before this PR (see Travis logs).
Merging.

PR #1022, opened 2019-07-10
Added XrdCmsRedirLocal as ofs.cmslib plugin. Allows redirector to red…

pkramp:
…irect clients to local filesystem, then using the XrdClLocalFileHandler. Builds on the LocalRedirect changes introduced in 4.8.
Developed for using XRootD on top of shared filesystem Lustre, but can be used for other HPC centers.

PR to add the plug-in as discussed with @abh3
Readme in header as discussed.

I used the same license header we also used for the XrdClLocalFileHandler, I hope thats okay.
Cheers,
Paul


abh3:
Sorry for taking so long to review this. There was a problematic use of the kXR_vernnn flag. This flag is meant to describe protocol versions not client abilities. So, to speed things along I introduced a mechanism to indicate that the client supports local files and thus 'file://" urls.

Ostensibly, one should check XrdOucErrInfo::ucap for XrdOucErrInfo::uLclF and, if set, local files are supported. If XrdOucErrInfo::uUrlOK is also set then you can redirect the client to a local file. This is all in commit 684411 which is in git head.

So, please remove the changes to XProtocol.hh and XrdClXRootDTransport.cc, and change your plugin code to test these flags.

pkramp:
Hey Andy,

I reverted what you mentioned and am now using the new client capabilities you added.

Cheers,
Paul

xrootd-dev:
Hi Paul,

Great. I’ll merge this as soon as I can, definitely within a day or two.

Andy

PR #1023, opened 2019-07-10
[XrdTpc] Add curl handle cleanup on redirections or errors

jthiltges:
We were seeing growing memory usage from curl_easy_init() calls in ProcessPullReq() and ProcessPushReq() on a local redirector.

This patch adds curl_easy_cleanup() calls to a few cases where it was needed, especially before RedirectTransfer().


bbockelm:
@ffurano - please review.

PR #1024, opened 2019-07-10
Fix redirection handling

bbockelm:
There were a few issues with redirection handling:
- Opaque data in the original URL was being added twice; once quoted and once unquoted.
- The second copy of opaque data duplicated the `?` character
- The `resourceplusopaque` member did not quote arguments in all cases.


bbockelm:
@jthiltges - could you patch this one into the redirector at Nebraska?

ffurano:
Michal, wait a bit please, the testbeds are crunching

PR #1035, opened 2019-08-07
Use common filesystem loader

### LINKED ISSUE #1034: libXrdHttpTPC does not support fallback for the fslib directive
The code in XrdTpcConfigure uses its own mechanism to load a custom OFS plugin library but it doesn't handle the fallback scenario when the name doesn't contain "-4".

Code is here:
https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcConfigure.cc#L163

Sample error:
```
Plugin loaded 
190802 16:37:33 30238 sysInitialize: Will load configuration for the TPC handler from /etc/xrd.cf.fst1
190802 16:37:33 30238 sysConfig: xrootd.fslib line successfully processed by TPC handler.  Base library: libXrdEosFst.so
=====> xrootd.fslib -2 libXrdEosFst.so
190802 16:37:33 30238 sysConfig: Failed to base plugin  libXrdEosFst-4.so libXrdEosFst-4.so: cannot open shared object file: No such file or directory
190802 16:37:33 30238 sysInitialize: Encountered a runtime failure when loading  Failed to configure the HTTP third-party-copy handler.
``

bbockelm:
When the TPC library was maintained out-of-tree, we had to stick to the public interfaces.  This forced the filesystem loader code to duplicate a lot of logic (not necessarily 100% correct).

With this change, the plugin will make a call directly into the internal function that does the SFS loading elsewhere in Xrootd.

@abh3 - if you are OK with the approach and implementation, there's another place where we can remove code in `XrdMacaroons`.

Fixes #1034

### DIFF
--- a/src/XrdTpc/XrdTpcConfigure.cc
+++ b/src/XrdTpc/XrdTpcConfigure.cc
@@ -8,6 +8,13 @@
 #include "XrdOuc/XrdOucPinPath.hh"
 #include "XrdSfs/XrdSfsInterface.hh"
 
+
+extern XrdSfsFileSystem * XrdXrootdloadFileSystem(XrdSysError *,
+                                                  XrdSfsFileSystem *,
+                                                  const char *, int,
+                                                   const char *, XrdOucEnv *);
+
+
 extern XrdSfsFileSystem *XrdSfsGetDefaultFileSystem(XrdSfsFileSystem *native_fs,
                                                     XrdSysLogger     *lp,
                                                     const char       *configfn,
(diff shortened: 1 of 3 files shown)


abh3:
Hi Brian,

I am generally OK with that though we should look to see if we can use the 
SFS interface that teh xrootd protocol loads as opposed to loading it 
twice. The bigger question is why are you using the SFS interface and not 
the oss interface?

Andy

bbockelm:
> can [we] use the SFS interface that the xrootd protocol loads

I suspect we could - is it obvious how to "pull that through" to the plugin?

> The bigger question is why are you using the SFS interface and not the oss interface?

This code also needs to work on redirectors.

abh3:
Ah, I see. We could if we could gaurantee that xrootd would always be 
loaded first. Unfortunately, people can subvert that by command line 
overrides so it's not a sure thing. Mind you, you will potentially be 
loading the SFS without any user specified wrappers.I assume you simply ar 
ere looking for the "xrootd.fslib" directive to load the correct thing, 
right?

Andy

bbockelm:
> I assume you simply are looking for the "xrootd.fslib" directive to load the correct thing, right?

Correct!

abh3:
I see that this pull request is still in draft form and cannot be merged. Do you have an estimate of when this will be ready?

bbockelm:
Oh!  I see that I missed Andy's last comment.  Marking as ready-to-review.

abh3:
Not quite ready, as there seem to be conflicts now. In any case, we likely need to re-review this in the context of stack-able plugins that appears in R5.

abh3:
Can we close this pull request as it doesn't seem applicable after all the other changes around this issue.

abh3:
Can we close this pull request. It has merge conflicts and I think this was already fixed in another way, right?

abh3:
Is this pull request even relevant now?

bbockelm:
I think we can close this right  now - can always come back as needed.

PR #1115, opened 2020-01-15
Initial code to internally proxy security credentials with a macaroon.

bbockelm:
This causes HTTP requests coming in using an X509 proxy to generate a new macaroon used internally.  Since the macaroon serializes the security session, this avoids problems to forward authorization if this is a gateway.

### DIFF
--- a/src/XrdHttp.cmake
+++ b/src/XrdHttp.cmake
@@ -17,12 +17,18 @@ if( BUILD_HTTP )
   # The XrdHttp library
   #-----------------------------------------------------------------------------
   include_directories( ${OPENSSL_INCLUDE_DIR} )
-  
+
+if( BUILD_MACAROONS )
+  set( MACAROON_FILES XrdMacaroons/XrdMacaroonsAuthz.cc XrdMacaroons/XrdMacaroonsAuthz.hh
+    XrdMacaroons/XrdMacaroonsConfigure.cc )
+endif()
+
   # Note this is marked as a shared library as XrdHttp plugins are expected to
   # link against this for the XrdHttpExt class implementations.
   add_library(
     ${LIB_XRD_HTTP_UTILS}
     SHARED
+    ${MACAROON_FILES}
     XrdHttp/XrdHttpProtocol.cc    XrdHttp/XrdHttpProtocol.hh
     XrdHttp/XrdHttpReq.cc         XrdHttp/XrdHttpReq.hh
                                   XrdHttp/XrdHttpSecXtractor.hh
@@ -36,12 +42,18 @@ if( BUILD_HTTP )
     MODULE
     XrdHttp/XrdHttpModule.cc )
 
+if( BUILD_MACAROONS )
+  include_directories(${MACAROONS_INCLUDES} ${UUID_INCLUDE_DIRS})
+endif()
+
   target_link_libraries(
     ${LIB_XRD_HTTP_UTILS}
     XrdServer
     XrdUtils
     ${CMAKE_DL_LIBS}
     pthread
+    ${MACAROONS_LIB}
+    ${UUID_LIBRARIES}
     ${OPENSSL_LIBRARIES}
     ${OPENSSL_CRYPTO_LIBRARY} )
 
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -43,6 +43,11 @@
 #include "XrdHttpSecXtractor.hh"
 #include "XrdHttpExtHandler.hh"
 
+// Includes only needed for macaroon signing.
+#include "XrdMacaroons/XrdMacaroonsAuthz.hh"
+#include "XrdOuc/XrdOucPinPath.hh"
+#include <dlfcn.h>
+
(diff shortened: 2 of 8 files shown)


bbockelm:
@ffurano - this is the code we had been discussing today: it should allow an EOS gateway to forward authorizations along as if the HTTP request came in with a token.

With this, the "HTTPS smoke tests" work when run in gateway mode!

simonmichal:
@bbockelm : you are linking now `LIB_XRD_HTTP_UTILS` against `MACAROONS_LIB` :

https://github.com/xrootd/xrootd/pull/1115/files#diff-db7a029d325861c5a1b6bacccc3f20edR55

but there is now guard checking if `MACAROONS_LIB` has been found in the first place, as a result if the HTTP component is being built and the macaroons were not found the build will fail, e.g. see:

https://travis-ci.org/xrootd/xrootd/jobs/637440081?utm_medium=notification&utm_source=github_status

this needs fixing before we can merge ;-)

djw8605:
If you just want to make the linking to the Macaroons library condition on `BUILD_MACAROONS`, I attached a patch here that should do that.

[0001-Conditional-link-Macaroons-lib.txt](https://github.com/xrootd/xrootd/files/4135879/0001-Conditional-link-Macaroons-lib.txt)
(it's a `git am` compatible patch file)

simonmichal:
The question is: after this PR gets merged will XrdHttp be still functional if libmacaroons are missing? this is to be answered by @ffurano and @bbockelm and possibly @abh3. 

(we are building for platforms were libmacaroons are not available, though I don't think anyone is using those platforms to run XrdHttp)

bbockelm:
Hi @simonmichal -

(Sorry for the delayed response - travel last week really knocked me out of commission)

Cleaning up the various build issues and ensuring things are conditionally compiled only on platforms with `libmacaroons` present are relatively easy, about an evening's work (in addition to the items @simonmichal pointed out, there are going to be a few `#ifdef`s in the C++ code.

However, I'm still waiting to hear back from @ffurano on whether this new feature is the way he wants to go with the code.  When we originally talked about this at CERN, he was reluctant about the whole approach given that he and Andy had previously discussed a different path based on SSI security instead.

If folks are willing to merge this, I can do the remaining untangling of the build...

Brian

ffurano:
Hi,

I think that your comment tells it all:

 "useful when an internal filesystem (such as XrdPss) does not have the ability to forward an XrdSecEntity but might understand macaroons."

The approach is solid, it's a pity that the internal theoretically orthodox approach (i.e. through XrdSec) has to be worked around in such ways, and I totally see the point for this kind of things.

I can merge it, will you please fix the build errors Brian?

abh3:
Well, that seems to backtrack on everything that we agreed on and that was 
not to branch at the security level as that would essentially create two 
copmplete and different services that would likely never converge. A total 
mess if you ask me. We also agreed, well thought we did, that we would use 
the SSS approach to carry through the the credentials as this was 
uniformaly the way it was expected to be done. Now I see that is off the 
table. Does that mean any more development for SSS is also off the table 
(i.e. not desired)? Obviously, I don't agree with this approach unless you 
all want to create a big mess. I'm not into big messes.

Andy

simonmichal:
For the record, the SSS approach is being tracked in #1138, @abh3 : could you give us a timeline for this development? I suppose if there is a consensus that #1138 makes this PR unnecessary we could close this it, @bbockelm , @ffurano  what's your opinion?

xrootd-dev:
Should be available in 5.1 if not sooner.

ffurano:
Hi,
this discussion is very positive and welcome indeed, and we all share the need of forwarding security credentials with a certain level of flexibility.

I understand the direction of willing to keep the xrootd protocol in the back of a gateway, and it makes sense to me. 

Even if I like the idea of a macaroon for this use case I'd also vote for the XrdSSS approach, and hope that it becomes able to forward the additional info that an XrdSecEntity instance can hold.

bbockelm:
I’m certainly happy to close this out (SSS is a more comprehensive approach if taken) if the 5.1 timeline works for EOS!  Maybe ping Elvin on it?

(At any rate, glad we forced the discussion and didn’t try to merge it into 4.11.2!)

simonmichal:
I've just crosschecked with Elvin and it seems he no longer needs a proxy for HTTP, hence from EOS side there's no desire for this PR.

xrootd-dev:
Yes, good that we settled things. Please take a look at the new SecEntity 
structure. It supports additional fields as well as arbitrary passive 
(i.e. key-value) and active attributes (i.e. pointer-object). At the 
moment, everything except active attributes will be relayed (mostly 
because there is no workable model for active attributes. Active 
attributes are meant to be used at the node that deals with the 
SecEntity structure -- for example, authorization that can use such an 
attribute to reuse authorization credentials without needing to rebuild 
them each time). If anything is missing, now is the time to speak up.

simonmichal:
I suppose we can close this PR?

bbockelm:
Yup!  Looks like I have the ability to close this ... thanks for the feedback everyone!

xrootd-dev:
Yes, I think you can.

PR #1133, opened 2020-02-10
Fix curl speed limit to be really around 10KB/s

vokac:
Curl speed limit is defined directly in KB/s (not total size transferred divided by seconds in time windows).


ffurano:
I merged it as you are both agreeing. My opinion however is that it's a bit too low...

PR #1136, opened 2020-02-14
Python: make rpath relative to the module

chaen:
This commit https://github.com/xrootd/xrootd/commit/fca51b27e2989b033b3ce889929e3d06b5258206 makes it very difficult to relocate the client.
That PR should allow it by specifying an RPATH relative to the python module


chaen:
Hum, not sure if the tests failure have anything to do with my changes

simonmichal:
@chaen : you've added: `import sysconfig` which is shadowing `from distutils import sysconfig`, but no worries I'll fix it after merging it

PR #1142, opened 2020-02-26
Add TCP statistics printing for closed sockets

djw8605:
With this pull request, we can print out the TCP statistics for closed sockets.   I am looking for feedback on how to conditionally enable this functionality with a configuration option.

### DIFF
--- a/src/Xrd/XrdLinkXeq.cc
+++ b/src/Xrd/XrdLinkXeq.cc
@@ -258,9 +258,49 @@ int XrdLinkXeq::Close(bool defer)
 // Close the file descriptor if it isn't being shared. Do it as the last
 // thing because closes and accepts and not interlocked.
 //
-   if (fd >= 2) {if (KeepFD) rc = 0;
-                    else rc = (close(fd) < 0 ? errno : 0);
-                }
+   if (fd >= 2) {
+      if (KeepFD) {rc = 0;}
+      else {
+
+         // Get the remotely connected address
+         char addr_str[512];
+         int addr_rc = Addr.Format(addr_str, 512);
+         if (addr_rc < 1) {
+            Log.Emsg("Error from addr Format", std::to_string(addr_rc).c_str());
+         }
+
+         // Get the TCP socket information, if available (getsockopt can fail)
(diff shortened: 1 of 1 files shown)


abh3:
First, what is the motivation (presumably some requirement) for this additional code? This kind of stuff really belongs the monitoring stream not in the log. That said, a) it's Linux specific so it needs to be if-def'd, b) please use sizeof() when referencing previously sized arrays, c) snprintf error return will always be negative (it really can't return 0 unless nothing was in the format string which isn't the case); the '< 1' should be '< 0' for clarity, and d) please use XrdSysE2T() when printing errors; printing just the errno is not particularly meaningful. Finally, the code is rather long and obscures what is being done in a rather complicated part of the code. It should be moved to a separate method. So, let's first discuss the need fand proper placement or this before we move forward.

djw8605:
The motivation is to measure the network behavior during a transfer.

I agree that this should go into the monitoring stream.  I will close this pull request and start another when I have a draft in the monitoring stream.

This type of information makes sense to be in the "disconnect" event in the monitor, but I'm worried about backward compatibility with the monitoring stream being a binary format.  Should I create a new type of monitoring event.  For example, a `tcp` event, on the same level as an disc or open event.

I can make the other changes easily enough.

xrootd-dev:
Hi Derek,


On Mon, 2 Mar 2020, Derek Weitzel wrote:

> The motivation is to measure the network behavior during a transfer.
First, of course,we have to define what a transfer is. I assume you are 
talking about xrdcp not just a run-of-the mill session that does some 
reads and writes. Right?

> I agree that this should go into the monitoring stream.  I will close 
> this pull request and start another when I have a draft in the 
> monitoring stream.
OK.

> This type of information makes sense to be in the "disconnect" event in 
> the monitor, but I'm worried about backward compatibility with the 
> monitoring stream being a binary format.  Should I create a new type of 
> monitoring event.  For example, a `tcp` event, on the same level as an 
> disc or open event.
That is a very good question. The answer to it will depend on what the 
answer is to "what do you mean by 'transfer'".

Andy

PR #1147, opened 2020-03-04
XrdMacaroons: Allow chaining user defined XrdAccAuthorize plugin

esindril:
This extends the current behavior and keeps full backwards compatibility with the way XrdMacaroons was loading the (default) authorization plugin or was chaining the XrdAccSciTokens library. I've just added support for specifying a custom library which provides a XrdAccAuthorization plugin to be chained to XrdMacaroons (both to the http and authz plugins).

This is done by using as configuration parameter the key "chain_authz=libCustomAcc.so". We need this since EOS for example does not use at all the default XrdAcc plugin but provides its own authorization plugin. This will be useful also for other projects that want to rely on XrdMacaroons but have their own authorization plugin.
It would be great if this gets merged into 4.11.3.
@bbockelm, could you please review it?
Thanks!

### DIFF
--- a/src/XrdMacaroons/XrdMacaroons.cc
+++ b/src/XrdMacaroons/XrdMacaroons.cc
@@ -1,4 +1,3 @@
-
 #include <stdexcept>
 #include <dlfcn.h>
 
@@ -9,6 +8,7 @@
 #include "XrdOuc/XrdOucPinPath.hh"
 #include "XrdSys/XrdSysError.hh"
 #include "XrdSys/XrdSysLogger.hh"
+#include "XrdSys/XrdSysPlugin.hh"
 #include "XrdHttp/XrdHttpExtHandler.hh"
 #include "XrdAcc/XrdAccAuthorize.hh"
 #include "XrdVersion.hh"
@@ -24,6 +24,80 @@ extern XrdAccAuthorize *XrdAccDefaultAuthorizeObject(XrdSysLogger   *lp,
                                                      const char     *parm,
                                                      XrdVersionInfo &myVer);
 
+//------------------------------------------------------------------------------
+//! Handle chained authorization plugins specified directly as a parameter in
+//! the parms variable or in the key=value format such as "chain_authz=lib.so"
+//!
+//! @param err error object user ofr logging
+//! @param config path to configuration file
+//! @param parms string representing the configuration options given to the
+//!        libXrdMacaroons library
+//!
+//! @return authorization plugin object to be chained
+//------------------------------------------------------------------------------
+XrdAccAuthorize* HandleChainedAuthz(XrdSysError* err, const char* config,
+                                    const char* parms)
+{
+  if (!parms) {
+    return NULL;
+  }
+
+  XrdOucString sparms = parms;
+  XrdOucString token = "chain_authz=";
+
+  if (sparms.beginswith(token.c_str())) {
+    sparms.erase(0, token.length());
+  }
+
+  err->Log(Macaroons::LogMask::Info, __FUNCTION__,
+           "Try to chain authz:", sparms.c_str());
+
+  // Trim whitespaces
+  while (sparms.length() && (sparms[0] == ' ')) {
+    sparms.erase(0, 1);
+  }
+
+  if (sparms.length() == 0) {
+    err->Log(Macaroons::LogMask::Error, __FUNCTION__,
+             "No authz lib specified in the chain option");
+    return NULL;
+  }
+
+  int pos = sparms.find(' ');
+  const char* new_parms = ((pos == STR_NPOS) ||
+                           (pos == sparms.length() -1)) ? NULL : &sparms[pos + 1];
+  XrdOucString lib_name(sparms, 0, pos);
+  char resolve_path[2048];
+  bool no_alt_path = false;
+
+  if (!XrdOucPinPath(lib_name.c_str(), no_alt_path, resolve_path,
+                     sizeof(resolve_path))) {
+    err->Log(Macaroons::LogMask::Error, __FUNCTION__,
+             "Failed to locate library path for", lib_name.c_str());
+    return NULL;
+  }
+
+  // Try to load the XrdAccAuthorizeObject provided by the given library
+  XrdAccAuthorize *(*authz_ep)(XrdSysLogger*, const char*, const char*);
+  XrdSysPlugin authz_plugin(err, resolve_path, "authz", &compiledVer, 1);
+  void* authz_addr = authz_plugin.getPlugin("XrdAccAuthorizeObject", 0, 0);
+  authz_plugin.Persist();
+  authz_ep = (XrdAccAuthorize * (*)(XrdSysLogger*, const char*,
+                                    const char*))(authz_addr);
+  XrdAccAuthorize* authz_obj = NULL;
+
+  if (authz_ep && (authz_obj = authz_ep(err->logger(), config, new_parms))) {
+    err->Log(Macaroons::LogMask::Info, __FUNCTION__,
+             "Successfully chained authz plugin from", resolve_path);
+    err->Log(Macaroons::LogMask::Info, __FUNCTION__, "Chained authz plugin "
+             "with params \"", (new_parms ? new_parms : ""), "\"");
+    return authz_obj;
+  } else {
+    err->Log(Macaroons::LogMask::Error, __FUNCTION__,
+             "Failed loading authz plugin from", resolve_path);
+    return NULL;
+  }
+}
 
 extern "C" {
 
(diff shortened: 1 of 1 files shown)


abh3:
We cannot let this go in Elvin. It completely conflicts with the way this is done in R5 and we can't have two different mechanisms competing in the same code base. I know you spent a lot oftme on this but it really completely undoes everything we did in R5 to solve this problem. In your particular case, either port the R5 back to 4.11 or please branch the code.

esindril:
Hi Andy,

I had a look at R5, more specifically at this commit with adds support for stacking different OFS plugins:
https://github.com/xrootd/xrootd/commit/a56e9b4ec0dd7784cafa87ddc01ccf5e009ebe38#diff-22c42d4e97f2203a846bb636eac80e77

but I don't quite see how this applies in this case where I would like to control the authorization plugin which is loaded by an XrdHttpExtHandler plugin namely XrdMacaroons?

I looked at the XrdMacaroons code in R5 and it's no different to what is now in 4.11. I'm happy to adapt my approach if you can tell me how I can have the XrdMacaroons load a different authz plugin here:
https://github.com/xrootd/xrootd/blob/stable-5.0.x/src/XrdMacaroons/XrdMacaroons.cc#L98

Thanks!

abh3:
Hi Elvin,

I largely based my comments on the fact that a) XrdAccAuthorize seemed to have been changed far

abh3:
Hi Elvin,

You are dealing with some pretty nasty code. I agree that the stacking mechanism in Macaroons is pretty ad hoc and we are trying to standardize how plugins are stacked. There seems to be precious little information on how macaroons fits into the whole authorization scheme. The idea of simply stacking the default authorization is bad and it's even worse by using the linking trick (I know you tried to correct that but the code got even more obscure). You are correct that there was no change to Macaroons but the whole stacking phenomena should follow the convention:

ofs.authlib ++ <stack this one>

See: https://xrootd.slac.stanford.edu/doc/dev50/ofs_config.htm#_Toc8244755

It's not clear to me where the initial load goes and what is the directive being used to load this plugin. But how it manages stacking is pretty bad (as I said) and I don't want it to get even worse by plastering it over with patches.

The other issue is that we normally do back-ports not forward ports. The pull request should have been against git head not a stable release, especially when it's in a release cycle. It also puts us in the position of  dropping it unless we port forwards.

So, what directive controls this Macaroons plugin? How does it relate to XrdAccAuthorize? It really seem pretty hacky the way it is now (which has nothing to do with your patches).

bbockelm:
FWIW - a lot of the direct linking code is due to the fact there's not a public API (and all this code originated as an external plugin).  That's an opportunity for code cleanup now that this plugin has access to the private Xrootd APIs.

From the `ofs.authlib` point-of-view, an example of how it's loaded now is:

```
ofs.authlib libXrdMacaroons.so libXrdAccSciTokens.so
```

So, authorizations are first examined by `libXrdMacaroons`; if the result is inconclusive, then it's passed to `libXrdAccSciTokens`; if that result is inconclusive, then it is passed to the default plugin.

@esindril - can you kindly remind me of the use case here?  I recall talking about it but don't think I wrote it down.  That is, if you have `libXrdAccEOS`, why doesn't the following work:

```
ofs.authlib libXrdMacaroons.so libXrdAccEOS.so
```
?

xrootd-dev:
OK, so in R5 all of this is handled automatically and no direct linking is 
required. To get the same result (with an arbitrary level of stacking),
assuming the base authorization plugin should be the default one, as 
follows:

ofs.authlib ++ libXrdAccSciTokens.so
ofs.authlib ++ libXrdMacaroons.so

Then libXrdMacaroons.so would be called first. If it doesn't like the 
thing it sees, it passes the call to libXrdAccSciTokens.so (it gets a 
pointer to the next stacked plugin). That, in turn, can pass it off to the 
default plugin if it wishes. You can, of course, declare one of those as 
the base plugin by not specifying "++". That keeps backward compatability.

It's all structured to allow one to write a plugin with only the pulic 
API's. So, yes, all of that should be cleaned up. The whole scheme is 
available in git head and documented in the XrdAccAuthorize.hh file on how 
to structure the code for base or push mode (or both).

Andy

bbockelm:
This sounds very promising!

(One aside on the manual text for this - I was very confused by this sentence in the manual:
```
++        The specified plug-in should stack on top of the existing plug-in or default. Once specified, it cannot be overridden by a subsequent directive,
```

I read this as saying "this can only be used once", which somewhat contradicts what you wrote above.  A clarification in the manual would be appreciated.)

The second unpleasant piece in this code is the fact that, within the HTTP handler, we need to have access to the `XrdAccAuthorize` object used by the OFS to generate the appropriate permissions for the Macaroon.

Is there such a way to do that?  This is currently accomplished by directly constructing the object and via a reimplementation of the `ofs.authlib` logic.  There's not much gain if we significantly clean up this usage only to add more complexity elsewhere.  What you point out only helps the case of making an authorization object for the `ofs` and not recreating it in the plugin.

Thanks!

abh3:
On Wed, 4 Mar 2020, Brian P Bockelman wrote:

> This sounds very promising!
Yes, it was written specifically to handle multi-mode authorization which 
SciTokens essentially introduces (didn'y know Macaroons did the same).

> (One aside on the manual text for this - I was very confused by this 
> sentence in the manual:
> ```
> ++        The specified plug-in should stack on top of the existing 
> plug-in or default. Once specified, it cannot be overridden by a 
> subsequent directive,
> ```
>
> I read this as saying "this can only be used once", which somewhat 
> contradicts what you wrote above.  A clarification in the manual would 
> be appreciated.)
OK, poor wording on my part. You can always override the base plugin with 
a subsequent directive. That's the one without a "++". You can't eally 
extend this o any plugin that was stacked with a "++". Once stacked, it 
reains stacked. All you can do is add another one to the stack. That's 
what I was trying to say. So, I'll reword that in all the places it 
appears. I say that because the syntax was extended to most plugins. Yes, 
you can stack all the plugins that matter at he mment and we can extend it 
to additional ones should the need arrise.

> The second unpleasant piece in this code is the fact that, within the 
> HTTP handler, we need to have access to the `XrdAccAuthorize` object 
> used by the OFS to generate the appropriate permissions for the 
> Macaroon.
Yes, I see the problem. I was going to add an "Access(const path[, mode])" 
to the whole thing so you can get that information. Then I had second 
thoughts as I didn't see who would use that. I guess I was wrong here. I 
can add that feature and it will be available via the bridge as an fsctl() 
operation. I assume that Access() as defined by POSIX will do the trick 
for you. Either it says yes or no or returns the mode bits allowed. Right?

> Is there such a way to do that?  This is currently accomplished by 
> directly constructing the object and via a reimplementation of the 
> ofs.authlib` logic.  There's not much gain if we significantly clean up 
> this usage only to add more complexity elsewhere.  What you point out 
> only helps the case of making an authorization object for the `ofs` and 
> not recreating it in the plugin.
Yes, I am trying to desparately avoid doing this and I think we can do 
that pretty easily.

> Thanks!
You're welcome.

Andy

abh3:
Hi Brian,

Just realized, your plugin doesn't know anything about the bridge, right? 
If so, I can plop a pointer to an authorization object in the XrdOucEnv 
object passed to the protocol plugin. Then you'll have access to all the 
methods using whatever the config was in the ofs. Will that be better for 
you?

Andy

esindril:
@bbockelm  Concerning your question:
> @esindril - can you kindly remind me of the use case here? I recall talking about it but don't think > I wrote it down. That is, if you have libXrdAccEOS, why doesn't the following work:
>
> ofs.authlib libXrdMacaroons.so libXrdAccEOS.so
> ?

I think you already provided the answer in you subsequent comment:
> The second unpleasant piece in this code is the fact that, within the HTTP handler, we need to
> have access to the XrdAccAuthorize object used by the OFS to generate the appropriate 
> permissions for the Macaroon.
>
> Is there such a way to do that? This is currently accomplished by directly constructing the object > and via a reimplementation of the ofs.authlib logic. There's not much gain if we significantly clean > up this usage only to add more complexity elsewhere. What you point out only helps the case of > making an authorization object for the ofs and not recreating it in the plugin.

I actually have a similar patch for the XrdSciTokens library, but I'll for the the things to settle on a cleaner solution for all this chaining business.

@abh3 
> If so, I can plop a pointer to an authorization object in the XrdOucEnv
> object passed to the protocol plugin. Then you'll have access to all the
> methods using whatever the config was in the ofs. Will that be better for you?

I think this looks reasonable and it's already cascaded from the XrdHttpProtocol to any XrdHttpExtHandler object. Anyhow, this will only be available in R5, right?

bbockelm:
@abh3:
> If so, I can plop a pointer to an authorization object in the XrdOucEnv object passed to the protocol plugin

I _think_ this would work; at least, would be willing to test for v5.

I do eventually need both a pointer to the authorization object and for the filesystem object (for `XrdHttpTPC`, not Macaroons).

For v4 versus v5: I'm OK with showing this works with v5 and seeing how much nasty code this allows us to remove -- and then making a decision about a backport.

bbockelm:
@abh3 - any updates on this?

abh3:
Yes, the env passed around will have three additional keys:

XrdAccAuthorize* pointer to the authorization plugin
XrdOss* Pointer to the storage system plugin
XrdSfsFileSystem* pointer to the file system plugin

The patches are already in R5 and will be back ported to 4.11.3.

bbockelm:
@abh3 -- I think #1159 has a better approach for Xrootd 5 (and keeps the backward compat).

So the question is - what should we do for the 4.11 branch?  @esindril - do you have any workarounds?

I would be OK with something somewhat hack-ish going in just to get EOS un-stuck, especially if we have something better going forward.

abh3:
Yes, I agree that 1159 is much cleaner. BTW you know that the pointer will be zero if authorization was not enabled. So, you may want to rethink how to handle that. I suppose I should have added yet another envar to tell you that. Is not configuring authorization something that is acceptable to the Macaroon support? As for a work around, I'm looking for suggestions. I guess you need the stacking framework to make it all workable, right?

PR #1155, opened 2020-03-15
Improve HTTP-TPC error messages.

bbockelm:
Importantly, if the passive endpoint provides a reasonable error messages, this is now passed along to the TPC client.


bbockelm:
@ffurano or @abh3 - I think this is ready to go.

PR #1157, opened 2020-03-16
Include a list of connections in the HTTP-TPC performance marker.

bbockelm:
This will provide the TPC monitoring with information about whether IPv4 and/or IPv6 are being used.

Example perf marker including connection info:

```
Perf Marker\n
Timestamp: 1537788010\n
Stripe Index: 0\n
Stripe Bytes Transferred: 238745\n
Total Stripe Count: 1\n
RemoteConnections: tcp:129.93.3.4:1234,tcp:[2600:900:6:1301:268a:7ff:fef6:a590]:2345\n
End\n
```

### DIFF
--- a/src/XrdTpc/XrdTpcMultistream.cc
+++ b/src/XrdTpc/XrdTpcMultistream.cc
@@ -264,7 +264,7 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
         time_t now = time(NULL);
         time_t next_marker = last_marker + m_marker_period;
         if (now >= next_marker) {
-            if (SendPerfMarker(req, current_offset)) {
+            if (SendPerfMarker(req, handles, current_offset)) {
                 return -1;
             }
             last_marker = now;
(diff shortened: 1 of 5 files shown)


abh3:
It's not at all obvious where this information goes. Hopefully not to the log file! You may or may not know that R5 introduced a generalized monitoring stream that allow such information to be sent to a collector, as defined in the config file. You are free to use whatever format you want, though to keep the variety down, JSON would be the obvious choice. If you'd like, we can pass a pointer to the monitoring object  via the HTTP protocol with the ENV object.

bbockelm:
Actually, this is what is returned by the server back to the TPC client (see https://twiki.cern.ch/twiki/bin/view/LCG/HttpTpcTechnical)... not part of the monitoring stream but rather is included in what the FTS client collects.

abh3:
Got it. Could you add a comment to that effect somewhere in the code so that the next time I (or anyone else) look at  it's obvious why it's there.

bbockelm:
@ffurano or @abh3 - I think this is ready to go.

PR #1158, opened 2020-03-20
Have HTTP-TPC reuse the SFS from the environment.

bbockelm:
The XRootD framework now exports the global SFS object to the environment (which is available to the HTTP-TPC plugin).  This switches HTTP-TPC to use that object instead of loading a separate copy of the filesystem.


bbockelm:
@ffurano / @abh3 - I think this one is ready to go forward.

PR #1159, opened 2020-03-20
When possible, use the built-in chaining featured in Xrootd 5.

bbockelm:
@esindril - I think this would work for EOS, at least for Xrootd5.

### DIFF
--- a/src/XrdMacaroons/XrdMacaroons.cc
+++ b/src/XrdMacaroons/XrdMacaroons.cc
@@ -5,6 +5,7 @@
 #include "XrdMacaroonsHandler.hh"
 #include "XrdMacaroonsAuthz.hh"
 
+#include "XrdOuc/XrdOucEnv.hh"
 #include "XrdOuc/XrdOucString.hh"
 #include "XrdOuc/XrdOucPinPath.hh"
 #include "XrdSys/XrdSysError.hh"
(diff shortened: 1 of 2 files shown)


esindril:
Why can't this be also ported to 4.11.3? Since I see the core part of exposing these pointers was already backported?
Otherwise, yes this look good enough for EOS.

esindril:
This is good enough for me. My only problem was to have the macaroon authz chain with the ofs authz. I think this is solved now with the env pointers, the rest I can manage from EOS code.

xrootd-dev:
The reason we could backport the pointer setting was because it was 
mindless and if we git it wrong nothing bad would happen other than the 
Macaroon stuff wouldn't work they way people envisioned. That could get 
fixed in 4.11.4. The other code is a basic change in the architecture and 
more involved.We generally don't backport that kind of complexity.

Andy

esindril:
Could you backport the env pointer stuff without the XrdAccAuthorizeAdd part which I assume is the R5 specific bit?

simonmichal:
Essentially what you need are only those 2 lines:
https://github.com/xrootd/xrootd/pull/1159/files#diff-62856a6e59b3e111c2b21dabdad9cb75R118-R119
?

esindril:
Yes.

simonmichal:
Then to streamline all this, with Brian's permission, I can commit those two lines to the master branch and back port the commit to 4.11.3. Brian then will have to remove them from the PR.

Let me know if that's OK with all of you!

bbockelm:
Yup, go for it!  I'll fix the conflict once it appears.

simonmichal:
Done: af4df60d21b0dd0188660c699e22969afc9ada2d

bbockelm:
Oh duh - there's not going to be a conflict unless you merge in branch-4 to master.  So, this can go in whenever.

@esindril - when you get the chance though, can you verify this worked?

Issue #1167, opened 2020-03-28
Wrong error code from Xrootd bridge for file exists


bbockelm:
When a file exists, the cmsd returns the wrong error code (or maybe just doesn't provide a distinct error code?) resulting in the wrong xrootd error code.

See the following example (note PID 2072999 is the `cmsd` and 2075105 is the `xrootd`):
```
200328 09:18:45 2072999 unknown.1:22@example do_Select: failed; Unable to create new file; file already exists. /prefix/test_folder4
200328 09:18:45 2075105 Receive ???????? 52 bytes on 3071
200328 09:18:45 2075105 Decode example gave unknown.1:22@example err -2 'Unable to create new file; file already exists.' /prefix/test_folder4
200328 09:18:45 2075096 unknown.1:22@example XrootdResponse: sending err 3011: Unable to create new file; file already exists.
```

I think what happens is the `-2` error code in the `cmsd` response is interpreted as a Unix `errno` by the `xrootd` daemon, which corresponds to `ENOENT` and then mapped to error code `3011`.

The wrong response (3011 is `kXR_NotFound` or file not found) is causing the HTTP protocol to return the wrong response and clients to take the wrong action.

The right fix isn't readily apparent -- it looks like giving the "right" error code from the `cmsd` would be a large-scale protocol break, which I'm not confident in doing immediately.  May have to do a short-term workaround based on the error message in the meantime.

abh3:
I left a comment in the pull request but I will post here as well.

I looked, indeed, the cmsd does not differentiate the between ENOENT and EEXISTS.
That can be fixed. I don't particularly like relying on message text as that tends
to change over time as people complain that the message isn't sufficiently
descriptive.

The correct error code should have been returned by the cmsd and, in fact, it's supposed to do that. I agree that it isn't obvious because error code handling becomes complicated when the cmsd attempts a retry and gets yet another error. However, in this case it's pretty straightforward as a retry is not possible. So, it's not too complicated to fix this.

That does that change your opinion on the fix?

bbockelm:
I spent about an hour tinkering with different approaches to fixing the return code from the cmsd.  However, it was quickly becoming a wholesale rewrite of all return codes between the server and client - it kept snowballing into pieces of the codebase I wasn't familiar with.

So, I agree with the idea that we really should fix the problem at the root.

[...]

PR #1168, opened 2020-03-28
Workaround an incorrect error code from the Xrootd bridge.

bbockelm:
This avoids the issue in #1167, which causes the wrong error code to be given for a known error text.  This simply looks for the known error text and corrects the HTTP status code.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2621,6 +2621,13 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
     {
 
       if (xrdresp != kXR_ok) {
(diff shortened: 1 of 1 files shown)


abh3:
Shouldn't we just simply fix the underlying problem? What error code should have been returned?

abh3:
I looked, indeed, the cmsd does not differentiate the between ENOENT and EEXISTS. That can be fixed. I don't particularly like relying on message text as that tends to change over time as people complain that the message isn't sufficiently descriptive. What's you take on this now?

bbockelm:
@abh3 fixed the error code in the R5 branch so this is no longer necessary.  Will close this and reopen a new PR targeting only the stable branch.

PR #1169, opened 2020-03-30
Workaround an incorrect error code from the Xrootd bridge.

bbockelm:
This avoids the issue in #1167, which causes the wrong error code
to be given for a known error text.  This simply looks for the known
error text and corrects the HTTP status code.

The root cause is already fixed in R5 -- this is just a minimal workaround for stable.


abh3:
I see that you substitute your own text for he error. You may wish to use the text that comes with the error code; especially if it's EEXIST (or kXR_ItExists). That message differentiates between files and directories.

abh3:
I think we have gotten a solution here add pull is already out dated anyway (it's for 4.11.x and are at 4.12.x). So, I am closing it.

PR #1171, opened 2020-03-31
Support PyPy

chrisburr:
Conda-forge is current adding support for [PyPy](https://www.pypy.org/features.html) to as many packages as it can. As I work on some projects where PyPy might be useful I took a look at fixing the XRootD build and it turned out to be very simple. This PR includes the necessary patch.

See also: https://github.com/conda-forge/xrootd-feedstock/pull/19

### DIFF
--- a/bindings/python/src/PyXRootDModule.cc
+++ b/bindings/python/src/PyXRootDModule.cc
@@ -80,7 +80,6 @@ namespace PyXRootD
 #endif
   {
     // Ensure GIL state is initialized
-    Py_Initialize();
     if ( !PyEval_ThreadsInitialized() ) {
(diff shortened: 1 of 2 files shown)


bindings/python/src/PyXRootDModule.cc
```diff
@@ -80,7 +80,6 @@ namespace PyXRootD
 #endif
   {
     // Ensure GIL state is initialized
-    Py_Initialize();
```
chrisburr: To the best of my knowledge this is only needed when embedding a Python interpreter inside your application: https://docs.python.org/3/c-api/init.html

PR #1179, opened 2020-04-20
Initialize the XrdAccEntityInfo structure.

bbockelm:
When one `vorg` is provided and no groups are provided, the `grup` data memeber may be uninitialized on use.  This has resulted in periodic segfaults on the SciTokens ACC provider; I'm guessing that `grup` quite often has an initial value of `nullptr`, even when uninitialized.

I suspect this hasn't been observed for the VOMS provider because all VOMS server implementations create at least one group; hence this case is never hit.

### DIFF
--- a/src/XrdAcc/XrdAccEntity.hh
+++ b/src/XrdAcc/XrdAccEntity.hh
@@ -42,7 +42,12 @@ struct XrdAccEntityInfo
        const char *vorg;
        const char *role;
        const char *grup;
-                   XrdAccEntityInfo() {}
+                   XrdAccEntityInfo() :
+                       name(NULL),
+                       host(NULL),
+                       vorg(NULL),
+                       role(NULL),
+                       grup(NULL) {}
                   ~XrdAccEntityInfo() {}
       };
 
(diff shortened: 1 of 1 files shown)


bbockelm:
Once we resolve this, @simonmichal please consider for backport.  This is an attempt to fix a periodic segfault observed in 4.11.3.

simonmichal:
@bbockelm : thanks a lot for the PR, could you please refactor the commit so the initialisation is done in the default constructors (`XrdAccEntityInfo()` and `EntityAttr()`), otherwise it won't build on SLC (GCC 4.4.7). I do know it's a pain! By any chance do you know if it's possible to use `devtoolset` with EPEL build?

bbockelm:
@simonmichal - update pushed.  Back down to C++03 :(

For devtoolset: I found some references online about doing this in EPEL7.  It's not clear whether that's an option for EPEL6.  Here's what we do in OSG:

https://github.com/opensciencegrid/xrootd-lcmaps/blob/master/rpm/xrootd-lcmaps.spec

bbockelm:
@simonmichal - build failures look unrelated?

simonmichal:
@bbockelm : yes that was me (the build is fixed now), I gave it try and it compiles like a charm now, it's ready to be merged.

PR #1183, opened 2020-04-22
Reload certificate store once an hour.

bbockelm:
This triggers a reload of the cert store used for chain building and certificate verification (CRLs) once an hour.

Fixes #750

PR #1186, opened 2020-04-28
Periodically reload verify cert store (stable)

bbockelm:
This is a backport of #1183 (plus some `#ifdef`'s so it can compile on SL6).

This uses a per-SSL-object certificate store (shared via ref counts) in order to provide verification functions.  A helper task periodically loads a fresh copy of the cert store.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -636,9 +636,17 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
           BIO_set_nbio(sbio, 1);
           ssl = SSL_new(sslctx);
 
+          // On newer versions of OpenSSL, we use the periodically
+          // updated store `verify_store` with the `x509_store_lock`
+          // held; older OpenSSL versions are missing `SSL_set1_*_cert_store`,
+          // so we simply rely on the global one in SSL_CTX.  The latter
+          // doesn't refresh the verify store, which causes issues with some
+          // plugins.
+#if OPENSSL_VERSION_NUMBER >= 0x010100000L
           XrdSysRWLockHelper scopedLock(x509_store_lock, true);
           SSL_set1_verify_cert_store(ssl, verify_store);
           SSL_set1_chain_cert_store(ssl, verify_store);
+#endif
 
         }
 
@@ -1693,6 +1701,7 @@ extern "C" int verify_callback(int ok, X509_STORE_CTX * store) {
 
 
 /// Initialization of the ssl security
+//  Only scheduled on sufficiently new OpenSSL.
 
 class XrdCertStoreJob : XrdJob
 {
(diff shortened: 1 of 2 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -633,6 +635,19 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
...
           BIO_set_nbio(sbio, 1);
           ssl = SSL_new(sslctx);
+
+          // On newer versions of OpenSSL, we use the periodically
+          // updated store `verify_store` with the `x509_store_lock`
+          // held; older OpenSSL versions are missing `SSL_set1_*_cert_store`,
+          // so we simply rely on the global one in SSL_CTX.  The latter
+          // doesn't refresh the verify store, which causes issues with some
+          // plugins.
+#if OPENSSL_VERSION_NUMBER >= 0x010100000L
```
abh3, line 645: Not so fast, the test should be '>' not '>=' because OpenSSL 1.01 does not have the set1 functions. I know from your comment that you did compile it on SL6 but that only worked because someone must have upgraded their OpenSSL (which of course they should have). Unfortunately, lots of sites use RH6 out of the box and it comes with 1.0.1e-fips and it definitely won't compile! I think all of teh test should change and hen you'll be fine.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1767,6 +1833,20 @@ int XrdHttpProtocol::InitSecurity() {
     }
   }
 
+  // Initialize a store for use on individual SSL objects and schedule
+  // a periodic update.  These separate stores are only usable on versions
+  // of OpenSSL 1.1.0 or later.
+#if OPENSSL_VERSION_NUMBER >= 0x010100000L
```
abh3, line 1839: Same here.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: I'm not sure I understand this?

`OPENSSL_VERSION_NUMBER >= 0x010100000L` should test for 1.1.0 or later.  That is, it only enables this code if you are at least on 1.1.0 -- that should exclude any variant of RHEL6.

(FWIW - the OpenSSL RPM I tested against was `openssl-1.0.1e-58.el6_10.x86_64`)

simonmichal: I'm not sure if it is intentional, but it will also exclude any variant of RHEL7, as e.g. both Centos7 and CC7 have openssl `1.0.2k`. From what I see `SSL_set1_verify_cert_store` & `SSL_set1_chain_cert_store` are available in OpenSSL 1.0.2 (see e.g. https://github.com/openssl/openssl/blob/OpenSSL_1_0_2k/ssl/ssl.h). If we would like to include RHEL7 we should test for `OPENSSL_VERSION_NUMBER >= 0x10002000L`.

bbockelm: Argh - sorry, I haven't been able to sit down and get to this in the last two days (which have each had >6hrs of Zoom meetings...).  You're right, this is too strict.

abh3:
Not to worry, we will appropriately change it.

simonmichal:
Already done: 5c73add419b208ab38e467d630b4c533fe960c05

Issue #1188, opened 2020-05-01
XRootD 5.0-rc2: Handshake fails with TLS turned on, but no client verification


djw8605:
I am testing the xrootd 5.0-rc2 and having issues when turning TLS on.  I have am using a cache -> origin architecture.

On the cache, I have no special configuration.  It is OSG's build: xrootd-server-5.0.0-0.rc2.2.osgup.el7.x86_64.  The cache has the proxy line removed, so it cannot use a proxy to force other authentication methods.

On the origin, I have the TLS lines:
```
xrd.tls /etc/grid-security/xrd/xrdcert.pem /etc/grid-security/xrd/xrdkey.pem
xrd.tlsca noverify
xrd.trace all
xrd.trace tls
xrd.protocol tls xrootd:1095 *
```

The origin also has `gsi` enabled.  Same version of xrootd as the cache.

### Cache Log
```
200501 00:14:32 13399 XrdInet: Accepted connection from 161@natnode
200501 00:14:32 13399 XrdProtLoad: matched port 8443 protocol http
200501 00:14:32 13399 anon:161@natnode XrdPoll: FD 161 attached to poller 0; num=1
200501 00:14:32 13399 XrootdBridge: unknown.1:161@natnode login as nobody
200501 00:14:32 13399 unknown.1:161@natnode ofs_stat:  fn=/hcc/PROTECTED/dweitzel-test/blah2.txt
[2020-05-01 00:14:33.148553 +0000][Error  ][XRootDTransport   ] [p0@origin:1095.0] No protocols left to try
[2020-05-01 00:14:33.148609 +0000][Error  ][AsyncSock         ] [p0@origin:1095.0] Socket error while handshaking: [FATAL] Auth failed
[2020-05-01 00:14:33.148660 +0000][Error  ][PostMaster        ] [p0@origin:1095] elapsed = 0, pConnectionWindow = 120 seconds.
[2020-05-01 00:14:33.148682 +0000][Error  ][PostMaster        ] [p0@origin:1095] Unable to recover: [FATAL] Auth failed.
[2020-05-01 00:14:33.148689 +0000][Error  ][XRootD            ] [p0@origin:1095] Impossible to send message kXR_stat (path: /hcc/PROTECTED/dweitzel-test/blah2.txt?authz=Bearer%20biglongbearertoken&oss.lcl=1&pss.tid=http, flags: none). Trying to recover.
[...]

PR #1195, opened 2020-05-17
Add new 'anonymous mode' for unix authentication.

bbockelm:
In anonymous mode, neither username nor group name is used in Unix authentication.  This is envisioned to be useful in cases where authorization happens later as part of a bearer token in the request.

This greatly simplifies the approach for #1188 as one can safely enable both `unix` and `gsi`.  Because this allows the admin to mark `unix`-based authentication as anonymous, it can safely be used in cases where the Authfile is loaded.

@djw8605

### DIFF
--- a/src/XrdSecunix/XrdSecProtocolunix.cc
+++ b/src/XrdSecunix/XrdSecProtocolunix.cc
@@ -34,6 +34,7 @@
 #include <stdlib.h>
 #include <strings.h>
 #include <sys/types.h>
+#include <sstream>
 
 #include "XrdVersion.hh"
 
@@ -43,6 +44,8 @@
 #include "XrdSys/XrdSysHeaders.hh"
 #include "XrdSys/XrdSysPthread.hh"
 #include "XrdSec/XrdSecInterface.hh"
+
+bool g_anonymous_mode = false;
 
 /******************************************************************************/
 /*              X r d S e c P r o t o c o l u n i x   C l a s s               */
@@ -61,8 +64,10 @@ friend class XrdSecProtocolDummy; // Avoid stupid gcc warnings about destructor
         XrdSecCredentials *getCredentials(XrdSecParameters  *parm=0,
                                           XrdOucErrInfo     *einfo=0);
 
-        XrdSecProtocolunix(const char *hname, XrdNetAddrInfo &endPoint)
-                          : XrdSecProtocol("unix")
+        XrdSecProtocolunix(const char *hname, XrdNetAddrInfo &endPoint,
+                           bool anonymous_mode)
+                          : XrdSecProtocol("unix"),
+                            m_anonymous_mode(anonymous_mode)
                           {Entity.host = strdup(hname);
                            Entity.name = (char *)"?";
                            epAddr      = endPoint;
@@ -80,6 +85,7 @@ private:
 
 XrdNetAddrInfo            epAddr;
 char                     *credBuff;      // Credentials buffer (server)
+bool                      m_anonymous_mode;
 };
 
 /******************************************************************************/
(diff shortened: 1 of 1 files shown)


abh3:
As you know, I hate anonymous mode except in a clearly defined code path. That said, I would prefer this be yet another authentication protocol. Say, XrdProtocolAnon. That way people know what they are getting into. While Unix is not any better I hate mixing those two protocols under one rubric. Anon mode really needs to be obvious so people can clearly grasp what is happening. 

Whatever happened to the authentication token identifier idea? That at least would stop people at the front door instead of leaving it wide open and crossing your fingers that intruders would eventually get caught (they won't, of course).

bbockelm:
Hi @abh3 -

Thought a bit more on the idea of `XrdProtocolAnon` over the weekend.  The primary problem I see is the lack of compatibility for older clients -- the whole reason I delved into this proposed "anonymous mode" over the weekend was I have a use case where:

1.  The network is secure (no MiTM).
2.  I want to support users with a variety of old CMSSW versions.

I still like the idea of having a separate token authentication protocol - but that's a touch more work (and will require a new client).

So, I see two potential routes here:
1. We go forward with this new flag for the existing `XrdSecunix` plugin.
2. We implement a `XrdSecAnon` plugin that advertises itself at the wire protocol level as `unix`.  One keeps compatibility with old releases but the local site admin has to very explicitly load it (can't operate in this mode by accident -- although I would guess the `sec.protocol unix anonymous` in the existing patch is pretty explicit...)

abh3:
Hi @bbockelm -
It's only software so we have a lot more options than two. I don't mind spending more time to get a consistent understandable implementation, especially when it deals with security. So...

a) The XrdProtocolanon (I made a previous mistake the 'A' letter should be lower case) fits into the scheme. It doesn't require a new client it merely requires that the plugin be available client-side. That may be easy or difficult. I thought that CMS does automatic distribution of various plugins (e.g. StashCache stuff) already so I thought this would just fall into that category. If I'm wrong, then this would be difficult.

b) I take a dim view of hacking the unix protocol. These protocols are supposed to be one-purpose only protocols anything more is just asking for trouble. Sorry if I take the purist approach.

c) One could do something else on the server. Since anon doesn't do anything then the server can do something. Suppose you allowed it to run through the protocols with non succeeding. Then the server could a) simply accept the login without authentication (horror on many faces), or b) redirect the client elsewhere. 

Option (b) comes with a lot  more flexibility on what to do but I don't see how this is going to solve anything given the case you outlined. 

Option (a) is workable but I would insist that if we do that, the server severely restricts what the client can do (e.g. open for reading only, perhaps a stat, read, and close). In fact, I was going to recommend that even with anon protocol (which essentially does the same thing behind the servers back).

At least in (a) we have some modicum of damage control. What do you think?

bbockelm:
Actually, I think I'm going to close this one for now.  Here's my thinking:
1.  Derek and I have a pretty narrow use case which requires us to support tokens in V4.  That's generally not a safe thing to do, hence I don't feel bad keeping this out-of-tree.
2.  For V5, we don't need backward compatibility since TLS mode itself isn't backward compatible.  In such a case, we can just do a completely new `anon` and/or `token` plugin.

Toward the latter - the experiments we are doing now are hopefully going to inform what's needed for a `token` plugin.  We're currently faking the client parts via an `XrdCl` plugin.

xrootd-dev:
Your call. However, it's pretty clear to me we will need an anon mode of 
some sort. So, restricted access is still on the table (much like ftp). 
Even so, I would prefer to always have some kind of authentication in any 
case. It doesn't have to be fancy; just enough to screen out the bad 
actors. At some point, I think expriments will want something at that 
level. I don't now how Alice manages. I guess they have been lucky so far.

Andy

adriansev:
ALICE use unix and then acc implemented by [alicetokenacc](https://github.com/cern-eos/xrootd-alicetokenacc/tree/xrootd4) that use [libTokenAuthz](https://github.com/cern-eos/libTokenAuthz)   
So, all accesses require an encrypted envelope for a specific operation, see example configuration [TkAuthz.Authorization](https://github.com/cern-eos/xrootd-alicetokenacc/blob/xrootd4/.authz/xrootd/TkAuthz.Authorization)

PR #1201, opened 2020-05-21
Make the xrdmapc help text and manpage match the code.

ellert:
The short option for --list is -l, not -1. The code that defines the
option is src/XrdApps/XrdMapCluster.cc line 481:

      {OPT_TYPE "list",      1, 0, (int)'l'},

$ xrdmapc -1
xrdmapc: Invalid option, '-1'.
$ xrdmapc -l
xrdmapc: '-l' argument missing.

Add a zero-width space (\&) before 'all' at the start of the line, so
that the apostrophe at the start of the line is not interpreted as a
macro.

Fixes rpmlint warning:

$ rpmlint xrootd-client
xrootd-client.x86_64: W: manual-page-warning /usr/share/man/man1/xrdmapc.1.gz 21: warning: macro `all'' not defined
1 packages and 0 specfiles checked; 0 errors, 1 warnings.

Also fixes broken rendering of the man page where the line with the
unknown macro is ignored.

Original:

OPTIONS

       -1 | --list [all|m|s]

       -q | --quiet
            Does not print error messages to cerr; errors appear inline.

Updated:

OPTIONS

       -l | --list [all|m|s]
            'all' lists managers and servers (default), 'm' lists only man‐
            agers and 's' lists only servers.

       -q | --quiet
            Does not print error messages to cerr; errors appear inline.

### DIFF
--- a/docs/man/xrdmapc.1
+++ b/docs/man/xrdmapc.1
@@ -16,9 +16,9 @@ The \fBxrdmapc\fR utility queries the XRootD redirector (status/subscribers/path
 when specified, uses <host>:<port> to determine the locations of path and does optional verification.
 
 .SH OPTIONS
-\fB-1\fR | \fB--list\fR [\fB\fR\fIall\fR|\fIm\fR|\fIs\fR]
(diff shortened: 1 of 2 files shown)


abh3:
@ellert Thanks for the patch!

Issue #1176, opened 2020-04-12
XrdHttp crash when dealing with incomplete headers


esindril:
This crash happened in one of the DOMA tests. Apparently, there an issue with the "state machine" represented by the CurrentReq.reqstate since its value is "-1" and no external handler is called for the HTTP request but also no Bridge object is created. Note there is an external handler that should deal with such GET requests but since reqstate is non-0 this never gets invoked. Below you can find a backtrace of the crash and other relevant information. This happens with XRootD 4.11.3. A full coredump is available here:

`lxfsre09a02.cern.ch:/var/spool/abrt/ccpp-2020-04-09-20:06:50-3949`

The backtrace:
```
(gdb) f 0
#0  0x00007f2cfdae8e8f in XrdHttpReq::ProcessHTTPReq (this=this@entry=0x7f2cdd8a9520) at /usr/src/debug/xrootd-4.11.3/src/XrdHttp/XrdHttpReq.cc:1240
1240                if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
(gdb) bt
#0  0x00007f2cfdae8e8f in XrdHttpReq::ProcessHTTPReq (this=this@entry=0x7f2cdd8a9520) at /usr/src/debug/xrootd-4.11.3/src/XrdHttp/XrdHttpReq.cc:1240
#1  0x00007f2cfdae1105 in XrdHttpProtocol::Process (this=0x7f2cdd8a9400, lp=<optimized out>) at /usr/src/debug/xrootd-4.11.3/src/XrdHttp/XrdHttpProtocol.cc:988
#2  0x00007f2d00845a59 in XrdLink::DoIt (this=0x7f2ca6385818) at /usr/src/debug/xrootd-4.11.3/src/Xrd/XrdLink.cc:441
#3  0x00007f2d00848def in XrdScheduler::Run (this=0x610e78 <XrdMain::Config+440>) at /usr/src/debug/xrootd-4.11.3/src/Xrd/XrdScheduler.cc:357
#4  0x00007f2d00848f39 in XrdStartWorking (carg=<optimized out>) at /usr/src/debug/xrootd-4.11.3/src/Xrd/XrdScheduler.cc:87
#5  0x00007f2d0080ea67 in XrdSysThread_Xeq (myargs=0x7f2caac5cde0) at /usr/src/debug/xrootd-4.11.3/src/XrdSys/XrdSysPthread.cc:86
[...]

PR #1203, opened 2020-05-25
Avoid the reqstate to go below zero in case of strange headers

ffurano:


ffurano:
Should fix https://github.com/xrootd/xrootd/issues/1176

PR #1205, opened 2020-05-25
Compilation fix

ffurano:


abh3:
Hi Fabrizio, I see you are using a more strict compiler. Mine only cares that the variable was referenced not that it was usefully referenced. You can re-post the pull request and simply delete the two occurrences or I can do that and close the pull request. Your choice!

simonmichal:
Hi Fabritzio, I haven't noticed your PR and fix the compilation problem in https://github.com/xrootd/xrootd/commit/a0d55409a47bcd5c6c95bcc1fcff56b0dcc9f256

I think we can simply close this one.

Cheers,
Michal

xrootd-dev:
Hi Michal,

No he has a point. I am willing to accept the PR because he is technically 
correct. The latest compilers acually check if variables are used on the 
right-hand side of a statement. Previously, they just required a variable
appear in a statement. So, Fabrizzio is correct. I just requested a better 
fix.

Andy

simonmichal:
Hi Andy,

I absolutely agree that Fabritzio is right, it's just I already fixed the problem yesterday in the evening in a0d55409a47bcd5c6c95bcc1fcff56b0dcc9f256 because I haven't noticed there is PR (but I did noticed our fedora rawhide build failing).

My fix from yesterday does exactly what Fabritzio'z PR ;-)

Cheers,
Michal

ffurano:
:-)

PR #1208, opened 2020-05-28
Updates to the libXrdVoms.1 man page:

ellert:
- replace libXrdSecgsiVOMS with libXrdVoms in the man page text
- end all .nf (no fill) sections with .fi (fill)
- correct the header file names in the FILES section
- use __VERSION__ in .TH as in other man pages


abh3:
Thanks @ellert

PR #1216, opened 2020-06-16
Update base plugin list

ellert:


abh3:
Hi Mattias,

Why was libXrdOssSIgpfsT.so removed from the plugin list?

Andy

ellert:
It was not removed. It was moved a few lines down since I sorted the list alphabetically. The list was almost sorted, but not exactly for some reason.

abh3:
Sorry. missed that :-)

PR #1223, opened 2020-06-24
Use existing library to extract DN from peer chain

bbockelm:
OpenSSL libraries don't understand proxy certificates and pull out the wrong subject for the use with the gridmap.  We can use the existing Xrootd libraries instead.

Fixes #1221

PR #1224, opened 2020-06-24
Client (proxy) certificate handing in XrdHttp

esindril:
This includes the changes in #1223, fixes the linking issues and exposes the functionality to the XrdHttp plugin. Moreover it also deals with simple client certificates that are not handled by SSL_get_peer_cert_chain.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -32,6 +32,8 @@
 #include "XrdOuc/XrdOucGMap.hh"
 #include "XrdSys/XrdSysTimer.hh"
 #include "XrdOuc/XrdOucPinLoader.hh"
+#include "XrdCrypto/XrdCryptoX509Chain.hh"
+#include "XrdCrypto/XrdCryptosslAux.hh"
 
 #include "XrdHttpTrace.hh"
 #include "XrdHttpProtocol.hh"
@@ -284,30 +286,29 @@ XrdProtocol *XrdHttpProtocol::Match(XrdLink *lp) {
 int XrdHttpProtocol::GetVOMSData(XrdLink *lp) {
   TRACEI(DEBUG, " Extracting auth info.");
 
-  X509 *peer_cert;
-
   // No external plugin, hence we fill our XrdSec with what we can do here
-  peer_cert = SSL_get_peer_certificate(ssl);
-  TRACEI(DEBUG, " SSL_get_peer_certificate returned :" << peer_cert);
+  STACK_OF(X509) *peer_chain = SSL_get_peer_cert_chain(ssl);
+  TRACEI(DEBUG, " SSL_get_peer_certificate returned :" << peer_chain);
   ERR_print_errors(sslbio_err);
 
-  if (peer_cert) {
+  if (peer_chain) {
     char bufname[256];
-    
+
     // Add the original DN to the moninfo. Not sure if it makes sense to parametrize this or not.
     if (SecEntity.moninfo) free(SecEntity.moninfo);
-    
-    // The original DN is not so univoque. In the case of a proxy this can be either
-    // the issuer name or the subject name
-    // Here we try to use the one that is known to the user mapping
-    
-    
-    
+
+    int mape;
+
+    XrdCryptoX509Chain chain;
+    int count = XrdCryptosslX509ParseStack(peer_chain, &chain);
+    if (!count) {
+           TRACEI(DEBUG, " No certificates found in peer chain.");
+    }
+    const char * dn = chain.EECname();
+    SecEntity.moninfo = strdup(dn);
+
+    TRACEI(DEBUG, " Subject name is : '" << SecEntity.moninfo << "'");
     if (servGMap) {
-      int mape;
-      
-      SecEntity.moninfo = X509_NAME_oneline(X509_get_issuer_name(peer_cert), NULL, 0);
-      TRACEI(DEBUG, " Issuer name is : '" << SecEntity.moninfo << "'");
       
       mape = servGMap->dn2user(SecEntity.moninfo, bufname, sizeof(bufname), 0);
       if ( !mape && SecEntity.moninfo[0] ) {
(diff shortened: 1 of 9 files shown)


ffurano:
The XrdHttp part seems reasonable to me. For the many changes to other parts I'd pass on to someone else

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -283,124 +285,100 @@ XrdProtocol *XrdHttpProtocol::Match(XrdLink *lp) {
...
-          TRACEI(DEBUG, " Mapping name: " << SecEntity.moninfo << " --> " << bufname);
-          if (SecEntity.name) free(SecEntity.name);
-          SecEntity.name = strdup(bufname);
-        }
-        else {
-          TRACEI(ALL, " Mapping name: " << SecEntity.moninfo << " Failed. err: " << mape);
-        }
-      }
-      
+  if (!dn) {
```
bbockelm, line 299: Is falling back to OpenSSL safe/reasonable?  That is, is there a setup where a peer certificate is available and works but the `XrdCrypto` library doesn't?

I mention this because this code will definitely do the wrong thing for proxies.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -283,124 +285,100 @@ XrdProtocol *XrdHttpProtocol::Match(XrdLink *lp) {
...
-    
   }
-  else return 0; // Don't fail if no cert
 
-  if (peer_cert) X509_free(peer_cert);
 
+  // If we could not find anything good, take the last 8 non-space characters of the main subject
+  if (!SecEntity.name) {
+    size_t j = 8;
+    SecEntity.name = strdup("unknown-\0"); // note it's 9 chars
```
bbockelm, line 372: Is it safe to automatically assign names here (and above)?

Recall that parts of the DN are user-controlled ... not sure if a clever user could control their mapped name.

What if an admin wants a dedicated "unmapped" account when the user is not in the gridmap?

src/XrdHttp/XrdHttpProtocol.cc
esindril: If we don't do this then simple certificates don't work. The EEC for a plain certificate is the issuer and what you want is the subject in this case.

src/XrdHttp/XrdHttpProtocol.cc
esindril: This is probably a better question for @ffurano.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: No... the EEC for a plain certificate is still the correct subject (at least, this was done correctly for my testing).

The algorithm for discovering the EEC is (basically) to order the chain and, starting at the end, walking up the chain.  The EEC is the last certificate discovered that is _not_ marked as a CA (a well-defined attribute within X509).

Did your testing reveal issues here?

esindril: Yes, for my plain simple certificate issued by the CERN GridCA the EEC was the CERN GridCA. There was only one certificate in the chain and this was the CA one. Also looking at the description of SSL_get_peer_cert_chain there is an explicit mentioning that one should use SSL_get_peer_certificate:
https://www.openssl.org/docs/man1.0.2/man3/SSL_get_peer_cert_chain.html

src/XrdHttp/XrdHttpProtocol.cc
ffurano: I'm not able to answer about the "clever user" mapping an "unknown-*" name to something, maybe in a badly configured site, as I assume that the machine is safe.
Seems still reasonable to me to fill that field in that way if any other attempt to get a name fails. This applies also to the unencrypted case

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Interesting!  I wonder what my test worked then.

Let me investigate.  I'd like to unify this to a single library function instead of splitting over two approaches.

bbockelm:
@esindril - I fixed a memory leak and got rid of the extra `SSL_get_peer_certificate` call here: https://github.com/esindril/xrootd/pull/2

@ffurano - when checking for leaks, I noticed that `xrdhttpvoms` always leaks a copy of the X509 certificate if there's no VOMS extension present.  See: https://gitlab.cern.ch/lcgdm/xrdhttpvoms/-/merge_requests/1.  I would also note your change for 0.2.6 hardcodes the XRootD 5 plugin extension (`-5.so`).  You might be better served by using the CMake macros which ship with the XRootD RPM as they define the `XROOTD_PLUGIN_VERSION` variable.  This way, the build always "does the right thing" according to the compile-time version of xrootd.

I think I'm ready for this to go in once the above PR is merged.

abh3:
Actualy, we are going to retire the XrdHttpVoms plugin as liXrdVoms 
plugin works as the 1-for-1 replacement and does all the new things in the 
right way. So, no urgent need to update it. It's available in the 4.12.x 
series as well as in R5.

bbockelm:
@abh3 - sounds good!  I just took a look at the `XrdVomsHttp` code and it doesn't appear to have the same memory leak.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Ok - I think I understand this code better now.  I didn't see before that it always had a prefix of `unknown-`.

bbockelm:
@esindril - I think I'm happy with this.  Looks ready to merge from my end.

esindril:
I can also confirm everything works as expected. Looking forward to the bug fix release. Thanks!

abh3:
I can't say this pull request makes me happy. We've been trying to consolidate all of the x509 stuff in as few places as possible and this pull requests further splatters x509 dependencies all over the http code. It's going to be a long-term maintenance nightmare. It's certainly makes consolidation monumentally more difficult. Sure, this was the quick way of doing this but leaves the maintainers holding the bag and that is not a good approach. For instance, why wasn't the parsing logic put in the VOMS extractor plugin as it's the one that has to deal with this? At least the dependencies would be in a single place. Also, I still don't fully understand what problem this is trying to solve as we've been using proxy certificates for the longest time with the code, at least via the VOMS plugin. So, it's not even clear what is being gained here. So, please some background and explanation would be warranted; especially for such a far reaching pull request.

esindril:
The initial problem was that the HTTP authentication part was not able to properly match a given certificate (it could also be a proxy certificate or multi-delegated proxy certificate) to the information from a gridmap-file and therefore extract the proper local user identity stored in the gridmap-file. Due to this, jobs submitted for example by ATLAS people through FTS that uses multi-delegated proxy certificates (NOTE: through HTTTP) always fail since they don't get mapped to the correct local identity. I agree this is not ideal, but I couldn't see a different way around it. Also the method were this matching is done (GetVOMSData) is a bit misleading and it doesn't necessarily have anything to do with VOMS.

Nevertheless, if there is any other way around this, by using any other piece of code, I am happy to try it out, but for the moment this is blocking ATLAS to submit FTS jobs using HTTP.

abh3:
OK, I now understand the issue. It's the bloody gridmap file processing that has haunted us from the day it was introduced by Globus (gotta love those folks).  What I really want is a maintainable patch. To help with this, could you resubmit this pull as a new one so we don't see all the amendments as it makes it very difficult to figure out where we can isolate the code. I'll close the old one once you do. Frankly, we don't need the history on this pull request. At that point I can recommend alternatives. Also, ATLAS has been living with this issue for some time. It's not like it's stopping them from doing TPC (albeit maybe not HTTP TPC) but that is not on a critical path. I am involved with that WLCG group and there are a lot of politics there and politics shouldn't make us deploy something that will be hard to maintain just because. 

Now, there are two more issues here 1) where is the R5 version of this patch? 2) have you tested that this does not break GSI?

Mind you, we cannot accept a backport-only patch that is deemed critical as this would make R5 irrelevant. So, let's just slow down a bit and get a workable solution across all releases.

abh3:
Thank you for resubmitting to the master branch!

PR #1226, opened 2020-06-24
Use multiple sequence numbers, one for each monitor destination

djw8605:
Adds late binding to the header creation.  The sequence number is individual to a monitor destination.  With this change, we should be able to use the sequence number to estimate the number of missing packets at the monitor destination.


abh3:
Hmmm, I should have gotten back to you much sooner with the answer about sequence numbers. It's too long to post here so I will email you and see if that changes things. I think it does as this may be done much more simply and might not even have to be done depending on what the need is.

abh3:
I'm closing this as commit 9baea65f addressed this issue in a simpler way.

PR #1231, opened 2020-06-30

esindril:
…icate and properly

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -32,7 +32,6 @@
 #include "XrdSys/XrdSysE2T.hh"
 #include "XrdSys/XrdSysTimer.hh"
 #include "XrdOuc/XrdOucPinLoader.hh"
-
 #include "XrdHttpTrace.hh"
 #include "XrdHttpProtocol.hh"
 
@@ -54,7 +53,6 @@
 #include <fcntl.h>
 
 #define XRHTTP_TK_GRACETIME     600
-
 
 
 /******************************************************************************/
@@ -91,8 +89,6 @@ char *XrdHttpProtocol::sslcafile = 0;
 char *XrdHttpProtocol::secretkey = 0;
 
 char *XrdHttpProtocol::gridmap = 0;
-XrdOucGMap *XrdHttpProtocol::servGMap = 0;  // Grid mapping service
-
 int XrdHttpProtocol::sslverifydepth = 9;
 BIO *XrdHttpProtocol::sslbio_err = 0;
 XrdHttpSecXtractor *XrdHttpProtocol::secxtractor = 0;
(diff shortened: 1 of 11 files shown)


esindril:
This pull request replaces #1224 and is done against the master branch.

abh3:
Thanks Elvin, that makes it easier. The pull request did raise some Travis failures which don't seem to be related to the request per se but now we need to investigate why Travis rejected the pull request, sigh.

abh3:
Please correct this error:
/home/travis/build/xrootd/xrootd/src/XrdHttp/XrdHttpProtocol.cc:301:8: error: 
      use of undeclared identifier 'myCryptoFactory'
  if (!myCryptoFactory->X509ParseStack()((void*) ssl, &chain)) {

esindril:
Yes, I fixed it now.

abh3:
Thanks Elvin. So, let's get the basics out of the way, Have you guaranteed that the Crypto patches did not break GSI? From my view gsi traverses some of the changes and we want to make sure those changes didn't break  gsi functionality.

esindril:
The basic tests that I've done with GSI work as before. The Crypto patches add just one new API (X509ParseStack) so in principle should not affect any of the previous behavior. Initially, I favored a more ad-hoc approach where all this would still be contained in the Http part but Brian wanted to have this functionality logically grouped in the Crypto part as conceptually that makes more sense.

abh3:
True. Brian has his own agenda, not that I'm quibling about this one. Good 
to know. So, I will put my focus on what we can do to make this future 
proof. However, I am at the end of my useful attention span (it is 3 AM) 
so it will have to wait for tomorrow.

Andy

esindril:
Hi Andy,

Just for me to understand, you propose to drop the changes in XrdCrypto and keep all this certificate handling business isolated in the XrdHttp part?

Thanks,
Elvin

abh3:
Hi Elvin,

Not at all. You can't do what you want to do without using the XrdCrypto 
library. I was saying that I'd like XrdHttpProtocol.cc refactored so that 
al the security related stuff, especially that deals with XrdCrypto be 
moved to a separate file for clarity. The changes to XrdCrypto are fine 
other than passing the ssl object to it; which can be easily avoided.

Apparently, my final "points" comments were not clear. If not, other than 
this, please tell me what is not still clear.

Andy

esindril:
Ok, I tried to touch on all the points you mentioned. I moved the code related to gridmap handling into it's own function and in the XrdHttpSecurity.cc file. But moving the InitSecurity is not that straight forward since the newly added XrdHttpProtoInfo namespace and the connected variables make all this quite difficult. I quickly found my self pulling  "Ariadnae's thread" so I stopped before going to far. The http security part is splattered all over the place and the global variables don't make it any easier to consolidate this. Let me know how this new pull request looks, and I can iterate some more on it.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1778,6 +1661,14 @@ int XrdHttpProtocol::Configure(char *parms, XrdProtocol_Config * pi) {
 /******************************************************************************/
 
 bool XrdHttpProtocol::InitSecurity() {
```
abh3, line 1663: Moving this should be pretty trivial so I don't understand why it seemed difficult. I guess because the sprinkling of variables is rather baroque. The key is understanding what InitSecurity() relies on (and it's not much):
a) You need to include XrdHttpTace.hh
b) You need to add
namespace XrdHttpProtoInfo
{
extern XrdTlsContext *xrdctx;
}
using namespace XrdHttpProtoInfo;
c) You need to include XrdTlsContext.hh
d) You need to include XrdOucGMap.hh which can then be deleted from XrdHttpProtocol.cc itself

That's pretty much it! So please try again.

abh3:
Ooops, forgot to post final comments. So, we are getting there. You are correct that the code is a total snarl. That's what we are trying to fix and now you're the unwitting part of that :-) So, I think one of the complications is the (faulty) assumption that gridmap needs to be called before xecxtractor. Honestly, secxtractor should not depend on any other plugin being called before it. Also, gridmap is the final arbiter of the dn. So, it doesn't matter what the secxtractor puts in the dn field. It gets over-ridden.  Mind you that the libXrdVoms incorrectly substitutes tabs for spaces in the dn. I have no idea why it does that but it's plain wrong especially when it doesn't even extract the name!

So, if you reverse the calls I think you will find that you can get rid of a lot of garbage code.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -0,0 +1,137 @@
...
+  XrdTlsPeerCerts* pc = lp->getPeerCerts();
+
+  if (myCryptoFactory && pc &&
+      !myCryptoFactory->X509ParseStack()(pc, &chain)) {
+    TRACEI(DEBUG, "No certificate found in peer chain.");
+    delete pc;
+    return;
+  }
+
+  delete pc;
```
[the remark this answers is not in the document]
esindril: There is (was) a delete for both the successful and the failure case, I don't see the leak here. Anyway not it's on the stack.

src/XrdHttp/XrdHttpProtocol.cc
esindril: Move in the new PR.

abh3:
As for my comment about reversing things. From my perspective, gridmap doesn't change any fields in the SecEntity other than "name".  If that is wrong then all bets are off. Note that GSI calls gridmap ahead on VOMS but doesn't fix-up any changes in the name field. I'm OK with that as long as it gets documented. VOMS should not be touching the "name" field.

esindril:
Unfortunately, the secxtractor does modify the name which is stored in the SecEntity and that is why the processing of the gridmap file must come before the secxtractor gets called. If this happens then the gridmap file contents is useless.

abh3:
It doesn't after I added my fix. It should not be replacing spaces with tabs and that is all it did. That code has been removed.

abh3:
Actually, which secxtractor are you talking about?

esindril:
Which fix are you referring to? We use this secxtractor and it does modify the name ...
https://gitlab.cern.ch/lcgdm/xrdhttpvoms/-/blob/develop/XrdHttpVoms.cc#L157

abh3:
Gees, that is no longer supported and you are on your own if you use that one. The official secxtractor is libXrdVoms.so and it works in the 4.12.x series. It is the only secxtractor that support multi-VO certs.

abh3:
Plus, any VOMS secxtrator that modifies the name field is plain wrong. The gridmap file is the source of record. It makes no sense to override it.

esindril:
Fine, I can switch to this one (libXrdVoms.so), I just need to test it out and make sure it does not bring any other surprises. :) Concerning this pull requests, what else needs to be updated?

abh3:
Don't know yet. I will review the changes once I get some sleep. Right now I'm pretty blurry eyed. Make sure you the one in git head (that is libXrdVoms.so). It's now part of the core xrootd package. It has the latest fix that people started to complain about. The fix should be in 4.12.4 as well (well if Michal sees my email before the release).
Sorry that it's taking longer that you would like but then again from what I understand the certs you are seeing are coming from an R&D project and it has nothing to do with production ATLAS. You could just as easily given that group a test server to play with.

esindril:
I will close this one in favor of two new pull requests one for R4 and one for R5.

PR #1234, opened 2020-07-02
Adding fbuff argument to monitoring to restrict maximum size of fstream packet

djw8605:
@ddavila0 can you test this change since you already have a testbed setup.

### DIFF
--- a/src/XrdXrootd/XrdXrootdConfig.cc
+++ b/src/XrdXrootd/XrdXrootdConfig.cc
@@ -1345,6 +1345,7 @@ int XrdXrootdProtocol::xlog(XrdOucStream &Config)
 
    Purpose:  Parse directive: monitor [all] [auth]  [flush [io] <sec>]
                                       [fstat <sec> [lfn] [ops] [ssq] [xfr <n>]
+                                      [fbuff <sz>]
                                       [ident <sec>] [mbuff <sz>] [rbuff <sz>]
                                       [rnums <cnt>] [window <sec>]
                                       dest [Events] <host:port>
@@ -1362,6 +1363,7 @@ int XrdXrootdProtocol::xlog(XrdOucStream &Config)
                             ssq    - computes the sum of squares for the ops rec
                             xfr <n>- inserts i/o stats for open files every
                                      <sec>*<n>. Minimum is 1.
+         fbuff  <sz>        size of message buffer for file stream monitoring.
          ident  <sec>       time (seconds, M, H) between identification records.
          mbuff  <sz>        size of message buffer for event trace monitoring.
          rbuff  <sz>        size of message buffer for redirection monitoring.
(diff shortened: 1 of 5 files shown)


abh3:
Before I comment on some of the coding problems in this PR, what is the motivation? Why would you ever want to sacrifice buffer size for forced packet frequency beyond what the flush specification gives you?

djw8605:
We have found that large UDP packets are fragmented and rarely make it to the destination, at least in the network conditions we are seeing.

src/XrdXrootd/XrdXrootdConfig.cc
```diff
@@ -1441,6 +1443,15 @@ int XrdXrootdProtocol::xmon(XrdOucStream &Config)
                    if (mrType) monRBval = static_cast<int>(tempval);
                       else     monMBval = static_cast<int>(tempval);
                   }
+         else if (!strcmp("fbuff", val))
+                {if (!(val = Config.GetWord()))
+                    {eDest.Emsg("Config", "monitor fbuff value not specified");
+                     return 1;
+                    }
+                 if (XrdOuca2x::a2sz(eDest,"monitor fbuff", val,
+                                             &tempval, 1024, 65536)) return 1;
```
abh3, line 1452: The upper limit should be 65472 not 65536 because it anyone specifies that the packet length in the header will become zero as it's a short. The lower limit should ideally be no less than 4096. If you network drops 4K datagram packets something is really wrong with your network.

src/XrdXrootd/XrdXrootdMonFile.cc
```diff
@@ -174,14 +175,15 @@ void XrdXrootdMonFile::Close(XrdXrootdFileStats *fsP, bool isDisc)
...
-void XrdXrootdMonFile::Defaults(int intv, int opts, int xfrcnt)
+void XrdXrootdMonFile::Defaults(int intv, int opts, int xfrcnt, int fbuff)
 {
 
 // Set the reporting interval and I/O counter
 //
    repTime = intv;
    xfrCnt  = xfrcnt;
    xfrRem  = xfrcnt;
+   fBuff   = fbuff;
```
abh3, line 186: This isn't going to work at all. If you wind up through the call stack your will notice that Defaults() can be called with fbuff having a value of zero. So this should be:
fBuff = (fbuff <= 0 ? 65472 : fbuff);

Also, why did you call something that is a "size" an fbuff? How about changing it to the more accurate fbsz and fBsz or anything else that tells you its a quantity.

src/XrdXrootd/XrdXrootdConfig.cc
djw8605: I agree that networks suck.  There are many articles on why fragmented UDP packets are difficult to use.  But, standard MTU is normally 1500, so just under 1500 usable.  We want to give the option to have smaller than MTU packets, to stop most fragmentation.

src/XrdXrootd/XrdXrootdMonFile.cc
djw8605: I took your suggestion, fbsz and fBsz.  And I added that if statement you suggested.  My pull request would crash when no fbsz is specified in the config without it.

djw8605:
@abh3 just another poke to merge this.

abh3:
Done!

PR #1238, opened 2020-07-02

esindril:
…icate and properly

This pull request is aimed at R5.

### DIFF
--- a/src/XrdCrypto/XrdCryptoFactory.cc
+++ b/src/XrdCrypto/XrdCryptoFactory.cc
@@ -315,6 +315,16 @@ XrdCryptoX509ParseFile_t XrdCryptoFactory::X509ParseFile()
 }
 
 //______________________________________________________________________________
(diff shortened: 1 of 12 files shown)


esindril:
There is still the issue that if the certificate is not retrieved and we fail then no token (macaroon, scitoken etc) access is possible. For the moment this is allowed, therefore tokens work with the PR. This happens here:
https://github.com/xrootd/xrootd/pull/1238/files#diff-7b19b9632974132e3ff4f2f2f1924e1aR108

esindril:
Hi Andy,

This is now ready from my side. Note this PR is done against the current master.

Thanks,
Elvin

abh3:
Hi Elvin,

Thanks! I hear that this will need to be backported. Do you think you 
could hustle up a version for 4.12.x stable?

Andy

esindril:
I am preparing that one now and this will be in #1239 done against the stable-4.12.x branch.

PR #1239, opened 2020-07-02

esindril:
…icate and properly

Conflicts:
	src/XrdHttp/XrdHttpProtocol.cc
	src/XrdHttp/XrdHttpProtocol.hh

This PR provides the same functionality as #1238 but is done against stable-4.12.x and doesn't use any Tls object since they don't exist in the R4 version.

### DIFF
--- a/src/XrdCrypto/XrdCryptoFactory.cc
+++ b/src/XrdCrypto/XrdCryptoFactory.cc
@@ -315,6 +315,16 @@ XrdCryptoX509ParseFile_t XrdCryptoFactory::X509ParseFile()
 }
 
 //______________________________________________________________________________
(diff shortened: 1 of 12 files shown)


bbockelm:
Did this get closed by accident?  The commit message in https://github.com/xrootd/xrootd/commit/e6a7b0b71e4e0693d846b6d941ec6438d3033ee9 closed this PR but clearly is unrelated.

abh3:
Inadvertent close.

xrootd-dev:
Hi Brian,

Thanks for pointing this out. The commit message referenced the wrong 
issue.

Andy

esindril:
This is also ready now.

abh3:
Thanks Elvin but you are staying up way too late!

PR #1242, opened 2020-07-03
[Xrootd HTTP] Allow redirect to local filesystem when using HTTP

pkramp:
With these changes, an HTTP-client can be redirected to the file://-protocol by the xrootd redirector employing the XrdCmsRedirLocal-plugin.

The plug-in must be configured to allow HTTP redirects to local filesystem via the xrdcmsredirlocal.httpredirect true option.
Otherwise, default redirect behaviour is used.

Only minimal changes to other sources than the plug-in necessary.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -661,6 +661,13 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   else
     redirdest = "Location: http://";
   
+  // port -1 signals redirect to local filesystem
+  // switch to file:// protocol
+  if (port == -1)
+  {
+    TRACE(REQ, " XrdHttpReq::Redir Switching to file:// ");
+    redirdest = "Location: file://";
+  }
   // Beware, certain Ofs implementations (e.g. EOS) add opaque data directly to the host name
   // This must be correctly treated here and appended to the opaque info
   // that we may already have
@@ -681,7 +688,7 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   else
     redirdest += hname;
 
-  if (port) {
+  if (port > 0) {
     sprintf(buf, ":%d", port);
     redirdest += buf;
   }
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -661,12 +661,14 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   else
     redirdest = "Location: http://";
   
-  // port -1 signals redirect to local filesystem
-  // switch to file:// protocol
-  if (port == -1)
+  // port < 0 signals switch to full URL
+  if (port < 0)
   {
-    TRACE(REQ, " XrdHttpReq::Redir Switching to file:// ");
-    redirdest = "Location: file://";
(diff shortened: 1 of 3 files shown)


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -661,6 +661,13 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   else
     redirdest = "Location: http://";
   
+  // port -1 signals redirect to local filesystem
+  // switch to file:// protocol
+  if (port == -1)
```
abh3, line 666: Ummm, I think the comment is wrong here. When port is < 0 then the target is a full URL not a host?token:port target. I presume you do set port to -1 and the target is a full URL with "file:://" as the protocol. But it might also be a regular full URL. The code here should take that into account and not merely assume it's a local file redirect. Technically, the check for '-1' is correct because it is documented as such. However, we are considering loosing it and allowing for any value less than zero to accommodate future needs. So, I'd suggest checking it's merely negative so we don't have to revisit this when it changes.

pkramp: Hi Andy, thanks.

do you merely want a change such as this:
```
  // port < 0 signals switch to full URL
  if (port < 0)
  {
    TRACE(REQ, " XrdHttpReq::Redir Switching to file:// ");
    redirdest = "Location: file://";
  }
```

And I want to make sure I understand you correctly:
Wouldn't checking for port < 0 make everything more ambiguous?
Because new use cases that would utilize other ports < 0, would trigger the local file redirect as well?

Or do you mean that there will only be the one use case for ports < 0, and that will remain to be local file redirect?
And I am unclear on the term "regular full URL". In what other cases does xrootd handle non host?token:port targets than in the file://-protocol?

Cheers,
Paul

abh3: Hi Paul,

The new definition is that if the port number is < 0 then the redirect target is a full URL. In your case, you need to check if the URL starts with "file://" and effect the redirection to the local file. Otherwise, the redirection should proceed as a normal full URL redirect for some other protocol. So, do not assume that just because you got a full URL it is for local file redirection.

pkramp: Hi Andy,

ok I see what you mean.
I hope the newest commit will be sufficient.
I previously did not return the file://-prefix from the plug-in, just the path.

In the root-protocol, a port of -1 was sufficient.
I can get started on changing this too, checking for the presence of "file://" there as well.

Cheers,
Paul

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -661,6 +661,15 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   else
     redirdest = "Location: http://";
   
+  // port < 0 signals switch to full URL
+  if (port < 0)
+  {
+    if (string(hname).find("file://") != string::npos)
```
abh3, line 667: Well, that is not a safe comparison as you may get a match even though it's not a file-local URL. You should do this

if (strncmp(hname, "file://", 7) == 0)

that's better than the "std" alternative 
if (string(hname).rfind("file://", 0) == 0)
as you don't have to allocate memory to do this simple comparison.

abh3:
I dislike some of the github the managerial stuff but could you click on the button that asks for a re-review to clear the warning that a requested change is pending? I know it's a hassle but then we get a clean record. Once that happens I'll merge the pull.

PR #1260, opened 2020-07-24
Prevent deadlock in Python bindings from XRootD.client.finalize.finalize

chrisburr:
Fixes an issue where if files are not closed before the Python interpreter exits `XRootD.client` causes the Python interpreter to deadlock.

This can be reproduced in both XRootD 4.12.3 and 5.0.0. If there is any intention to make further releases for the 4.x series it would be nice to have this back ported.

You can find some more details about this below and in: https://github.com/scikit-hep/uproot/issues/504

## Reproducer

1. Create a file called `xrootd_bug.py` containing:

```python
import XRootD.client
f = XRootD.client.File()
f.open('root://eospublic.cern.ch//eos/opendata/lhcb/AntimatterMatters2017/data/B2HHH_MagnetDown.root')
```

2. Run `python xrootd_bug.py`

## Technical details

This is caused by the `atexit` function shutting down the worker thread pool:

https://github.com/xrootd/xrootd/blob/912673bfbc75ff0d9f661556716238f65e671a6a/bindings/python/libs/client/finalize.py#L28-L35

The debug output shows that it shuts down all the threads and then sends the call to close the file:
```
***** Calling client.__XrdCl_Stop_Threads()
[2020-07-24 08:38:40.953763 +0200][Debug  ][JobMgr            ] Stopping the job manager...
[2020-07-24 08:38:40.954512 +0200][Debug  ][JobMgr            ] Job manager stopped
[2020-07-24 08:38:40.954574 +0200][Debug  ][TaskMgr           ] Stopping the task manager...
[2020-07-24 08:38:40.954863 +0200][Debug  ][TaskMgr           ] Task manager stopped
[2020-07-24 08:38:40.954921 +0200][Debug  ][Poller            ] Stopping the poller...
***** Finished client.__XrdCl_Stop_Threads()
[2020-07-24 08:38:40.990681 +0200][Debug  ][File              ] [0x84000b80@root://xrootd-lhcb.cr.cnaf.infn.it:1094//storage/gpfs_lhcb/lhcb/disk/LHCb/Collision18/LEPTONIC.MDST/00077054/0001/00077054_00019123_1.leptonic.mdst?xrdcl.requuid=17781476-5b88-4609-98f3-6b4d78e188eb] Sending a close command for handle 0x0 to xrootd-lhcb.cr.cnaf.infn.it:1094
[2020-07-24 08:38:40.990769 +0200][Debug  ][ExDbgMsg          ] [xrootd-lhcb.cr.cnaf.infn.it:1094] MsgHandler created: 0x83d0e9c0 (message: kXR_close (handle: 0x00000000) ).
```

As the call to send the close command is after the thread pool has been shut down there is there is nothing to actually process the request and the process hangs forever.

## Solution in this PR

This PR works around the problem by looping over all objects known to Python and closing any `XRootD.client.file.File` objects. There may be other ways this bug can manifest, someone more familiar with the code should think about if there are any other commands added to the thread pool when objects are deleted.

### DIFF
--- a/bindings/python/libs/client/finalize.py
+++ b/bindings/python/libs/client/finalize.py
@@ -21,9 +21,14 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 #------------------------------------------------------------------------------
+from __future__ import absolute_import
 
+import gc
 import atexit
+
 from pyxrootd import client
+from .file import File
+
 
 @atexit.register
 def finalize():
@@ -32,4 +37,12 @@ def finalize():
         no Python APIs are called after the Python Interpreter gets
         finalized.
      """
+     # Ensure there are no files left open as calling their destructor will
+     # cause "close" commands to be sent.
+     # If this isn't done, there will be no running threads to process requests
+     # after this function returns so the interpreter will deadlock.
+     for obj in gc.get_objects():
+         if isinstance(obj, File) and obj.is_open():
+             obj.close()
+
      client.__XrdCl_Stop_Threads()


jpivarski:
@jpivarski Linking myself here so that I see when it's done.

bbockelm:
@oshadura - you may be interested in watching this one.  It could explain some of the issues you were seeing when dask workers were scaling down.

chrisburr:
To add, it might be worthwhile fixing this somewhere deeper down on the C++ side in the destructors `pyxrootd.client.File` or `XrdClFile`.

PR #1273, opened 2020-08-27
Fix 32 bit compilation

ellert:
/builddir/build/BUILD/xrootd-5.0.1/src/XrdPfc/XrdPfc.cc:649:50: error: format '%lu' expects argument of type 'long unsigned int', but argument 9 has type 'size_t' {aka 'unsigned int'} [-Werror=format=]


ellert:
Hi!

Could you merge this to stable-5.0.x?
I had to keep this patch for the xrootd 5.0.2 build.

Issue #1293, opened 2020-09-28
XrdTpc does not encode the redirection URL


esindril:
Seen in ATLAS tests where the davix client can not parse the redirection URL from an EOS instace since the TPC code does not encode the redirection information. The following code should be dealing with this:
https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L140

As an example davix will fail due to unescaped characters in the URL:
```
 export SRC=https://golias100.farm.particle.cz/dpm/farm.particle.cz/home/atlas/atlasscratchdisk/1M
$ export DST=https://eosatlas.cern.ch:443/eos/atlas/atlasscratchdisk/SAM/x
$ export TSRC=$(curl --silent --cert /tmp/x509up_u$(id -u) --key /tmp/x509up_u$(id -u) --cacert /tmp/x509up_u$(id -u) --capath /etc/grid-security/certificates -X POST -H 'Content-Type: application/macaroon-request' -d '{"caveats": ["activity:DOWNLOAD"], "validity": "PT30M"}' "$SRC" | jq -r '.macaroon')
$ export TDST=$(curl --silent --cert /tmp/x509up_u$(id -u) --key /tmp/x509up_u$(id -u) --cacert /tmp/x509up_u$(id -u) --capath /etc/grid-security/certificates -X POST -H 'Content-Type: application/macaroon-request' -d '{"caveats": ["activity:UPLOAD,DELETE,LIST"], "validity": "PT30M"}' "$DST" | jq -r '.macaroon')

$ davix-cp --trace all --capath /etc/grid-security/certificates --copy-mode pull -H 'Credentials: none' -H "Authorization: Bearer $TDST" -H "TransferHeaderAuthorization: Bearer $TSRC" $SRC $DST
...
[...]

PR #1298, opened 2020-10-04
URI-quote outgoing opaque data for HTTP-TPC

bbockelm:
Ensure that we URI-quote the outgoing opaque data when we redirect as part of a HTTP-TPC.

This matches the same quoting strategy for normal HTTP redirects.

Fixes: #1293


bbockelm:
@esindril - can you take a look at this?  I took your advice in the end and utilized `curl`'s built in URI escape mechanism.

PR #1319, opened 2020-11-02
Pass the environment pointer internally in XrdCksManOss

bbockelm:
With this, if the `XrdCksData` object has an `XrdOucEnv` pointer then we pass it down to the `XrdCksManOss::Calc` interface.

### DIFF
--- a/src/XrdCks/XrdCksManOss.cc
+++ b/src/XrdCks/XrdCksManOss.cc
@@ -111,7 +111,7 @@ int XrdCksManOss::Calc(const char *Lfn, XrdCksData &Cks, int doSet)
 
 /******************************************************************************/
   
-int XrdCksManOss::Calc(const char *Pfn, time_t &MTime, XrdCksCalc *csP)
+int XrdCksManOss::Calc(const char *Pfn, time_t &MTime, XrdCksCalc *csP, XrdOucEnv *envP)
 {
    class inFile
         {public:
@@ -127,9 +127,12 @@ int XrdCksManOss::Calc(const char *Pfn, time_t &MTime, XrdCksCalc *csP)
    size_t ioSize, calcSize;
    int    rc;
 
+   if (envP) {eDest->Emsg("XrdCksManOss::Calc", "Got envP");}
+   else {eDest->Emsg("XrdCksManOss::Calc", "No envP");}
+
 // Open the input file
 //
-   if ((rc = In.fP->Open(Lfn,O_RDONLY,0,openEnv))) return (rc > 0 ? -rc : rc);
+   if ((rc = In.fP->Open(Lfn,O_RDONLY,0, envP ? *envP : openEnv))) return (rc > 0 ? -rc : rc);
 
 // Get the file characteristics
 //
--- a/src/XrdCks/XrdCksManOss.hh
+++ b/src/XrdCks/XrdCksManOss.hh
@@ -61,7 +61,7 @@ virtual int         Ver( const char *Lfn, XrdCksData &Cks);
 virtual            ~XrdCksManOss() {}
 
 protected:
-virtual int         Calc(const char *Lfn, time_t &MTime, XrdCksCalc *CksObj);
+virtual int         Calc(const char *Lfn, time_t &MTime, XrdCksCalc *CksObj, XrdOucEnv *envP);
 virtual int         ModTime(const char *Pfn, time_t &MTime);
 
 private:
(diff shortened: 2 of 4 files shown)


abh3:
OK, I see what you are trying to do but the approach breaks ABI. So, it's a non-starter. I will come up with a ABI-compatible solution which I know I can do. Thank you for bringing the fact that we don't completely pass down the environment to the actual computation. In fact, there is still one more step that needs to get done here and that is to pass that environment to the actual Oss layer that does the physical I/O. I'll put that in also.

bbockelm:
I don't think this breaks ABI -- the `Calc` method override that I changed really should be named `CalcImpl` to make it clear it's not defined by `XrdCks`.

abh3:
While it technically doesn't break the ABI per se the problem is that the ManOss inherits the Manager API and the change reverts the the override of Calc to its default implementation. While that is counteracted by the change to the default manager implementation it doesn't help should someone replace the manager plugin or one of the Calc's the default manager implementation would use. In other words, the change is a semantic change that makes it difficult to know what the end result will be when site-specific plugins are specified to override various default implementations. There  is a more consistent way of doing this that keeps the overt semantics the same but yet passes the env to the Oss open() as needed. That's what I mean to say. Let me work on a solution that doesn't change the override behavior.

abh3:
Well, try as much as I want the correct environment cannot be passed irrespective of any changes to the XrdCks objects. Why? Because the problem is far more fundamental. The checksum calculation occurs in the background and the background task is not provided with the SecEntity object because, well, at that point it isn't needed. In your particular case you want it because you need to set the uid/gid. With some machinations I could do that but the setup wasn't created to pass pointers to complex objects. In R5 the correct mechanism is to set the correct uid/gid in the SecEntity object either via the authentication handler or, more likely in your case, in the post authentication handler plugin. That way the uid/gid is determined once and that information is easy to pass around. I will need more time to figure out how best to accomplish this. In any case, the revised solution did not touch any public headers nor changed any semantics; unfortunately, it also did not accomplish what was needed.

abh3:
OK, the sad story is the way we are approaching this is forbidden in the XRootD framework. Background tasks must be self-contained. They are forbidden to rely on the existence of any objects outside of their environment. This is because objects outside their environment may disappear. For instance, the object that launched a background task may go away because the client decided to close the connection. 

So, the solution here is to make sure the uid/gid is correctly established in the post-authentication plugin (which is what you really want to do in any case). Then I can pass a stripped down version of the SecEntity object with that uid/gid in it so you can use it for setfsuid/fsgid (btw those are deprecated now so they may not be there in the future).

This does not address a similar problem for bridged protocols (e.g. HTTP). Here I would have to also provide a plugin mechanism for you to decorate the SecEntity object when the protocol logs into the bridge. It's doable and the interface would not be any different from the one the xroot protocol uses.

Let me know how you want to proceed.

bbockelm:
> OK, the sad story is the way we are approaching this is forbidden in the XRootD framework

Yes - this is what I was suspecting as I was looking at the potential `XrdSecEntity` object lifetime issues.

> Then I can pass a stripped down version of the SecEntity object with that uid/gid in it so you can use it for setfsuid/fsgid.

If we are going to be serializing the uid/gid portions of the `SecEntity`, what's the added cost to serialize the whole thing (or at least make a copy of it)?  Doesn't SSS have to do something similar?  I'd be surprised if there was no serialization code existing already for the `XrdSecEntity` that can be used to pass data to the background task.

> (btw those are deprecated now so they may not be there in the future).

They were marked as deprecated in 2013 because of changes to signal handling in Linux 2.0 (1996).  However, while the original use case is now obsolete, I see no replacement option for having a per-thread FS ID and it being a valid use case.  Given there's no other alternate and Linux's history of carrying these forward forever ... I'm not too worried.

abh3:
Well, I was trying  to kill several birds with one stone. I was looking at the OSG plugin and it does a getpwnam() every single time which is rather ridiculous. That should be done once. Plus, if the plugin that you are using is this one:
https://github.com/opensciencegrid/xrootd-multiuser/blob/master/src/multiuser.cpp
it just doesn't follow the preferred plugin rules in R5 which allow you to get rid of as lot of code. Yes, will work but incredibly poorly. Frankly, instead of looking at putting in a lot of extra code on the xroot side to support this plugin, you would be far better off to just rewrite it to do something more reasonable. Once you do that  then we could potentially integrate it with the main repo.

As far as serializing the SecEntity object, yes sss does that but only once per login. Here we would potentially be serializing it  more than once. While cost of the fixed fields is not that high, the cost of the active fields could be quite high and not always appropriate. Frankly, you would be better off to simply fix the plugin to do something more reasonable.

abh3:
I am closing this pull request as it is no longer relevant. The issue is not what it appears to be. In fact, the multiuser plugin launches the checksum code with the desired uid/gid. Passing additional information is superfluous as it's already running with the desired uid/gid. So, this pull request is not going to fix anything. What we need is the explicit messages issued in the execution sequence to see what went wrong. Checking to code this should have worked fine as it is.

PR #1328, opened 2020-11-16
Update 4.x -> 5.0 in plugin version table

vokac:
XrdHttpGetSecXtractor 4.0 -> 5.0
XrdHttpGetExtHandler 4.8 -> 5.0

Fix issue with 5.0.3 and dmlite where plugin doesn't load with error:
```
Plugin version XrdHttpProtocolTest v5.0.3 is incompatible with domexrdhttp v5.0.2 (must be >= 4.8.x) in exthandlerlib /usr/lib64/libdome-5.so
------ HTTP protocol initialization failed.
201113 13:09:07 3734 XrdProtocol: Protocol XrdHttp could not be loaded
```

### DIFF
--- a/src/XrdVersionPlugin.hh
+++ b/src/XrdVersionPlugin.hh
@@ -100,8 +100,8 @@
         XrdVERSIONPLUGIN_Rule(Optional,  5,  0, XrdPfcGetDecision       )\
         XrdVERSIONPLUGIN_Rule(DoNotChk,  5,  0, XrdgetProtocol                )\
         XrdVERSIONPLUGIN_Rule(Required,  5,  0, XrdgetProtocolPort            )\
-        XrdVERSIONPLUGIN_Rule(Required,  4,  0, XrdHttpGetSecXtractor         )\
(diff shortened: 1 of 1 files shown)


abh3:
The pull request that you submitted would indeed solve the immediate problem. The
issue we had was that it wouldn't solve the essential problem.

There was an edge case that caused certain combinations of "required" vs
"acceptable" versions to fail the check. I really want to thank you for bringing up
this issue and without your observation we would have never caught it. The solution
was far more complicated, unfortunately, but it should provide a more consistent
version checking scenario (well, I hope this is the end of it).

Why is this complicated? Version checking is a compendium of not only what version is
required but of what version are asking for the requirement. The checking logic is
trying to ensure that the version of a shared library does not exceed its own
version. So, not only does a version check the "static" version requirements it also
ensures that the chain of plugins being loaded are actually compatible with each
other based on each one of those requirements.

Unfortunately, the checking logic was faulty when it came down to the minor version.
It was far too restrictive. The patch takes into account what the major version
requirements were (it did not do that).

I know this is more information that you wanted but I wanted to point out that
realistic version checking is far more complicated than what one would expect.

Issue #1336, opened 2020-11-24
Xcache segfault during purge


smithdh:
Hi,

With Pfc of current master (7f614bd).

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f3d248f1700 (LWP 19214)]
XrdPfc::File::DeltaStatsFromLastCall (this=0x0) at /usr/src/debug/xrootd/xrootd/src/XrdPfc/XrdPfcFile.cc:145
145	   Stats delta = m_last_stats;
(gdb) where
#0  XrdPfc::File::DeltaStatsFromLastCall (this=0x0) at /usr/src/debug/xrootd/xrootd/src/XrdPfc/XrdPfcFile.cc:145
#1  0x00007f3d25f38143 in XrdPfc::Cache::copy_out_active_stats_and_update_data_fs_state (this=this@entry=
    0xb5e5d0) at /usr/src/debug/xrootd/xrootd/src/XrdPfc/XrdPfcPurge.cc:567
#2  0x00007f3d25f3ba3c in XrdPfc::Cache::Purge (this=0xb5e5d0)
    at /usr/src/debug/xrootd/xrootd/src/XrdPfc/XrdPfcPurge.cc:781
#3  0x00007f3d25f2e901 in PurgeThread () at /usr/src/debug/xrootd/xrootd/src/XrdPfc/XrdPfc.cc:59
#4  0x00007f3d2d5f2977 in XrdSysThread_Xeq (myargs=0xb61950)
    at /usr/src/debug/xrootd/xrootd/src/XrdSys/XrdSysPthread.cc:86
#5  0x00007f3d2d19eea5 in start_thread () from /lib64/libpthread.so.0
#6  0x00007f3d2c49f96d in clone () from /lib64/libc.so.6
```

Opening PR with one possible fix.

osschar:
I merged your fix but I'll investigate this further as I don't think this was supposed to happen :)

Was this running under any special conditions? I assume you have not enabled dirstats ... they are off by default and I just realized I have been always testing with this on: `pfc.dirstats maxdepth 4`

osschar:
Oh, I see how this happens, null entry in m_active_map means the File is currently being opened so two threads don't try to create two File objects for the same path. Yours was the correct fix.

It's an old bug but it didn't show before because dirstat collection was completely off and now it is on at least for the root directory so we can collect some overall statistics for summary monitoring and for g-stream (when I get to implementing this).

Thanks again!

[...]

PR #1337, opened 2020-11-24
Xcache: segfault during purge and plugin name

smithdh:
The plugin loader needs the .so suffix for libXrdOssCsi.so.

I saw a segfault during purge, I've also put one possible fix commit in this PR (see issue #1336).


osschar:
Thank you!

PR #1339, opened 2020-11-26
Fix case where 0 streams is requested.

snafus:
Fix case where stream_req is 0, which should be interpreted as a single stream. This fixes what is currently a typo in the ternary comparison.


bbockelm:
@simonmichal - this one's been lingering but would go into the "would be nice" category for 4.12.6.

xrootd-dev:
Thanks for pointing it out - noted :-)

Michal

PR #1351, opened 2020-12-07
1 ms = 1000000 ns

ellert:
Fix ms to ns conversion.

### DIFF
--- a/src/XrdSys/XrdSysPthread.hh
+++ b/src/XrdSys/XrdSysPthread.hh
@@ -175,7 +175,8 @@ inline int TimedLock( int wait_ms )
 {
   struct timespec wait, cur, dur;
   get_apple_realtime(wait);
-  wait.tv_nsec += wait_ms * 1000000;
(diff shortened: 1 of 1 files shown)


ellert:
Will XrdSysMutex::TimedLock( int wait_ms ) ever be called with wait_ms > 3294 ms?
3295 · 1000000 + 999999999 = 4294999999 > 4294967295 = 2³² - 1
The overflow can be avoided if
wait.tv_nsec += wait_ms * 1000000;
is replaced by
wait.tv_sec += wait_ms / 1000;
wait.tv_nsec += (wait_ms % 1000) * 1000000;

simonmichal:
@ellert : thanks for the PR, it seems to me you are right, could you update the PR so it also fixes the overflow?

ellert:
> @ellert : thanks for the PR, it seems to me you are right, could you update the PR so it also fixes the overflow?

Added commit to avoid overflow.

PR #1355, opened 2020-12-08
Use correct size for struct

ellert:
The fix for the gcc 11 compiler warning in commit
8f36753d46c183aa89058219e36e9a2980afd250 is incorrect.
Placing the sizeof statement in parenthesis does not correct the bug,
it just obfuscates the code so that the compiler doesn't see it.

The correct fix is to give the unnamed struct a name so that the
sizeof the struct can be used in the calculation.

### DIFF
--- a/src/XrdOuc/XrdOucPsx.cc
+++ b/src/XrdOuc/XrdOucPsx.cc
@@ -703,7 +703,7 @@ bool XrdOucPsx::ParseSet(XrdSysError *Eroute, XrdOucStream &Config)
 {
     char  kword[256], *val;
     int   kval, noGo;
-    static struct {const char *Sopt; const char *Copt; int isT;} Sopts[]  =
+    static struct sopts {const char *Sopt; const char *Copt; int isT;} Sopts[] =
        {
          {"ConnectTimeout",        "ConnectionWindow",1},    // Default  120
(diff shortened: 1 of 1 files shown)


simonmichal:
Indeed you're right, this not an array of pointers, thanks for point it out!

Could you make it: `sizeof(Sopts)/sizeof(*Sopts)` ?

ellert:
That would work. However, I think the current version is more consistent with the rest od the file:
```
$ grep -n 'sztab' src/XrdOuc/XrdOucPsx.cc
336:   struct sztab {const char *Key; long long *Val;} szopts[] =
342:   int i, numopts = sizeof(szopts)/sizeof(struct sztab);
$ grep -n 'traceopts' src/XrdOuc/XrdOucPsx.cc
794:    static struct traceopts {const char *opname; int opval;} tropts[] =
800:    int i, trval = 0, numopts = sizeof(tropts)/sizeof(struct traceopts);
$ grep -n 'sopts' src/XrdOuc/XrdOucPsx.cc
706:    static struct sopts {const char *Sopt; const char *Copt; int isT;} Sopts[] =
735:    int i, numopts = sizeof(Sopts)/sizeof(struct sopts);
```

PR #1362, opened 2020-12-14
Allow 0 issuers in SciTokens config

djw8605:
0 issuers is ok for SciTokens.  No token will be authorized,
but it shouldn't be an error causing xrootd to not start.  A
log message is still reported when 0 issuers are detected.


abh3:
Thanks Derek,

I will merge everything as soon as slow Travis finishes up its checks.

Andy

PR #1363, opened 2020-12-15
Adding ObjAdd to list of scitokens functions exported

djw8605:
- [x] Built
- [ ] Tested

(Still building)

### DIFF
--- a/src/XrdSciTokens/configs/export-lib-symbols
+++ b/src/XrdSciTokens/configs/export-lib-symbols
@@ -1,6 +1,7 @@
 {
 global:
   XrdAccAuthorizeObject*;
+  XrdAccAuthorizeObjAdd*;
 
 local:
   *;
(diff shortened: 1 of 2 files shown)


abh3:
Given that this causes problems, could you explain why this pull request was summited? I see no functional reason for it.

simonmichal:
@djw8605 : could you please instead of merging the master branch into your devel branch, rebase it on master? Merging it as it is now to master would create havoc in history.

djw8605:
Rebased to master.  

This pull request was requested by the latest testing errors that the OSG experienced.  The exact error was that it could not find the object `XrdAccAuthorizeObjAdd`, which I confirmed with `nm -D`.  The next error in the xrootd log was a failure to find version information for the `XrdAccAuthorizeObjAdd` function.

simonmichal:
Thanks!

PR #1364, opened 2020-12-15
Add UUID library to build xrdpfc_print

djw8605:
This was required in order to build master just now.

### DIFF
--- a/src/XrdPfc.cmake
+++ b/src/XrdPfc.cmake
@@ -77,7 +77,8 @@ target_link_libraries(
   xrdpfc_print
   XrdServer
   XrdCl
-  XrdUtils )
+  XrdUtils
+  ${UUID_LIBRARY} )
 
 #-------------------------------------------------------------------------------
 # Install


osschar:
@djw8605 @simonmichal why is this needed? Apparently it is pulled in through libXrdCl.

libXrdPfc is also linked against libXrdCl and not libuuid.

And so is xrdfs.

abh3:
Correct, these two pull requests failed to apply due to new depdnencies 
that were not fully resolved. So, we are not applyingthem until we can 
figure out why these were submitted?

Andy

abh3:
Could provide a reason of why this dependency is needed when we never needed it before?

simonmichal:
@djw8605 : on which platform is this needed? From what I see all our CI builds are green: 
https://gitlab.cern.ch/dss/xrootd/-/pipelines/2182085

djw8605:
I think I may have caught master at just the wrong time?  The UUID requirement was needed when I was building on EL7 at commit 7c190900b679d4327cb74b043affa880a74a6e7d.  But I can't replicate the need for uuid when I build from the current master, so I'm closing this pull request as unreproducible.

PR #1367, opened 2020-12-18
Fix compilation on Debian kFreeBSD

ellert:
This PR makes the code compile on Debian kFreeBSD.

In addition, a minor fix to the GNU/Hurd dirent handling in XrdPosixPreload.


simonmichal:
@ellert : thanks for the PR, there are quite some files involved, together with @abh3 we will review it, though probably at at a slow pace.

ellert:
Compilation is successful on all Debian architectures with this patch added:
https://buildd.debian.org/status/package.php?p=xrootd

PR #1376, opened 2021-01-05
krb5: Release lock if initialization fails

hahnjo:
The call to krbClientContext.Lock() was moved in commit 6e93d111c6
("krb5: fixes after later changes for memory leaks"), but the error
paths did not correctly UnLock().


hahnjo:
ping @simonmichal

abh3:
Sorry that it took so long for the review. The patch looks good.

PR #1382, opened 2021-01-14
Refactor how attributes are passed to chained auth plugins with SciTokens

djw8605:
This is also requires a change in xrootd-multiuser, which is proposed in https://github.com/opensciencegrid/xrootd-multiuser/pull/17

@abh3 Please review and criticize.  This is meant as the first pass.  I lightly tested this that it authorizes things correctly.


abh3 (approved):
Minor nit, otherwise OK. I will merge but at some point you or I will address the nit that doesn't affect anything. Thanks for the fast turnaround.

PR #1391, opened 2021-01-30
XrdTPC: Refactor stream writes

bbockelm:
This removes problematic recursive invocations of the stream writes and fixes some multistream errors that caused too-high concurrency.

The underlying cause of the multistream issue was allowing partial writes of the buffers.  The buffer occupancy is a feedback mechanism to the multistream code to not launch additional requests; the partial write optimization allowed a buffer to look empty when really there was more coming from the request.

Because we aren't allowing partial writes, this also had to allow partial accepts if the incoming curl buffer spanned two internal buffers.  With this, multistream stays megabyte aligned and should still work on RAL's Echo.

Finally, this adds some error messages on conditions considered internal logic errors -- should help with future debugging.

### DIFF
--- a/src/XrdTpc/XrdTpcStream.cc
+++ b/src/XrdTpc/XrdTpcStream.cc
@@ -67,10 +67,14 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
     m_log.Emsg("Stream::Write", ss.str().c_str());
     DumpBuffers();
 */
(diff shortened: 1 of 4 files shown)


abh3:
Thanks Brian. I didn't see anything wrong but there are enough changes that I think we may need an RC7 to verify this. Be on the lookout.

PR #1392, opened 2021-02-03
XrdTpc: Add a crude timeout mechanism.

bbockelm:
This does not enforce a minimum bandwidth (N MB/s) but rather
aborts the transfer if zero data has moved within a given timeframe.
By default, the transfer has 120 seconds to move the first byte of
data and it must transfer at least 1 byte every 60 seconds thereafter.

This purposely does not do bandwidth minimums as that's built into
future versions of `curl` and I'd prefer to not re-implement it.

The default timeouts can be overridden by setting:

```
tpc.timeout timeout_val [first_byte_timeout_val]
```

where the default is equivalent to:

```
tpc.timeout 60
```

If the `first_byte_timeout_val` is not set, it defaults to 2x the
`timeout_val`.

(Note: I am _not_ targeting the 5.1.0 release for this)

### DIFF
--- a/src/XrdTpc/XrdTpcConfigure.cc
+++ b/src/XrdTpc/XrdTpcConfigure.cc
@@ -4,6 +4,7 @@
 #include <dlfcn.h>
 #include <fcntl.h>
 
+#include "XrdOuc/XrdOuca2x.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 #include "XrdOuc/XrdOucStream.hh"
 #include "XrdOuc/XrdOucPinPath.hh"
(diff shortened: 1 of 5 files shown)


abh3:
@bbockelm There is an error in this pull request that needs to be corrected:
/home/travis/build/xrootd/xrootd/src/XrdTpc/XrdTpcMultistream.cc: In member function ‘int TPC::TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq&, TPC::State&, size_t, std::vector<TPC::State*>, TPC::TPCHandler::TPCLogRecord&)’:
/home/travis/build/xrootd/xrootd/src/XrdTpc/XrdTpcMultistream.cc:415:39: error: ‘mres’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
            << curl_multi_strerror(mres);
                                       ^
src/CMakeFiles/XrdHttpTPC-5.dir/build.make:75: recipe for target 'src/CMakeFiles/XrdHttpTPC-5.dir/XrdTpc/XrdTpcMultistream.cc.o' failed

bbockelm:
It's a reasonable call by the compiler ... not likely to have ever been triggered but reasonable to fix!

abh3:
Thanks Brian, waiting for the CI to finish.

PR #1398, opened 2021-02-10
XrdSecztn: Ensure we own the memory location for Entity.name.

bbockelm:


abh3:
I truly apologize for the hiccup in our CI that rejected this pull request. It had nothing to do with the pull request but was a bizarre interaction that we have yet to understand. Given the critical nature of the fix we rolled it into another fix that was equally important. So, unfortunately, you will not be given credit for solving this problem. I hope that doesn't deter you from submitting future patches as we always appreciate patches. It's just one of those one of things that happens. Again, sorry; but the fix has been applied so rest easy.

Issue #1407, opened 2021-02-22
Server wait indefinitely for close


smithdh:
While load testing an xcache (v5.0.3) it was noticed that after the end of a period of heavy load there were a number of files left open on the server (several minutes after all clients had terminated). There were also a number of threads left on the server in this position:

```
Thread 1853 (Thread 0x7fb6ee7ff700 (LWP 278761)):
#0  0x00007fb73d3a8b3b in do_futex_wait.constprop.1 () from /lib64/libpthread.so.0
#1  0x00007fb73d3a8bcf in __new_sem_wait_slow.constprop.0 () from /lib64/libpthread.so.0
#2  0x00007fb73d3a8c6b in sem_wait@@GLIBC_2.2.5 () from /lib64/libpthread.so.0
#3  0x00007fb73d830596 in XrdLink::Serialize() () from /lib64/libXrdUtils.so.3
#4  0x00007fb73dad65d8 in XrdXrootdProtocol::do_Close() () from /lib64/libXrdServer.so.3
#5  0x00007fb73d833ec9 in XrdLinkXeq::DoIt() () from /lib64/libXrdUtils.so.3
#6  0x00007fb73d83746f in XrdScheduler::Run() () from /lib64/libXrdUtils.so.3
#7  0x00007fb73d8375b9 in XrdStartWorking(void*) () from /lib64/libXrdUtils.so.3
#8  0x00007fb73d7f6a57 in XrdSysThread_Xeq () from /lib64/libXrdUtils.so.3
#9  0x00007fb73d3a2ea5 in start_thread () from /lib64/libpthread.so.0
#10 0x00007fb73c6a39fd in clone () from /lib64/libc.so.6
```

The clients which were generating the test load had logged a number of failures of XrdCl::File::Close reported as "[ERROR] Socket error". In the cases examined the client had received all the requested data correctly. Some investigation pointed to a failure of XrdXrootdAioReq::Read() on the first read() of a request, due to no available XrdXrootdAio objects. I think this is due to XrdXrootdAioReq::Alloc() sometimes returning a request object without the minimum number of XrdXrootdAio being available. I'll open a possible pull request in a moment.

smithdh:
see PR #1408

abh3:
I merged your fix. Should we close this issue?

smithdh:
Thank you. Yes, I'll close now.

PR #1408, opened 2021-02-22
[Server] Make sure each aio request has the minimum number of aio objects available

smithdh:
This is a suggested patch for the situation described in issue #1407


smithdh:
If somebody sets the max outstanding operations per request to 1 (xrootd.async maxsegs 1), reads will still usually fail with an error; this is the case before this patch too. I don't now if it's wanted to make a (probably) larger patch to cover that case; I leave this is it is for now.

abh3:
Hi David,

Thanks for the patch, I'll review it but likely it's fine. It will make it 
to 5.1.1. As for setting maxsegs to 1, that patch could simply say we 
don't allow maxsegs to be set below say 8 or whatever number you think 
would avoid the problem.

Andy

Issue #1404, opened 2021-02-19
XrdTpc PULL checksum issues


esindril:
Are there any known issues with TPC PULL where the final checksum does not match the user provided one?
We are running EOS on top of XRootD 4.12.7 and LHCb started using HTTP TPC in production but some of the TPC PULL transfers fail eventually since the final checksum does not match the user provided one. FTS later retries using probably TPC push and the transfers succeeds with the correct checksum.

In the EOS logs we don't see any errors whatsoever. For example on the disk node when such a file with the wrong checksum is written, we get a clean and successful transfer log:

```
210219 05:10:34 26159 TPC_PullRequest: event=PULL_START, local=/eos/lhcb/grid/prod/lhcb/MC/2012/RXCHAD.STRIP.DST/00122396/0000/00122396_00001840_1.rxchad.strip.dst, remote=https://lhcbwebdav-kit.gridka.de:2880/pnfs/gridka.de/lhcb/LHCb-Disk/lhcb/MC/2012/RXCHAD.STRIP.DST/00122396/0000/00122
396_00001840_1.rxchad.strip.dst, user=(anonymous); Starting a push request
210219 05:10:34 time=1613707834.930317 func=open                     level=INFO  logid=6a336346-7268-11eb-a7ac-a4bf0114cb20 unit=fst@st-048-bbead64b.cern.ch:1095 tid=00007f80735fb700 source=XrdFstOfsFile:120              tident=? sec=      uid=0 gid=0 name= geo="" path=/eos/lhcb/grid/prod/lhcb/MC/2012/RXCHAD.STRIP.DST/00122396/0000/00122396_00001840_1.rxchad.strip.dst info=cap.sym=<...>&cap.msg=<...>&mgm.logid=6a336346-7268-11eb-a7ac-a4bf0114cb20&mgm.replicaindex=0&mgm.replicahead=0&mgm.etag="342790846266998784:00000000"&mgm.id=4c1d6756&authz=<...> open_mode=201
[...]

PR #1409, opened 2021-02-23
XrdTpc: Buffer starting at the right offset.

bbockelm:
When we saved the partial buffer, we were saving from the beginning instead of the offset bytes.

#1404


abh3:
Hmm, this comment is not reassuring. What does it really mean?

// Empty buffer cannot accept?!?

bbockelm:
> Hmm, this comment is not reassuring. What does it really mean?
> 
> // Empty buffer cannot accept?!?

That `if` statement should be precluded by the logic in the prior `while` loop.  If that conditional is reachable, there was an internal logic error.

Another alternate would be to `assert` the process in that case -- but I rather just to spit out debugging information and fail the transfer.

PR #1412, opened 2021-02-24
[Server] Make sure stalled aio get written with the right offset and length

smithdh:
Make sure stalled aio get written with the right offset and length.


smithdh:
I didn't open an issue ticket for this: initially noticed on inspection, and subsequently could produce a corrupted file on a server (5.0.3) by xrdcp-ing a large file (1GB, so it would take a few seconds) and suspending the command a few times in the shell, and waiting a few seconds before resuming.

PR #1419, opened 2021-03-04
[XrdCl] Avoid partially reading status message data too early

smithdh:


abh3:
I will leave it to Michal to approve though he won't be back in action until Monday. However, it you could provide what the erroneous result is that would be helpful (i.e. what kind of error results).

smithdh:
I've added an issue to describe the original problem, what was seen etc. Issue #1420. Thanks.

abh3:
Hi David,

Superlative analysis, as always. Thanks!

Andy

simonmichal:
@smithdh : good catch, thanks a lot!

PR #1431, opened 2021-03-18
Overhaul curl's usage of CAs.

bbockelm:
For the NSS backend, this PR will periodically parse all CAs in the CA directory and concatenate the unique CAs into a single file.

The result is:
- Drastically fewer memory leaks in the libnss backend for libcurl.
- Far faster parsing of CAs (1 file to read versus ~1200).
- Avoids a bug where NSS tries to parse CRLs as a CA.

@esindril - I think these will solve all (most?  many?) of your NSS woes.  Would like to hear from you before it gets merged in.


abh3:
While I haven't looked at the actual update I do have one concern. What happens when a site has manually merged all the files and specified that it has done so (e.g. specifying a file instead of a directory). Does this update accommodate that?

bbockelm:
> What happens when a site has manually merged all the files and specified that it has done so (e.g. specifying a file instead of a directory). Does this update accommodate that?

No, it doesn't accommodate that.  However, I don't believe that worked with the former code either.  There's two options you can set in the curl library - directory or filename.  The prior code unilaterally set the directory.

Would be happy to add this feature though.

If we think the general approach is usable, then I would also like to incorporate CRL handling before a final merge.

Finally - note the tweaks I did in XrdCryptosslAux.  Not sure if that was the right way to go about it... seems a lot of work in order to simply keep OpenSSL calls out of `XrdTpc`.

abh3:
Well, I agree that gsi probably didn't do that correctly not withstanding 
that it claims to be able to it that via the --cafile option. 
Regardless, the XRootD TLS stack (and by extension HTTPS) does handle it 
and allows a site to manually combine the files.

So, for consistency it would be good to maintain that functionality even 
if, in this case, it means extending the gsi code to actually do it.

esindril:
Looks good for my side. Maybe one though would be not to mix http cadir/cafile directives with the ones used by TPC curl, since I guess the vanilla http part still needs the cadir to point to something relevant. Also another thought would be to maybe honour the **(XRDTPC_)CURLOPT_CAPATH/CURLOPT_CAFILE** env variables if they are present in the env - like this it's easier to control the TCP behavior without messing up to much with HTTP part.

esindril:
Actually the first part is ok as you bundle them on the fly - I noticed that we already have a bundle file with all the certs on the machines and I was thinking of a workaround - how to have it working with the current production version without any modification to the code.

bbockelm:
@esindril - the last push added an `XRDTPC_CADIR` override that explicitly sets the CADIR to be used with only the TPC component and disables the NSS hack.  Is that what you were aiming for?

@abh3 - I noticed the compilation is failing due to a deprecation warning for `readdir_r`; do you know what the alternative is?

esindril:
Yes, exactly. Thanks!

For the readdir there is not alternative just that it is/will be thread safe from some posix specification on. I guess a `-Wno-deprecated-declarations` flag on the file would be enough to make everyone happy.

abh3:
Actually, for consistency, the cadir/cafile really needs to be coordinated 
with the cadir and cafile in xrd.tls directive otherwise we are just 
getting a lot of confusion. Let's think about how we go about doing that!

xrootd-dev:
The way it's now handled is to use the explicit directory functions. That 
is, opendir() to get a file descriptor and then readdir() using that file 
descriptor, as noted in the following man page:

https://man7.org/linux/man-pages/man3/readdir_r.3.html

That said, before we start adding all of this, we should look at how to 
coordinate these specifications. As it stands we now have potentially 4 
places where the samee information is provided which is really really bad.
To wit:

1) gsi --cadir/--cafile
2) xrd.tls cadir/cafile
3) http.cadir/cafile
4) tpc cadir/cafile

In practice all of these will be the same unless there is a special 
casethat neeeds an override which I can't really envision (can you?). This 
is creatting a big mess. Let me come up with a way to make sense out all 
of this so that sites only have to specify it once and be done with it and 
we don't have to go chasing down conflictng specifications.

Andy

abh3:
Ooops, sent this to the wrong ticket....

Ah, that doesn't mean that I am not open to suggestions on how to
standardize the way the specifications get propogated. By all means,
suggest a policy that perculates the specification across all of the uses.

Andy

esindril:
A good post (actually see the comment) why readdir_r is considered deprecated
https://lwn.net/Articles/696474/

@abh3 I definitely agree we should not have 3/4 places to specify the same thing but being able to override some of them with env variables does seem a sensible approach to me. This might come in handy when debugging various things without the need to rebuild the code to get what you want.

xrootd-dev:
Agreeed, we do allow overrides the way it is now. It just would be could 
to consolidate for the common use case.

bbockelm:
Looking at the readdir manpage, it _is_ sufficiently threadsafe for the glibc implementation (we can't share the same `DIR*` across multiple threads).  I'll change it back to the standard one.

> That said, before we start adding all of this, we should look at how to coordinate these specifications. As it stands we now have potentially 4 places 

I'm a bit indifferent to the requested env var from @esindril; I can see where it's useful as a hidden feature to help test this but I'd be reluctant to document it outside Elvin's testing.  I'll let you two figure out what you want.

However - beyond the env var, I didn't touch any of the configuration code.  This is all done at the XrdHTTP layer and passed through to the extension handlers.  The change here is simply making sure the certfile is passed through internally as well.

abh3:
We need to have these two problems fixed @bbockelm 
(*)(DIR*)’ {aka ‘int (*)(__dirstream*)’} [-Werror=ignored-attributes]
  206 |     std::unique_ptr<DIR, decltype(&closedir)> dirp(fdopendir(fddir), &closedir);
      |                                             ^
make[2]: *** [src/CMakeFiles/XrdHttpTPC-5.dir/build.make:92: src/CMakeFiles/XrdHttpTPC-5.dir/XrdTpc/XrdTpcNSSSupport.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:941: src/CMakeFiles/XrdHttpTPC-5.dir/all] Error 2
make: *** [Makefile:144: all] Error 2

/home/travis/build/xrootd/xrootd/src/XrdTpc/XrdTpcNSSSupport.cc:206:45: error: ignoring attributes on template argument ‘int (*)(DIR*)’ {aka ‘int (*)(__dirstream*)’} [-Werror=ignored-attributes]
  206 |     std::unique_ptr<DIR, decltype(&closedir)> dirp(fdopendir(fddir), &closedir);
      |                                             ^
make[2]: *** [src/CMakeFiles/XrdHttpTPC-5.dir/build.make:92: src/CMakeFiles/XrdHttpTPC-5.dir/XrdTpc/XrdTpcNSSSupport.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:941: src/CMakeFiles/XrdHttpTPC-5.dir/all] Error 2
make: *** [Makefile:144: all] Error 2

Please fix!

bbockelm:
> We need to have these two problems fixed

For the record - I usually am a fan of even-slightly-misguided compiler warnings and the value they provide to society -- but man, those are really stupid warnings from GCC.

Fixed for now.

src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -485,9 +516,9 @@ int XrdCryptosslX509ParseFile(const char *fname,
       rewind(fcer);
       RSA  *rsap = 0;
       if (!PEM_read_RSAPrivateKey(fcer, &rsap, 0, 0)) {
-         DEBUG("no RSA private key found in file "<<fname);
+         DEBUG("no RSA private key found in file");
```
abh3, line 519: Why is it that you are dumbing down the error message? Frankly, I would really want to know what file it is talking about. This error is now rather useless.

src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -485,9 +516,9 @@ int XrdCryptosslX509ParseFile(const char *fname,
...
       } else {
-         DEBUG("found a RSA private key in file "<<fname);
+         DEBUG("found a RSA private key in file");
```
abh3, line 521: Same here.

src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -538,9 +569,6 @@ int XrdCryptosslX509ParseFile(const char *fname,
       }
    }
 
-   // We can close the file now
-   fclose(fcer);
-
```
abh3, line 543: Where does this file get actually closed now?

src/XrdTpc/XrdTpcNSSSupport.hh
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * libcurl with the NSS backend has significant memory leaks around the CA handling
+ * code.  We have discovered that the memory leaks are *smallest* when NSS is given
+ * all the CA certificates in a single file (as opposed to many files in a directory).
+ * 
+ * This class takes a traditional grid CA directory, parses its contents, and creates
+ * a single file.
+ * 
+ * Each restart of the server this temporary file is created; further, every hour a
+ * new copy of the CAs is made.
```
abh3, line 50: This really needs to be configurable. CA's rarely change every hour and a site might just want to have a better fit for it schedule like every 12 hours.

src/XrdTpc/XrdTpcNSSSupport.cc
```diff
@@ -0,0 +1,262 @@
...
+    if (!version_info->ssl_version) {return true;}
+    std::string ssl_version(version_info->ssl_version);
+
+    auto iter = ssl_version.find("NSS");
+    return iter != std::string::npos;
+}
+
+
+bool
+XrdTpcNSSSupport::Maintenance()
```
abh3, line 189: I know you are going to hate it but there is a large potential of leaking file descriptors all over this code. Please use the functions in XrdSysFD.hh which makes sure that file descriptors are not passed to forked processes.

src/XrdTpc/XrdTpcNSSSupport.cc
```diff
@@ -0,0 +1,262 @@
...
+    : m_fd(fd), m_fname(fname)
+    {}
+
+
+XrdTpcNSSSupport::XrdTpcNSSSupport(XrdSysError *err, std::string ca_dir)
+    : m_log(*err),
+      m_ca_dir(ca_dir),
+      m_plugin(err, &g_ver, "Crypto Support", "libXrdCryptossl.so")
+{
+    if (!(m_x509_to_file_func = m_plugin.Resolve("XrdCryptosslX509ToFile")) ||
```
abh3, line 165: Why in the world are you manually resolving these? The functions are available in standard shared libraries that you can link with. Presumably, this code is so critical to prevent HttpTpc from blowing up that you would really want it to work instead of just skipping all of this if you can't find the methods. Plus it makes no sense anyway since the libraries where you added these methods will be distributed with the code that uses them. So, why do this?

abh3 (changes requested):
Some suggested changes and some needed changes.  I may not have caught all the issues here but it's an iterative process.

src/XrdCrypto/XrdCryptosslAux.cc
bbockelm: Well, the `fname` variable no longer exists.  I can make it a requirement for logging-only, if you'd like.

src/XrdCrypto/XrdCryptosslAux.cc
bbockelm: The caller owns the `FILE*` input and it's responsible for closing (as above).

src/XrdTpc/XrdTpcNSSSupport.hh
bbockelm: Why?  Do we save significant I/O by going from hourly to every 12 hours?  In my tests, loading a CA directory was ~100-200ms.

There's a cost to adding a configuration knob; I question if it's worthwhile to give configuration options that allow one to save something like 2 seconds of CPU time per day.

One hour was selected to be so low no one could complain it is too infrequent (i.e., picks up RPM updates in a reasonable amount of time) and high enough to make the cost negligible.

src/XrdTpc/XrdTpcNSSSupport.cc
bbockelm: Sure, good idea.  Will work on that.

src/XrdTpc/XrdTpcNSSSupport.cc
bbockelm: Actually, it appears we cannot link to this - that's why I had to go this direction (trust me, already tried!  It's a pain to load like this and muck around with function pointers).  CMake cowardly refuses to link to a shared module (as opposed to a shared library).

Another option would be to compile + link an extra copy of those functions just for HttpTpc; I don't particularly like that either.

I like the idea of doing a hard fail instead of falling back.  Will work on that.

abh3:
After discussing this with a couple of people we think that the auto cafile generation would be a useful *general* addition so it can be used in more than the XrdTpc context. We've has some issues with the cadir being in StashCache and it goes away if it's made into a file. What are the chances of making this a standalone class so we can put it in the XrdTls package so that it's generally useful. It would certainly lower the overhead for OpenSSL as well.

bbockelm:
> What are the chances of making this a standalone class so we can put it in the XrdTls package so that it's generally useful. It would certainly lower the overhead for OpenSSL as well.

Are you sure it's necessary for `XrdTls` as well?

I ask that because I had assumed `XrdTls` does lazy loading of CA certificates - that is, instead of parsing the entire directory, it just pulls in the exact hashes it needs.

You are right that this would be useful for libcurl with the OpenSSL backend as well.  For NSS, it's about correctness -- but for OpenSSL, this would bring a noticeable speed improvement.

Additionally, the natural follow-up here is to put together a CRL file; libcurl _only_ supports concatenated CRL files and doesn't work with CRLs in a CA directory.  I figured I was going to delay enabling this for OpenSSL until the second round.

abh3:
Well, I'm not too sure about the lazy load because, if you recall, we do x509_store refreshes to pickup new CA's and CRL's. So, somehow I think OpenSSL reads the whole bloody thing when we reinitialize the store. If you were going to abstract this function it's likely easier to do it now than later when it will likely get encumbered with libcurl specifics. Think about it.

src/XrdTpc/XrdTpcNSSSupport.hh
abh3: I'm not going to insist. I thought it would be a nice feature to have. It can always be added later if sites really want it.

src/XrdCrypto/XrdCryptosslAux.cc
abh3: Oh, didn't catch the deletion of fname as it was indirect. If the fname is constant and there is another message referring to it then there is no need. If it's always changing (e.g. the name of each crl or ca) then likely it's good to pass it in if only for debugging purposes.

abh3:
I don't understand why you cannot link to this. It's a shared library with 
exposed functions and that is all you need. So, please tell me what error 
you are seeing and I will solve it for you.

Andy

bbockelm:
@abh3 - precisely, it's not a shared library in CMake terms, it's a shared module.  Here's the error you get if you try to link to it:

```
CMake Error at src/XrdTpc.cmake:47 (target_link_libraries):
  Target "XrdCryptossl-5" of type MODULE_LIBRARY may not be linked into
  another target.  One may link only to STATIC or SHARED libraries, or to
  executables with the ENABLE_EXPORTS property set.
Call Stack (most recent call first):
  src/CMakeLists.txt:60 (include)
```

I suppose we could factor out just the XrdCryptosslAux functions into a standalone library.

src/XrdTpc/XrdTpcNSSSupport.cc
bbockelm: Done.  In the end, there were only two points where file descriptors were opened (the rest were just converting to things like `FILE*` and `DIR*`.

Necessitated the addition of a new `XrdSysFD_Openat` function; from a quick check of manpages, appears to be portable across the BSDs and Solaris so I kept it out of the Linux-specific block.

src/XrdCrypto/XrdCryptosslAux.cc
bbockelm: Resolved in latest commit.

src/XrdCrypto/XrdCryptosslAux.cc
bbockelm: Latest commit added back the debugging output as before.  Resolving.

src/XrdTpc/XrdTpcNSSSupport.hh
bbockelm: Ok, given time limitation, I'm going to put this as an item for the future if we're asked.

src/XrdTpc/XrdTpcNSSSupport.cc
bbockelm: We now do a hard fail when the initialization fails.

Resolving this conversation because the discussion about how to best include the `XrdCryptosslAux` functions continued on below.

bbockelm:
@abh3 - refactored this into a standalone class in XrdTls.

Since you wanted to take a look at avoiding the dynamic load, instead of migrating the plugin code over to XrdTls, I removed it and put in a temporary hack to just compile the minimal set of `XrdCryptossl` files directly into `XrdUtils` (which, it turns out, is not quite so "minimal"!).

bbockelm:
@abh3 - ready for you to take another look at this.

I added the support for CRLs to the CA parser; this results in one temp file for CAs and one for CRLs (I could theoretically mix the two into a single file but I felt it'd be clearer to keep them separate).  Here's the approximate sizes:

```
-rw------- 1 bbockelm cse496 752K Mar 28 16:14 /tmp/xrootd_crl_file.6IHfi3.pem
-rw------- 1 bbockelm cse496 175K Mar 28 16:14 /tmp/xrootd_ca_file.VpRt22.pem
```

The one outstanding item was the one I was hoping to get you to look at: how to link in `libXrdCryptossl-5.so` into the XrdTls class.  Once we have an approach for that, we need to remember to revert the commits adding the files directly to `XrdUtils`.

A thought on that: we are using a really awfully small part of `XrdCrypto` here - effectively just `PEM_read_X509`, `PEM_write_X509`, `PEM_read_X509_CRL`, and `PEM_write_X509_CRL`.  It might cause a modest amount of code duplication, but we can just call the OpenSSL functions directly and drop the `XrdCrypto` baggage.

Thoughts?

src/XrdTls/XrdTlsTempCA.hh
```diff
@@ -0,0 +1,113 @@
...
+/******************************************************************************/
+
+#include <atomic>
+#include <string>
+#include <memory>
+
+// Forward dec'ls.
+class XrdSysError;
+
+namespace XrdTls {
```
abh3, line 36: We don't use namespaces in this way. If the class is named "XrdTls<whatever>" then it already has placed itself in an implicit flat namespace. If a class that would have been in a namespace, say XrdTls, it won't call itself XrdTls<whatever> but simply <whatever>. This avoids  redundant references to XrdTls::XrdTls<whatever> which is rather annoying. No, you cannot use "using" to avoid that (though obviously it does avoid it). I suggest simply not putting the class in the *new* XrdTls namespace as it's simply not needed and actually makes it inconvenient to debug.

src/XrdUtils.cmake
```diff
@@ -67,6 +67,25 @@ add_library(
   XrdTls/XrdTlsNotaryUtils.icc  XrdTls/XrdTlsNotaryUtils.hh
   XrdTls/XrdTlsPeerCerts.cc     XrdTls/XrdTlsPeerCerts.hh
   XrdTls/XrdTlsSocket.cc        XrdTls/XrdTlsSocket.hh
+  XrdTls/XrdTlsTempCA.cc        XrdTls/XrdTlsTempCA.hh
+
+  #-----------------------------------------------------------------------------
+  # XrdCrypto: linking against a few functions that are useful for XrdTls; avoids
+  # linking against all of libXrdCryptossl in XrdUtils
```
abh3, line 74: OK, I see the problem. I really don't know why this was made a plug-in to begin with. It would probably take too much time to figure out the history. My presumption is that at the time gsi was written there could be a choice of which crypto functions could be used. Practically, I don't think there  is a choice any more so it needn't be a plug-in. That said, not making it a plugin would then require changes in gsi, sigh. So, let's live with it this way. Note that we will have to mollify EPEL in that these files will be compiled twice, once for the plugin and once for XrdUtils.so, ughhh.

src/XrdSys/XrdSysFD.hh
```diff
@@ -151,6 +151,10 @@ inline int  XrdSysFD_Socketpair(int domain, int type, int protocol, int sfd[2])
                  }
 #endif
 
+// openat is part of POSIX.1-2008; in Linux, BSD, and Solaris
```
abh3, line 154: You mentioned in the pull comments that you used SysFD where possible which implies there are places which would have been good to use SysFD but were omitted allowing for a possible FD leak. Is that really the case or did this addition fix it?

abh3 (changes requested):
OK, how does all of this integrate? That is what gets the auto-generation of the cafile started? My original thought was to provide a directive something like:
Original: xrd.tlsca noverify | {certdir | certfile} path ...
Modified: xrd.tlsca noverify | {certdir | certfile [auto]} path

So, you would specify certdir and then also specify "certfile auto path" which means that the certfile would be automatically generated from certdir and preferentially used. This is important since I don't see how we can safely come up with a path to something we create without assistance from the config. It would also allow uniform distribution of that information (which is currently done if you specify both). Plus, if we run into problems autogenerating this, we can turn it off. My assumption is that you want this done all the time which would mean this better work perfectly through all of time which is unlikely. Remember that some sites may already be using their own cafile and we can't displace that. Any thoughts on this?

Finally, only one change requested to align namespace usage with the rest of the repo.

src/XrdUtils.cmake
bbockelm: > My presumption is that at the time gsi was written there could be a choice of which crypto functions could be used

My understanding is this was written "just in case" some other backend (like `libnss`) would come along and be a pluggable replacement to OpenSSL.  Personally, I think making a generic plugin mechanism "just in case" is more effort than just writing two implementations... anyhow, past is past.

I don't think there's any problem in EPEL to compile twice here -- with EPEL, the usual concern is "vendoring" where you compile in a 3rd party package instead of depending on it as a separate RPM.

Since there's nothing to do currently, I'll just resolve this thread.

src/XrdSys/XrdSysFD.hh
bbockelm: Oh - that was just imprecise english.

Unless I missed something, the remaining "open"-like calls are really just things like "fdopen" (that convert a file descriptor into a handle).

If you see any places where I open directly in this patchset, will be happy to fix.

src/XrdTls/XrdTlsTempCA.hh
bbockelm: Ok, will collapse this into the main namespace.  Easy enough.

abh3:
One last outstanding thing -- the comments I posted with the review a) how it all gets started, and b) integration with the config file. Both of these are important as I have nothing to go on when documenting this and itreally does need documentation instead of being a Tesla-like black box solution.

bbockelm:
@abh3 - to be clear, I only hooked up the `XrdTpc` pieces.  Here's the relevant locations:
- New `XrdTlsTempCA` object created: https://github.com/xrootd/xrootd/pull/1431/files#diff-c88246e4a0d4f2c51589397e8524f801b01a0c8003d011722c8901b960d1937eR82
- CA Guard handle obtained here: https://github.com/xrootd/xrootd/pull/1431/files#diff-2b4addbbda3815f5b12ae79ea8c5b7a9a4c89bfc7fbd160528a2811b091c70fcR107 (the reference has to be kept alive until libcurl is guaranteed to be done with it).

If we want to do the same thing for the other consumers of XrdTls, we'd need something that controls lifetime -- for XrdTpc, this is self-container because we know the libcurl handle will last precisely no longer than the synchronous function call.  XrdHttp could, for example, have a static `shared_ptr` that is updated from a common `XrdTlsTempCA` object every hour.  Do be careful about lifetime restrictions: if the shared pointer ever falls out of scope, then it's possible the OS will clean up the CA/CRL files.

For the configuration language -- this is essential for at least the NSS backend, so I hadn't planned to expose it as an option.  I can see that it really should be an option for other users of XrdTls (where NSS isn't as problematic!).

I like this variant you suggest:

```
xrd.tlsca noverify | {certdir | certfile [auto]} path
```

Note in the current implementation the auto-generated CA file *only* works with the `certdir` and is superseded by `certfile` if set.  Do you think we should do this regardless for XrdTpc, even if not explicitly requested?

src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -0,0 +1,315 @@
...
+
+}
+
+
+std::unique_ptr<XrdTlsTempCA::TempCAGuard>
+XrdTlsTempCA::TempCAGuard::create(XrdSysError &err) {
+    char ca_fname[] = "/tmp/xrootd_ca_file.XXXXXX.pem";
+    int ca_fd = mkstemps(ca_fname, 4);
+    if (ca_fd < 0) {
+        err.Emsg("TempCA", "Failed to create temp file:", strerror(errno));
```
abh3, line 203: Referencing line 209; both messages are identical eventhough they actually refer to different files. To avoid confusion I suggest the message be:
"Failed to create CA temp file:"

src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -0,0 +1,315 @@
...
+        return std::unique_ptr<TempCAGuard>();
+    }
+    char crl_fname[] = "/tmp/xrootd_crl_file.XXXXXX.pem";
+    int crl_fd = mkstemps(crl_fname, 4);
+    if (crl_fd < 0) {
+        err.Emsg("TempCA", "Failed to create temp file:", strerror(errno));
```
abh3, line 209: Referencing line 203; both messages are identical even though they actually refer to different files. To avoid confusion I suggest the message be:
"Failed to create CRL temp file:"

abh3:
Well, this is going to be a long post (10 points and all are relevant). So, here it goes:

1) For standard framework TLS, if you specify cadir and cafile we feed that to OpenSSL and it does the documented thing. The nice thing about autogenerated cafile is that you can dispense giving openssl the cadir; thus avoiding directory lookups that will fail as you know everything in cadir is in cafile and there is no point in reverting to cadir ever.

2) Assuming you agree with (1) then the directive could be augmented as:
xrd.tlsca noverify | {certdir | certfile [auto [only]]} path
meaning: 'auto only' generate the cafile and only use that. The reason for not forcing this is to have a bypass should there be a bug in collecting certs from the cadir. The meaning of "path" for "auto" cafile was my hope that this could be done transparently and that the most current version of the cafile would exist at the specified path.  The way to possibly do this is explained in point (7).

3) The code in XrdTlsTempCA generates a new one on demand when needed. Since it uses a unique tmp name there indeed needs to be an external mechanism to keep that file alive and deep-six it should a new one be generated after all uses of the current one (hence the shared pointer). However, that approach assumes the file is not always kept open by curl/nss and might be used for refer-backs (it would be surprising if either case was true but then again....). I suppose it's also the case there could be multiple invocations of curl and you want to use the same cafile; though I don't see any harm in moving to a new one at any time. In the end while I understand the use of the shared pointer, it might be overkill in this context.

4) With respect to (3) since cafile is regenerated upon demand it means the caller has to wait until that has been done and it certainly could take some time to do that (though it's likely under a minute). Depending on the client's timeout setting it could trigger a retry which is not exactly great. Certainly, this is an argument for generating these in the background and having a fresh one always ready.

******
5) Question! I did not see anywhere in the code how other accesses subsequent to the one that triggered the regeneration are forced to wait. In fact, if I read the code correctly, multiple such files may be created until the first one completes and the timer is reset. If that's the case, could that be fixed some how? It's implicitly fixed if we go with background regeneration.
******
6) Question! I see that the code also generates a composite CRL file and, indeed, curl let's you specify it and presumably it's used to verify the cert. The openssl command also allows you to do that. However, nowhere in the openssl API is it documented on how to provide the CRL list to the library. My presumption is that the library does not know how to use CRL's and it was implemented as an outboard feature in the command itself. Do you know anything more about that? Since the CRL file is available is would be nice to be able to actually use it for openssl.
******

7) Indeed doing an autogenerated cafile is a bear especially when you don't know how the consumer of the file actually uses it. In OpenSSL the file is read in its entirety once and you need to reset the verify locations in a new context to get a refresh. A real pain. I would assume curl/nss does roughly the same but it's easier since it has to be done for every connection. So usage of an auto-cafile in HTTP/XrootD differs from how it is used in XrdTpc. Trying to align both might be possible since we could potentially marry it with the x509_store refresh code. However, it would need to be done in a transparent way. So, here is another proposal (yeah, like you need one at this point :-) So, consider what appears to be a very simple augmentation...

a) The cafile is regenerated every x minutes in the background much the same way it is done now (simple but not trivial). However, if XrdTlsTempCA  is given a specific path then it needs to atomically create a symlink to the regenerated file using that path so that the name of the latest cafile is accessible by a static name  (e.g. ln -s to path.tmp and then mv path.tmp to path). You can use the framework's timed scheduler to drive this; though I will likely have to do some mods when I interface it with x509_store refresh to keep the timing in sync. That's a future thing.

b) The code in XrdTpc doesn't need to change at all though it would be nice if it used the static path for consistency.

c) Now all we need is a mechanism to feed XrdTlsTempCA.cc the path to use (see the next point).

8) I would propose the following new method in XrdTlsTempCA  
Init(const char *cadirpath, const char *cafilepath, bool autogen, int interval);

If autogen is true generate cafile from cadirpath and if cafilepath is not nil we want to always have a symlink
cafilepath->tempcafile. If autogen is false we want you to use cafilepath  as the cafile if it's not nil. You can ignore interval -- that's for the future. This would be called by the framework during the course of TLS configuration.

9) Should this be done no matter what? Well, it's not all that straightforward to answer. So, here are the constraints that I think should be used to guide the answer:

a) If a site specified cafile *without* "auto" then auto-generation should be disabled and the specified cafile should be used. Is that correct? Maybe yes maybe not. It certainly is correct if the site atomically regenerates the file (which some sites do) and we do a corresponding change to XrdTlsTempCA  which would make the choice possible. That is, it's told to use that cafile and not regenerate (see proposed new method in point 8).
b) While some sites might switch to making the XRootD framework auto-regenerate; I doubt those that regenerate themselves will go with that feature. Why? Because their auto cafile is used by many other applications outside xroot/http and is typically placed in a root-only writable directory. So doing it wouldn't work for them. That said, they are the ones that would likely specify non-auto cafile anyway (with all the attendant problems).
c) What if there is no cafile specification? This is the tricky part because we all know that libnss needs it and openssl would benefit as well. Common security practice says don't do security related things behind an admin's back. This feature is relatively benign until something goes wrong and then, quite reasonably, the security hats will jump all over it leading to a very bad situation.
d) Compromises: 1) print a rather wordy long warning in the log saying you really should allow this feature, 2) allow a plugin to tell the framework to turn on auto-regeneration if not specifically disabled. This would be specific to XrdTpc and could be done by it setting an envar (say "XRD_WANT_AUTO_CAFILE"). The issue is that it's a one way request and it might not be possible to turn it on. However, I think the XrdTpc code has a fallback when it can't get a cafile (like using cadir), right? Plus, if it's a fixed cafile it should get that one (after some code changes in XrdTlsTempCA).
e) I'm totally open to any other suggestion as the answer is not at all obvious.

10) Finally! Now all that said, I noticed an unfortunate nit-picking copy-past issue whose details are posted with the pull request. Could you correct that?

abh3:
One more thing which I forgot. Creating the file in /tmp is sort of bad. The proposed init method should get an extra parameter of the directory where the cafile should be created. In the framework we have a defined place for such files and they are normally not in /tmp and outside the scope of OS cleanup. So, I think the notion of cleanup is moot once XrdTlsTempCA adopts that mechanism.

simonmichal:
@bbockelm & @esindril : I have more of a general question regarding the TPC component: wouldn't it make more sense to use davix that provides a friendly abstraction rather than implementing everything from scratch on top of curl? Maybe it's worth considering ;-)

abh3:
I must admit this is a novel way of addressing competition for CA file maintenance and creating new files on a regular basis. My assumption is that you are still working out the other issues/observations I posted earlier, yes?

bbockelm:
Hi @abh3 -

Did a lot of studying of the OpenSSL code and how XrdTls implements it.  The good news is that there's likely no way that the TempCA implementation could beat it -- for a remote host, the existing XrdTls code accesses a single file and loads a single CA from it.  For the TempCA code, while it also loads a single file, there's still a need to parse linearly through all the available CAs.

This goes back to the question about the `libcurl` backend: OpenSSL assumes that the CA directory is hashed according to a specific hash function, allowing a cheap lookup.  However, `libcurl` makes no assumption about an OpenSSL backend and assumes the CA directory is just an unorganized directory full of CAs -- which then causes all the I/O problems.

So, I don't see any way we can improve what XrdTls does in general and right now the only case I see for TempCA is in the XrdTpc code.  That said, I _do_ like the current location of the code (it's all about TLS handling, might as well put it in the XrdTls directory).

I believe several of your points are addressed by the latest commit, which moves things out to a separate thread.  That reduces the maximum cost of getting a reference to a good set of CAs to the refcount increase on the shared pointer -- instead of potentially a scan of the directory.

Given the TempCA piece is necessary for correctness in the libnss backend - and an important reduction in I/O for openssl - instead of making this a daemon-wide config option, I think we should go back to having this on-and-required by default.

Toward another question above: libcurl only can create new connections while its handle is available.  Thus, as long as the shared pointer lives for the entirety of the libcurl transaction, you're guaranteed that the CA lifetimes are long enough.

Finally, there's the question of the location of the CA temporary file.  I chose the `/tmp` as I can rely on `mkstemp` to create the file and there's a plethora of daemons that make the (reasonable) assumption small temporary files can be kept there.  It works nicely as a configuration-free way to proceed.  If you had a good suggestion for another location, I'm quite interested.

For example, a good improvement would be a directory assumed to be owned by the `xrootd` user: with that, we can avoid security issues that cause us to rely on `mkstemp` in the current code and use non-changing, well-known filename for the CA file.  Then, we don't need to refcount the CA usage as libcurl can always open the exact same file.

abh3:
Thanks for looking into this. That simplifies a lot of decision making. As for the file location, in xrootd we have what's called an "adminpath" that a site gets to specify. All special files are created starting at that location. You can find out what the admin path is via envar XRDADMINPATH. So, I would suggest creating a directory ".tempca" rooted there (capitalization of your choice). You should set appropriate privs for that directory. Then it's all yours to do with as you like. While mkstmp works, I've heard that it too can mess up. There are any number of ways to get a unique file name with little effort. The one I use most often is start-up time (decimal or hex), period,  and a monotonically increasing number (a static starting at 0 will do). Since it's guaranteed to be your directory you can't conflict with anything else.

Also, don't forget to fix the two identical messages for two very different errors.

bbockelm:
Alright - working on this.  Have some OSG fires to take care of tomorrow, will attempt for tomorrow night.

abh3:
Great! You likely will want to remove any files in that directory when you start up.

abh3 (approved):

PR #1440, opened 2021-04-09
delete chunk buffer for VectorReadInfo after copying into python string

### LINKED ISSUE #1400: Memory leak in `vector_read` for pyxrootd
There seems to be a memory leak in the `vector_read` function in the python bindings. To reproduce:

```python
# debug_vector_read.py
import gc
import pyxrootd.client

filename = "root://eospublic.cern.ch//eos/opendata/atlas/OutreachDatasets/2020-01-22/4lep/MC/mc_345060.ggH125_ZZ4lep.4lep.root"

def test():
    f = pyxrootd.client.File()
    res = f.open(filename)
    if not res[0]["ok"]:
        raise IOError(res)
    chunksize = 1024 * 256
    chunks = [(i * chunksize, chunksize) for i in range(5)]
    res = f.vector_read(chunks)
    print(res[0]["message"])
    f.close()

for i in range(20):
    print(i)
    test()
    gc.collect()
```

When running this through [memory-profiler](https://pypi.org/project/memory-profiler/) with

```
mprof run debug_vector_read.py
mprof plot
```

i get this plot:

memory_leak

(Tested with xrootd 5.0.3 from conda-forge)

It seems the problem are these buffers which are never deleted:

https://github.com/xrootd/xrootd/blob/551521ea7368e3b2a4c7964325e2e7e3fe7ce047/bindings/python/src/PyXRootDFile.cc#L611

Not sure if or when it is allowed to delete them, but if i collect the buffers into an array and delete them after the call to `self->file->VectorRead(...)` the memory leak seems to be gone:

no_memory_leak

nikoladze:
This should fix #1400. I added a delete after a chunk buffer is copied into a python string in the conversion function for `VectorReadInfo`, analogous to what is done here for `ChunkInfo`:

https://github.com/xrootd/xrootd/blob/551521ea7368e3b2a4c7964325e2e7e3fe7ce047/bindings/python/src/Conversions.hh#L248-L257

### DIFF
--- a/bindings/python/src/Conversions.hh
+++ b/bindings/python/src/Conversions.hh
@@ -270,6 +270,8 @@ namespace PyXRootD
 
           PyObject *buffer = PyBytes_FromStringAndSize( (const char *) chunk.buffer,
                                                         chunk.length );
+          delete[] (char*) chunk.buffer;
+
           PyList_SET_ITEM( pychunks, i,
               Py_BuildValue( "{sOsOsO}",
                   "offset", Py_BuildValue( "k", chunk.offset ),


nikoladze:
The memory leak in the example in #1400 is indeed gone

image

simonmichal:
Thanks a lot!

PR #1447, opened 2021-04-22
[XrdTls] Memory leak fixes in XrdTlsTempCA

jthiltges:
- Free XrdCryptoX509Chain when finished
- Duplicate and close file descriptors to free memory

### DIFF
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -100,10 +100,10 @@ CASet::processFile(file_smart_ptr &fp, const std::string &fname)
     XrdCryptosslX509ParseFile(fp.get(), &chain, fname.c_str());
 
     auto ca = chain.Begin();
-    // Note we purposely leak the outputfp here; we are just borrowing the handle.
-    FILE *outputfp = fdopen(m_output_fd, "w");
-    if (!outputfp) {
+    file_smart_ptr outputfp(fdopen(XrdSysFD_Dup(m_output_fd), "w"), &fclose);
+    if (!outputfp.get()) {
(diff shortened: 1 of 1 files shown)


src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -100,10 +100,10 @@ CASet::processFile(file_smart_ptr &fp, const std::string &fname)
     XrdCryptosslX509ParseFile(fp.get(), &chain, fname.c_str());
 
     auto ca = chain.Begin();
-    // Note we purposely leak the outputfp here; we are just borrowing the handle.
-    FILE *outputfp = fdopen(m_output_fd, "w");
-    if (!outputfp) {
+    file_smart_ptr outputfp(fdopen(dup(m_output_fd), "w"), &fclose);
```
bbockelm, line 103: Can you replace the direct `dup` call with the wrapper (which sets `O_CLOEXEC`): https://github.com/xrootd/xrootd/blob/master/src/XrdSys/XrdSysFD.hh#L54

bbockelm:
@jthiltges - ran these through Valgrind.  All looks good!  I'd suggest to merge.

abh3:
This will be merged after JT updates the pull request to use SysFD which hasn't happened yet. JT? Please be aware we are on a tight timeline for releasing 5.2 and it would be good to have this patch in that release but it won't happen until this change occurs.

src/XrdTls/XrdTlsTempCA.cc
jthiltges: Good idea, thank you!

jthiltges:
PR updated with Brian's suggestions and squashed. Thank you for your time and attention, and please let me know if there's any further changes needed.

PR #1464, opened 2021-06-01
Clear OpenSSL error queue ahead of SSL_read/write/accept()

jthiltges:
- Clear the OpenSSL error queue to avoid reporting errors from other SSL operations, such as reading CRLs
- Log queued OpenSSL errors if present

PR #1465, opened 2021-06-08
Clear the OpenSSL error queue in XrdCryptosslX509ParseBucket()

jthiltges:
Regarding the OpenSSL `ERR_clear_error()` overhead introduced in #1464:

This PR should clear the OpenSSL error queue in `XrdCryptosslX509ParseBucket()`, along with reverting the `ERR_clear_error()` calls in `XrdTlsSocket`.

So far, I'm unable to reproduce the TLS error on our local xcache with this patch (on the latest experimental, xrootd-5.2.1-0.experimental.2679300.76d03f61). This certainly could use some thorough testing to confirm. @osschar

### DIFF
--- a/src/XrdCl/XrdClXRootDTransport.cc
+++ b/src/XrdCl/XrdClXRootDTransport.cc
@@ -2241,13 +2241,11 @@ namespace XrdCl
                     "[%s] Authenticated with %s.", hsData->streamName.c_str(),
                     protocolName.c_str() );
 
-        if( info->encrypted || ( info->serverFlags & kXR_gotoTLS ) ||
-            ( info->serverFlags & kXR_tlsLogin ) )
-          //--------------------------------------------------------------------
-          // Clear the SSL error queue of the calling thread, as there might be
-          // some leftover from the authentication!
-          //--------------------------------------------------------------------
-          Tls::ClearErrorQueue();
+        //--------------------------------------------------------------------
+        // Clear the SSL error queue of the calling thread, as there might be
+        // some leftover from the authentication!
+        //--------------------------------------------------------------------
+        Tls::ClearErrorQueue();
 
         return XRootDStatus();
       } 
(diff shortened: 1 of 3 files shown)


xrootd-dev:
Well, yes. I am also confused about this patch. On the client side, 
error_clear_err() is called after authentication so we know that the error 
queue, if one was left over from gsi, is gone. On the server-side we now 
do the same. Adding additional calls within gsi is a no-win situation as 
there are so many places gsi can generate an error without clearing it the 
mind boggles and it's likely all of them have not been caught. That's why 
treating gsi as a black box and doing an external clear is a better 
option.

Now that said, I think the other server-side issue may be that https also 
uses parts of gsi as well as other openSSL routines that may leave junk 
in the error queue. Finding all the places is also a hopeless case (at 
least for now).

There is one additional fix I can try and that is when a thread is 
returned to the thread pool the pool manager simply clear the error queue 
before the thread is allowed to be reused. That is a sledge hammer 
approach but we might as well try it as it has the least impact and pretty 
well puts a nail in this coffin.

Shall I?

Andy

jthiltges:
My goal was to get rid of the `ERR_clear_error()` calls in `XrdTlsSocket`. If the error queue clearing in f74d4531 is sufficient, then this PR can be ignored. I'll do some testing.

xrootd-dev:
Thanks JT. If it's not we go with plan B and try to clear them in the 
server at thread return time. This is getting to be a bear, sigh.

osschar:
Andy, remember, John is also seeing these errors in XCache when auth is enabled. I don't know if http access is enabled there as well.

xrootd-dev:
Hi Matevz,

Yes, the unfortunate part is that https is enabled and is happily uusing 
gsi as well. JT is going through the tedious steps of eliminating each 
issue as we add additional stuff as we need. He is a good trooper!

Andy

simonmichal:
@jthiltges : ahh, ok, I see your point, the error queue is cleared only if the channel is encrypted, however since multiple channels share the same event-loop thread it still may interfere with another channel that is encrypted

Could you try removing the if statement:
https://github.com/xrootd/xrootd/blob/bbf477b876818485ac9250f97f53464a6a38f6a6/src/XrdCl/XrdClXRootDTransport.cc#L2314-L2315
and see if this helps?

jthiltges:
* Reverted the changes to `XrdCryptosslX509ParseBucket()`
    - Handling the OpenSSL error queue close to where the errors happen still appeals to me
* Added a `Tls::ClearErrorQueue()` call to `CleanUpAuthentication()`
    - This seemed necessary if the authentication returns early due to an error
* Removed the `if` from the two places `Tls::ClearErrorQueue()` is called
    - I'm not certain that both places need the change

I cannot reproduce the TLS errors with this revised PR.

abh3:
Thanks JT for all the work! When we meet in person again, I'll buy you a couple of beers :-)

simonmichal:
Looks solid :-) Thanks a lot, great job!

PR #1505, opened 2021-09-05
Link libXrdServer to libatomic on riscv64 

ellert:
Link libXrdServer to libatomic on riscv64 to avoid undefined references:
```
/usr/bin/ld: libXrdServer.so.3.0.0: undefined reference to `__atomic_fetch_or_1'
/usr/bin/ld: libXrdServer.so.3.0.0: undefined reference to `__atomic_fetch_sub_1'
/usr/bin/ld: libXrdServer.so.3.0.0: undefined reference to `__atomic_fetch_and_1'
/usr/bin/ld: libXrdServer.so.3.0.0: undefined reference to `__atomic_fetch_add_1'
collect2: error: ld returned 1 exit status
```

PR #1506, opened 2021-09-05
src/XrdServer.cmake: Use ${CMAKE_THREAD_LIBS_INIT} instead of pthread

AdrianBunk:
This passes -pthread instead of -lpthread, fixing a build failure on RISC-V.

@ellert This is an alternative to #1505

### DIFF
--- a/src/XrdServer.cmake
+++ b/src/XrdServer.cmake
@@ -185,11 +185,14 @@ add_library(
   XrdDig/XrdDigConfig.cc        XrdDig/XrdDigConfig.hh
   XrdDig/XrdDigFS.cc            XrdDig/XrdDigFS.hh )
 
+set( THREADS_PREFER_PTHREAD_FLAG TRUE )
+find_package( Threads )
+
 target_link_libraries(
   XrdServer
   XrdUtils
   ${CMAKE_DL_LIBS}
-  pthread
+  ${CMAKE_THREAD_LIBS_INIT}
   ${EXTRA_LIBS}
   ${SOCKET_LIBRARY} )
 


simonmichal:
@AdrianBunk : thanks for the PR and sorry I haven't noticed it before merging #1505, yours seems to be slightly more cleaner, let me think what to do about it ;-)

ellert:
To be consistent, shouldn't all linking to -lpthread be replaced with -pthread? Not just for libXrdServer.
That way the compiler can avoid linking to libpthread on systems with new glibc where the pthread functions are included in glibc.
Those systems have an empty libpthread so that code using -lpthread doesn't fail to build, but maybe that backward compatibility thing goes away in a decade or so. 
(I know this is a bigger change going beyond the scope of just fixing the riscv64 build.)

AdrianBunk:
> To be consistent, shouldn't all linking to -lpthread be replaced with -pthread? Not just for libXrdServer.

Yes, this would be more consistent.

> That way the compiler can avoid linking to libpthread on systems with new glibc where the pthread functions are included in glibc.
> Those systems have an empty libpthread so that code using -lpthread doesn't fail to build, but maybe that backward compatibility thing goes away in a decade or so.

glibc is not one of these projects that break backwards compatibility, they are always working hard to ensure that even binaries from the 1990s still work (which is important for many users).

Linking with `libpthread` is required by POSIX, and the empty `libpthread.a` glibc >= 2.34 provides for supporting it is really trivial.

simonmichal:
@AdrianBunk & @ellert : thanks a lot for highlighting and addressing this issue! I will close this one as I just merged the #1509, which also includes the changes proposed in this PR.

Issue #1507, opened 2021-09-06
Transfers to Xcache via memcache proxy errors with "numerical argument out of domain" in 5.3.X


snafus:
Hi,
  We are observing transfer failings to an Xcache endpoint with error returned as "Error response: numerical argument out of domain".
Oxford runs an Xcache pointing at the RAL endpoint. Oxford runs with 5.3.1. When RAL had 5.2.0 these errors did not appear, but in testing with 5.3.1 we see issues. 
I've managed to reproduce the problem with a local Xcache instance, and also with a xrootd server using local storage (e.g. oss.localroot /xroot-test, and not using the XrdCeph plugin). 
It seems to occur in the last read of the file, and for files with a non 2**x size. e.g. 64MiB seems to work ok, but 64MiB+4bytes fails. 

The flow is  Xcache <-  memcache proxy <- server
If I remove the memcache proxy (e.g. Xcache <- server), then the problem disappears !

The client sees:
```
[2021-09-06 14:47:37.062940 +0000][Dump   ][XRootD            ] [ceph-dev-gw1.gridpp.rl.ac.uk:1094] Got a kXR_error response to request kXR_read (handle: 0x00000000, offset: 67108864, size: 33554432) [3019] Unable to read /root:/ceph-dev-gw1.gridpp.rl.ac.uk:1096/xrootd-test/test/jwalder/test_64MiBp1; numerical argument out of domain
[2021-09-06 14:47:37.063005 +0000][Debug  ][XRootD            ] [ceph-dev-gw1.gridpp.rl.ac.uk:1094] Handling error while processing kXR_read (handle: 0x00000000, offset: 67108864, size: 33554432): [ERROR] Error response: numerical argument out of domain.
[2021-09-06 14:47:37.063043 +0000][Debug  ][ExDbgMsg          ] [ceph-dev-gw1.gridpp.rl.ac.uk:1094] Calling MsgHandler: 0xef5370 (message: kXR_read (handle: 0x00000000, offset: 67108864, size: 33554432) ) with status: [ERROR] Error response: numerical argument out of domain.
[...]

PR #1510, opened 2021-09-07
[Server] Use length returned by read and fix an alignment check in pgRead

smithdh:
Changes XrdOssDF and XrdOucCacheIO pgRead to use the length returned by read when calculating the checksums, rather than the requested read size. Believed to fix issue #1507 (due to XrdOucCacheIO pgRead).

Also removes an alignment check which may have been inadvertently re-added to XrdOucCacheIO::pgRead during a rebase.

### DIFF
--- a/src/XrdOuc/XrdOucCache.cc
+++ b/src/XrdOuc/XrdOucCache.cc
@@ -44,12 +44,6 @@ int XrdOucCacheIO::pgRead(char                  *buff,
 {
    int bytes;
 
-// Make sure the offset is on a 4K boundary and the size is a multiple of
-// 4k as well (we use simple and for this).
-//
-   if ((offs & XrdSys::PageMask)
-   || (rdlen & XrdSys::PageMask)) return -EINVAL;
-
 // Read the data into the buffer
 //
    bytes = Read(buff, offs, rdlen);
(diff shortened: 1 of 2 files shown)


src/XrdOuc/XrdOucCache.cc
```diff
@@ -44,12 +44,6 @@ int XrdOucCacheIO::pgRead(char                  *buff,
 {
    int bytes;
 
-// Make sure the offset is on a 4K boundary and the size is a multiple of
-// 4k as well (we use simple and for this).
-//
-   if ((offs & XrdSys::PageMask)
-   || (rdlen & XrdSys::PageMask)) return -EINVAL;
-
```
abh3, line 52: How this check is being removed? While I agree the test for requiring all reads to be a multiple of 4K is likely to strict, for the ache all reads must be page aligned. So, at least that check should have remained.

abh3 (changes requested):
Thank you David for catching this. An obvious and stupid mistake. That said, I did ask about the removal the the test for pgRead() as it seem at least the offset check should stay.

smithdh:
(some discussion going on by email with Andy & Matevz about the alignment check(s) in XrdOucCacheIO::pgRead)

abh3 (approved):
You are absolutely correct. Somewhere along the way we started using the CacheIO object for all IO requests, cache or not. Since pgRead and pgWrite were extended to be able to do unaligned reads and writes the change was indeed necessary.

Issue #1516, opened 2021-09-24
openssl 3 plans?


adriansev:
Hi! Are there plans for supporting the very fresh openssl 3?
Thanks!

simonmichal:
@adriansev : yes, we do have to support it as it is now available on Fedora rawhide (and we are obliged to built on Fedora rawhide as we release XRootD in EPEL).

abh3:
We have a compatible compilation version in 5.4.0. Since there are few sites using xrootd with openSSL 3 we have no idea whether this works to everyone's satisfaction. Anyway, I am closing this ticket as the support is there even if untested.

rob-c:
@abh3 should I correct in saying that v5.4.x should be compiling with openssl v3.0.x (vanilla from the project gh) with gcc10+ on a centos host?
I am getting a lot of compilation errors, initially just some const-casting but then some more compilation errors about deprecation headers and such which is where I stopped debugging.

Is there a different branch/compilation-flag I'm supposed to use for testing this? From what I can tell most changes are wrapped behind ifdef version checks.
or,
Should the openssl libraries be compiled in a certain way as of v3 for best xrootd support?

PR #1517, opened 2021-09-26
Fix compilation on Fedora 36 with openssl 3.0.0

ellert:
Still compiles on RHEL/CentOS 7 with openssl 1.0.2k

Fixes: #1516

This is a big change that requires testing.

### DIFF
--- a/src/XrdCrypto/XrdCryptoLite_bf32.cc
+++ b/src/XrdCrypto/XrdCryptoLite_bf32.cc
@@ -40,7 +40,7 @@
 #include <netinet/in.h>
 #include <cinttypes>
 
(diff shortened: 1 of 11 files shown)


simonmichal:
Hi @ellert : thanks a lot for the PR! Could you change it so it goes to the `openssl3` branch instead of `master`? For now, we will make those changes only available on fedora rawhide as we don't have enough manpower to test it properly.

ellert:
> Hi @ellert : thanks a lot for the PR! Could you change it so it goes to the `openssl3` branch instead of `master`? For now, we will make those changes only available on fedora rawhide as we don't have enough manpower to test it properly.

Done.

PR #1528, opened 2021-10-06
Be nice to older (less clever) compiler versions

ellert:
error: parameter declared 'auto'

### DIFF
--- a/src/XrdCl/XrdClFileStateHandler.cc
+++ b/src/XrdCl/XrdClFileStateHandler.cc
@@ -1445,7 +1445,7 @@ namespace XrdCl
     uint32_t fstpglen = fLen;
 
     time_t start = ::time( nullptr );
-    auto h = ResponseHandler::Wrap( [=]( auto* s, auto* r ) mutable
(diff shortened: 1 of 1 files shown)


simonmichal:
Hi @ellert, thanks for the PR. The generic lambdas are part of C++14 standard, which is now officially required for XRootD. What is the platform / use case your are training to satisfy with those changes?

ellert:
> What is the platform / use case your are training to satisfy with those changes?

Until now building with -std=gnu++1y using the default compiler in RHEL/CentOS 7 worked.
The C++14 support in this compiler i incomplete, but this was the first time xrootd failed to build because of that.
I don't know if building only with the subset of C++14 supported by the default RHEL 7 compiler is expected to work, but until now it did work. Though that might have been just accidental.

simonmichal:
@ellert : I will merge this as this is minor change and I suppose it will make your life easier, that said be prepared that we might use C++14 features (e.g. `make_unique`, lambda capture expression, etc.)

PR #1529, opened 2021-10-06
Include Start/End time in logs for TPC transfer

juztas:
@bbockelm


juztas:
@bbockelm Can you comment on this and can we get this finally merged, so it reaches 5.4 xrootd release?

abh3 (changes requested):

PR #1547, opened 2021-11-02
1543 emptry crl file fails http tpc pull transfer

### LINKED ISSUE #1543: [XrootdHttpTPC] Empty CRL file fails a HTTP TPC pull transfer
Hello,

After the submission of an HTTP TPC Pull transfer, I got the following error client side:

```
* Connection #0 to host xrootd-ccaffy-dev01.cern.ch left intact
failure: Failed to load CRL file (path? access rights?, format?)[xrootddev@xrootd-ccaffy-dev01 xrootd-run]$ 
```
On server side:

```
211027 15:06:03 32354 TPC_PullRequest: event=TRANSFER_FAIL, local=/tmp/tpcfile1, remote=https://eospps.cern.ch/eos/user/ccaffy/file1, user=(anonymous), bytes_transferred=0; HTTP library failure: Failed to load CRL file (path? access rights?, format?)
```
Here is the configuration file I have:

```
all.export /tmp nolock
all.adminpath  /var/spool/xrootd
xrd.port 1095
ofs.tpc autorm ttl 7 15 xfr 9 pgm /home/xrootddev/xrootd-run/my_xrdcp.sh
xrd.tls /etc/grid-security/xrd/xrdcert.pem /etc/grid-security/xrd/xrdkey.pem
xrd.tlsca certdir /etc/grid-security/certificates
http.exthandler xrdmacaroons libXrdMacaroons-5.so
macaroons.secretkey /etc/xrootd/macaroon-secret
ofs.authlib libXrdMacaroons-5.so
if exec xrootd
  xrd.protocol http:1095 libXrdHttp-5.so
  http.exthandler xrdtpc libXrdHttpTPC-5.so
  http.secxtractor libXrdVoms-5.so
fi
all.sitename xrootddev
```

After debugging the xrootd process, I could see that curl complains with "Failed to load CRL file (path? access rights?, format?)"
when the `CURLOPT_CRLFILE` option points to an empty file.

The workaround I found was to set the following environment variable:

```
export XRDTPC_CADIR=/etc/grid-security/certificates/
```
This prevents the class  `XrdTlsTempCA` from being instanciated and therefore prevents to
set the `CURLOPT_CRLFILE` option (done by the method `TPCHandler::ConfigureCurlCA()`).

In my opinion, this can be problematic for users. The directory where the certificates
are located is passed via the configuration of the server. If a user has no CRL file on the certificate directory, XRootD should
just ignore it and should not try to set the `CURLOPT_CRLFILE` curl option.

In production, everything works fine because the concatenated CRL file is not empty:

```
[root@eospps-fe1 (mgm:master mq:master) ~]$ cat /tmp/mgm/.xrdtls/crl_file.pem | wc -l
12518
```

My first question is, is it me who wrongly configured the server?

Otherwise, should we add a check that verifies that the concatenated CRL file is not empty before assigning it to `CURLOPT_CRLFILE` ?

@abh3 , @bbockelm what is your opinion about this issue?

Thanks in advance for your answers :)

Cheers,
Cedric

ccaffy:
Fixes #1543

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,5 +58,3 @@ xrootd.spec
 dist
 *.egg-info
 bindings/python/VERSION
-cmake-*
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,5 @@ xrootd.spec
 dist
 *.egg-info
 bindings/python/VERSION
+cmake-*
+.idea/
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -177,7 +177,6 @@ private:
 bool
 CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
 {
-    bool atLeastOneValidCRLFound = false;
     file_smart_ptr outputfp(fdopen(dup(m_output_fd), "w"), &fclose);
     if (!outputfp.get()) {
         m_log.Emsg("CAset", "Failed to reopen file for output", fname.c_str());
@@ -194,7 +193,7 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
         if (!hash_ptr) {
             continue;
         }
-        atLeastOneValidCRLFound = true;
+        m_atLeastOneValidCRLFound = true;
         auto iter = m_known_crls.find(hash_ptr);
         if (iter != m_known_crls.end()) {
             //m_log.Emsg("CRLset", "Skipping known CRL with hash", fname.c_str(), hash_ptr);
@@ -209,8 +208,6 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
             return false;
         }
     }
-    if(!m_atLeastOneValidCRLFound)
-        m_atLeastOneValidCRLFound = atLeastOneValidCRLFound;
     fflush(outputfp.get());
 
     return true;
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -47,6 +47,7 @@
 
 #include <sstream>
 #include <vector>
+#include <atomic>
 
 namespace {
     
@@ -138,7 +139,8 @@ class CRLSet {
 public:
     CRLSet(int output_fd, XrdSysError &err)
     : m_log(err),
-      m_output_fd(output_fd)
+      m_output_fd(output_fd),
+      m_atLeastOneValidCRLFound(false)
     {}
 
     /**
@@ -153,6 +155,12 @@ public:
      * Returns true on success.
      */
     bool processFile(file_smart_ptr &fd, const std::string &fname);
+    /**
+     * Returns true if a valid CRL file has been
+     * found during the execution of the
+     * processFile(...) method, false otherwise
+     */
+    bool atLeastOneValidCRLFound() const;
 
 private:
     XrdSysError &m_log;
@@ -162,12 +170,14 @@ private:
         // one only once.
     std::unordered_set<std::string> m_known_crls;
     const int m_output_fd;
+    std::atomic<bool> m_atLeastOneValidCRLFound;
 };
 
 
 bool
 CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
 {
+    bool atLeastOneValidCRLFound = false;
     file_smart_ptr outputfp(fdopen(dup(m_output_fd), "w"), &fclose);
     if (!outputfp.get()) {
         m_log.Emsg("CAset", "Failed to reopen file for output", fname.c_str());
@@ -184,6 +194,7 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
         if (!hash_ptr) {
             continue;
         }
+        atLeastOneValidCRLFound = true;
         auto iter = m_known_crls.find(hash_ptr);
         if (iter != m_known_crls.end()) {
             //m_log.Emsg("CRLset", "Skipping known CRL with hash", fname.c_str(), hash_ptr);
@@ -198,9 +209,15 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
             return false;
         }
     }
+    if(!m_atLeastOneValidCRLFound)
+        m_atLeastOneValidCRLFound = atLeastOneValidCRLFound;
     fflush(outputfp.get());
 
     return true;
+}
+
+bool CRLSet::atLeastOneValidCRLFound() const {
+    return m_atLeastOneValidCRLFound;
 }
 
 }
@@ -411,6 +428,7 @@ XrdTlsTempCA::Maintenance()
     //    new_file->getCRLFilename().c_str());
     m_ca_file.reset(new std::string(new_file->getCAFilename()));
     m_crl_file.reset(new std::string(new_file->getCRLFilename()));
+    m_atLeastOneCRLFound = crl_builder.atLeastOneValidCRLFound();
     return true;
 }
 
--- a/src/XrdCrypto/XrdCryptosslX509Crl.cc
+++ b/src/XrdCrypto/XrdCryptosslX509Crl.cc
@@ -44,7 +44,6 @@
 
 #include <openssl/bn.h>
 #include <openssl/pem.h>
-#include <sstream>
 
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
 #define X509_REVOKED_get0_revocationDate(x) (x)->revocationDate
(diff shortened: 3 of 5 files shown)


bbockelm:
Does this actually fix #1543 though?  In the last comment there, you mention it fails if the CRL file is non-empty and contains no valid CRLs.  Shouldn't we use some of the internal XRootD routines to parse the file and see if there's one CRLs present?

abh3:
Yes, this is something to conisder. However, we should also make sure that 
malicious injection of an empty/invalid CRL file does not completely 
eliminate CRL checking for all time. This is a big security loophole. As 
much as I dislike CRL's they do serve a purpose. So, the solution is not 
completely simple.

bbockelm:
> However, we should also make sure that malicious injection of an empty/invalid CRL file does not completely eliminate CRL checking for all time

This is a file periodically generated by XRootD in the admin directory.  I figure if the attacker owns the admin directory, then it's already game over.

It's also an aggregate of all CRLs available - if the attacker controls all your CRLs, then it's probably also game over.

bbockelm:
Indeed, that's a fantastic idea!  I expect this to be a pretty rare occurrence so a warning message seems warranted.

@ccaffy - does this sound good to you?

ccaffy:
Hi all, thanks for your comments!

To sum things up, you would like to disable the CRL checking and display a warning message if there is at least one wrong CRL encountered during the CRL processing?

That's sounds a good plan to me :)

ccaffy:
> Does this actually fix #1543 though? In the last comment there, you mention it fails if the CRL file is non-empty and contains no valid CRLs. Shouldn't we use some of the internal XRootD routines to parse the file and see if there's one CRLs present?

I thought that checking that the aggregated-CRL file is not empty would be sufficient as curl already checks the validity of its content. In my opinion, failing a transfer due to an empty CRL file is not acceptable, but failing a transfer due to a wrong content is acceptable.
Hence why I only added a check on the emptiness of the file ;)

bbockelm:
> To sum things up, you would like to disable the CRL checking and display a warning message if there is at least one wrong CRL encountered during the CRL processing?

Not quite -- I'm thinking that if there are *no* CRLs in the CRL file then we disable CRL checking.  Here, I'm willing to define "valid" as simply parseable as a CRL (not looking closely at the contents).  Particularly, we should have a routine like [this one](https://github.com/xrootd/xrootd/blob/master/src/XrdTls/XrdTlsTempCA.cc#L169-L203); instead of looping through all CRLs in a file, return successful as soon as a single one is present.

Actually, that suggests a simpler approach.  Since we are writing the CRLs to this file in the first place, why not keep track of the number written and, if there were zero, simply don't invoke the corresponding curl option?

ccaffy:
> Actually, that suggests a simpler approach. Since we are writing the CRLs to this file in the first place, why not keep track of the number written and, if there were zero, simply don't invoke the corresponding curl option?

Yes, for me this is fine :) I can implement that :+1:

ccaffy:
I updated the code yesterday :) Please tell me if this is OK for you so that this fix can be merged :)

src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -198,11 +209,17 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
             return false;
         }
     }
+    if(!m_atLeastOneValidCRLFound)
+        m_atLeastOneValidCRLFound = atLeastOneValidCRLFound;
```
abh3, line 213: Doesn't this mean that once a valid CRL is found a subsequent scan that doesn't find one will not be registered as such? Won't that be a problem?

abh3 (commented):
A couple of questions and perhaps needed changes before we merge this. Sorry for taking so long.

.gitignore
```diff
@@ -58,3 +58,5 @@ xrootd.spec
 dist
 *.egg-info
 bindings/python/VERSION
+cmake-*
+.idea/
```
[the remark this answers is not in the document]
ccaffy: Hi Andy,

No problem :)

Because I use CLion as IDE with local code deployed on a VM. The cmake-* directories is the equivalent of the `build` directory.
The .idea directory is created by CLion.
I just do not want these directories to be pushed to the project :)

Now that you said this, it looks like I can put these excludes in my local `.git/info/exclude` file. I will therefore remove these two lines from the .gitignore file

src/XrdTls/XrdTlsTempCA.cc
ccaffy: I only modify the variable `m_atLeastOneValidCRLFound` if this variable is false. So, once a valid CRL is found, the `m_atLeastOneValidCRLFound` will be set to true and therefore will never be assigned anymore during the `XrdTlsTempCA::Maintenance()` loop.

That way, we only enable CRL checking if there is at least one valid CRL found during the Maintenance run :)

abh3: What about this little issue that doesn't seem quite right.

abh3: Yes, but you should consider that the next time the maintenance loop runs there may be no valid CRLs. I know that's unlikely but unlikely things happen. I think you should always update the current state of the maintenance loop otherwise that look does half the j0ob.

ccaffy: If during the next time the maintenance loop runs there is no valid CRL file, do we agree that the CRL checking must be disabled?

ccaffy: Ok, I see what you mean. I just need to reset this boolean to false at the beginning of the Maintenance run :+1:

abh3: I think that if the current CRL file is replaced with a bad CRL file then, yes. It should be disabled. This is dynamic because the maintenance check tries to create a CRL file and if it fails then a) if the previous CRL file is still enforce then we need do nothing, or b) it gets replaced with a null file then we need top disable CRL checking. So, in the end it depends on the state the maintenance run leaves the CRL file (something I have not researched).

abh3:
Well, not really because it is set false at the start (well initial 
state). Let me explain the state:

a) we find a CRL, the state is set to true.

b) second scan and we don't find a CRL.

c) What do we do? Two cases:
    1) The previous file is kept so it's OK to say use the CRL.
    2) The previpus file is replaced with a null file -- not OK.

I think you see the conundrum. It really depends on what happens to the 
compisite CRL file; i.e. is it kept or replaced. So, given that I don't 
know (as a reviewer) it seems safest to just reflect the current state. 
Now, you can elucidete what the case is. However it should be made 
painfully clear in the comments ahead of the test; otherwise we will be 
revisting this for years to come.

Andy

ccaffy:
I thought we wanted the composite CRL file to reflect what is in the `/etc/grid-security/certificates` directory and not trying to see if in the previous run, a CRL that was found is not found anymore or is replaced by a null file in the next run.

Here are the tests I did:

First, best case scenario, there are two valid pem files in the `/etc/grid-security/certificates` directory:

```
//Server side
211201 13:42:09 23886 TPC_TempCA: Reloading the list of CAs and CRLs in directory
```

Composite CRL file:

```
-----BEGIN X509 CRL-----
[...] CRL1 file content
-----END X509 CRL-----
-----BEGIN X509 CRL-----
[...] CRL2 file content
-----END X509 CRL-----
```

Now, I empty the CRL2 file content:
```
root@xrootd-ccaffy-dev01 xrootd-run]# cat /etc/grid-security/certificates/my_crl_file_2.crl
[root@xrootd-ccaffy-dev01 xrootd-run]#
```

... and after the next run of the Maintenance:

```
-----BEGIN X509 CRL-----
[...] CRL1 file content
-----END X509 CRL-----
```

The composite CRL file is re-generated. Containing only the first CRL file as I emptied the second CRL file.

Now I will modify the CRL1 file content to break the CRL correctness of this file:

```
[root@xrootd-ccaffy-dev01 xrootd-run]# truncate -s 500 /etc/grid-security/certificates/my_crl_file.crl
```

Then the composite CRL file is re-generated and contains nothing:

```
[xrootddev@xrootd-ccaffy-dev01 xrootd-run]$ cat /var/spool/xrootd/.xrdtls/crl_file.pem
[xrootddev@xrootd-ccaffy-dev01 xrootd-run]$
```

A TPC transfer will then have the CRL check disabled:

```
211201 13:56:48 23872 TPC_TpcHandler: No valid CRL file has been found in the file /var/spool/xrootd/.xrdtls/crl_file.pem. Disabling CRL checking.
```

Is this the behaviour we expect? Because for me, the composite CRL file should be an image of what is in the `/etc/grid-security/certificates` directory.

To sum things up, the composite CRL file is completely re-generated at each Maintenance run :)

Sorry for the time spent on this topic, but I really need to understand and I want to be sure that you have the full picture of why I did this implementation like this ;)

xrootd-dev:
I don't think that would work all that well because there may be a 
transfer during the maintenance run and then CRLs would not be checked 
eventhough there still is the previous file. Franklly, I would just get 
rid of the if statement gaurd.

Andy

ccaffy:
Alright then, I get rid of the if statement. Thanks for your comments :)

Issue #1544, opened 2021-11-01
[XrdThrottle] basic_string::_S_construct null not valid


dellaric:
Dear experts,

I noticed a reproducible crash when enabling  XrdThrottle (together with XrdHttpTPC, XrdAccSciTokens, XrdMacaroons, XrdMultiuser). It always happens as

211101 13:52:44 22855 XrootdBridge: unknown.24:41@etf-01.cern.ch login as nobody
211101 13:52:44 22855 multiuser_UserSentry: Anonymous client; no user set, cannot change FS UIDs
211101 13:52:44 22855 multiuser_UserSentry: Switching FS uid for user prdcms001
INFO in AuthzFun: Lookup with key /DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=sciaba/CN=430796/CN=Andrea Sciaba::cms:/cms/Role=production,/cms,/cms/ALARM,/cms/GGUSExpert,/cms/TEAM,::
INFO in AuthzFun: Using cached entity with username prdcms001
211101 13:52:47 22907 multiuser_UserSentry: Anonymous client; no user set, cannot change FS UIDs
211101 13:52:47 22855 TPC_PullRequest: event=PULL_START, local=/store/unmerged/SAM/se_webdav_20211101_125243_etf-01.mcpy, remote=https://cmsxrd.ts.infn.it:1094/store/unmerged/SAM/se_webdav_20211101_125243_etf-01.mwrt, user=(anonymous); Starting a push request
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid

and similarly for Push requests. The problem disappears when disabling XrdThrottle.

The output from strace looks like this (hopefully these are the relevant lines):

[pid 22855] writev(2, [{iov_base="211101 13:52:47 22855 ", iov_len=22}, {iov_base="TPC_", iov_len=4}, {iov_base="PullRequest", iov_len=11}, {iov_base=": ", iov_len=2}, {iov_base="event=PULL_START, local=/store/u"..., iov_len=220}, {iov_base="\n", iov_len=1}], 6) = 260
/usr/lib64/libc-2.17.so(writev+0x60) [0xf55d0]
/usr/lib64/libXrdUtils.so.3.0.0(XrdSysLogger::Put(int, iovec*)+0x177) [0x4a577]
[...]

PR #1548, opened 2021-11-02
[XrdThrottle] do not initialise a string with a null pointer

dellaric:
Fix for https://github.com/xrootd/xrootd/issues/1544


bbockelm:
👍 looks good tome.

PR #1560, opened 2021-11-21
 Use 512 bits to generate the prime value for DH handshake

gganis:
Increase the number of bits used in the generation of the prime value for the Diffie-Hellman handshake to increse its 'safeness' and to prevent warning/error messages issues by new versions of the Java Bouncy Castle security library .

### DIFF
--- a/docs/PreReleaseNotes.txt
+++ b/docs/PreReleaseNotes.txt
@@ -6,25 +6,9 @@ Prerelease Notes
 ================
 
 + **New Features**
-  **[Server]** Add new class for gathering config data from config file.`
-  **[Server]** Implement firefly network flow monitoring.
-  **[Server]** Allow prefunctory redirect based on client's net attributes.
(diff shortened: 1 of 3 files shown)


ellert:
Until the old pre-openssl3 code is removed, any change to code that exists in two version must be changed twice.
This PR changes the file:
src/XrdCrypto/XrdCryptosslCipher.hh
It must therefore also introduce the same changes to the file:
src/XrdCrypto/openssl3/XrdCryptosslCipher.hh
In order for the two implementations to be equivalent.
The openssl3 version of the code is compilable also with older openssl versions, and should at some point become the only version, but until this happens, changes must be implemented twice.

simonmichal:
@ellert : noted, I will take care of it later this week.

simonmichal:
@gganis : as it does not merge cleanly I took the liberty and pushed the commit myself: dd54932bfca32acab5c34d269bce81ff38f8aaa5

PR #1568, opened 2021-12-07
[XrdSciTokens] Avoid double slashes in the final path rule that can r…

esindril:
…esult from

concatenating the base path with a path from a name_mapfile rule.

Exmaple:
/etc/xrootd/scitokens.cfg
-------------------------
...
base_path=/
name_mapfile=/etc/xrootd/mapfile.json
...

/etc/xrootd/mapfile.json
-------------------------
[{"sub": "abbaabba", "path":/some/path", "response": "someuser"}]

The resulting rule against which the maching will be attempeted will use the path
"//some/path" which will will never match a legitimate path.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -15,6 +15,7 @@
 #include <fstream>
 #include <unordered_map>
 #include <tuple>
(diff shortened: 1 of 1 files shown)


abh3 (approved):
Not sure this is the best way of handlinng it but it does solve the problem.

PR #1571, opened 2021-12-11
Help older compiler find the right template.

ellert:
../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc: In member function 'bool XrdAccSciTokens::GenerateAcls(const string&, uint64_t&, {anonymous}::AccessRulesRaw&, std::string&, std::string&, std::string&, std::vector<{anonymous}::MapRule>&, std::vector<std::basic_string<char> >&)':
../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:652:83: error: no matching function for call to 'regex_replace(std::string&, std::regex, const char [2])'
                                                              std::regex("//"), "/");
                                                                                   ^
../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:652:83: note: candidates are:
In file included from /usr/include/c++/4.8.2/regex:62:0,
                 from ../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:18:
/usr/include/c++/4.8.2/bits/regex.h:2162:5: note: template<class _Out_iter, class _Bi_iter, class _Rx_traits, class _Ch_type> _Out_iter std::regex_replace(_Out_iter, _Bi_iter, _Bi_iter, const std::basic_regex<_Ch_type, _Rx_traits>&, const std::basic_string<_Ch_type>&, std::regex_constants::match_flag_type)
     regex_replace(_Out_iter __out, _Bi_iter __first, _Bi_iter __last,
     ^
/usr/include/c++/4.8.2/bits/regex.h:2162:5: note:   template argument deduction/substitution failed:
../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:652:83: note:   deduced conflicting types for parameter '_Bi_iter' ('std::basic_regex<char>' and 'const char*')
                                                              std::regex("//"), "/");
                                                                                   ^
In file included from /usr/include/c++/4.8.2/regex:62:0,
                 from ../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:18:
/usr/include/c++/4.8.2/bits/regex.h:2182:5: note: template<class _Rx_traits, class _Ch_type> std::basic_string<_Ch_type> std::regex_replace(const std::basic_string<_Ch_type>&, const std::basic_regex<_Ch_type, _Rx_traits>&, const std::basic_string<_Ch_type>&, std::regex_constants::match_flag_type)
     regex_replace(const basic_string<_Ch_type>& __s,
     ^
/usr/include/c++/4.8.2/bits/regex.h:2182:5: note:   template argument deduction/substitution failed:
../xrootd-5.4.0/src/XrdSciTokens/XrdSciTokensAccess.cc:652:83: note:   mismatched types 'const std::basic_string<_Ch_type>' and 'const char [2]'
                                                              std::regex("//"), "/");
                                                                                   ^

### DIFF
--- a/src/XrdMacaroons/XrdMacaroonsHandler.cc
+++ b/src/XrdMacaroons/XrdMacaroonsHandler.cc
@@ -1,5 +1,4 @@
 
-#include <regex>
 #include <cstring>
 #include <string>
 #include <iostream>
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -15,7 +15,6 @@
 #include <fstream>
 #include <unordered_map>
 #include <tuple>
-#include <regex>
 
 #include "INIReader.h"
 #include "picojson.h"
@@ -648,8 +647,10 @@ private:
             for (auto path : config.m_base_paths) {
                 auto path_rule = rule;
                 path_rule.m_path_prefix = path + rule.m_path_prefix;
-                path_rule.m_path_prefix = std::regex_replace(path_rule.m_path_prefix,
-                                                             std::regex("//"), "/");
+                auto pos = path_rule.m_path_prefix.find("//");
+                if (pos != std::string::npos) {
+                    path_rule.m_path_prefix.erase(pos + 1, 1);
+                }
                 map_rules.emplace_back(path_rule);
             }
         }


simonmichal:
@ellert : thanks for the PR, on which platform / compiler are you getting this problem? Looking at `regex_replace` signatures I see:

```
template <class traits, class charT>
  basic_string<charT> regex_replace (const charT* s,
          const basic_regex<charT,traits>& rgx,
          const charT* fmt,
          regex_constants::match_flag_type flags = regex_constants::match_default);
```

this should much well a `C string`.

Secondly, instead of creating a temporary `std::string("/")` could you please do
```
static const std::string slash( "/" );
path_rule.m_path_prefix = std::regex_replace( path_rule.m_path_prefix, std::regex("//"), slash );
```

ellert:
@simonmichal : The compiler is your old favourite gcc 4.8 on RHEL/CentOS 7.

I have looked into this a bit more, and there is a really sneaky issue here.
Even though the regex_replace with the signature with std::string exists in the version of libstdc++ that accompanies the compiler, and the code compiles with the proposed change, it doesn't work.
Looking into the headers, there is a comment that says:
~~~
   * @todo Implement this function.
~~~
One would have hoped that this would rather fail to compile, or at least result in a compilation warning. But no - it successfully compiles and you have to read the comments in the headers to know it is not working.
So this PR will not fix it. So I suggest not merging it.

I do have a question though. What is the intention of this replacement? I see that it wants to replace 2 consecutive slashes with one, but what is supposed to happen with 3 or more consecutive slashes? Currently 3 or 4 consecutive slashes become 2. 5 or 6 consecutive slashes become 3. And so on. Was this intended? It seems odd. Or was the intention to replace any number of consecutive slashes with a single one. In that case the regex should be "/+", not "//".

esindril:
@ellert you are right, a better regex would be "/+" to be replaced with "/". In this particular case, the individual paths are already in canonical format, so it was just the concatenation which could create paths with double slashes. I can drop the use of std::regex and can use the already provided `MakeCanonical` function from the XrdSciTokensAccess.cc file. Let me know what you prefer.

ellert:
> I can drop the use of std::regex and can use the already provided `MakeCanonical` function from the XrdSciTokensAccess.cc file. Let me know what you prefer.

Yes, please!

ellert:
@esindril : If there can only be one set on consecutive slashes and there can never be more than 2 slashes in this set, the code can be simplified. I have updated th PR with a possible simplification (not using std::regex).

abh3:
Looking at the code for MakeCanonical() we are far better off using 
regex(). So, I very much would prefer *not* to use MakeCanonical() it's 
dreadfully ineffecient.

abh3:
No\, please do not do this. MakeCanonical() is dreadful!

bbockelm:
@esindril -

You know, it strikes me that double-nested for loop is generating strings that depend solely on the configuration -- i.e., you could hoist the code to the configuration functions and generate them once instead of on every authorization.

In terms of `MakeCanonical`: if we don't like the current implementation of `MakeCanonical`, let's make a better implementation.  This is a place where we need a canonical path and, instead of using the helper function to do exactly that, we wrote a single line that fails to make the path canonical in various cases (notice the merged fix handles `//` correctly but not `///`; does not handle `..` or `.`).

xrootd-dev:
As for make canonical, xroot already does this for every incomming path 
and does handle any number of slashes and single dots. What it refuses to 
do is handle double dot because that effectively is a relative path which 
aren't allowed. Never heard a peep. I am not convinced that use cases that 
will never (or should never) happen need to be covered with anything other 
than rejection. Usually solutions for those are complex enough to be 
easily coded in a buggy fashion that won't be obvious until it creates a 
security problem and then all hell breaks loose. I put dot dot in that 
category as it's created nightmare situations in the past, even in Unix.

Andy

Issue #1538, opened 2021-10-25
Add mapfile support for XrdVoms


matyasselmeci:
We'd like to be able to create a mapfile for voms authentication such that we can choose the mappings between FQANs and Unix users; this is needed for multiuser to work. VOMS mapfiles are made of lines like

    "FQAN-PATTERN" USERNAME

where USERNAME is a Unix user and FQAN-PATTERN is a glob.

For GSI, mapfile support is built in, but we've been told it's better to write a post-auth plugin for VOMS instead.

bbockelm:
Hey Wei,

I think the VOMS mapfile tackles the case where there's an underlying storage system that really needs a "user name" that can't be derived algorithmically via the FQAN.

Examples I know about are EOS and xrootd-multiuser (the former does the "VOMS mapfile" by converting a VOMS FQAN to a long list of DNs and maps those to a single username... that kinda works but is problematic for other reasons).

Brian

abh3:
Brian kindly provided a patch for this, so I am closing this ticket. This should appear in 5.5.0.

PR #1572, opened 2021-12-12
Add support for VOMS mapfile

bbockelm:
Add a mapfile for VOMS FQANs.  This allows the VOMS plugin to map the contents of the VOMS extension to a name in the `XrdSecentity`.

This adds two new configuration parameters, `voms.mapfile` and `voms.trace`.  Here's an example invocation:

```
voms.mapfile /etc/grid-security/voms-mapfile
voms.trace debug
```

Where the mapfile is in the format used by XrdLcmaps:

```
"/cms/Role=lcgadmin/Capability=NULL" lcgadmin
"/cms/*" cmsuser
"/des/Role=Analysis/Capability=NULL" des
```

At the `debug` level, things are pretty chatty - including dumping each line during parsing and each mapping invocation.

The mapfile is re-parsed every 30 seconds.

### DIFF
--- a/src/XrdHttp/XrdHttpSecurity.cc
+++ b/src/XrdHttp/XrdHttpSecurity.cc
@@ -165,7 +165,7 @@ XrdHttpProtocol::HandleGridMap(XrdLink* lp)
       TRACEI(DEBUG, " Mapping name: '" << SecEntity.moninfo << "' --> " << bufname);
       if (SecEntity.name) free(SecEntity.name);
       SecEntity.name = strdup(bufname);
-      SecEntity.eaAPI->Add("gridmap.name", bufname, true);
+      SecEntity.eaAPI->Add("gridmap.name", "1", true);
     }
     else {
       TRACEI(ALL, " Mapping name: " << SecEntity.moninfo << " Failed. err: " << mape);
--- a/src/XrdHttp/XrdHttpSecurity.cc
+++ b/src/XrdHttp/XrdHttpSecurity.cc
@@ -26,6 +26,7 @@
 #include "XrdCrypto/XrdCryptoX509Chain.hh"
 #include "XrdCrypto/XrdCryptosslAux.hh"
 #include "XrdCrypto/XrdCryptoFactory.hh"
+#include "XrdSec/XrdSecEntityAttr.hh"
 #include "XrdTls/XrdTlsPeerCerts.hh"
 #include "XrdTls/XrdTlsContext.hh"
 #include "XrdOuc/XrdOucGMap.hh"
@@ -164,6 +165,7 @@ XrdHttpProtocol::HandleGridMap(XrdLink* lp)
       TRACEI(DEBUG, " Mapping name: '" << SecEntity.moninfo << "' --> " << bufname);
       if (SecEntity.name) free(SecEntity.name);
       SecEntity.name = strdup(bufname);
+      SecEntity.eaAPI->Add("gridmap.name", bufname, true);
     }
     else {
       TRACEI(ALL, " Mapping name: " << SecEntity.moninfo << " Failed. err: " << mape);
--- a/src/XrdVoms/XrdVomsHttp.cc
+++ b/src/XrdVoms/XrdVomsHttp.cc
@@ -29,7 +29,6 @@
 
 // This code is based on
 
-#include "XrdVomsMapfile.hh"
 #include "XrdVomsFun.hh"
 
 /** @brief This code is based on the basic architecture shown in
@@ -143,9 +142,7 @@ XrdHttpSecXtractor *XrdHttpGetSecXtractor(XrdHttpSecXtractorArgs)
 
 // Now return the interface object
 //
-   auto base = static_cast<XrdHttpSecXtractor *>(new XrdVomsHttp(eDest, *vomsFun));
-   auto wrapper = static_cast<XrdHttpSecXtractor *>(XrdVomsMapfile::Configure(eDest, base));
-   return wrapper ? wrapper : base;
+   return (XrdHttpSecXtractor *)new XrdVomsHttp(eDest, *vomsFun);
 }
 
 /******************************************************************************/
--- a/src/XrdVoms/XrdVomsHttp.cc
+++ b/src/XrdVoms/XrdVomsHttp.cc
@@ -29,6 +29,7 @@
 
 // This code is based on
 
+#include "XrdVomsMapfile.hh"
 #include "XrdVomsFun.hh"
 
 /** @brief This code is based on the basic architecture shown in
@@ -142,7 +143,9 @@ XrdHttpSecXtractor *XrdHttpGetSecXtractor(XrdHttpSecXtractorArgs)
 
 // Now return the interface object
 //
-   return (XrdHttpSecXtractor *)new XrdVomsHttp(eDest, *vomsFun);
+   auto base = static_cast<XrdHttpSecXtractor *>(new XrdVomsHttp(eDest, *vomsFun));
+   auto wrapper = static_cast<XrdHttpSecXtractor *>(XrdVomsMapfile::Configure(eDest, base));
+   return wrapper ? wrapper : base;
 }
 
 /******************************************************************************/
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -263,7 +263,7 @@ XrdVomsMapfile::Apply(XrdSecEntity &entity)
     // set when the mapfile is used to generate the name.
     std::string gridmap_name;
     auto gridmap_success = entity.eaAPI->Get("gridmap.name", gridmap_name);
-    if (gridmap_success) {
+    if (gridmap_success && gridmap_name == "1") {
         return 0;
     }
 
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -400,6 +400,12 @@ XrdVomsMapfile::MaintenanceThread(void *myself_raw)
            myself->m_is_valid = false;
            continue;
        }
+       // Use ctime here as it is solely controlled by the OS (unlike mtime,
+       // which can be manipulated by userspace and potentially not change
+       // when updated - rsync, tar, and rpm, for example, all preserve mtime).
+       // ctime also will also be updated appropriately for overwrites/renames,
+       // allowing us to detect those changes as well.
+       //
        if ((myself->m_mapfile_ctime.tv_sec == statbuf.st_ctim.tv_sec) &&
            (myself->m_mapfile_ctime.tv_nsec == statbuf.st_ctim.tv_nsec))
        {
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -428,11 +428,11 @@ XrdVomsMapfile::MaintenanceThread(void *myself_raw)
            if (rval == EINTR) continue;
            else break;
        } else if (rval == 0) { // timeout!  Let's run maintenance.
+           next_update = monotonic_time_s() + m_update_interval;
            struct stat statbuf;
            if (-1 == stat(myself->m_mapfile.c_str(), &statbuf)) {
                myself->m_edest->Emsg("XrdVomsMapfile", errno, "Error checking the mapfile",
                    myself->m_mapfile.c_str());
-               next_update = monotonic_time_s() + m_update_interval_failure;
                myself->m_mapfile_ctime.tv_sec = 0;
                myself->m_mapfile_ctime.tv_nsec = 0;
                myself->m_is_valid = false;
@@ -443,16 +443,13 @@ XrdVomsMapfile::MaintenanceThread(void *myself_raw)
            {
                myself->m_edest->Log(LogMask::Debug, "Maintenance", "Not reloading VOMS mapfile; "
                    "no changes detected.");
-               next_update = monotonic_time_s() + m_update_interval;
                continue;
            }
            memcpy(&myself->m_mapfile_ctime, &statbuf.st_ctim, sizeof(decltype(statbuf.st_ctim)));
 
            myself->m_edest->Log(LogMask::Debug, "Maintenance", "Reloading VOMS mapfile now");
-           if ( (myself->m_is_valid = myself->ParseMapfile(myself->m_mapfile)) ) {
-               next_update = monotonic_time_s() + m_update_interval;
-           } else {
-               next_update = monotonic_time_s() + m_update_interval_failure;
+           if ( !(myself->m_is_valid = myself->ParseMapfile(myself->m_mapfile)) ) {
+               myself->m_edest->Log(LogMask::Error, "Maintenance", "Failed to reload VOMS mapfile");
            }
        } else { // FD ready; let's shutdown
            if (fds.revents & POLLIN) {
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -228,8 +228,11 @@ XrdVomsMapfile::MakePath(const XrdOucString &group)
     XrdOucString entry;
     std::vector<std::string> path;
     path.reserve(4);
-    while ((from = group.tokenize(entry, from, '/')) != -1) {
-        if (entry.empty()) continue;
+        // The const'ness of the tokenize method as declared is incorrect; we use
+        // const_cast here to avoid fixing the XrdOucString header (which would break
+        // the ABI).
+    while ((from = const_cast<XrdOucString&>(group).tokenize(entry, from, '/')) != -1) {
+        if (entry.length() == 0) continue;
         path.emplace_back(entry.c_str());
     }
     return path;
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -72,18 +72,6 @@ XrdVomsMapfile::XrdVomsMapfile(XrdSysError *erp,
     const std::string &mapfile)
     : m_mapfile(mapfile), m_edest(erp)
 {
-    // Setup communication pipes; we write one byte to the child to tell it to shutdown;
-    // it'll write one byte back to acknowledge before our destructor exits.
-    int pipes[2];
-    if (-1 == XrdSysFD_Pipe(pipes)) {
-        m_edest->Emsg("XrdVomsMapfile", "Failed to create communication pipes", strerror(errno));
-        return;
-    }
-    if (-1 == XrdSysFD_Pipe(pipes)) {
-        m_edest->Emsg("XrdVomsMapfile", "Failed to create communication pipes", strerror(errno));
-        return;
-    }
-
     struct stat statbuf;
     if (-1 == stat(m_mapfile.c_str(), &statbuf)) {
         m_edest->Emsg("XrdVomsMapfile", errno, "Error checking the mapfile", m_mapfile.c_str());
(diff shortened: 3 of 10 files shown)


bbockelm:
Oh!  I should mention that the configuration parameters are of the variety "last time set wins", making it easier for administrators to provide an override file.

brianhlin:
Addresses #1538

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -0,0 +1,359 @@
...
+    if (tried_configure) {
+        auto result = mapper.get();
+        if (result) {
+            result->SetExtractor(xtractor);
+            result->SetErrorStream(erp);
+        }
+        return result;
+    }
+
+    tried_configure = true;
```
abh3, line 305: Why do we need this at all. The assumption, code-wise, is that this will be called during initialization and hence will be called only once. So, it's very confusing why this code exists here at all. Is there something we don't know about?

src/XrdVoms/XrdVomsHttp.cc
```diff
@@ -142,7 +143,9 @@ XrdHttpSecXtractor *XrdHttpGetSecXtractor(XrdHttpSecXtractorArgs)
 
 // Now return the interface object
 //
-   return (XrdHttpSecXtractor *)new XrdVomsHttp(eDest, *vomsFun);
+   auto base = static_cast<XrdHttpSecXtractor *>(new XrdVomsHttp(eDest, *vomsFun));
+   auto wrapper = static_cast<XrdHttpSecXtractor *>(XrdVomsMapfile::Configure(eDest, base));
+   return wrapper ? wrapper : base;
```
abh3, line 148: I have a problem with this a) it totally ignores any errors that may occur during configuration and this has always been a site complaint about silent errors that should be fatal, and b) this could have been done in a much more straightforward way by simply inserting a test for mapping in XrdVomsHttp::GetSecData() instead using an indirect wrapper. Though, frankly, I would do it the other way around and pass a mapper to the voms function itself (or a null pointer) so that this works for http as well as xrootd. Then we have a chance to make it consistent across all protocols.

src/XrdVoms/XrdVomsgsi.cc
```diff
@@ -57,7 +59,11 @@ int XrdSecgsiVOMSFun(XrdSecEntity &ent)
 {
 // Make sure we were initialized. If so, invoke the function and return result.
 //
-   return (vomsFun ? vomsFun->VOMSFun(ent) : -1);
+   int retval = (vomsFun ? vomsFun->VOMSFun(ent) : -1);
+   if (retval == -1) {return retval;}
+
+   auto mapfile = XrdVomsMapfile::Get();
+   return mapfile ? mapfile->Apply(ent) : retval;
```
abh3, line 66: As per my previous comments, we really don't need a change here. The mapping should be part of the core VomsFun and the interface to the xroot-side initialization should pretty much do what the http side does. That way it's all in one place and you clear can see it's part of the core logic. Plus it guarantees that the map will be applied if the "voms.mapfile" directive is in the config file irrespective of the protocol.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -0,0 +1,359 @@
...
+    if (now > m_last_update.load(std::memory_order_relaxed) + 30) {
+        retval = ParseMapfile(m_mapfile);
+        m_last_update.store(now, std::memory_order_relaxed);
+    }
+    return retval;
+}
+
+
+bool
+XrdVomsMapfile::ParseLine(const std::string &line, std::vector<std::string> &entry, std::string &target)
```
abh3, line 109: We will need documentation on the format of the mapfile as well as the rules that are used for the actual mapping. Can you provide such a document so that it can be included in the security reference.

abh3 (changes requested):
Good start but there are various things that really need to be changed. I likely did not find all of them but that's what iteration is all about. I also look at this as a way to address another problem you raised is how do we make this consistent across all protocols and do not require each one to specify the same thing using a different directive.

src/XrdOuc/XrdOucString.cc
```diff
@@ -257,7 +257,7 @@ int XrdOucString::form(XrdOucString &str, const char *fmt, ...)
 #endif
 
 //______________________________________________________________________________
-int XrdOucString::find(const char c, int start, bool forward)
+int XrdOucString::find(const char c, int start, bool forward) const
```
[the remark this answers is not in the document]
bbockelm: Rats.  I thought `const`'ness of the symbol wasn't part of the GCC C++ ABI... turns out it's not part of the GCC C ABI.

No problem, will revert and do ugly `const_cast` workarounds at the call sites.

src/XrdVoms/XrdVomsHttp.cc
bbockelm: I did the wrapper because I thought that's what we agreed to in the phone call -- keep the code outside the existing XrdVoms class.  Will be happy to put it inside the `XrdVomsHttp` class itself if you're OK with that.

I am reluctant to create a new version of the plugin protocol as that would probably require me to add even more arguments to the GSI configuration, which is already much too long.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: Sure.  How about I also add it to a module-specific README?  The parsing rules are taken from here:

https://argus-documentation.readthedocs.io/en/stable/misc/grid_map_file.html

(except I will decline to support generic 8-bit ASCII or generic unicode escape sequences as those are not permitted in IGTF... would rather someone really convince me unicode support in DNs is a good idea!)

I wasn't able to find a document for the wildcard / matching rules so I took them from the existing implementation in order to allow sites to reuse their existing voms-mapfiles.  Would be good to have them written down.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: It's necessary in the current setup: the two protocols are configured independently.  Hence, `XrdVoms` is invoked twice whereas i want the mapper to be a singleton.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -0,0 +1,359 @@
...
+    if (retval) return retval;
+
+    return Apply(entity);
+}
+
+
+int
+XrdVomsMapfile::Apply(XrdSecEntity &entity)
+{
+    Reconfigure();
```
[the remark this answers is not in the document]
abh3: The alternative solution is to extend the time to something reasonable but keep track of the mtime and reparse only if the mtime changed within that interval. This also works but at the cost but random slowness in doing the mapping when you have to reparse. The thread solution avoids this problem.

src/XrdVoms/XrdVomsMapfile.cc
abh3: OK, could you please add a comment for why the code is actually there?

src/XrdVoms/XrdVomsHttp.cc
abh3: Two comments:
1) Indeed we did agree to that but that was before issue 1577 which requested we converge the voms mapfile support. That is a good idea but for a clean convergence the code would have to be moved into the plugin. However, that's fairly easy to do given the wrapper which is used for xroot and http. So, that's why I did a turn-around. 

2) Actually, you really don't have to change the plugin protocol, just enhance it. So, if you return MAP_FAILED instead of zero then the caller will know there was a fatal error while zero (still used by old plugins) would mean to ignore the plugin for whatever reason. Note that MAP_FAILED is defined in sys/mman.h and specifically is used to indicate that a memory address could not be established due to am error.

src/XrdVoms/XrdVomsMapfile.cc
abh3: A READ_ME file is OK and I suppose I will have to convert it to reference text as people defining the map file generally do not install the source. Pointing then to the argus page won't help either because, well as you said, you don't support all of it and completely agree with you we don't need to support all of it.

abh3:
@bbockelm Addressed your comments. I think we are converging.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: The README file is added in the latest set of commits.  Resolving.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: Comment is added in the latest set of commits.  Resolving.

src/XrdVoms/XrdVomsMapfile.cc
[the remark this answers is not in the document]
bbockelm: Decided in the end to go with the separate thread as you originally suggested.  See latest set of commits.

src/XrdOuc/XrdOucString.cc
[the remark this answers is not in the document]
bbockelm: All is reverted.

src/XrdOuc/XrdOucString.hh
```diff
@@ -280,24 +280,25 @@ public:
...
+   bool          empty() const { return length() == 0; }
    char         &operator[](int j);
-   int           find(const char c, int start = 0, bool forward = 1);
-   int           find(const char *s, int start = 0);
-   int           find(XrdOucString s, int start = 0);
-   int           rfind(const char c, int start = STR_NPOS)
+   int           find(const char c, int start = 0, bool forward = 1) const;
+   int           find(const char *s, int start = 0) const;
+   int           find(XrdOucString s, int start = 0) const;
+   int           rfind(const char c, int start = STR_NPOS) const
```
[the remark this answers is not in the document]
bbockelm: All is reverted.

src/XrdVoms/XrdVomsHttp.cc
bbockelm: Latest commit refactors this class to be a core part of the plugin as requested.  Resolving.

src/XrdVoms/XrdVomsgsi.cc
bbockelm: Resolving conversation - this is now done inside the VOMSFun core.

bbockelm:
@abh3 - I believe the latest refactor has addressed all your comments.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -367,8 +413,70 @@ XrdVomsMapfile::Configure(XrdSysError *erp, XrdHttpSecXtractor *xtractor)
...
+{
+    auto myself = static_cast<XrdVomsMapfile*>(myself_raw);
+
+   auto now = monotonic_time_s();
+   auto next_update = now + m_update_interval;
+   while (true) {
+       now = monotonic_time_s();
+       auto remaining = next_update - now;
+       struct pollfd fds;
+       fds.fd = myself->m_maintenance_pipe_r;
```
abh3, line 424: My take is this is only for the case when the object is destroyed and you want to stop the maintenance thread. OK, I get that but the object will never be destroyed as far as I can see. Once the mapfile is started it continues for the life of the server. So, I am confused as to what problem this is trying to solve.

abh3 (changes requested):

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -164,6 +165,7 @@ XrdHttpProtocol::HandleGridMap(XrdLink* lp)
       TRACEI(DEBUG, " Mapping name: '" << SecEntity.moninfo << "' --> " << bufname);
       if (SecEntity.name) free(SecEntity.name);
       SecEntity.name = strdup(bufname);
+      SecEntity.eaAPI->Add("gridmap.name", bufname, true);
```
abh3, line 168: I see this consistently set but who is actually going to look at that? I thought the whole idea is that the entity.name would hold this value. So, I must be mistaken. Can you explain the reasoning here (I am sure it will be obvious once you do).

bbockelm: It's not obvious!

It's a bit of a problem of order of precedence.  For the final value of `entity.name`, want the following order:
1.  Explicit entries in the grid mapfile.
2. Explicit entries in the VOMS mapfile
3. Whatever the default behavior is for the GSI plugin as a whole (e.g., potentially an auto-generated name).

Unfortunately, the naive solution results in an ordering of (1), (3), then (2).  Thus, we add any explicit usage of the grid mapfile to the entity so we can distinguish a positive entry from a default fallback.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: It is a bit overkill - I only triggered it during the destruction of the object when the startup configuration failed.

I figured I'd err toward meticulous cleanup (the inspiration is in the temp CA handler code) instead of just assuming nothing needs to be torn down.

src/XrdVoms/XrdVomsMapfile.hh
```diff
@@ -73,14 +77,35 @@ private:
...
-    std::atomic<time_t> m_last_update;
+    // Pipes to allow the main thread to communicate shutdown events to the maintenance
+    // thread, allowing for a clean shutdown.
+    int m_maintenance_pipe_r{-1};
+    int m_maintenance_pipe_w{-1};
+    int m_maintenance_thread_pipe_r{-1};
+    int m_maintenance_thread_pipe_w{-1};
+
+        // After success, how long to wait until the next mapfile check.
+    static constexpr unsigned m_update_interval = 30;
```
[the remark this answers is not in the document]
bbockelm: I based this on experience -- how long does the sysadmin wait for a refresh before complaining on Slack that their change isn't taking effect?

Don't have strong feelings though.

src/XrdVoms/XrdVomsMapfile.hh
```diff
@@ -73,14 +77,35 @@ private:
...
+        // After failure, how long to wait until the next mapfile check.
+    static constexpr unsigned m_update_interval_failure = 3;
```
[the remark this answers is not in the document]
bbockelm: Ok, sounds good.

src/XrdVoms/XrdVomsMapfile.hh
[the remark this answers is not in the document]
abh3: I see your point. Maybe this should be parametrized so that everyone gets to specify their own version of anticipation.

src/XrdVoms/XrdVomsMapfile.hh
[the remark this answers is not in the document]
abh3: OK, I guess the easiest change is to make it the same value as the default check time. Didn't see the change here.

abh3 (commented):
OK, I think we are down to one alternate proposal which you may or may not want to make and one that you agreed to but wasn't yet made. It's all in the new comments.

src/XrdHttp/XrdHttpSecurity.cc
bbockelm: The VOMS plugin runs after the GSI authentication logic; the GSI authentication logic that generates the `entity.name` may be based on the mapfile _or_ the default logic.  What's added here allows one to differentiate those two cases.

We need to know the difference between the default and explicit mapping because, in one case (the default), the VOMS mapfile overrides the `entity.name` value.  In the other case, the VOMS mapfile should keep the `entity.name` untouched.

bbockelm:
Ok, I think I addressed the two items!  Assuming that my comment about the order-of-logic between the GSI & VOMS plugins makes sense, i think it's ready to go.

src/XrdVoms/XrdVomsMapfile.hh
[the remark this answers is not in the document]
bbockelm: Just used the default refresh time in the latest commits as suggested.  Resolving!

src/XrdVoms/XrdVomsMapfile.hh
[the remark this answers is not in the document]
bbockelm: I suspect we'll be good for now -- if an admin has a reasonable argument to make this a parameter, we can do that in a follow-up PR.

bbockelm:
@abh3 - quick ping on this for next week.  I think all is ready (or, at least, ready for the next round of input!).

src/XrdHttp/XrdHttpSecurity.cc
abh3: OK, I now get it. However, couldn't we accomplish the same thing with a one bit flags to indicate the type of ID in the the name. Then Voms could do the right thing. That would be more straightforward than recording the full name as it looks like it going to be used for some other purpose that a simple "type" check. BTW we do have bit in the secentity structure. What do you think?

abh3 (commented):
OK, one final comment on how to make the code less obtuse when it comes to deciding on how to handle the name filed. I'd like your comments. Everything else looks fine.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -0,0 +1,417 @@
...
+       if (-1 == stat(myself->m_mapfile.c_str(), &statbuf)) {
+           myself->m_edest->Emsg("XrdVomsMapfile", errno, "Error checking the mapfile",
+               myself->m_mapfile.c_str());
+           myself->m_mapfile_ctime.tv_sec = 0;
+           myself->m_mapfile_ctime.tv_nsec = 0;
+           myself->m_is_valid = false;
+           continue;
+       }
+       if ((myself->m_mapfile_ctime.tv_sec == statbuf.st_ctim.tv_sec) &&
+           (myself->m_mapfile_ctime.tv_nsec == statbuf.st_ctim.tv_nsec))
```
abh3, line 404: Are you sure you want ctime? I suppose you are banking on the fact that the file will be effectively recreated with say an "mv", right? Can we depend on that or should be ignore changes unless that has actually happened. Either way a comment would be appreciated to that effect.

abh3 (commented):
Missed handling when to reprocess the map file. I think we only need a comment to explain why ctime is used on not mtime.

OK, one final comment on how to make the code less obtuse when it comes to deciding on how to handle the name filed. I'd like your comments. Everything else looks fine.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: Yup - `ctime` can't be manipulated by userspace tools, making it more reliable for detecting changes rolled out by some external tooling or renames.  Added the comment on this!

abh3:
Thanks for adding the comment. Now, what about using a type flag for the name instead of recording the actual name? I propose this because it would be very apparent what is being done here in terms of ordering the name mapping by using a type flag.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -420,51 +385,33 @@ XrdVomsMapfile::MaintenanceThread(void *myself_raw)
...
+       {
+           myself->m_edest->Log(LogMask::Debug, "Maintenance", "Not reloading VOMS mapfile; "
+               "no changes detected.");
+           continue;
+       }
+       memcpy(&myself->m_mapfile_ctime, &statbuf.st_ctim, sizeof(decltype(statbuf.st_ctim)));
 
-   return nullptr;
```
dellaric: The return nullptr line wants to stay:

/tmp/xrootd-5.4.1-rc1/src/XrdVoms/XrdVomsMapfile.cc: In static member function static void* XrdVomsMapfile::MaintenanceThread(void*)':
/tmp/xrootd-5.4.1-rc1/src/XrdVoms/XrdVomsMapfile.cc:423:1: error: no return statement in function returning non-void [-Werror=return-type]
 }
 ^

Regards,
Giuseppe.

src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -0,0 +1,423 @@
...
+               "no changes detected.");
+           continue;
+       }
+       memcpy(&myself->m_mapfile_ctime, &statbuf.st_ctim, sizeof(decltype(statbuf.st_ctim)));
+
+       myself->m_edest->Log(LogMask::Debug, "Maintenance", "Reloading VOMS mapfile now");
+       if ( !(myself->m_is_valid = myself->ParseMapfile(myself->m_mapfile)) ) {
+           myself->m_edest->Log(LogMask::Error, "Maintenance", "Failed to reload VOMS mapfile");
+       }
+   }
```
abh3, line 422: Apparently we neeed a return statement here. Suggestion is

return nullptr;

See comment from Giuseppe.

abh3 (changes requested):
Giuseppe found an apparent missing return statement.

src/XrdVoms/XrdVomsMapfile.cc
bbockelm: Addressed in latest commits.

bbockelm:
@abh3 - I think I have the last two items cleaned up.  We now return nullptr where necessary and switch to recording a single bit (as `1`) instead of a redundant name.

abh3:
Thank you Brian.

PR #1578, opened 2021-12-16
Save token subject in the XrdSecEntity extended attribute API

bbockelm:
Hi @esindril !

Your question the other day about the token subject got me thinking and I was able to identify a use case where the subject was useful independent of the (authorization system): we can utilize it for the throttling plugin.

Mind you, I still think it's a bad idea to feed this to EOS... but with this it's at least available in the `token.subject` xattr for the incoming request.

Brian

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -212,11 +212,11 @@ struct IssuerConfig
 class XrdAccRules
 {
 public:
-    XrdAccRules(uint64_t expiry_time, const std::string &username, const std::string &token_username,
+    XrdAccRules(uint64_t expiry_time, const std::string &username, const std::string &token_subject,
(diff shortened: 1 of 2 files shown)


esindril:
Hi @bbockelm,

I think that properly modifying the WLCG JWT compliance tests and the expectations in terms of token scopes make this requirement obsolete for EOS. Believe me, I am also not of fan of sending such raw token info down the line to EOS, but I was under the impression that the compliance tests were "set in stone" and agreed upon by the wider community - and at that point passing this info to EOS was the only way it could work.

Nevertheless, it's good to know this exists ... who knows what the future will bring! 

Thanks,
Elvin

bbockelm:
Ping on this PR ... would be nice to return to this in the new year.

xrootd-dev:
It's on the list. The delay was due to our mandatory two week recharge 
holiday. Never logged in which is what SLAC wanted.

PR #1583, opened 2022-01-05
[Question] Missing return when dns error in delegation

chaen:
Hi,

we have seen this error happening but the operation still succeeded. I happened to notice that there is no `return` statement. No idea if it is by design or not.


abh3:
Actually, it is by design. The message in question is merely a warning. Perhaps a bit too aggressive since the config said that DNS was supposed to be used but we thought silent bypasses weren't such a good idea. As you should see, this situation simply forbids delegation but otherwise if we can accomplish the required authentication without delegation then all will be fine. By putting in the return statement that possibility is foreclosed. So, in fact, the pull request is not what we really want here.

PR #1587, opened 2022-01-11
Update to XrdCl and XrdEC

wyang007:
1. add pgread/pgwrite
2. avoid 'qdl' delay during read
3. avoid servers in ServerPending state during write
4. store a version number to an Xattr in order to identify (and ignore) old zip archived left by previous deletion
5. set chunk names inside a zip archive to "obj.1.0", etc (not tied to the file name) so that we can rename
6. Set env XRDCL_EC so that other code can act accordingly
7. bug fix


simonmichal:
I had a look and it looks good, just one comment, could you use this version of `SetXAttr`:
https://github.com/xrootd/xrootd/blob/7c6dfb6b74ea30121e92e257c909dd6b9eed9bf6/src/XrdCl/XrdClFileOperations.hh#L959-L962

this way we won't need two separate requests to set the extended file attributes.

PR #1586, opened 2022-01-11
fix: Install Python bindings with `pip` if available

matthewfeickert:
* This is part two of two PRs that will address Issue #1579.
* Requires PR #1585 to go in first for atomicity, and then will rebase on top of it.
* Resolves #1579
   - There might be a few more things to _really_ resolve this, but I think this is good enough for now.

---

As directly invoking `python setup.py` is considered deprecated behavior by the PyPA, use `python -m pip install` to install the Python bindings if the Python interpreter has a valid version of pip associated with it.

If there is a valid pip, build and install the Python bindings with `python -m pip install`.

If there is no valid pip, warn the user and give them instructions on how to get pip on their machine. Attempt to still install the Python bindings by falling back on the deprecated `python setup.py build` and `python setup.py install` which will also invoke Easy Install if `setuptools` `v0.60.0+` is used.

Both branches have been tested in CI (well, both have been tested on a branch on my fork where I force there to be no `pip` available on one. In the CI that ran for this PR all situations have `pip` available.)

<details>
<summary>(and in Docker image builds):</summary>

## Building with `python -m pip install`

```console
$ docker run --rm -ti xrootd/xrootd:build-pip-pass 
[root@f6701ee26ac7 /]# python -c "import pyxrootd; print(pyxrootd)"
<module 'pyxrootd' from '/usr/local/venv/lib/python3.8/site-packages/pyxrootd/__init__.py'>
[root@f6701ee26ac7 /]# python -c "import XRootD; print(XRootD)"
<module 'XRootD' from '/usr/local/venv/lib/python3.8/site-packages/XRootD/__init__.py'>
[root@f6701ee26ac7 /]# command -v xrootd
/usr/local/venv/bin/xrootd
[root@f6701ee26ac7 /]# command -v xrdcp
/usr/local/venv/bin/xrdcp
[root@f6701ee26ac7 /]# python
Python 3.8.11 (default, Aug  2 2021, 20:01:57) 
[GCC 8.3.1 20190311 (Red Hat 8.3.1-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from XRootD import client
>>> from XRootD.client.flags import DirListFlags, OpenFlags, MkDirFlags, QueryCode
>>> myclient = client.FileSystem('root://someserver:1094')
>>> myclient
<XRootD.client.filesystem.FileSystem object at 0x7fb564b22fa0>
>>> 
[root@f6701ee26ac7 /]#
```

## Falling back to `python setup.py install`

```console
[...]

PR #1585, opened 2022-01-11
fix: Use `shutil.which` over `distutils.spawn.find_executable` when possible

matthewfeickert:
This is part one of two PRs that will address Issue #1579.

As `distutils` and Easy Install are deprecated

c.f.:
* https://setuptools.pypa.io/en/stable/deprecated/distutils-legacy.html
* https://setuptools.pypa.io/en/stable/deprecated/easy_install.html

it is preferable convert use  of `distutils` to either use existing CPython libraries or use `setuptool`'s versions of them, as advocated by the PyPA.

* [`distutils.sysconfig` -> CPython's `sysconfig`](https://discuss.python.org/t/understanding-site-packages-directories/12959/4)
* `distutils.setup` -> `setuptools.setup`
* `distutils.core.Extension` -> `setuptools.Extension`
* `distutils.spawn.find_executable` -> [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which) 

However, this PR only applies

* `distutils.spawn.find_executable` -> `shutil.which`

for the packaging scripts. (@henryiii pointed out that `shutil.which` (added in Python 3.3) is preferable to `setuptools._distutils.spawn.find_executable`.)

This PR was meant to also change:

* `distutils.setup` -> `setuptools.setup`
* `distutils.sysconfig` -> CPython's `sysconfig`

but it seems that these are still required given the current state of the build and will need to be revised in a later PR.

The change

`distutils.core.Extension` -> `setuptools.Extension`

could be applied in:

* `bindings/python/setup_pypi.py`
* `bindings/python/setup_standalone.py`

but isn't as these files are obsolete and should be removed (c.f. Issue #1594).

To ensure the RPM build continues to succeed `python2-setuptools` and `python3-setuptools` are added as `BuildRequires`.

To make the packaging scripts as easy to work with as possible for future releases of `pip` and `setuptools` (c.f. Issue #1579), have the `packaging/wheel/install.sh` script use `pip` to perform a local install if `shutil.which` is available (falling back to the deprecated `python setup.py install` if not) and have `packaging/wheel/publish.sh` attempt to use `build` to build the `sdist` (falling back to the deprecated `python setup.py sdist` if it can't access PyPI).

---

[...]

Issue #1579, opened 2021-12-21
Python wheel `install.sh` is using deprecated instructions as of setuptools v0.60.0


matthewfeickert:
## Problem

Hi. :wave: With the release of [`setuptools` `v0.60.0` on 2021-12-19](https://pypi.org/project/setuptools/60.0.0/), the Python [wheel `install.sh`](https://github.com/xrootd/xrootd/blob/af217b9a5c196f921bdc460930390828a9f927f4/packaging/wheel/install.sh) script will produce warnings (here shown for Python 3.9) of the form 

```pytb
/usr/local/venv/lib/python3.9/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/local/venv/lib/python3.9/site-packages/setuptools/command/easy_install.py:156: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
```

and the resulting directory structure in `site-packages` will change from installing `pyxrootd` in `site-packages`'s top level (correctly)

```console
root@d68489115a42:/# ls -lhtr /usr/local/venv/lib/python3.9/site-packages/
total 48K
drwxr-xr-x 2 root root 4.0K Dec 21 05:23 wheel-0.37.0.dist-info
drwxr-xr-x 5 root root 4.0K Dec 21 05:23 wheel
drwxr-xr-x 5 root root 4.0K Dec 21 05:23 pip
drwxr-xr-x 2 root root 4.0K Dec 21 05:23 pip-21.3.1.dist-info
drwxr-xr-x 7 root root 4.0K Dec 21 06:22 setuptools
drwxr-xr-x 6 root root 4.0K Dec 21 06:22 pkg_resources
-rw-r--r-- 1 root root  152 Dec 21 06:22 distutils-precedence.pth
drwxr-xr-x 3 root root 4.0K Dec 21 06:22 _distutils_hack
drwxr-xr-x 2 root root 4.0K Dec 21 06:22 setuptools-59.8.0.dist-info
-rw-r--r-- 1 root root  252 Dec 21 06:24 xrootd-v5.4.0-py3.9.egg-info
drwxr-xr-x 3 root root 4.0K Dec 21 06:24 pyxrootd
drwxr-xr-x 4 root root 4.0K Dec 21 06:24 XRootD
```

[...]

PR #1591, opened 2022-01-14
ci: Add GitHub Actions based CI

### LINKED ISSUE #1589: Add PR facing CI with GitHub Actions
> Regarding the CI, we run our builds in: https://gitlab.cern.ch/dss/xrootd/-/pipelines
> 
> We currently don't have a build that handles PR, we used to use Travis for this purpose but this has been discontinued.

> Here's the job running python tests: https://github.com/xrootd/xrootd/blob/7c6dfb6b74ea30121e92e257c909dd6b9eed9bf6/.gitlab-ci.yml#L564-L570
> 
> Here are instructions on how to run a build on CC7:
https://github.com/xrootd/xrootd/blob/7c6dfb6b74ea30121e92e257c909dd6b9eed9bf6/.gitlab-ci.yml#L74-L79

_Originally posted by @simonmichal in https://github.com/xrootd/xrootd/issues/1585#issuecomment-1011171419_

As there is no PR facing CI this makes it hard to check that everything is passing tests before asking a maintainer for a review. Would there be interest in replacing the now defunct Travis CI with a GitHub Actions based CI that would run in the same base images that the GitLab CI does? If so, I have this already working on my fork and can contribute a PR.

matthewfeickert:
Resolves #1589 

Use GitHub Actions to add a CI workflow that runs on:

* push events to master or on tags
* pull request events targeting master
* a weekly CRON job that runs every Sunday at 01:23 UTC
* on GitHub releases
* on demand against any branch through workflow dispatch

Using concurrency groups if an event triggers another run of the workflow while a previous run in ongoing, the older run will be automatically canceled to avoid wasting resources.

The jobs target CentOS based workflow to start off with:

* Building with CMake like a user would and then running some libXrdCl tests. This build is a shorter build.
* Building with CMake like a user would but with Python 3 pip, setuptools, and wheel updated to the latest releases to check that users running with modern releases won't hit breaking changes.
* Building in full RPM.
* Building a Python sdist with the publishing workflow scripts.
* Building a Python sdist with the publishing workflow scripts, but with Python 3 pip, setuptools, and wheel updated to the latest releases to check that users running with modern releases won't hit breaking changes.

At the moment only CentOS based jobs are added as Debian based jobs will fail for installation of the Python bindings given the Issues related to Issue #1579 and PR #1585 and #1586. Debian based jobs can be added along with fixes to these.


.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+    types: [published]
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
```
matthewfeickert, line 24: This concurrency group will automatically cancel previous runs if a run on the _same branch_ is triggered.

.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+
+jobs:
```
matthewfeickert, line 26: GitHub Actions is a bit verbose, so there isn't much of a way to simplify this at the moment.

.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+            which \
+            python3-devel \
+            python3-setuptools \
+            git \
+            cppunit-devel
+        yum clean all
+
+    # Need to use v1 of action as image Git is too old
+    - name: Clone repository now that Git is available
+      uses: actions/checkout@v1
```
matthewfeickert, line 56: The faster/nicer [`v2` of this action](https://github.com/actions/checkout/tree/v2.4.0) requires Git 2.18, and CentOS 7 ships `v1.8.3.1`, so fall back to the old version here which still works.

.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+
+    - name: Run libXrdCl tests
+      run: |
+        cd ../build/tests
+        ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/UtilsTest/'
+        ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/SocketTest/'
+        ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/PollerTest/'
+        # ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/PostMasterTest/'
+        # ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/FileSystemTest/'
+        # ./common/test-runner ./XrdClTests/libXrdClTests.so 'All Tests/LocalFileHandlerTest/'
```
matthewfeickert, line 97: These tests failed, so I left them in here as a TODO of sorts.

.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+            systemd-devel \
+            zlib-devel \
+            devtoolset-7-gcc-c++ \
+            which \
+            python3-devel \
+            python3-setuptools \
+            git \
+            cppunit-devel
+        yum clean all
+        python3 -m pip --no-cache-dir install --upgrade pip setuptools wheel
```
matthewfeickert, line 124: I've added these `updated-python` jobs as it is important to check for changes that have happened in the `pip` and `setuptools` world that happen since CentOS 7 ships very old releases. Future Debian jobs will do a better job of checking this, as they have newer CPython releases while CentOS's Python 3.6 is capped out at [`setuptools` `59.6.0`](https://pypi.org/project/setuptools/59.6.0/) as [`setuptools` `59.7.0`+ is Python 3.7+](https://pypi.org/project/setuptools/59.7.0/).

matthewfeickert:
@simonmichal This is ready for review. To see this run in full please take a look at https://github.com/matthewfeickert/xrootd/runs/4819530943?check_suite_focus=true on my fork 

run_all

where I've just forced things to run

```diff
$ git diff origin/ci/add-build-ci-workflow 86d8ef3
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
index 2a93fd474..2959276fe 100644
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,10 @@ name: build
 
 on:
   push:
-    branches:
-    - master
-    tags:
-    - v*
+    # branches:
+    # - master
+    # tags:
+    # - v*
   pull_request:
     branches:
     - master
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,321 @@
...
+
+jobs:
+
+  cmake-centos7:
+
+    runs-on: ubuntu-latest
+    container: gitlab-registry.cern.ch/linuxsupport/cc7-base:latest
+
+    steps:
+    - name: Install external dependencies with yum
```
matthewfeickert, line 34: I'll note that these jobs fail sometimes infrequently because a mirror fails to be found in time. If these get restarted they pass.

Issue #1556, opened 2021-11-15
Prime value used for Diffie-Hellman handshake is too small


alrossi:
Because of security fixes made to it, we (dCache) have decided to move to a more recent version of the Java Bouncy Castle security library (1.67).  We are not back-porting this update (for the moment), but have only applied it to our master branch.

With it, we have experienced the following issue.

When our TPC client attempts to communicate with an xrootd server source endpoint, we get this error:

```
Remote endpoint sent: P = 250300537505961931441816061639185717883, G = 5, L = 0,
15 Nov 2021 09:51:01 (dcatest04-5) [] Cryptographic issues encountered during cert step: unsafe p value so small specific l required
15 Nov 2021 09:51:01 (dcatest04-5) [] Unable to complete gsi authentication to fndcatemp1.fnal.gov:1095, channel ef39d639, stream 1, session 02000000A67500001900000002000000: org.dcache.xrootd.core.XrootdException: Could not complete cert step: an error occurred during cryptographic operations..
```

The Bouncy Castle code has become less tolerant as to the prime number (P) value used for the DH exchange.

Recognizing a potential backward compatibility issue, however, the developers also added the possibility of overriding this check.  If we set the Java security property:

```
org.bouncycastle.dh.allow_unsafe_p_value=true
```

on the pools, then we no longer get the error and the transfer proceeds normally.

We can of course live with the override of the safety check for the moment, but this should be understood as a workaround and not a permanent fix.  The better thing to do is to have xrootd emit a longer/safer P value on the GSI handshake.  dCache, for instance, has used (for quite a while now) a 512-bit OpenSSL generated 'safe' number.  Here is the comparable log statement from our TPC client when a dCache door is the source:

```
Remote endpoint sent: P = 8810252053191919589359420599637641654529526533255167910139019997260323899182669360368264455955601543214780189083983437542290860708561316883421676261825939, G = 2, L = 0
```

Thanks,  Al

simonmichal:
It should be fixed in dd54932bfca32acab5c34d269bce81ff38f8aaa5, which is available in 5.4.0-rc3, could you give it a try?

alrossi:
I will update the client and try it out later this morning ... thank you!  -Al

________________________________________________
Albert L. Rossi
Senior Software Developer
Scientific Computing Division, Scientific Data Services, Distributed Data Development
FCC 229A
Mail Station 369 (FCC 2W)
Fermi National Accelerator Laboratory
Batavia, IL 60510
[...]

PR #1595, opened 2022-01-19
[XrdCrypto] Generate DH parameters on first call to XrdCryptosslCipher

jthiltges:
This should address the increased load issue discussed in #1556. I've _very lightly_ tested this change in production, and it seems OK. I'm planning more thorough tests tomorrow.

We'll also need to update `openssl3/XrdCryptosslCipher.cc` to match, but it should be fairly straightforward.

### DIFF
--- a/src/XrdCrypto/XrdCryptosslCipher.cc
+++ b/src/XrdCrypto/XrdCryptosslCipher.cc
@@ -32,6 +32,7 @@
 /*                                                                            */
 /* ************************************************************************** */
 #include <cstring>
+#include <assert.h>
 
 #include "XrdSut/XrdSutRndm.hh"
 #include "XrdCrypto/XrdCryptosslTrace.hh"
@@ -459,7 +460,7 @@ XrdCryptosslCipher::XrdCryptosslCipher(bool padded, int bits, char *pub,
    // Constructor for key agreement.
    // If pub is not defined, generates a DH full key,
    // the public part and parameters can be retrieved using Public().
-   // The number of random bits to be used in 'bits'.
+   // 'bits' is ignored (DH key is generated once)
    // If pub is defined with the public part and parameters of the
    // counterpart fully initialize a cipher with that information.
    // Sets also the name to 't', if different from the default one.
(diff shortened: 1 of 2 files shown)


abh3:
Looks good, thanks JT. Let me know when it passes your tests and I will merge it.

ellert:
Please note that this PR updates a file that exists in both an old pre openssl 3 version

src/XrdCrypto/XrdCryptosslCipher.cc

and a new openssl 3 compatible version

src/XrdCrypto/openssl3/XrdCryptosslCipher.cc

Until the new code (that also works on older openssl versions) replaces the old code, all changes must be done to both versions to keep them in sync.

abh3:
Yes, we are aware of this. We are not merging this until a) final tests have been performed, and b) the openssl3 update is included in the pr. This is just a preliminary pr.

jthiltges:
At Nebraska, we've been running with this patch in production on our local redirector for about 27 hours now. Redirector CPU use looks OK, CMS ETF and HammerCloud look OK, and we haven't received reports of issues.

Since things look promising, I'll look at getting the equivalent openssl3 changes into this PR. Though at present, we don't have a suitable environment for testing.

abh3:
Sounds good. I think we can manage verifying the SSL3 part with our Ubuntu collaborators.

jthiltges:
This adds openssl3 code that should be equivalent to the pre-openssl3 code. I'd greatly appreciate feedback from @ellert.

It also adds a bit of cleanup. Time permitting, I'll put it into production at Nebraska tomorrow.

abh3:
@ellert Could you check if the openssl3 changes are not problematic? We really need to push this out ASAP.

simonmichal:
@abh3 : if the PR is ready to be merged and @ellert don't have time to have a look at it, then please go ahead with the merge and once done I will port the changes to openssl3

xrootd-dev:
You don't have to port anything the PR already contains the port for 
openssl3. We just wanted to make sure it was OK as we can't even compile 
it. So, I will merge it and we go from there.

PR #1605, opened 2022-01-28
fix: Sanitize Python version to be PEP 440 compliant

matthewfeickert:
To avoid errors in modern versions of `pip`, ensure that the version name created for the Python bindings is sanitized to be compliant with [PEP 440](https://www.python.org/dev/peps/pep-0440/)

c.f.:
* https://www.python.org/dev/peps/pep-0440/
* https://github.com/pypa/pip/issues/8368

This essentially means that the version number needs to pass `pip._vendor.packaging.version.Version()`

Example:

```python
from pip._vendor.packaging.version import Version

Version("v20220128-34c8a39")  # pip._vendor.packaging.version.InvalidVersion: Invalid version: 'v20220128-34c8a39'
Version("2022.128")  # <Version('2022.128')>
Version("2022.128+a28a91c")  # <Version('2022.128+a28a91c')>
Version("6.0.0")  # <Version('6.0.0')>

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -336,7 +336,8 @@ jobs:
         . /opt/rh/devtoolset-7/enable
         cp packaging/wheel/* .
         ./publish.sh
-        python3 -m pip --verbose install dist/xrootd-*.tar.gz
+        cd ..  # Move xrootd.egg-info off PYTHONPATH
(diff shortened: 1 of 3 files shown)


matthewfeickert:
@simonmichal this is ready for review. I'm happy to answer any questions on it.

Issue #1614, opened 2022-02-10
XrdMacaroon segfaults on client auth (RHEL8 host)


mdoidge:
OS: Rocky Linux release 8.5 (Green Obsidian) 4.18.0-348.12.2.el8_5.x86_64
Versions: libmacaroons-0.3.0-6.el8.x86_64
                xrootd-server-5.4.0-2.el8.x86_64

Hello,
Whilst attempting to setup an xrootd server with http tpc support at our site we ran into an issue where the xroot service would segfault every time an http request was made against it[1]. We ran the same config on a CentOS7 test box and had no issues.

Colleagues in the UK tracked the issue to this segment of the code:
https://github.com/xrootd/xrootd/blob/master/src/XrdMacaroons/XrdMacaroonsHandler.cc#L571-L572

It's been the long time since I've done much C/C++ work and I'm mainly paraphrasing here, but AIUI the action of preallocating a vector with space from reserve then pointing to the start of a reserve to create a buffer does not work under the stricter checking of found in RHEL8 (and by extension RHEL8 clones, and probably RHEL9).

Suggested by the same learned colleagues in the UK, a fix could look like:
- std::vector<char> macaroon_resp; macaroon_resp.reserve(size_hint);
- if (macaroon_serialize(mac_with_date, &macaroon_resp[0], size_hint, &mac_err))
+ std::unique_ptr<char[]> macaroon_resp(new char[size_hint]);
+ if (macaroon_serialize(mac_with_date, macaroon_resp.get(), size_hint, &mac_err))
(And the same for the later references to &macaroon_resp[0] on line 584).

This issue has makes running xrootd RHEL8 clones impossible until fixed - and would likely extend to RHEL9. As many sites in the UK are planning to migrate to a filesystem fronted by an xrootd+http server this is of great interest, especially as many are installing on a RHEL8 clone as default practice.

Many thanks,
Matt
Lancaster Site Admin

[1] Stacktrace example for a crashing thread(logs just abruptly stop):                                                                
                                                                #0  0x00007f42f000737f raise (libc.so.6)
                                                                #1  0x00007f42efff1db5 abort (libc.so.6)
                                                                #2  0x00007f42e54283c8 _ZSt20__replacement_assertPKciS0_S0_ (libXrdMacaroons-5.so)
                                                                #3  0x00007f42e5426079 _ZNSt6vectorIcSaIcEEixEm (libXrdMacaroons-5.so)
[...]

PR #1630, opened 2022-02-28
Modify vector's size instead of capacity to avoid bounds-checking failures

simpsonst:
There are three commits, each to a distinct file.  All involve using `vector::resize` instead of `reserve`, and `vector::size` instead of `capacity`, and the unmodified code produces a bounds-checking assertion failure in `vector::operator[]` when compiled with `-D_GLIBCXX_ASSERTIONS`.  A [smoke test](https://github.com/paulmillar/http-tpc-utils/blob/master/bin/smoke-test.sh) was applied to the first two commits to verify that they eliminate a problem.
The first (in `XrdMacaroonsHandler.cc`) specifically fixes xrootd/xrootd#1614.  Although already closed, I think this change would be more readable.
The second (in `XrdTpcStream.hh`) is in code that was not engaged/discovered during the smoke test until the first was fixed.
The third (in `XrdThrottleManager.cc`) was found just by a scan of the code, and is not engaged by our configuration.  However, adding `xrootd.fslib throttle default` does engage it, and causes the assertion failure as soon as the process starts up without the fix.

### DIFF
--- a/src/XrdMacaroons/XrdMacaroonsHandler.cc
+++ b/src/XrdMacaroons/XrdMacaroonsHandler.cc
@@ -568,8 +568,7 @@ Handler::GenerateMacaroonResponse(XrdHttpExtReq &req, const std::string &resourc
 
     size_t size_hint = macaroon_serialize_size_hint(mac_with_date);
 
-    std::vector<char> macaroon_resp; macaroon_resp.reserve(size_hint);
-    macaroon_resp.push_back((char)0);
+    std::vector<char> macaroon_resp; macaroon_resp.resize(size_hint);
     if (macaroon_serialize(mac_with_date, &macaroon_resp[0], size_hint, &mac_err))
     {
         printf("Returned macaroon_serialize code: %lu\n", (unsigned long)size_hint);
--- a/src/XrdThrottle/XrdThrottleManager.cc
+++ b/src/XrdThrottle/XrdThrottleManager.cc
@@ -48,10 +48,10 @@ XrdThrottleManager::Init()
 {
    TRACE(DEBUG, "Initializing the throttle manager.");
    // Initialize all our shares to zero.
-   m_primary_bytes_shares.reserve(m_max_users);
-   m_secondary_bytes_shares.reserve(m_max_users);
-   m_primary_ops_shares.reserve(m_max_users);
-   m_secondary_ops_shares.reserve(m_max_users);
+   m_primary_bytes_shares.resize(m_max_users);
+   m_secondary_bytes_shares.resize(m_max_users);
+   m_primary_ops_shares.resize(m_max_users);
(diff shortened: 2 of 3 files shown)


abh3 (changes requested):
It would appear that this pull request is based on a previous version. Specifically, line 572 was added to XrdMacaroonsHandler.cc and should have appeared as

macaroon_resp.push_back((char)0);

and does not appear in this pull request. Could you rebase your branch to the latest master and resubmit the pull request?

simonmichal:
@simpsonst : thanks a lot for your PR! My understanding is that those 3 commits are yours:

[Set vector's logical size to avoid bounds-checking failure.](https://github.com/xrootd/xrootd/pull/1630/commits/989e088ca3375d95b035abd0130e78689d4d28c9)
[Vector must be resized to prevent bounds-checking assertion failure.](https://github.com/xrootd/xrootd/pull/1630/commits/666d8acaf162a49a1851cab220a5c5b4740dc776)
[Set vector's logical size to avoid bounds-checking failure.](https://github.com/xrootd/xrootd/pull/1630/commits/989e088ca3375d95b035abd0130e78689d4d28c9)

Now, we need to rework the PR so it contains only those 3 commits, unfortunately in the current form we cannot merge it because it pollutes the history, which in turn makes it very difficult to cut bugfix releases. What I would suggest is that you either:
- remove the redundant commits from your branch and then rebase on master, or
- create a new branch from master, `cherry-pick` the 3 commits and then replace the PR branch

Let me know if you need any help!

simpsonst:
Thanks @simonmichal, I figured I might have butchered it.
(The third commit is bac553b75df1427e119699345c9a766f5792ddb3; the third you listed I think is the same as the first.)
Going for the second option, I've done:
```
git checkout master
git checkout -b fix-vector-reserve
git cherry-pick 989e088ca3375d95b035abd0130e78689d4d28c9 bac553b75df1427e119699345c9a766f5792ddb3 666d8acaf162a49a1851cab220a5c5b4740dc776
```
That's produced a conflict on `src/XrdMacaroons/XrdMacaroonsHandler.cc`, so I've resolved that, and continued with:
```
git cherry-pick --continue
```
...and used the default commit message. I presume the next step is to push my local `fix-vector-reserve` over the `fix-macaroons-reserve` on the fork.  From the new branch, I guess it's:
```
git push --force origin fix-macaroons-reserve
```
But please advise!

simonmichal:
@simpsonst : I think you first need to replace the local version of `fix-macaroons-reserve`:
- `git checkout fix-macaroons-reserve`
- `git reset --hard fix-vector-reserve`
- `git push -f`

simonmichal:
Perfect, thanks a lot! I think it can be merged now :-)

abh3 (approved):

PR #1640, opened 2022-03-10
Have the XrdHttp extraction logic match GSI.

bbockelm:
The GSI security protocol defaults to the "trymap" logic which, according to documentation, is "try to map the DN but if unsuccessful use the hash of the client’s DN as the user identifier (username)". With this change, the XrdHttpSecurity interface will follow the same logic.

Since some sites may have special setups which rely on the old mechanism, one can get the old behavior by setting the optional new `compatNameGeneration` configuration in the `http.gridmap` setting.  For example:

```
http.gridmap compatNameGeneration /etc/xrootd/grid-mapfile
```

Would restore the old behavior.

With this change, a user can effectively match the `nomap`, `trymap`, and `usemap` settings between the XRootD and HTTP protocols and the two protocols have the same default.  Without the change, only the `usemap` setting could be matched with HTTP.  Having the defaults the same greatly decreases the "surprise" factor of using both protocols; the prior HTTP default of guessing a name from the DN is undocumented.

VOMS-based mappings (and mapfile) behavior is unaffected.

@olifre - you have one of the more complex security configs that I could think of.  Would you be unaffected by this change?

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -90,6 +90,7 @@ char *XrdHttpProtocol::secretkey = 0;
 
 char *XrdHttpProtocol::gridmap = 0;
 bool XrdHttpProtocol::isRequiredGridmap = false;
+bool XrdHttpProtocol::compatNameGeneration = false;
 int XrdHttpProtocol::sslverifydepth = 9;
 BIO *XrdHttpProtocol::sslbio_err = 0;
 XrdHttpSecXtractor *XrdHttpProtocol::secxtractor = 0;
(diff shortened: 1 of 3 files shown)


olifre:
@bbockelm Indeed, we have been using hashes of client DNs until about two weeks ago, but limited ourselves to GSI only to simplify things. Given this change, for sure the "surprise" factor which also affected me initially ("why doesn't this work with HTTP in the same way?") would be cured :+1: . 

That being said, we could finally simplify our configuration a bit, since the VO for which we needed the more complex configuration is now (finally) migrating away from manual fiddling towards Rucio and FTS. So we would be unaffected now, and would even have been unaffected before (since in the end, we only used this for GSI to simplify things). But this change is very welcome, since I won't exclude other VOs (or projects) requiring a more complex configuration especially in initial testing phases will pop up :+1:

PR #1644, opened 2022-03-12
Populate XrdSciTokens with more detailed log messages.

bbockelm:
This introduces configuration-file-controlled tracing of SciTokens-related messages and a number of detailed log messages.

As typical, this is controlled by a new config variable:
```
scitokens.trace [level] [level] ...
```

Where level can be one of `all`, `none`, `debug`, `info`, `warning`, or
`error`.

Roughly, the levels imply:
- Error: fatal issue in plugin configuration
- Warning: Potentially-problematic configuration or token that is invalid
- Info: Informational messages about the plugin.  Shouldn't have any logging per-request
- Debug: Verbose, per-request details about the tokens.

The default is error + warning; debug is not recommended for continuous production settings but is immensely helpful in solving problems.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -1,7 +1,7 @@
 
 #include "XrdAcc/XrdAccAuthorize.hh"
 #include "XrdOuc/XrdOucEnv.hh"
-#include "XrdOuc/XrdOucStream.hh"
+#include "XrdOuc/XrdOucGatherConf.hh"
 #include "XrdSec/XrdSecEntity.hh"
 #include "XrdSec/XrdSecEntityAttr.hh"
 #include "XrdSys/XrdSysLogger.hh"
@@ -39,6 +39,30 @@ enum LogMask {
     Error = 0x08,
     All = 0xff
 };
+
+std::string LogMaskToString(int mask) {
+    if (mask == LogMask::All) {return "all";}
+
+    bool has_entry = false;
+    std::stringstream ss;
+    if (mask & LogMask::Debug) {
+        ss << "debug";
+        has_entry = true;
+    }
+    if (mask & LogMask::Info) {
+        ss << (has_entry ? ", " : "") << "info";
+        has_entry = true;
+    }
+    if (mask & LogMask::Warning) {
+        ss << (has_entry ? ", " : "") << "warning";
+        has_entry = true;
+    }
+    if (mask & LogMask::Error) {
+        ss << (has_entry ? ", " : "") << "error";
+        has_entry = true;
+    }
+    return ss.str();
+}
 
 typedef std::vector<std::pair<Access_Operation, std::string>> AccessRulesRaw;
 
@@ -811,26 +835,22 @@ private:
         if (!XrdOucEnv::Import("XRDCONFIGFN", config_filename)) {
             return false;
         }
-        XrdOucStream stream(&m_log, getenv("XRDINSTANCE"));
-
-        int cfg_fd;
-        if ((cfg_fd = open(config_filename, O_RDONLY, 0)) < 0) {
-            m_log.Emsg("Config", errno, "open config file", config_filename);
+        XrdOucGatherConf scitokens_conf("scitokens.trace", &m_log);
+        int result;
+        if ((result = scitokens_conf.Gather(config_filename, XrdOucGatherConf::trim_lines)) < 0) {
+            m_log.Emsg("Config", -result, "parsing config file", config_filename);
             return false;
         }
-        stream.Attach(cfg_fd);
-        char *var;
+
+        char *val;
         std::string map_filename;
-        while ((var = stream.GetMyFirstWord())) {
-            if (strcmp(var, "scitokens.trace")) {
-                continue;
-            }
-            auto val = stream.GetWord();
-            if (!val || !val[0]) {
-                m_log.Emsg("Config", "SciTokens logging level not specified");
+        while (scitokens_conf.GetLine()) {
+            m_log.setMsgMask(0);
+            scitokens_conf.GetToken(); // Ignore the output; we asked for a single config value, trace
+            if (!(val = scitokens_conf.GetToken())) {
+                m_log.Emsg("Config", "scitokens.trace requires an argument.  Usage: scitokens.trace [all|error|warning|info|debug|none]");
                 return false;
             }
-            m_log.setMsgMask(0);
             do {
                 if (!strcmp(val, "all")) {m_log.setMsgMask(m_log.getMsgMask() | LogMask::All);}
                 else if (!strcmp(val, "error")) {m_log.setMsgMask(m_log.getMsgMask() | LogMask::Error);}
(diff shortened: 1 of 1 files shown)


src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -339,14 +412,20 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
...
                 } else {
+                    m_log.Log(LogMask::Warning, "Access", "Failed to generate ACLs for token");
                     return OnMissing(Entity, path, oper, env);
                 }
+                if (m_log.getMsgMask() & LogMask::Debug) {
+                    m_log.Log(LogMask::Debug, "Access", "New valid token", access_rules->str().c_str());
+                }
             } catch (std::exception &exc) {
-                m_log.Emsg("Access", "Error generating ACLs for authorization", exc.what());
+                m_log.Log(LogMask::Info, "Access", "Error generating ACLs for authorization", exc.what());
```
abh3, line 422: Why is this message being relegated to info? It seems hat one would want to know this has happened, yes?

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -716,34 +802,77 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         return true;
     }
 
+
+    bool Config() {
```
abh3, line 806: I think you should replace all of this with the new way of getting the plug-in config. Look at XrdOucGatherConf.hh which pretty much replaces all the manual handling of the config file. I'd prefer to start using this for all new code.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -1,6 +1,7 @@
 
 #include "XrdAcc/XrdAccAuthorize.hh"
 #include "XrdOuc/XrdOucEnv.hh"
+#include "XrdOuc/XrdOucStream.hh"
```
abh3, line 4: If you use XrdOucGatherConf.hh then there is no need to include this file. See detailed notes in this at "Config()".

abh3 (changes requested):
Looks OK. Two things: 1) possible mislabeling a message, and 2) I'd prefer you use the new way of compactly handling the config file so we don't have code splatter.

bbockelm:
(Just as a quick update - no disagreements with the PR feedback above.  I was just on vacation for a week and haven't had time to cycle back to this.)

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: Done in latest commits.

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: Good idea.  Done in the latest commits.

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: Sounds reasonable.  Fixed in the latest.

bbockelm:
Ok, I think the various requested changes are in.  Please re-review!

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -296,7 +391,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         pthread_rwlock_init(&m_config_lock, nullptr);
```
abh3, line 391: You should consider using XrdSysRWLock as this would be very helpful in debugging lock/unlock problems in the future.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -465,7 +574,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         pthread_rwlock_unlock(&m_config_lock);
```
abh3, line 574: XrdSysRWLock

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -602,7 +712,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         auto enf = enforcer_create(token_issuer.c_str(), &m_audiences_array[0], &err_msg);
         pthread_rwlock_unlock(&m_config_lock);
```
abh3, line 713: XrdSysRWLock

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -636,15 +746,15 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         auto iter = m_issuers.find(token_issuer);
         if (iter == m_issuers.end()) {
             pthread_rwlock_unlock(&m_config_lock);
```
abh3, line 748: XrdSysRWLock

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -636,15 +746,15 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
...
-            m_log.Emsg("GenerateAcls", "Authorized issuer without a config.");
+            m_log.Log(LogMask::Warning, "GenerateAcls", "Authorized issuer without a config.");
             scitoken_destroy(token);
             return false;
         }
         const auto &config = iter->second;
         value = nullptr;
         if (scitoken_get_claim_string(token, "sub", &value, &err_msg)) {
             pthread_rwlock_unlock(&m_config_lock);
```
abh3, line 756: XrdSysRWLock

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -942,7 +1093,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
         }
 
         if (issuers.empty()) {
-            m_log.Emsg("Reconfig", "No issuers configured.");
+            m_log.Log(LogMask::Warning, "Reconfig", "No issuers configured.");
         }
 
         pthread_rwlock_wrlock(&m_config_lock);
```
abh3, line 1099: XrdSysRWLock

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -87,6 +122,46 @@ XrdAccPrivs AddPriv(Access_Operation op, XrdAccPrivs privs)
     return static_cast<XrdAccPrivs>(new_privs);
 }
 
+const std::string OpToName(Access_Operation op) {
```
abh3, line 125: Likely this should be made a generic utility in XrdAcc as another implementation of the same thing exists there. Something to put in the todo list.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -29,6 +32,38 @@ XrdVERSIONINFO(XrdAccAuthorizeObjAdd, XrdAccSciTokens);
 
 namespace {
 
+enum LogMask {
```
abh3, line 35: This has gotten a life of its own and I see the same implementation in three  other modules (XrdMacaroonsHandler.hh, XrdTpcTPC.hh, and XrdVomsMapfile.hh) making this the fourth sych implementation. Shouldn't this be abstracted into a utility function?

abh3 (approved):
Nothing that would stop the merge but three suggestions to improve the code which I would appreciate you revisit and apply the suggested improvements.

PR #1647, opened 2022-03-14
python: add env var for install customizations

adriansev:
add XROOTDPY_INSTALL_OPTS env var for customization of install step of python bindings


adriansev:
@matthewfeickert have a much better https://github.com/xrootd/xrootd/pull/1648 so i will close this one

PR #1651, opened 2022-03-16
Force usage of c++17 when building bindings

ktf:


ktf:
For some  reason on my setup (macOS Monterey with a custom python) the provided changes are required to get the python binding building. Without it, "clang" is actually used, without any flags, and that obviously fails because of the C++11isms. I also noticed that the RPATH for the bindings are set wrongly, using the linux $ORIGIN variable which AFAICT does not exists on macOS. Adding the proper RPATH via install_name_tool fixes the issue.

adriansev:
ping @simonmichal while this was unfortunately submitted against stable-5.4.x could you take a look and maybe cherry-pick also in master (to be available in the future versions?). It seems that this is required for macos systems.. maybe this could be fixed deeper in the chain? (as this seems to fix only the symptom not the cause). 
Also, maybe some python expert(s) have a insight on this? :D @matthewfeickert :D

bindings/python/CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ if ( NOT ${VALID_PIP_EXIT_CODE} EQUAL 0 )
   install(
     CODE
     "EXECUTE_PROCESS(
-      COMMAND ${PYTHON_EXECUTABLE} ${SETUP_PY} install \
+    COMMAND /usr/bin/env CC=c++ CFLAGS=\"-std=c++17\" ${PYTHON_EXECUTABLE} ${SETUP_PY} install \
```
matthewfeickert, line 81: You definitely do not want to hardcode this anywhere. This should be set through CMake at build configure time with the `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`. c.f. https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html for more details.

matthewfeickert (changes requested):

bindings/python/CMakeLists.txt
matthewfeickert: Can probably just do something similar to [what ROOT does with `CMAKE_CXX_STANDARD`](https://github.com/root-project/root/blob/a17c8472ee6257a2937db46e465e43cdc6f71638/cmake/modules/CheckCompiler.cmake#L144-L169).

ktf: You misunderstood my point. The issue is the setup.py which decides to go for the CC / CFLAGS in order to compile a C++11 python module. Notice this affects also some ubuntu versions.

ktf: I can change it to use CMAKE_CXX_FLAGS, but I think something fishy is going on.

matthewfeickert: @ktf I would suggest that you make a GitHub Issue to discuss this and to clearly describe what the Issue is, what you want to have happen, what in the codebase needs to get changed, and what you plan to implement. 

Also, as @adriansev already pointed out, your PR is targeting the wrong branch. If you refactor this PR to provide a solution that has been agreed upon in the GitHub issue then you should have it target the `master` branch.

adriansev: @matthewfeickert  i did https://github.com/xrootd/xrootd/issues/1660 to discuss there the problem. thanks a lot!

simonmichal:
@ktf : I think wee can close this PR as we now have: https://github.com/xrootd/xrootd/commit/fa83acfdb5d7101bf8ac50b61a1018a366fa2d7a, could you please confirm?

PR #1654, opened 2022-03-16
Also grant Readdir when token grants read permission.

olifre:
Before this change, directory listing is impossible with a read claim,
leading to the confusing situation that trying to access files in a directory
may return 404 but listing the directory returns permission denied.


olifre:
Note: This irritating behaviour has been found by @federicaagostini in https://github.com/indigo-iam/wlcg-jwt-compliance-tests/issues/34#issuecomment-1069229524 . Thanks!

bbockelm:
Hi @olifre -

Just had a minute to look at this. I think adding directory listing permission is fine here.  Let's merge it!

Brian

PR #1658, opened 2022-03-21
chore: Use `command -v` over `which` for portability

matthewfeickert:
`command -v` is a POSIX safe way to find an executable program. While `which` is very common, it still requires installation on many Linux operating systems, and so `command -v` can offer a more portable alternative with a very low amount of additional cognitive overhead on the maintainers.

Additionally, use a subshell, `$()`, over backticks for command substitution as `$()` is recommended for multiple reasons including readability.

c.f.:
* https://stackoverflow.com/a/4708569/8931942
* [Why is $(...) preferred over `...` (backticks)?](http://mywiki.wooledge.org/BashFAQ/082)
* [POSIX spec for Command Substitution](https://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03)

I made the decision to leave `which` in 

https://github.com/xrootd/xrootd/blob/62d11709ca6247cec3c40596be7b9948f885f9e3/docker/builds/DockerfileCentos7#L6

and 

https://github.com/xrootd/xrootd/blob/62d11709ca6247cec3c40596be7b9948f885f9e3/docker/builds/DockerfileCentos8#L6

while no longer required, as I figured that these serve as debugging environments where having `which` handy might be nice for alternative reasons, just as how `vim` is installed in them as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,7 +44,6 @@ jobs:
             systemd-devel \
             zlib-devel \
             devtoolset-7-gcc-c++ \
-            which \
             python3-devel \
             python3-setuptools \
             git \
@@ -116,7 +115,6 @@ jobs:
             systemd-devel \
             zlib-devel \
             devtoolset-7-gcc-c++ \
-            which \
             python3-devel \
             python3-setuptools \
             git \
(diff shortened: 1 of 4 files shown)


matthewfeickert:
@simonmichal This is ready for review. I didn't open an Issue to discuss this as it seems small enough/low enough change burden to not be controversial, but I can appreciate that this could also be viewed as stylistic and so not acceptable. If so, my apologies, and feel free to close. :+1:

PR #1669, opened 2022-04-04
[XrdHttp] Use 405 for mkcol/mkdir EEXIST

### LINKED ISSUE #1663: Error 500 (server error) on mkdir over HTTPS when directory exists
Hi,

I've noticed with newer versions of xrootd (tested on v5.4.2) the return code for mkdir (mkcol) over HTTP(S) seems to be 500 if the directory already exists (quite possibly related to issue  #1360). This upsets gfal-mkdir as it seems to expect an error code of 405 in this case, an error code 500 is seen as an I/O error:

```
# Directory doesn't exist, no error...
$ gfal-mkdir https://test.host:1094/data/testdir
# Directory now exists, expect "File exists" error, but get I/O error instead:
$ gfal-mkdir https://test.host:1094/data/testdir
gfal-mkdir error: 5 (Input/output error) - HTTP 500 : Unexpected server error: 500  with url https://test.host:1094/data/testdir
```

The error I would expect is "File exists":
```
$ gfal-mkdir https://test.host:1094/data/testdir
gfal-mkdir error: 17 (File exists) - HTTP 405 : Method Not Allowed, File Exist  with url https://test.host:1094/data/testdir
```

I think RFC4918 specifies a return code of 405 in this case to indicate that the directory already exists (although I may well be reading the wrong RFC). I'm using a transfer tool that's built on-top of gfal and this difference in code is critical: An I/O error stops any further operations, whereas it knows to continue if it gets "file exists" on an mkdir command.

I was able to get the behaviour I expected with this patch, although I don't know if this has any other negative effects on the other http methods:
```
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -965,6 +965,9 @@ void XrdHttpReq::mapXrdErrorToHttpStatus() {
       case kXR_InvalidRequest:
         httpStatusCode = 405; httpStatusText = "Method is not allowed";
         break;
+      case kXR_ItExists:
+        httpStatusCode = 405; httpStatusText = "Resource exists";
+        break;
       default:
         break;
     }
```

Would it be possible for something like this patch be merged to fix this issue?

Regards,
Simon

sfayer:
Hi,

This is the patch to use return code 405 for mkcol if the directory already exists as discussed in #1663.

Resolves #1663

Regards,
Simon

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2692,8 +2692,12 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
     {
 
       if (xrdresp != kXR_ok) {
-        prot->SendSimpleResp(httpStatusCode, NULL, NULL,
-                             httpStatusText.c_str(), httpStatusText.length(), false);
+        if (xrderrcode == kXR_ItExists) {
+          prot->SendSimpleResp(405, NULL, NULL, (char *) "Method is not allowed; resource already exists.", 0, false);
+        } else {
+          prot->SendSimpleResp(httpStatusCode, NULL, NULL,
+                               httpStatusText.c_str(), httpStatusText.length(), false);
+        }
         return -1;
       }
 


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2692,8 +2692,12 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
     {
 
       if (xrdresp != kXR_ok) {
-        prot->SendSimpleResp(httpStatusCode, NULL, NULL,
-                             httpStatusText.c_str(), httpStatusText.length(), false);
+        if (xrderrcode == kXR_ItExists) {
+          prot->SendSimpleResp(405, NULL, NULL, (char *) "Method is not allowed", 0, false);
```
abh3, line 2696: Hmmm, can you also add "; resource already exists." to the text so that at least the person getting the message knows why it cannot be done?

abh3 (changes requested):

src/XrdHttp/XrdHttpReq.cc
sfayer: Sure, I've updated the return text.

Issue #1660, opened 2022-03-21
xrootd::python::macos wheel usage of clang does not set the correct cxx standard


adriansev:
So, when building the wheel pip does not the the correct cxx flags. for us (ALICE) a quick fix was to do https://github.com/xrootd/xrootd/pull/1651 but would be better to find the cause of pip not using the correct flags (the same used also for building the rest of the software).
The actual errors (building v5.4.2) can be seen [here](https://asevcenc.web.cern.ch/asevcenc/xrootd_py_macos_err/log)
For the moment (because we want 5.4.2) we build our fork that contains our workable fix, but would be great to find the solution upstream (for the next tag)

adriansev:
@ktf @matthewfeickert FYI, thanks guys!

matthewfeickert:
> So, when building the wheel pip does not the the correct cxx flags.

What CXX flags are you passing now? Not like in PR #1651, but what CXX flags are you passing in at CMake build configure time?

> for us (ALICE) a quick fix was to do #1651 but would be better to find the cause of pip not using the correct flags (the same used also for building the rest of the software). The actual errors (building v5.4.2) can be seen [here](https://asevcenc.web.cern.ch/asevcenc/xrootd_py_macos_err/log)

Can you also detail what the macOS software environment is that you're using? I don't have much experience building C++ software with macOS in CI, but it seems to be a bit of a more unstable mess around versions with things getting pulled in from both XCode and homebrew at times. As you're building this in CI can you link to the GitLab Pipelines file and section, or if private, share the relevant build config? In short, wow can someone reproduce the macOS build environment in CI?

adriansev:
Hi @matthewfeickert ! So the env vars used are these https://github.com/alisw/alidist/blob/master/defaults-o2.sh#L1
then in the xrootd recipe we have: https://github.com/alisw/alidist/blob/master/xrootd.sh#L53

On the macos information i cannot help (unless you give me some commands), these days was my first time logging in a macos account :)
We do have a CI but the general building is the same i described to you last time (with aliBuild ..etc..) and no matter the platform the building procedure is the same (and with the same recipes that take care of adaptation to underlying platform).
So, on the xcode/homebrew stuff maybe Giulio Eulisse can comment (weird that i cannot mention him)

matthewfeickert:
[...]

PR #1672, opened 2022-04-07
[FIX] Fix import syntax to enable shutil.which check

### LINKED ISSUE #1668: python :: pipy install on ubuntu 20.04 does not find the libs
Hi! @matthewfeickert sorry to bother you again :) any idea why this still does not work?
AFAIR this was solved, but i just hit this up on a ubuntu 20.04 .. the output is [here](https://asevcenc.web.cern.ch/asevcenc/xrootd_ubuntu/output.txt)

matthewfeickert:
Resolves #1668

In PR #1585 a syntax error was introduced  as `shutil.which` is a method of the Python 'shutil' module and not a module, so the proper import is `from shutil import which` as regardless of the CPython version `import shutil.which` will result in a

```pytb
ModuleNotFoundError: No module named 'shutil.which'; 'shutil' is not a package
```

This error was causing all runs of `packaging/wheel/install.sh` to use the legacy behavior of `python setup.py install` regardless if `shutil.which` was available or not.

As all CPython versions that have `shutil.which` (Python 3.3+) are able to run `packaging/wheel/setup.py` then they will be able to take advantage of having pip be able to create a valid wheel during install time. In the event that a very old version of pip is used, an egg distribution will be installed but it will still work fine so there is no real reason to rename it in a manner that mocks the dist-info of a wheel.

Also a GHA job to build on Ubuntu to add as an additional check.

This PR also avoids finding multiple directories that have 'python' in the name under `/usr/local/lib/` on the GitHub Actions runners, use the `maxdepth` option for `find` to limit the search to directories of the pattern `/usr/local/lib/python3.*`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -641,3 +641,57 @@ jobs:
         python3 -c 'import XRootD; print(XRootD)'
         python3 -c 'import pyxrootd; print(pyxrootd)'
         python3 -c 'from XRootD import client; print(client.FileSystem("root://someserver:1094"))'
+
+  sdist-ubuntu:
+
+    # Use of sudo as https://github.com/actions/virtual-environments requires it
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Install external dependencies with apt-get
+      run: |
+        sudo apt-get update -y
+        DEBIAN_FRONTEND=noninteractive sudo apt-get install -y \
+          g++ \
+          git \
+          cmake \
+          uuid-dev \
+          dpkg-dev \
+          libssl-dev \
+          libx11-dev \
+          python3 \
+          python3-pip \
+          python3-venv \
+          python3-dev \
+          pkg-config \
+          tree
+        sudo apt-get autoclean -y
+        python3 -m pip --no-cache-dir install --upgrade pip setuptools wheel
+        python3 -m pip list
+
+    - name: Clone repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Build sdist using publishing workflow
+      run: |
+        cp packaging/wheel/* .
+        ./publish.sh
+        python3 -m pip --verbose install ./dist/xrootd-*.tar.gz
+        python3 -m pip list
+
+    - name: Show site-pacakges layout for XRootD modules
+      run: |
+        find $(python3 -c 'import XRootD; import pathlib; print(str(pathlib.Path(XRootD.__path__[0]).parent))') \
+          -type d \
+          -iname "*xrootd*" | xargs tree
+
+    - name: Verify Python bindings
+      run: |
+        python3 --version --version
+        python3 -m pip list
+        python3 -m pip show xrootd
+        python3 -c 'import XRootD; print(XRootD)'
+        python3 -c 'import pyxrootd; print(pyxrootd)'
+        python3 -c 'from XRootD import client; print(client.FileSystem("root://someserver:1094"))'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -282,8 +282,8 @@ jobs:
     - name: Verify Python bindings
       run: |
         export LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}"
-        export PYTHON_VERSION_MINOR=$(_tmp=$(find /usr/local/lib/ -type d -iname "python*") && echo ${_tmp: -3}; unset _tmp)
-        export PYTHONPATH="/usr/local/lib/python${PYTHON_VERSION_MINOR}/site-packages:${PYTHONPATH}"
+        export PYTHON_VERSION_MINOR=$(python3 -c 'import sys; print(f"{sys.version_info.minor}")')
+        export PYTHONPATH="/usr/local/lib/python3.${PYTHON_VERSION_MINOR}/site-packages:${PYTHONPATH}"
         python3 -m pip list
         python3 -m pip show xrootd
         python3 -c 'import XRootD; print(XRootD)'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -305,7 +305,6 @@ jobs:
           krb5 \
           ossp-uuid \
           libxml2 \
-          git \
           openssl@3
 
     - name: Install necessary Python libraries
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -252,7 +252,7 @@ jobs:
         python3 -m pip --no-cache-dir install --upgrade pip setuptools wheel
 
     - name: Clone repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
@@ -313,7 +313,7 @@ jobs:
         python3 -m pip list
 
     - name: Clone repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
@@ -445,7 +445,7 @@ jobs:
         dnf clean all
 
     - name: Clone repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
@@ -504,7 +504,7 @@ jobs:
         sudo apt-get autoclean -y
 
     - name: Clone repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
--- a/packaging/wheel/install.sh
+++ b/packaging/wheel/install.sh
@@ -40,7 +40,7 @@ fi
 cd ../bindings/python
 
 # Determine if shutil.which is available for a modern Python package install
-${6} -c 'import shutil.which' &> /dev/null  # $6 holds the python sys.executable
+${6} -c 'from shutil import which' &> /dev/null  # $6 holds the python sys.executable
 shutil_which_available=$?
 if [ "${shutil_which_available}" -ne "0" ]; then
     ${6} setup.py install ${3}
--- a/packaging/wheel/install.sh
+++ b/packaging/wheel/install.sh
@@ -40,6 +40,8 @@ fi
 cd ../bindings/python
 
 # Determine if shutil.which is available for a modern Python package install
+# (shutil.which was added in Python 3.3, so any version of Python 3 now will have it)
+# TODO: Drop support for Python 3.3 and simplify to pip approach
 ${6} -c 'from shutil import which' &> /dev/null  # $6 holds the python sys.executable
 shutil_which_available=$?
 if [ "${shutil_which_available}" -ne "0" ]; then
@@ -57,28 +59,3 @@ fi
 
 cd $startdir
 rm -r xrootdbuild
-
-# TODO: Remove all of the below and build a wheel using PyPA tools
-
-# convert the egg-info into a proper dist-info
-egginfo_path=$(ls $1/xrootd-*.egg-info)
-core="${egginfo_path%.*}"
-core="${egginfo_path#$1/}"
-sufix="${core#xrootd-*.*.*_-}"
-core="${core%_-*}"
-if [[ "$core" == "$sufix" ]]
-then
-    distinfo_path="${egginfo_path%.*}.dist-info"
-else
-    distinfo_path="$1/$core-$sufix"
-fi
-echo $distinfo_path >> /tmp/out.txt
-mkdir $distinfo_path
-mv $egginfo_path $distinfo_path/METADATA
-echo -e "Wheel-Version: 1.0\nGenerator: bdist_wheel (0.35.1)\nRoot-Is-Purelib: true\nTag: py2-none-any\nTag: py3-none-any" > $distinfo_path/WHEEL
-touch $distinfo_path/RECORD
-distinfo_name=${distinfo_path#"$1"}
-find $1/pyxrootd/      -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/$distinfo_name -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/XRootD/        -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/pyxrootd/ -type l | awk '{print$1 ",,"}' >> $distinfo_path/RECORD


matthewfeickert:
@simonmichal @adriansev this is ready for review.

.github/workflows/build.yml
```diff
@@ -282,7 +282,7 @@ jobs:
     - name: Verify Python bindings
       run: |
         export LD_LIBRARY_PATH="/usr/local/lib:${LD_LIBRARY_PATH}"
-        export PYTHON_VERSION_MINOR=$(_tmp=$(find /usr/local/lib/ -type d -iname "python*") && echo ${_tmp: -3}; unset _tmp)
+        export PYTHON_VERSION_MINOR=$(_tmp=$(find /usr/local/lib/ -maxdepth 1 -type d -iname "python3.*") && echo ${_tmp: -3}; unset _tmp)
```
matthewfeickert, line 285: This fixes a depth of find bug on the runners that was exposed when the runner environments updated.

packaging/wheel/install.sh
```diff
@@ -57,28 +59,3 @@ fi
...
-echo $distinfo_path >> /tmp/out.txt
-mkdir $distinfo_path
-mv $egginfo_path $distinfo_path/METADATA
-echo -e "Wheel-Version: 1.0\nGenerator: bdist_wheel (0.35.1)\nRoot-Is-Purelib: true\nTag: py2-none-any\nTag: py3-none-any" > $distinfo_path/WHEEL
-touch $distinfo_path/RECORD
-distinfo_name=${distinfo_path#"$1"}
-find $1/pyxrootd/      -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/$distinfo_name -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/XRootD/        -type f -exec sha256sum {} \; | awk '{printf$2 ",sha256=" $1 "," ; system("stat --printf=\"%s\" "$2) ; print '\n'; }' >> $distinfo_path/RECORD
-find $1/pyxrootd/ -type l | awk '{print$1 ",,"}' >> $distinfo_path/RECORD
```
matthewfeickert, line 84: Unless there is some specific reason to _need_ to rename things to act like a wheel instead of an egg for Pythons that have old pips then this should get removed to simplify things and to avoid files not existing causing errors during the build.

If you have Python 3.3+ and a modern pip then `pip` can handle the build and install of a wheel from the sdist. Otherwise (that is you're on Python 3.2 (which is very very EOL so please don't) or using a pip that is quite old) you'll get a totally fine egg and can be used or uninstalled normally, so there's no clear motivation (to me) to go through attempting to renamed its `egg-info` dir to `dist-info`.

packaging/wheel/install.sh
```diff
@@ -40,7 +40,9 @@ fi
 cd ../bindings/python
 
 # Determine if shutil.which is available for a modern Python package install
-${6} -c 'import shutil.which' &> /dev/null  # $6 holds the python sys.executable
+# (shutil.which was added in Python 3.3, so any version of Python 3 now will have it)
+# TODO: Drop support for Python 3.3 and simplify to pip approach
+${6} -c 'from shutil import which' &> /dev/null  # $6 holds the python sys.executable
```
matthewfeickert, line 45: This is the relevant fix. 

## Wrong behavior
```pytb
>>> import shutil.which
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'shutil.which'; 'shutil' is not a package

```

## Correct behavior
```python
>>> from shutil import which
```

packaging/wheel/install.sh
```diff
@@ -40,7 +40,9 @@ fi
 cd ../bindings/python
 
 # Determine if shutil.which is available for a modern Python package install
-${6} -c 'import shutil.which' &> /dev/null  # $6 holds the python sys.executable
+# (shutil.which was added in Python 3.3, so any version of Python 3 now will have it)
+# TODO: Drop support for Python 3.3 and simplify to pip approach
```
matthewfeickert, line 44: In a follow up PR (to keep scope simple) if people are okay with dropping support for Python 3.1 and Python 3.2 ([EOL was October 11, 2014](https://peps.python.org/pep-0392/)) then this can all get simplified down to just using `pip`. As the wheels are assuming you're using Python 3 anyway this should not be a problem for anyone.

matthewfeickert:
@simonmichal @adriansev this is ready for review.

@adriansev if you can also check out this PR's branch and do your own testing for ALICE that would be great. :+1:

.github/workflows/build.yml
```diff
@@ -642,3 +642,56 @@ jobs:
...
+    - name: Clone repository
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Build sdist using publishing workflow
+      run: |
+        cp packaging/wheel/* .
+        ./publish.sh
+        python3 -m pip --verbose install ./dist/xrootd-*.tar.gz
```
matthewfeickert, line 681: I haven't checked on `--user` installs, but the follow up PR I mentioned will be able to handle this if this doesn't.

.github/workflows/build.yml
```diff
@@ -642,3 +642,57 @@ jobs:
...
```
matthewfeickert, line 682: @adriansev I just tested this locally in a `ubuntu:20.04` contaienr using a `--user` install and that works as well, as expected. :+1:

packaging/wheel/install.sh
adriansev: @matthewfeickert i believe that the lowest baseline is centos 7 that have python 3.6.8 and pip 9.0.3 (setuptools 39.2.0 wheel 0.31.1) so anything lower than this can go

matthewfeickert: Great! That's what I assumed. CentOS 7 is going to hold people at Python 3.6 for a long time, but I don't know of any real world cases where people are still using any Python 3.X version older than Python 3.6. :+1:

adriansev:
> @simonmichal @adriansev this is ready for review.
> 
> @adriansev if you can also check out this PR's branch and do your own testing for ALICE that would be great. +1

@matthewfeickert  it will be a little bit tricky as we need also the https://github.com/xrootd/xrootd/issues/1660  ... is there a way to apply 2 PRs to a branch?

matthewfeickert:
> is there a way to apply 2 PRs to a branch?

I would just cherry-pick anything that you need onto the branch that you get from my fork, as they don't have any overlap. You can also just look at what is in this PR that matters and manually make commits on your forks as they are very short.

adriansev:
@matthewfeickert i just tried on centos 7 and the output is nice and tidy :)
```
2022-04-18@23:29:35:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Processing ./bindings/python
2022-04-18@23:29:35:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Preparing metadata (setup.py): started
2022-04-18@23:29:36:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Preparing metadata (setup.py): finished with status 'done'
2022-04-18@23:29:36:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Building wheels for collected packages: xrootd
2022-04-18@23:29:36:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Building wheel for xrootd (setup.py): started
2022-04-18@23:29:50:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Building wheel for xrootd (setup.py): finished with status 'done'
2022-04-18@23:29:50:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Created wheel for xrootd: filename=xrootd-2022.407+2a11c8e-cp36-cp36m-linux_x86_64.whl size=1478778 sha256=79c7ad5226bb8cab345a13ec74f2377cde975c21d1b198c8b7a68711e7dc79da
2022-04-18@23:29:50:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Stored in directory: /software/physics-tools/alicebuild/xjalienfs_xrootd/tmp/pip-ephem-wheel-cache-4e02pmj6/wheels/06/6c/53/feed371522f56f0815276316902d0cb9aa48442567446f7c9d
2022-04-18@23:29:50:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Successfully built xrootd
2022-04-18@23:29:52:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Installing collected packages: xrootd
2022-04-18@23:29:52:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Attempting uninstall: xrootd
2022-04-18@23:29:52:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:     Found existing installation: xrootd 5.4.2
2022-04-18@23:29:52:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:     Uninstalling xrootd-5.4.2:
2022-04-18@23:29:55:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:       Successfully uninstalled xrootd-5.4.2
2022-04-18@23:29:55:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Successfully installed xrootd-2022.407+2a11c8e
```

i will get back also with ubuntu result

adriansev:
also ubuntu 20.04 looks similarly nice:
```
2022-04-19@00:06:00:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Processing ./bindings/python
2022-04-19@00:06:00:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Preparing metadata (setup.py): started
2022-04-19@00:06:01:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Preparing metadata (setup.py): finished with status 'done'
2022-04-19@00:06:01:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Building wheels for collected packages: xrootd
2022-04-19@00:06:01:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Building wheel for xrootd (setup.py): started
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Building wheel for xrootd (setup.py): finished with status 'done'
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Created wheel for xrootd: filename=xrootd-2022.407+2a11c8e2d-cp38-cp38-linux_x86_64.whl size=1470471 sha256=d0a959e5a6bf536891976c900388e5c02c313649034a5d667162e4a025482aea
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Stored in directory: /home/adrian/alisw/xjalienfs_xrootd/tmp/pip-ephem-wheel-cache-m8mfitbp/wheels/54/63/2d/c6b7b652204ef99c3377d9ac514211b1e7a086333565a86711
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Successfully built xrootd
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Installing collected packages: xrootd
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:   Attempting uninstall: xrootd
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:     Found existing installation: xrootd 5.4.2-
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:     Uninstalling xrootd-5.4.2-:
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd:       Successfully uninstalled xrootd-5.4.2-
2022-04-19@00:06:11:DEBUG:xjalienfs:XRootD:xjalienfs_xrootd: Successfully installed xrootd-2022.407+2a11c8e2d
```

.github/workflows/build.yml
adriansev: erm, for minor of python version why not `python3 -c 'import sys; print(sys.version_info.minor)'` ?

matthewfeickert: Yeah
```yaml
export PYTHON_VERSION_MINOR=$(python3 -c 'import sys; print(f"python{sys.version_info.major}.{sys.version_info.minor}")')
```

is more stable then using `find`

matthewfeickert: Applied in the rebase :+1: 

```diff
-        export PYTHON_VERSION_MINOR=$(_tmp=$(find /usr/local/lib/ -type d -iname "python*") && echo ${_tmp: -3}; unset _tmp)
-        export PYTHONPATH="/usr/local/lib/python${PYTHON_VERSION_MINOR}/site-packages:${PYTHONPATH}"
+        export PYTHON_VERSION_MINOR=$(python3 -c 'import sys; print(f"{sys.version_info.minor}")')
+        export PYTHONPATH="/usr/local/lib/python3.${PYTHON_VERSION_MINOR}/site-packages:${PYTHONPATH}"
```

matthewfeickert:
I went ahead and also removed installing `git` from the macOS job as there was conflicts between the brew install and the GitHub Action runner environment that is shipped.

.github/workflows/build.yml
```diff
@@ -305,7 +305,6 @@ jobs:
           krb5 \
           ossp-uuid \
           libxml2 \
-          git \
```
matthewfeickert, line 308: Needs to be removed as the brew install conflicts with what the GitHub Action runner macOS environment already has installed and the job fails.

matthewfeickert:
> however if instead of `python3 -m pip install` I do `pip3 install` it does work like a charm.
> 
> Any idea why? (pardon my ignorance if this is a trivial question or expected behaviour)

No worries @simonmichal. :+1: The problem there is that

```
python3 -m pip install dist/xrootd-6.0.0.tar.gz 
```

also puts your cwd on `sys.path` and as as things _currently_ stand there is a `xrootd.egg-info` there, so `pip` "sees" the `xrootd.egg-info` and says "I already have a distribution of `xrootd` `v6.0.0` on `PYTHONPATH` so there is nothing to install. So there are two alternatives:

1. Easiest: Just use `--upgrade` even though you don't have anything installed yet
```
python3 -m pip install --upgrade ./dist/xrootd-6.0.0.tar.gz
```

2. Change directories so `xrootd.egg-info` is off PATH. This is how things are done in CI:

https://github.com/xrootd/xrootd/blob/9db73ec71becb468045d73862048bf91726264e8/.github/workflows/build.yml#L634-L635

though I should maybe update that in the future to approach 1 which is easier to understand I think.

Directly calling `pip3` doesn't have this behavior as you're not invoking it as a Python module, but then you're not guaranteed to have the right Python do the installation. Here though of course there is only one Python so there is no issue.

You also don't show enabling `devtoolset-7` as you're on CentOS 7

https://github.com/xrootd/xrootd/blob/9db73ec71becb468045d73862048bf91726264e8/.github/workflows/build.yml#L631

but seems you didn't need it? 

This (avoiding PATH issues) is one of the reasons that `src` distribution layouts for Python packages exist, but that's a discussion for another Issue.

simonmichal:
Thanks for the explanation :-) I think we are ready to merge, aren't we? :-)

> You also don't show enabling devtoolset-7 as you're on CentOS 7
> but seems you didn't need it?

actually the `install.sh` script enables devtoolset automatically:
https://github.com/xrootd/xrootd/blob/9db73ec71becb468045d73862048bf91726264e8/packaging/wheel/install.sh#L15-L17

matthewfeickert:
The build errors for `rpm-centos7` and `dpkg-ubuntu` were introduced in dd6fd8a7689491030725e60601376196b110bddc and/or 9db73ec71becb468045d73862048bf91726264e8 that has exit codes of

```
install: cannot stat 'src/XrdClHttp/config/http.client.conf.example': No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.JpH7CG (%install)
    Bad exit status from /var/tmp/rpm-tmp.JpH7CG (%install)
```

simonmichal:
> The build errors for rpm-centos7 and dpkg-ubuntu were introduced in https://github.com/xrootd/xrootd/commit/dd6fd8a7689491030725e60601376196b110bddc and/or https://github.com/xrootd/xrootd/commit/9db73ec71becb468045d73862048bf91726264e8 that has exit codes of

yes, that's absolutely right, I'll fix this today ;-)

matthewfeickert:
Thanks! :) `master` is looking good now with CI passing.

PR #1685, opened 2022-04-22
Check if we need libatomic to use atomic operations in the C++ code

ellert:
This is required for using 64 bit atomics on some 32 bit architectures.


matthewfeickert:
The failing CI jobs are fixed in PR #1672, so those can be ignored here.

PR #1688, opened 2022-04-22
[XrdCrypto] Modify XrdCryptosslNameOneLine to allow commas in DNs

jthiltges:
Possible solution to xrootd#1681. I've not yet tested it, so submitting as a draft. If it seems like a valid option, I'm happy to get it built and tested. @bbockelm

### DIFF
--- a/src/XrdCrypto/XrdCryptosslAux.cc
+++ b/src/XrdCrypto/XrdCryptosslAux.cc
@@ -762,13 +762,13 @@ void XrdCryptosslNameOneLine(X509_NAME *nm, XrdOucString &s)
 
 #ifndef USEX509NAMEONELINE
    BIO *mbio = BIO_new(BIO_s_mem());
-   X509_NAME_print_ex(mbio, nm, 0, XN_FLAG_COMPAT);
+   X509_NAME_print_ex(mbio, nm, 0, XN_FLAG_SEP_MULTILINE);
    char *data = 0;
    long len = BIO_get_mem_data(mbio, &data);
    s = "/";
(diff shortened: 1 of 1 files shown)


bbockelm:
Ah, this is a clever approach!

Definitely much simpler than "doing it right" and properly escaping meta characters.  I think there's a quite low likelihood that there's a DN we want to accept with a `\n` embedded.

bbockelm:
@jthiltges - were you able to test this one in the end?

jthiltges:
Looked fine in testing.

The DN string came in with newlines as expected:
```
(gdb) print s
$1 = (XrdOucString &) @0x7fc45c023688: {_vptr.XrdOucString = 0x614c10 <vtable for XrdOucString+16>, str = 0x7fc45c01c680 "/DC=org\nDC=incommon\nC=US\nST=California\nL=La Jolla\nO=University of California, San Diego\nCN=stashcache.t2.ucsd.edu\nCN=563638112", len = 126, siz = 127, static blksize = -1}
```

Then replaced with slashes:
```
(gdb) print s
$5 = (XrdOucString &) @0x7fc45c023688: {_vptr.XrdOucString = 0x614c10 <vtable for XrdOucString+16>, str = 0x7fc45c01c680 "/DC=org/DC=incommon/C=US/ST=California/L=La Jolla/O=University of California, San Diego/CN=stashcache.t2.ucsd.edu/CN=563638112", len = 126, siz = 127, static blksize = -1}```

bbockelm:
@abh3 - I think this one is ready to go.  John mentioned he's tested this out successfully in production.

PR #1691, opened 2022-04-26
[CI] Simplify sdist install jobs and logging

matthewfeickert:
Following up @simonmichal's https://github.com/xrootd/xrootd/pull/1672#issuecomment-1109472205 this removes unnecessary sourcing of `devtoolset-7` in the sdist build jobs. With `devtoolset-7` no longer providing a compiler to the environment at detection time in `setup.py` also add `gcc-c++` to the list of packages to install with `yum`.

Also simplify installs (c.f. https://github.com/xrootd/xrootd/pull/1672#issuecomment-1108914342) by installing with

```
python3 -m pip --verbose install --upgrade ./dist/xrootd-*.tar.gz
```

(Note the use of `--upgrade` is only necessary on tags where there is properly formatted SemVer version info in the `xrootd.egg-info`. A dev version like '20220426.post7797547' doesn't need the --upgrade, but it is easier to just always use it.)

For better logging and debugging info in CI:
* Show python version used in CI with `python --version --version`.
* Post installation for sdist install builds, show the site-packages layout for the installed XRootD modules. This introduces a dependency on `tree` so also ensure that is installed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -422,7 +422,6 @@ jobs:
 
     - name: Build sdist using publishing workflow
       run: |
-        . /opt/rh/devtoolset-7/enable
         cp packaging/wheel/* .
         ./publish.sh
         ls -lhtra dist/
@@ -559,6 +558,7 @@ jobs:
         yum update -y
         yum install --nogpg -y \
             cmake3 \
+            gcc-c++ \
             make \
             krb5-devel \
             libuuid-devel \
@@ -580,7 +580,6 @@ jobs:
 
     - name: Build sdist using publishing workflow
       run: |
-        . /opt/rh/devtoolset-7/enable
         cp packaging/wheel/* .
         ./publish.sh
         cd ..  # Move xrootd.egg-info off PYTHONPATH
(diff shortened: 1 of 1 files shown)


matthewfeickert:
@simonmichal this is ready for review.

matthewfeickert:
@simonmichal While e6c10b79848d88597668a919e9cdc2d6de7ad372 has the CentOS based build failing, this PR was passing all of the CI before that commit was introduced, and so is still ready for review.

Once this PR is in I can add a PR that I have staged that would fully resolve Issue #1579 by moving to moving off `distutils` to modern `setuptools` for Python 3.7+ builds (so CentOS 7 being stuck on Python 3.6 would still use `distutils` but everything modern would advance).

adriansev:
@matthewfeickert Could you please into account the situation where on centos 7, a modern python is built and used? or this is already detected? Thanks a lot!

matthewfeickert:
> Could you please into account the situation where on centos 7, a modern python is built and used? or this is already detected? Thanks a lot!

OS info is not in the logic for this future PR, so there is nothing to worry about RE: CentOS 7. But let's avoid talking too much about other future PRs on this one so that we don't get sidetracked. :wink: If you'd like I can open up a new version of Issue #1579 for discussion on what I'm planning.

PR #1704, opened 2022-05-10
Add a "python3-xrootd" provides to the Python 3 bindings in EL7 builds too

matyasselmeci:
This is already being done in EL8 (%python3only case) but not in EL7 builds.


matyasselmeci:
Yes, that should do the trick as well. Is the spec file change included in 5.4.3rc4?

simonmichal:
This has been scheduled for `5.5.0`, we really want to finalise `5.4.3` and don't introduce more changes so the release is not being delayed any more. We will start working towards `5.5.0` shortly after `5.4.3` is released with should happen early next week. I hope this works for you!

matyasselmeci:
Sure, that's fine. I will close this, then.

PR #1724, opened 2022-06-20
[XrdPosix] Fix a case where delayed destroy would never re-try to close the file

smithdh:
While testing XrdCl under network problems and heavy load it was seen that sometimes XrdPosix was losing a file, e.g.

220616 19:35:07 10284 Posix_DDestroy: in use 1 timeout closing root://pmpe01:1294//mockdat/file_tag85707784944_fnum4105.dat_22748801_0 2029 objects lost

this was with a small test client program, using the XrdPosix interface. Investigation showed only 1 attempt to close() the file, rather than the expected 6 (or 7) attempts. I think there is a problem with the reference count accounting, and - according to what I think it happening - I've attached a suggested change.

### DIFF
--- a/src/XrdPosix/XrdPosixXrootd.cc
+++ b/src/XrdPosix/XrdPosixXrootd.cc
@@ -270,7 +270,7 @@ int XrdPosixXrootd::Close(int fildes)
    EPNAME("Close");
    XrdCl::XRootDStatus Status;
    XrdPosixFile *fP;
-   bool ret;
+   bool dnocb, ret;
 
 // Map the file number to the file object. In the prcess we relese the file
 // number so no one can reference this file again.
@@ -284,7 +284,7 @@ int XrdPosixXrootd::Close(int fildes)
 // the caller will get a zero return code should we delay the close.
 //
    fP->Ref();
-   if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP) && fP->Refs() < 2)
+   if ((dnocb = fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) && fP->Refs() < 2)
       {if ((ret = fP->Close(Status))) {delete fP; fP = 0;}
           else if (DEBUGON)
                   {std::string eTxt = Status.ToString();
@@ -293,6 +293,11 @@ int XrdPosixXrootd::Close(int fildes)
       } else {
        ret = true;
       }
+
+// If Detach indicates it completed without the need to call the callback
+// take the reference count down as there will be no callback to do it
+//
+   if (fP && dnocb) fP->unRef();
 
 // If we still have a handle then we need to do a delayed delete on this
 // object because either the close failed or there is still active I/O


smithdh:
There is a problem will null fP. Fixing now.

osschar:
Detach() is no longer supposed to be called multiple times. If it returns true, the cache IO object has been "stopped" and deleted. Otherwise it promises to call the passed callback when it was able to complete all outstanding IO (from prefetching) (https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfcIO.cc#L36).

Maybe the lost file was still in the process of being closed by the cache?

Detach() is never called when there is known outstanding IO requested from cache by the Posix layer itself -- at least it shouldn't be. Pfc does not check for this as at the time this mechanism was designed it wasn't really easy for Pfc to do so ... now that all reads are asynchronous it would be a trivial check to add.

smithdh:
Hi. This wasn't with the Pfc; it was a small client using posix layer, i.e. XrdPosixXrootd::Read() etc, towards a server.

abh3:
OK, so the issue is that the code may or may not use the Pfc since the fP pointer may either point to the actual file or, if a cache is being used, to the cached file (wonders of virtual objects). So, I think the question is what happens when it actually points to a Pfc object that doesn't expect Detach() to be called ore than once but the code may do so? Matevz do you know what would happen?

osschar:
Well, you promised it would not be called again :) The first time around, if Pfc needs more time, it will keep scheduling a job to check if IO is active and then call detach-finalize and invoke the CB you passed into original Detach. If you call detach again it will redo the ioactive check and potentially do what the scheduled job is supposed to do ... and the scheduled job will then land on a deleted object. So, nothing good.

What do you expect to happen to the CB you've passed in to the first / original Detach?

Why would you actually need to call Detach multiple times? You told me that when you call detach on your side the posix object is as good as gone so if a new request comes for the same file a new posix object will be created.

abh3:
Well, my question is based on the patch. So, it would seem that the patch does not make it possible to call Detach() more than once. While he patch appears to be technically correct, it seems like winds up involving delayed destroy when it need not and that was somewhat confusing relative to Detach(). David did you consider this variation to avoid delayed destroy?

fP->Ref();
if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) fP->Unref();
if (fP->Refs() < 2)
   {if ((ret = fP->Close(Status))) {delete fP; fP = 0;}
.... Rest of the code is the same, ending with
   } else {
     ret = true;
   }

Note that we no longer need dnocb nor the unref should it be true as this is taken care of inline and allowing the close to proceed if we are the only reference. However, it will still leave the file permanently in the queue if the close() fails as there will be no one to try it again. The original code had the same problem but I think it was done that way because it's unavoidable because XrdCl will leave that file in an undefined state so we can't close it again. Other than that, does this variation look right?

smithdh:
Hi Andy, Matevz,

Thanks for considering the patch. Sorry I think I didn't my goal well at first: my idea is not to allow XrdPosixXrootd::Close() (and hence Detach()) to be called multiple times.

I was in the situation where the first and only call of fP->Close(Status) on line XrdPosixXrootd.cc:288 fails. The file is then entered for Delayed destroy, I believe as intended, at least it was it was doing, and XrdPosixFile::DelayedDestroy() then periodically considers closing the file at line XrdPosixFile.cc:224 up to XrdPosixGlobals::ddMaxTries (6) times. There is a question of whether it will ever succeed - if not then the current action of leaving the file lost seems the best that can be done - but in my case it was possible that the retry would work.

What I saw in my test was that the Close() from line XrdPosixFile.cc:224 is never attempted (seen by instrumenting XrdCl). I understand the reason for that is that the condition at line XrdPosixFile.cc:223 is not true, i.e. Refs() is not zero, but = 1. I don't think anything was using the file.

So my idea with the patch was to make sure that the ref count of the file will eventually become zero (assuming that nothing is using the file). It seemed to me this didn't work with the current code when Detach() returned true (i.e. synchronous, and so never calling the callback which would have done an unRef() itself) .

Andy suggested:

```
fP->Ref();
if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) fP->Unref();
if (fP->Refs() < 2)
```

I think there's a problem in that the 3rd line, Refs()<2, is no longer the right condition. The believe that the condition wanted is that Detach() was synchronous and nothing is now using the file. Avoiding adding another variable, perhaps:

```
ret = true;
fP->Ref();
if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) {
   fP->unRef();
   if (fP->Refs() == 0) {
      if ((ret = fP->Close(Status))) {delete fP; fP = 0;}
         else if (DEBUGON)
                     {std::string eTxt = Status.ToString();
                      DEBUG(eTxt <<" closing " <<fP->Origin());
                      }
   }
}

// ....
if (fP) XrdPosixFile::DelayedDestroy(fP);
```

abh3:
I did understand after rechecking DelayedDestroy. Detach() can only be called once, no matter what. However, the code alternative is equivalent to the simpler original suggestion. Here is what the equivalence is:
a) fP->Ref();
    if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) fP->Unref();
    if (fP->Refs() < 2)
At this point  is Refs() is 1 then it must be due to the Ref() we just issued. So, it must be less than 2. Hence close can occur because we are the only ones holding the ref count.

b) fP->Ref();
    if (fP->XCio->Detach((XrdOucCacheIOCD&)*fP)) {
       fP->unRef();
       if (fP->Refs() == 0) {
At this point if Refs() is 0 it must be due to the Unref() we just issued (i.e. the count was 1 and <2 and now is 0). Hence close can occur because no one is holding the ref count (minor semantic quibble).

So, in fact, both code path so exactly the same thing if there is no error. However, there is a big difference when close returns an error. The object then gets on the delayed destroy queue with the ref count being 0.  That means delayed destroy will try to re-close the file. Since the file object is in an undefined state it's not clear whether or not this will have any bad side-effects. Perhaps @simonmichal can comment on this.

smithdh:
Hi Andy,

Thanks a lot for looking at this: This hasn't been reported as causing a problem in any production use; it was only something that caused me to make a separate investigation during a test. So I'm not sure it is worth pursuing this given there's some discussion around it: I think it's perfectly fine if you'd like to either leave it as is, or change it in anyway you'd prefer. (If you wanted a certain behaviour you could also let me know, and I'll try to prepare an appropriate patch, if you like). For instance if the conclusion is that delayed destroy is only that, not a kind of retry, we could avoid ever adding files which fail their initial close to the delayed list but directly to the lost list, or exit the attempted close+delete early from XrdPosixFile::DelayedDestroy() (before the ddMaxTries attempts) if a close is attempted but fails. Or, If indeed this a balance of doing the least bad thing in the face of a potentially inconsistent state we just leave it as is, since apparently it is not causing real problems.

Yours,
David

osschar:
There seems to be another issue nearby ... but it has to do with Pfc and async reads.

It can happen that async posix read handler decides to drop the outstanding reads when they are delayed beyond certain timeout and call this Close() regardless of that.

This should not really happen when in pfc mode as the outstanding async read calls will eventually invoke the pfc callbacks (either successful or failed / timed out) on a potentially deleted IO object.

It is unclear if the reference count check should be performed at the async delayed request handler before Close() is called.

XrdPosixXrootd::Close() calling Detach() before checking the actual reference count potentially send IO object into a too early shutdown.

simonmichal:
> So, in fact, both code path so exactly the same thing if there is no error. However, there is a big difference when close returns an error. The object then gets on the delayed destroy queue with the ref count being 0. That means delayed destroy will try to re-close the file. Since the file object is in an undefined state it's not clear whether or not this will have any bad side-effects. Perhaps @simonmichal can comment on this.

If the `XrdCl::File` is in an error state the close will be failed right away (without sending the request):
https://github.com/xrootd/xrootd/blob/38112a935133ff23fce302d61e166b2d4194449a/src/XrdCl/XrdClFileStateHandler.cc#L976-L980

Now, the only problem with deleting a `XrdCl::File` object which cannot be closed is that there might be some outstanding requests on that object. I could add a method for you that allows to check if there are any in-the-flight requests, or I could add a method that allows you to register a callback to be called once all the outstanding requests have been cleared.


> It can happen that async posix read handler decides to drop the outstanding reads when they are delayed beyond certain timeout and call this Close() regardless of that.

In the presence of outstanding read requests the close will fail by definition.

smithdh:
Hi. Thanks to everybody for the input. I'll close this pr, without merge, to discard it. Matevz, if you have in mind an issue for Pfc in this area (i.e. concerning async reads), I suppose you can open a dedicated issue or pr.

PR #1741, opened 2022-07-18
Create Job on failed Run method in XrdClOperations

Nikipedia:
Until now, the failed request resulted in the responsehandler being called immediately. 
This meant that no new thread was created and the reading mutex would be blocking forever (even when a pipeline is started as Async).
By replacing it with a job a new thread is created for the answer.


simonmichal:
@Nikipedia : thanks a lot for the PR, it looks good :-)

PR #1742, opened 2022-07-18
Length check in the LFH constructor

Nikipedia:
A maximum length is given to the constructor to detect corruption and prevent reading over the borders of the LFH.
In case the lengths read from the buffer are larger than the buffer itself, bad_data is thrown.


simonmichal:
@Nikipedia : thanks for the PR. 

AFAIU the intention is that if the optional argument `bufferSize` is not provided the checks are omitted, which provides backwards compatibility (that's good :-).

I added 2 minor comments about moving `throw` to next line in if statement, could you please address them?

simonmichal (approved):

PR #1745, opened 2022-07-19
MaxSize in CDFH constructor

Nikipedia:
Added a maximum length parameter to the parsing of the CDFH to detect corruptions in the CDFH length parameters (e.g. filenameLength) and to avoid reading over the end of the file.
Default argument is zero to allow for compatibility. 
The CDFH "Parse" method passes the new constructor argument.


simonmichal:
@Nikipedia : thanks for the PR, it looks good :-)

PR #1746, opened 2022-07-19
Metadata check of CDFHs 

Nikipedia:
To detect corruptions when opening the archive, the "offset", "uncompressedSize" and "compressedSize" values of the CDFH objects are checked against the total archive size.
Additionally, the sum of the sizes is checked against the total archive size.
In case of corruption, bad_data is thrown.


src/XrdCl/XrdClZipArchive.cc
```diff
@@ -430,6 +430,19 @@ namespace XrdCl
                                         else
                                           std::tie( cdvec, cdmap ) = CDFH::Parse( buff, eocd->cdSize, eocd->nbCdRec );
                                         log->Dump( ZipMsg, "[0x%x] CD records parsed.", this );
+                                        uint64_t sumSize = 0;
```
simonmichal, line 433: My understanding is you are summing the "compressed" and "uncompressed" size and making sure this does not exceed the archive size, in order to make sure we are not reading past the end of the buffer. Now the buffer will contain either compressed data (in case the archive was compressed) or uncompressed data (in case it was not compressed), so it doesn't make sense to check both. If it is compressed check only the sum of "compressed" file sizes, and if it is uncompressed the sum of "uncompressed" file sizes. In practice (AFAIR), if the file is NOT compressed the value of "compressed" size and "uncompressed" size will be the same (please double check in the documentation: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT), so in the end you can get away with always only checking "compressed" size.

src/XrdCl/XrdClZipArchive.cc
```diff
@@ -430,6 +430,19 @@ namespace XrdCl
...
+										uint64_t sumCompSize = 0;
+										for (auto it = cdvec.begin(); it != cdvec.end(); it++) {
+											sumSize += (*it)->uncompressedSize;
+											sumCompSize += (*it)->compressedSize;
+											if ((*it)->offset > archsize || (*it)->offset + (*it)->compressedSize > archsize
+													|| (*it)->offset + (*it)->uncompressedSize > archsize)
```
simonmichal, line 439: Similarly as in the previous comment, you should only check the "compressed" size.

simonmichal (approved):

Issue #1740, opened 2022-07-18
pgrw configuration server side


Jo-stfc:
I've tested an upgrade from xrootd 5.3.3 to 5.4.3 on a machine with a ceph storage backend. This caused an immediate drop in performance, which I've tracked to the fact that on recent xrootd versions, pgrw is the default transfer operations client side. Changes on the XrdCeph plugin (like features) didn't seem to be picked up correctly, as by the time the plugin converts the request into aio_read, they're already split into 64kb chunks, which is the main reason for the slowdown. 
Due to the fact that small read sizes might not be optimal for some storage types (ceph included), would it be possible to have this be configurable server-side, by having a configuration parameter determine whether the server supports pgrw, rather than it being tied to the protocol version, as it is currently?

abh3:
It would seem that the issue here is that async is allowed and for the pgread case it was meant to be used in that context for streaming across the network. So, I see the problem here you don't want to turn off async I/O to Ceph (note that in 5.4 we do turn async off now unless ithe I/O is going across the network). So, I think what you want is to be able to do is turn off async pgread but allow normal reads to use async I/O. Right?

abh3:
OK, so we have a conflict of what we want to accomplish here. We tried to detect transmission error and obviously that does affect how we interact with the storage. However, the brute force "turn it off" disables transmission errors which in HL-LHC will be a big deal. So, that's not a good solution. I won't say we hit it right but the proposed solution isn't right either. I appreciate your immediate concern but we need to find a god transition because we need to address this issue that will plague HL-LHC.  So, let's figure out what the right solution is for Ceph based storage.

Jo-stfc:
The main problem is the small individual read size. either changing the read size or buffering multiple requests into a single read request to ceph might improve things

abh3:
Well, that was what I was talking about, If you can't support a 64K read size you need to turn it off which in this case means turning off pgread. The evil thing about this is that it will revert to TLS which is still a 64K read and you will be sunk nonetheless. Hey. we are trying to verify the data that was sent. So, we need to find the compromise between network integrity and file system integrity/. In general we thought Ceph solved that problem but apparently we missed it.

Jo-stfc:
[...]

PR #1749, opened 2022-07-25
reduce maxCSSZ to avoid readv EINVAL error

Jo-stfc:
if the reported value goes above 2093056, clients crash due to readv having iovcnt > 1024.
The original value of 1022 * page size of 4096 generated dlen of 4186112. These were then resolved client side in XrdCl/XrdClAsyncPageReader.hh to ::readv calls of iovcnt 2044, triggering the einval error. 

fix for #1740


simonmichal:
@Jo-stfc : thanks a lot for your PR and for your investigation and understanding the root cause of the problem, it has been fixed in the client: 8319e5392f5c47296bd00627792fa1419782fa3b

PR #1750, opened 2022-07-27
ret type error on compilation

Jo-stfc:
type comparison - max_iovcnt() is of type int


simonmichal:
@Jo-stfc : thanks a lot for your PR, this is already fixed in 26eb65159521fd4bfe6b557f933ee1cc9bd34411

PR #1759, opened 2022-08-10
[XrdCl] Avoid race by using TimeOutSID in single place

smithdh:
Hello,

This is part of a set of 3 PRs for a set of changes to avoid various problems seen while investigating some EOS crashes or issues. There may well be discussion over these PRs!

Currently a SID (stream ID) may be

(i) marked as timed-out.
(ii) Later (usually quite soon) it may be marked as timed-out again.

It may also be released, either after (i) but before (ii) or after (ii).

If the release happens before (ii) the current SID Manager keeps the free SID marked as timed-out. Later it can be alloced again to a request, but when the response to the request is received the response handler doesn't get called with the usual result, as would be the case with a timed out request - but there is not failure or time-out notification sent either. One result seen in some tests is that the client's (sync) call might wait and never unblock.

e.g. (i) happens at XRootDMsgHandler::OnStreamEvent() (XrdClXRootDMsgHandler.cc:900). (ii) may happen at XRootDMsgHandler::HandleResponse() (XrdClXRootDMsgHandler.cc:1196). The release may happen at XRootDTransport::MessageReceived() (XrdClXRootDTransport.cc:1535). 

The approach taken in this PR was to change the SID manager, rather than to change the ordering or way the SIDs are alloated/timed-out/released. Changes are made that a SID can be allocated, marked as timed out. But to be made available again must be both released and marked as not-timed out. The sequence is indicated in the comments on the sidmanager patch.

### DIFF
--- a/src/XrdCl/XrdClXRootDMsgHandler.cc
+++ b/src/XrdCl/XrdClXRootDMsgHandler.cc
@@ -827,8 +827,7 @@ namespace XrdCl
           log->Debug( XRootDMsg, "[%s] Wait time is too long, timing out %s",
                       pUrl.GetHostId().c_str(),
                       pRequest->GetDescription().c_str() );
-          pStatus   = Status( stError, errOperationExpired );
-          HandleResponse();
+          HandleError( Status( stError, errOperationExpired) );
         }
         return;
       }
@@ -889,14 +888,6 @@ namespace XrdCl
 
     if( pTimeoutFence.load( std::memory_order_relaxed ) )
       return 0;
-
-    if( pSidMgr && pMsgInFly && ( event == Timeout
-        || status.code == errOperationExpired
-        || status.code == errOperationInterrupted ) )
-    {
-      ClientRequest *req = (ClientRequest *)pRequest->GetBuffer();
-      pSidMgr->TimeOutSID( req->header.streamid );
-    }
 
     HandleError( status );
     return RemoveHandler;
@@ -1189,10 +1180,8 @@ namespace XrdCl
     if( pSidMgr && finalrsp )
     {
       ClientRequest *req = (ClientRequest *)pRequest->GetBuffer();
-      if( !status->IsOK() && pMsgInFly &&
-          ( status->code == errOperationExpired || status->code == errOperationInterrupted ) )
-        pSidMgr->TimeOutSID( req->header.streamid );
-      else
+      if( status->IsOK() || !pMsgInFly ||
+          !( status->code == errOperationExpired || status->code == errOperationInterrupted ) )
         pSidMgr->ReleaseSID( req->header.streamid );
     }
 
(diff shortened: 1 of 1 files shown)


simonmichal:
@smithdh : thanks a lot for this PR and for describing this corner-case scenario!!!

Now I would like to add some additional insight:

In the past, in case of socket errors and timeouts the completion handler used to be called in the event-loop thread (this was deadlock prone and has been changed). At that time we used to have only the (ii) call to `TimeOutSID`. After moving all completion handler invocations to the thread-pool we had to mark the `SID` as timed-out at an earlier stage, because otherwise we would risk a race-condition with the server response, hence the (i) call to `TimeOutSID`.

It is worth noticing that the `TimeOutSID` routine is excessively called in case of: `errOperationExpired` and `errOperationInterrupted`. I had a look and checked where the status is set to those 2 error codes:

1) XrdClXRootDMsgHandler.cc : 832
We are setting the status to `errOperationExpired` because we got a `kXR_wait` response and taking into account the wait time we would exceed the `RequestTimeout` hence we are time-outing the request right away.
In this case we are calling `HandleResponse`, hence (ii) call to `TimeOutSID` is responsible for marking the `SID` as timed-out.

2) XrdClStream.cc : 381
For a given request in the outgoing-queue the `RequestTimeout` has been exceeded. This basically results in calling `XRootDMsgHandler::OnStatusReady` -> `XRootDMsgHandler::HandleError` -> `XRootDMsgHandler::HandleRspOrQueue` -> `XRootDMsgHandler::HandleResponse`. Again, the (ii) call to `TimeOutSID` is responsible for marking the `SID` as timed-out.

3) XrdClInQueue.cc : 172
For a given request in the incoming-queue the `RequestTimeout` has been exceeded. This results in calling `XRootDMsgHandler::OnStreamEvent`, hence (i) call to `TimeOutSID` is responsible for marking the `SID` as timed-out.

4) XrdClChannel.cc : 168
The user is enforcing a channel disconnect, hence we are using `errOperationInterrupted`. Now this boils down to calling `XRootDMsgHandler::OnStatusReady` for requests in outgoing-queue and `XRootDMsgHandler::OnStreamEvent` for incoming-queue requests and is equivalent to 2) and 3).


All that said, I would suggest to change the patch so we are using the `TimeOutSID` routine only in one place, rather than changing the `SIDManager` class. So here are some suggestions:

1) Move following code (i):
https://github.com/xrootd/xrootd/blob/3756f1b43931e782c4917ab0edc443743893162c/src/XrdCl/XrdClXRootDMsgHandler.cc#L893-L899
to the beginning of `XRootDMsgHandler::HandleError`.

2) Remove the (ii):
https://github.com/xrootd/xrootd/blob/3756f1b43931e782c4917ab0edc443743893162c/src/XrdCl/XrdClXRootDMsgHandler.cc#L1194
It’s worth noticing that we still need the `if` statement because we can call the `ReleaseSID` only for responses that have not been timed-out.

3) We need to replace those 2 lines:
https://github.com/xrootd/xrootd/blob/3756f1b43931e782c4917ab0edc443743893162c/src/XrdCl/XrdClXRootDMsgHandler.cc#L830-L831
with:
`HandleError( Status( stError, errOperationExpired) );`

In case of `OnStreamEvent` and `OnStatusReady` we are anyway calling the `XRootDMsgHandler::HandleError` routine, hence we will mark respective `SID`s as timed-out.

This way all the scenarios that include setting the error to `errOperationExpired` or `errOperationInterrupted` will go through the `HandleError` routine that will mark the `SID` as timed-out. I think this will make the code less complex.

Please let me know your thoughts!

smithdh:
Hi Michal. Thanks that seems like a good idea to simplify the change, while still avoiding the problem. I've revised the patch to use that approach; but I've moved the PR back to draft just for now. I'll let some tests complete before moving the patch back.

smithdh:
Looks fine; if you're happy I think it can be merged. I'll move PR back to 'ready for review'.
Thank you!

simonmichal:
@smithdh : it looks great, thanks a lot!

PR #1760, opened 2022-08-10
[XrdCl] Avoid situation where client does not read all of a network message

smithdh:
Hello,

This is part of a set of 3 PRs for a set of changes to avoid various problems seen while investigating some EOS crashes or issues. There may well be discussion over these PRs!

This PR primarily tries to avoid de-synconisation of the reading of (valid) streams from the server when a request times out. There is

(i) An addition to AsyncMsgReader::Read such that is a kXR_status message is detected, but there is no MsgHandler for the streamid, the correct number of bytes is still read from the network, so that subsequent messages can be received. The situation can happen during a timeout, where e.g. a pgread response is received but the request has timed-out.
(i - a) to allow (i) Stream::InspectStatusRsp() was modified so that the caller can differentiate the situation where there is no MsgHandler and the situation where not-raw readout was wanted.

(ii) Changes in Stream::OnIncoming(). (a) The MsgHandler inMsgHelper is definitely cleared, previously there were some situations where it was not. The current (possibly partial) message has been received so it should be appropriate to clear the handler from the helper. Conversely if it is not cleared, and the next message on the network was for a different SID it may be passed to the wrong MsgHandler, which could read the wrong number of bytes read (and loss of sync) or put them into the response of a different request. (b) Call pTransport->MessageReceived() only for not-partial messages. The MessageReceived() function includes a check for completed requests to release the timeout-mark (which should be applied only to the final response), and decrements the strmqueues statistics-counters, which I believe are for whole messages. (c) Clears the timeout-fence for requests that are moved from the inMsgHelper back to the InQueue (in case of stream errors). This could avoid user sync requests blocking indefinitely.

(iii) Changed the Stream::ForceError() to keep the lock while all sub-streams are moved, not only the first stream.

### DIFF
--- a/src/XrdCl/XrdClAsyncMsgReader.hh
+++ b/src/XrdCl/XrdClAsyncMsgReader.hh
@@ -199,36 +199,96 @@ namespace XrdCl
               //----------------------------------------------------------------
               // Now check if there are some additional raw data to be read
               //----------------------------------------------------------------
-              if( !inhandler )
+              if( inhandler )
               {
-                uint16_t action = strm.InspectStatusRsp( substrmnb,
-                                                         inhandler );
-
-                if( action & MsgHandler::Corrupted )
-                  return XRootDStatus( stError, errCorruptedHeader );
-
-                if( action & MsgHandler::Raw )
-                {
-                  //------------------------------------------------------------
-                  // The next step is to read the raw data
-                  //------------------------------------------------------------
-                  readstage = ReadRawData;
-                  continue;
-                }
-
-                if( action & MsgHandler::More )
-                {
-                  //------------------------------------------------------------
-                  // The next step is to read the additional data in the message
-                  // body
-                  //------------------------------------------------------------
-                  readstage = ReadMsgBody;
-                  continue;
-                }
+                //--------------------------------------------------------------
+                // The next step is to finalize the read
+                //--------------------------------------------------------------
+                readstage = ReadDone;
+                continue;
               }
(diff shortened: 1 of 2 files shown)


smithdh:
Hi,

I wrote:

> (ii) Changes in Stream::OnIncoming(). (a) The MsgHandler inMsgHelper is definitely cleared, previously there
> were some situations where it was not.

Actually I believe those situations (i.e. when MessageReceived() could return DigestMsg when inMsgHelper.handler is set) are eliminated with PRs 1758 + 1759 so probably this PR could be slightly simplified by not changing these lines of the existing code base, i.e. keeping the two 'mh.Reset();' in Stream::OnIncoming() at their current places. I'll let you consider it.

smithdh:
Hi Michal,

I wrote:

> I'll let you consider it.

I mean I believe the diff could be slightly simplified, if you'd like me to adjust the patch let me know. I won't start changing the pr now as you're probably reviewing/thinking about it.

simonmichal:
@smithdh : thanks a lot for the PR, the error scenarios you presented are absolutely valid!

Regarding (i), I think it would be better to refactor the `AsyncMsgReader` so `kXR_status` is handled separately like we do for `kXR_attn`. We could then handle clearly the 3 cases: read additional body (pgwrite), read additional raw data (pared) and t/o request (which is equivalent to reading everything into the message body). That said, I think this is the best thing me can have for `5.5.0`, so it's better to merge it now and do the refactoring latter.

Issue #1752, opened 2022-08-01
Missing HTTP error code mapping for storage.create


vokac:
Release candidate 5.5.0-rc1 is still [not fully compliant](https://ci.cloud.cnaf.infn.it/job/wlcg-jwt-compliance-tests/job/master/525/artifact/reports/reports/latest/se-prague-xrootd/log.html) with WLCG JWT profile `storage.create:/` scope. Attempt to overwrite existing file with just `storage.create:/` scope fails with HTTP 500 while it should fail with HTTP 403.

Current error mapping in the XRootD HTTP module is not sufficient
https://github.com/xrootd/xrootd/blob/af9b7e416aed422fcfa6cba7cbc9ead04f8f2ce3/src/XrdHttp/XrdHttpReq.cc#L986-L1009
because overwriting existing file fails with
```
220730 17:20:36 818 unknown.77:26@ui1 Xrootd_Response: 0000 sending err 3018: Unable to create /wlcg/test.create; file exists
```
and 3018 (kXR_set) error code mapping is not present => default HTTP 500 is used
https://github.com/xrootd/xrootd/blob/af9b7e416aed422fcfa6cba7cbc9ead04f8f2ce3/src/XrdHttp/XrdHttpReq.cc#L983-L984

Full server side log can be found in https://vokac.web.cern.ch/vokac/tmp/xrootd-5.5.0rc1-storage-create-http-500.log

ffurano:
Hi,

  I see that the code misses the mapping for

  kXR_ItExists,        // 3018

  given that the error is "file exists" I'll map it to 409

  The server log (thanks!) is pretty clear about this:

220730 17:20:36 818 scitokens_Access: Grant authorization based on scopes for operation=excl_create, path=/wlcg/test.create
220730 17:20:36 818 scitokens_Access: Request username wlcgprd
220730 17:20:36 818 ofs_open: ***@***.*** Unable to create /wlcg/test.create; file exists
220730 17:20:36 818 ***@***.*** Xrootd_Response: 0000 sending err 3018: Unable to create /wlcg/test.create; file exists
220730 17:20:36 818 http_Req:  XrdHttpReq::Error
220730 17:20:36 818 ***@***.*** http_Req: PostProcessHTTPReq req: 4 reqstate: 0
220730 17:20:36 818 ***@***.*** http_Req: PostProcessHTTPReq mapping Xrd error [3018] to status code [500]
[...]

PR #1763, opened 2022-08-15
Return permission denied when overwrites are not allowed

bbockelm:
For the new "exclusive create" permission (meaning "never overwrite existing data"), if the file cannot be created due to the destination existing, we should return a "permission denied".

This also fixes a relatively serious issue in the built-in AuthDB module where it may incorrectly authorize the creation of files due to an uninitialized memory read.

Fixes: #1752

### DIFF
--- a/docs/ReleaseNotes.txt
+++ b/docs/ReleaseNotes.txt
@@ -5,6 +5,83 @@ XRootD
 Release Notes
 =============
 
+-------------
+Version 5.5.0
+-------------
+
++ **New Features**
+  **[XrdApps]** Provide command line too to manipulate checksum xattr.
+  **[XrdApps]** xrdreplay: support quoted columns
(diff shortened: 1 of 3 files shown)


bbockelm:
@simonmichal - this definitely needs to be in 5.5.0 (at least the patch that incorrectly give folks 'create' access).  However, for PRs, do you want me to target `stable-5.5.x` like I've done or do you want me to target `master` and then cherry-pick?

simonmichal:
@bbockelm : thanks for letting me know, please do the PR against master, I will sync the stable branch after it gets merged

simonmichal:
@bbockelm : could you please drop the changes to `docs/ReleaseNotes.txt`. I think this is just a left over after changing the target branch from stable to master, and it will give me a conflict when merging to stable. Thanks!

simonmichal:
@bbockelm : sorry I overlooked it's overwritten by #1765, please disregard my previous comment

PR #1765, opened 2022-08-17
Fix overwrite return code (master branch version)

bbockelm:
This is a new copy of #1763 - only difference is it's based on the master branch and not the stable-5.5.x.


bbockelm:
FWIW -- build failures on macos looks unrelated (something in XrdPfcFile?)

PR #1767, opened 2022-08-22
Fix self-move.

bbockelm:
In what appears to be a refactor gone wrong, the `username` variable is moved to itself.  On RHEL7's compiler, this appears to be a no-op.  In RHEL8 (probably correctly) it clears out the string, un-setting the username completely.

@simonmichal - please also backport to the 5.5.x branch.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -776,10 +776,11 @@ private:
         token_subject = std::string(value);
         free(value);
 
+        std::string tmp_username;
         if (config.m_map_subject) {
-            username = token_subject;
+            tmp_username = token_subject;
         } else {
-            username = config.m_default_user;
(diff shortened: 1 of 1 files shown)


simonmichal:
@bbockelm : thanks for the PR! It makes me wonder, do we actually need the `move`? wouldn't it be enough just to drop the line 859:

https://github.com/xrootd/xrootd/blob/dc680e5f8b82aa7e90e23d48d7ea22cbe18d6402/src/XrdSciTokens/XrdSciTokensAccess.cc#L859

In the end the `username` variable is being set to the right value, isn't it?

bbockelm:
I would prefer keeping the `move` -- I don't want to touch any of the output variables until we know the rest of the function is successful.  I dislike inconsistent state on failure.

abh3:
OK, we see the reasoning here and it's pretty sound. Thanks for explaining.

Issue #1768, opened 2022-08-23
Staged installation of Python bindings not working after migration to pip


amadio:
Original bug report from Gentoo Linux: https://bugs.gentoo.org/861452

Bisection with git points to commit https://github.com/xrootd/xrootd/commit/31e3d717ad191e966ddbd4802f85982bfb5618bd as the problematic change.

How to reproduce the problem:
```sh
$ # Configure with Python enabled and install prefix in the system (e.g. /usr)
$ cmake -S $PWD -B build -DENABLE_PYTHON=1 -DPYTHON_EXECUTABLE=/usr/bin/python3 -DCMAKE_INSTALL_PREFIX=/usr
$ cmake --build build
$ # Perform a staged installation into a DESTDIR which is not empty (e.g. to a directory install in the current directory):
$ env DESTDIR=$PWD/install cmake --build build --target install
```
What I get:
```
... (many removed lines of output)
-- Processing: /home/amadio/src/xrootd/install//usr/share/man/man1/xrdcp.1
-- Processing: /home/amadio/src/xrootd/install//usr/share/man/man1/xrdmapc.1
-- Installing: /home/amadio/src/xrootd/install/usr/lib64/libXrdClHttp-5.so
-- Set runtime path of "/home/amadio/src/xrootd/install/usr/lib64/libXrdClHttp-5.so" to ""
Processing ./bindings/python
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'done'
Building wheels for collected packages: xrootd
  Building wheel for xrootd (setup.py): started
  Building wheel for xrootd (setup.py): finished with status 'done'
  Created wheel for xrootd: filename=xrootd-2022.822+dc680e5f8-cp310-cp310-linux_x86_64.whl size=711034 sha256=ca7888c135d15b360c8f8df1f7c05222855e3bf39bd7e0105ced715581f59d67
  Stored in directory: /tmp/pip-ephem-wheel-cache-k0enpw1z/wheels/4f/bf/c3/60b496a0b3d3cc1bbe87053fb501e707645f2f8833a13d0c18
Successfully built xrootd
Installing collected packages: xrootd
  Attempting uninstall: xrootd
    Found existing installation: xrootd 5.4.3
    Uninstalling xrootd-5.4.3:
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/usr/lib/python3.10/site-packages/XRootD/__init__.py'
[...]

PR #1769, opened 2022-08-23
Fix staged installation of Python bindings with pip

amadio:
When performing a staged installation with pip, it will attempt to uninstall a previous installation of XRootD if it exists, so if XRootD is already installed into `/usr`, it may try to remove files installed by the system package manager.

Fixes: 31e3d717ad191e966ddbd4802f85982bfb5618bd
Closes: #1768

See also: https://bugs.gentoo.org/861452

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -95,7 +95,7 @@ else()
     CODE
     "EXECUTE_PROCESS(
       COMMAND /usr/bin/env ${XROOTD_PYBUILD_ENV} ${PYTHON_EXECUTABLE} -m pip install \
-                ${PIP_OPTIONS} \
+                --ignore-installed ${PIP_OPTIONS} \
                 ${CMAKE_CURRENT_BINARY_DIR}
       )"
   )


simonmichal:
@amadio : thanks a lot for the PR, maybe @matthewfeickert could have a look at it

bindings/python/CMakeLists.txt
```diff
@@ -95,7 +95,7 @@ else()
     CODE
     "EXECUTE_PROCESS(
       COMMAND /usr/bin/env ${XROOTD_PYBUILD_ENV} ${PYTHON_EXECUTABLE} -m pip install \
-                ${PIP_OPTIONS} \
+                --ignore-installed ${PIP_OPTIONS} \
```
matthewfeickert, line 98: `--ignore-installed` is almost NEVER what you want to do and certainly NOT what you want to do by default. This should not go in as is.

matthewfeickert (changes requested):

matthewfeickert:
I'll let @amadio confirm that it works for him, but this PR can get closed given https://github.com/xrootd/xrootd/issues/1768#issuecomment-1224217117.

bindings/python/CMakeLists.txt
matthewfeickert: Sorry the all caps are a bit much &mdash; I was in a rush while typing this and didn't think to leave a more thoughtful response. The reason why I would suggest trying to always avoid `--ignore-installed` is laid out in the docs which I quote there. As it name suggests, you're literally telling `pip` to act like you're installing into a blank directory space that has just been created. This can cause obvious pathologies as you're now subject to having potentially anything in the the package space that can lead to confusing and undesirable behavior at runtime.

If you intentionally are using `--ignore-installed` for a situation where [you know what you're doing and there really isn't any better alternative](https://github.com/matthewfeickert/cvmfs-venv/blob/afd316674c954b5f709bf7a0181899b8045f3463/README.md#disadvantages) then that's obviously fine, but this should definitely not be the default behavior for anything.

This is why I suggested in https://github.com/xrootd/xrootd/issues/1768#issuecomment-1224217117 to just use the `PIP_OPTIONS` CMake option which was added for situations like this.

amadio: No worries. I actually sort of agree with the above, and that's [what I did](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=769b1f2e6faabebc6d0900142cf6d671db1e9ed3) in Gentoo to get the package fixed. However, I think that maybe having an alternative to using the pip install method altogether would be better. If someone just wants to make a binary tarball of XRootD to install elsewhere and tries the usual `make DESTDIR=xrootd-5.x/ install`, that may, as a side effect, just remove the currently installed version of XRootD from their system. This is probably not a good idea as default behavior and why I thought that having that `--ignore-installed` as default would be better (also since `make install` usually has as default behavior installation without asking/uninstalling anything). That being said, I didn't see an option that can negate it by adding to `PIP_OPTIONS` so maybe a different solution is preferable. One idea would be to have `PIP_OPTIONS="--user"` by default, for example, to avoid the situation I describe above.

amadio:
> Package managers don't install to pip's site-packages and if they do then they are doing it wrong and this isn't a problem with XRootD.

The build system of XRootD is calling `pip` to install files, not the package manager. Package managers rely on the install targets of the respective packages. In the case of Gentoo, portage is just calling the equivalent of `env DESTDIR=$STAGING_AREA cmake --build $BUILD_DIR --target install`, and that's trying to delete files outside of the build area (i.e. in /usr), causing a sandbox violation ([bug 861452](https://bugs.gentoo.org/861452)). The package manager installs to a staging area, then moves installed files into the final destination in a second step, keeping track of what's installed, etc. The installation phases are described [here](https://devmanual.gentoo.org/ebuild-writing/functions/).

bindings/python/CMakeLists.txt
matthewfeickert: > If someone just wants to make a binary tarball of XRootD to install elsewhere and tries the usual make `DESTDIR=xrootd-5.x/` install, that may, as a side effect, just remove the currently installed version of XRootD from their system.

If that's the situation then [`pip`'s `--target` option](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-t) should be used

> Install packages into `<dir>`. By default this will not replace existing files/folders in `<dir>`. Use `--upgrade` to replace existing packages in `<dir>` with new versions.

Example:

```console
$ docker run --rm -ti python:3.10 /bin/bash
root@0abcc4f646be:/# python -m venv venv && . venv/bin/activate
(venv) root@0abcc4f646be:/# python -m pip --quiet install --upgrade pip
(venv) root@0abcc4f646be:/# python -m pip install 'numpy<1.23.0'
Collecting numpy<1.23.0
  Downloading numpy-1.22.4-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (16.8 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 16.8/16.8 MB 23.1 MB/s eta 0:00:00
Installing collected packages: numpy
Successfully installed numpy-1.22.4
(venv) root@0abcc4f646be:/# mkdir /example
(venv) root@0abcc4f646be:/# python -m pip --quiet install --target /example --upgrade numpy
(venv) root@0abcc4f646be:/# python -m pip show numpy
Name: numpy
Version: 1.22.4
Summary: NumPy is the fundamental package for array computing with Python.
Home-page: https://www.numpy.org
Author: Travis E. Oliphant et al.
Author-email: 
License: BSD
Location: /venv/lib/python3.10/site-packages
Requires: 
Required-by: 
(venv) root@0abcc4f646be:/# ls /example/
bin  numpy  numpy-1.23.2.dist-info  numpy.libs
(venv) root@0abcc4f646be:/#
```

Trying that instead of the use of `--prefix` might be a reasonable approach.

> That being said, I didn't see an option that can negate it by adding to `PIP_OPTIONS` so maybe a different solution is preferable.

Just use `DPIP_OPTIONS="--ignore-installed"` for your particular use case.

> One idea would be to have `PIP_OPTIONS="--user"` by default, for example, to avoid the situation I describe above.

That is also something that should not be done. `--user` should not be used unless there are extreme situations as you should not be installing packages with `pip` globally.

amadio: I had looked into the `--root` option only, I think indeed the `--target` option is what we need here. I will update my pull request using that instead of `--ignore-installed`.

matthewfeickert: > I had looked into the `--root` option only, I think indeed the `--target` option is what we need here. I will update my pull request using that instead of `--ignore-installed`.

No. You will need to add it as an option in [`bindings/python/CMakeLists.txt`](https://github.com/xrootd/xrootd/blob/ef4f041db001d791adb14d42cbc51d84bcd82117/bindings/python/CMakeLists.txt) and not as a global default.

amadio:
Closing since #1768 has been closed.

PR #1778, opened 2022-09-01
[XrdHttp] The HTTP server certificate is properly renewed every time the XrdTlsContext::Refresh thread runs

ccaffy:
After that fix, I started the XrdHttp server with an expired certificate and updated it to a valid certificate. The transfers were failing before the run of the Refresh thread and were successful after the run.


src/XrdTls/XrdTlsContext.cc
```diff
@@ -860,32 +860,30 @@ void *XrdTlsContext::Session()
...
 #else
-   SSL_CTX_set1_cert_store(pImpl->ctx, newX509);
+   SSL_CTX_set1_cert_store(xrdctxnew->pImpl->ctx, newX509);
 #endif
 
-// Save the generated context and clear it's presence
-//
-   XrdTlsContext *ctxold = pImpl->ctxnew;
-   pImpl->ctxnew = 0;
-
```
abh3, line 880: OK, I see there is a misunderstanding here. What the original code actually does is grab the X509 store from the new context and sets it to be used by the current context. It then simply deletes the new context and sets the pointer to it to zero, indicating that there is no new x509 store so that the x509 transfer executes only once and allows a new x509 store to be created. The reason we need to do that is because an x509 store creation also requires the creation of a context but we never want to use that newly created context. We must continue to use the current context because we don't know how the current context is being used (e.g. there may be an outstanding pointer to it). So, we must always use the current context to create SSL objects. The changes starting on line 876  undo all of that and I would say that it's luck that it worked. Hence, all of the original code starting at line 876 should not have been changed. The code above that line was indeed incorrect and your changes fixed that, thanks!.

abh3 (changes requested):
Please read the comment explaining why some of the changes need to be undone. Specifically, the original code starting at line 876 to the end of Session() should have been changed.

ccaffy:
Hi Andy,

Thanks a lot for your comments. I now understand what you were trying to do...
Unfortunately, your suggestion would have worked if it was possible to exchange the certificate store between SSL_CTX. This is only possible for openssl with a version >= 1.1.1 (with the `SSL_CTX_set1_cert_store()` macro) (documentation: https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_get_cert_store.html)
.
As we are running Centos 7 that comes with OpenSSL 1.0.2, this macro does not exist and cannot be used.

I have discussed this with @simonmichal and I believe we have found a way around it.

I'll keep you updated as I progress...

Cheers,
Cedric

ccaffy:
Hi @abh3 ,

I have made some changes that work with HTTP, but as I touched the implementation of the `XrdTlsContext` class, I would also like to see if I did not break anything with XRootD TLS transfers.

I may make a fool of myself here, but I do not manage to have `roots` transfers working against my XRootD server (with and without my changes). What should be configured on the server side to have `xrdcp` working with a `roots://` URL ?

My current server config contains the following:

```
all.export /tmp nolock
all.adminpath  /var/spool/xrootd
xrd.trace tls
xrd.port tls 1095
xrd.tls /etc/grid-security/xrd/xrdcert.pem /etc/grid-security/xrd/xrdkey.pem
xrd.tlsca certdir /etc/grid-security/certificates refresh 15s log failure
if exec xrootd
  xrd.protocol http:1096 libXrdHttp-5.so
fi
```

Thanks in advance !

abh3:
Hi Cedric,

The problem here is that 'xrd.port tls 1095' directive. It really only works for http because https can deal with a port already using ssl. In xroot protocol we always start off not using ssh and hen upgrade to using ssh when needed. So, there is another directive that tells the server when that should happen. Specifically, xrootd.tls
https://xrootd.slac.stanford.edu/doc/dev55/xrd_config.htm#_Toc88514013

In order to always requite tls for everything use 'xrootd.tls all' and the server will tell the client it has to use tls whether the protocol is 'root' or 'roots'. If you want to be able to control it on the client side via the protocol specification then simply drop the 'tls' option of the xrd.port directive and omit the xrootd.tls directive.

ccaffy:
Hi @abh3 ,

Many thanks for your answer. I removed that `xrd.port tls 1095` directive. It indeed did not work. Now I still have the problem of the server having a TLS error when I use `roots`: 

```
220912 09:48:46 7273 anon.0:23@localhost TLS_Accept: Accepting a TLS connection...
220912 09:48:46 7273 XrdTLS: anon.0:23@localhost TLS error rc=0 ec=1 (error_ssl) errno=0.
220912 09:48:46 7273 XrdTLS: anon.0:23@localhost 140102374799104:error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:s3_pkt.c:1493:SSL alert number 48

220912 09:48:46 7273 XrdLinkXeq: TLS connection from anon.0:23@localhost failed; error_ssl
220912 09:48:46 7273 XrootdXeq: Unable to enable TLS for anon.0:23@localhost
220912 09:48:46 7286 anon.0:26@localhost TLS_Accept: Accepting a TLS connection...
220912 09:48:46 7273 XrootdXeq: anon.0:23@localhost disc 0:00:00
220912 09:48:46 7286 XrdTLS: anon.0:26@localhost TLS error rc=0 ec=1 (error_ssl) errno=0.
220912 09:48:46 7286 XrdTLS: anon.0:26@localhost 140102338623232:error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca:s3_pkt.c:1493:SSL alert number 48

220912 09:48:46 7286 XrdLinkXeq: TLS connection from anon.0:26@localhost failed; error_ssl
220912 09:48:46 7286 XrootdXeq: Unable to enable TLS for anon.0:26@localhost
220912 09:48:46 7286 XrootdXeq: anon.0:26@localhost disc 0:00:00
```

Again, no problem with HTTP with the following curl command:

```
curl -v -X GET https://localhost:1096/tmp/testFile 2>&1 >/dev/null
```

Thanks again!

abh3:
I presume you pushed the changes. Let me look at them and then let's discuss why it works for http but not for xroots. I will be at CERN and specifically at the IT building starting this Monday.

ccaffy:
Hi @abh3 ,

I did not push anything yet. OK, I'll push only fixes for HTTP.

The reason it does not work with xroots is because the SSL_CTX given to SSL_accept() is always the same. Also, the Refresh thread is not started with xroots while it is the case for HTTP :)

Great news you are at CERN this week, we can discuss about all of this around a coffee :)

Cheers,
Cedric

abh3:
I looked at the patch and it's starting to become even more complicated. I really want to find the simplest solution for code maintainability purposes. This is especially import as 1.02 OpenSSL has very limited lifetime. So, I re-read the comments and one struck me that now does not make sense. The supposition here is that SSL_CTX_set1_cert_store() is used for all version of OpenSSL (i.e. the note was that it only exists for 1.1.1 and above) and that is true. But that is why that call  is replaced by two equivalent calls SSL_CTX_set1_verify_cert_store() and SSL_CTX_set1_chain_cert_store() when dealing with older versions. So it should not matter if we are dealing with 1.0.1 or any version above it. Please see

https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set1_verify_cert_store.html

ccaffy:
Thanks for your comment Andy, I'm happy to discuss this with my laptop at CERN. Please pass by my office whenever you are available :)

abh3:
We gone through the issues here and will revert to the most simple solution. Since we'd like to start with new history on this issue, I am closing this pull request.

abh3:
Hi Cedric,

Yjis is an issue with the root CA which is causing cert validati9on to 
fail. If you seach for "unknown ca:s3_pkt.c:1493" you will many issues 
similar, if not identical, to this along with possible solutions. The 
particular solution is dependint on how the certs were issued.

Andy

PR #1798, opened 2022-10-10
[XrdHttp + XrdHttpExtHandler] Fixes a segmentation fault happening when a client sends a first line starting with a space

ccaffy:
Hi Andy,

During normal EOS operations, we noticed that one storage node crashed in the XrdHttpExtHandler.cc:93 line.
@smithdh (many thanks again!) could reproduce the issue by starting an EOS FST and submitting, via netcat: 

```
nc -Cv localhost 11000
Ncat: Version 7.50 ( https://nmap.org/ncat )
Ncat: Connected to ::1:11000.
 GET // HTTP/1.1
```

Notice the first line starting with a space. This space, at the beginning of the first line would prevent the `XrdHttpReq::parseFirstLine()` method to call the `XrdHttpReq::parseResource()` one.

Once the constructor of an `XrdHttpExtReq` is called, the line `headers["xrd-http-fullresource"] = req->resourceplusopaque.c_str();` will segfault as the resource was not parsed. Indeed, the `resourceplusopaque` XrdOucString has not been asigned and its the c_str() method of an will therefore return a NULL pointer.

Hence my two fixes.

Cheers,
Cedric

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -90,7 +90,8 @@ verb(req->requestverb), headers(req->allheaders) {
   int envlen = 0;
   
   headers["xrd-http-query"] = req->opaque?req->opaque->Env(envlen):"";
-  headers["xrd-http-fullresource"] = req->resourceplusopaque.c_str();
+  const char * resourcePlusOpaque = req->resourceplusopaque.c_str();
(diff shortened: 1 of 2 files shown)


abh3:
Thanks for the fix. I will merge it as soon as Michal checks why the CentOS 7 test script got a SEGV.

simonmichal:
It must have been just a hiccup in the CI, I had a look but couldn't find the problem nor reproduce it. I will go ahead and merge it.
 (BTW we have to update the CI script so we can extract the coredump or at least a stack trace.)

PR #1802, opened 2022-10-15
Implement XrdSciTokensHelper interface for macaroons.

bbockelm:
In order to use ZTN with a token library, the library must provide a XrdSciTokensHelper object.  This implements the interface inside the Macaroons::Authz object, allowing ZTN authentication to be utilized with a macaroon-based token.

This can be enabled with the following configuration directive:
```
sec.protocol -tokenlib libXrdMacaroons-5.so
```

### DIFF
--- a/src/XrdMacaroons/XrdMacaroons.cc
+++ b/src/XrdMacaroons/XrdMacaroons.cc
@@ -27,6 +27,7 @@ extern XrdAccAuthorize *XrdAccDefaultAuthorizeObject(XrdSysLogger   *lp,
                                                      const char     *parm,
                                                      XrdVersionInfo &myVer);
 
+XrdSciTokensHelper *SciTokensHelper = nullptr;
 
 extern "C" {
 
@@ -38,7 +39,9 @@ XrdAccAuthorize *XrdAccAuthorizeObjAdd(XrdSysLogger *log,
 {
     try
     {
-        return new Macaroons::Authz(log, config, chain_authz);
+        auto new_authz = new Macaroons::Authz(log, config, chain_authz);
+        SciTokensHelper = new_authz;
+        return new_authz;
     }
     catch (std::runtime_error &e)
     {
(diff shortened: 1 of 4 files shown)


bbockelm:
The motivation here is the Coffea-Casa analysis facility at Nebraska: the facility internally uses macaroons between the client and XCache.

The current authentication setup predates the existence of the ztn protocol and requires both a server-side and client-side plugin.

With this server-side change, we'd be able to use ztn out-of-the-box on the client as introduced in 5.2.0 and drop the customization to use Macaroons as a session token with XCache.

wyang007:
Do this means we can use a macaroon in root protocol (via ZTN) ?

bbockelm:
@wyang007 - yes, with this change you can use a macaroon via ZTN.

A logical followup of this PR would be a module that allows both macaroon or SciTokens to be accepted in a single config.  I purposely decided to do this piece first and want to work on a joint module next.

wyang007:
I am thinking of macaroon and scitoken as something similar to signed URL in the sense that it allows someone a temporary access. and macaroon is issued by the storage admin and scitoken is issued by a VO. 

The current mechanism for a storage admin to obtain a macaroon is only available in http protocol (which is OK) and require either an x509 or scitoken authentication. I wonder if there are other "local" authentication methods (such as krb5, password, sss, etc.) on the http side that allow an admin obtain a macaroon . Andy says that "password" may work in http but never tested.

bbockelm:
Hi - what needs to be done to get this merged?  Would really like to have this for the analysis facility at Nebraska.

Thanks!

wyang007:
at the minimum, Andy's back from vacation :-)

abh3:
As part of fixing the macaroon stuff could you also look at ticket

https://github.com/xrootd/xrootd/issues/1811

it seems like a related item and likely an easy fix.

Issue #1804, opened 2022-10-19
Xrootd 5.5.1 failed to build when -DFORCE_ENABLED=ON -DXRDCL_ONLY=TRUE is used


smuzaffar:
Hi,
  I am trying to build xrootd 5.5.1 client for CMS and it failed with error [a] . Note that the cmake options we use to build includes
```
-DFORCE_ENABLED=ON -DXRDCL_ONLY=TRUE
```

I think https://github.com/xrootd/xrootd/blob/master/cmake/XRootDDefaults.cmake#L31-L32 should be set to `FALSE` for `XRDCL_ONLY=TRUE`

[a]
https://cmssdt.cern.ch/SDT/jenkins-artifacts/pull-request-integration/PR-f4ea1a/28381/external/xrootd/5.5.1-85b58a07b6557565b78005c924a2a88a/log

```
-- Found KERBEROS5: /usr/lib64/libkrb5.so;/usr/lib64/libcom_err.so  
-- Looking for curl_multi_wait
-- Looking for curl_multi_wait - found
CMake Error at /cvmfs/cms-ib.cern.ch/nweek-02755/el8_amd64_gcc10/external/cmake/3.18.2-9266e9e57a6fac9dc658f5e66a697c43/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find Macaroons (missing: MACAROONS_INCLUDES MACAROONS_LIB)
Call Stack (most recent call first):
  /cvmfs/cms-ib.cern.ch/nweek-02755/el8_amd64_gcc10/external/cmake/3.18.2-9266e9e57a6fac9dc658f5e66a697c43/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindMacaroons.cmake:20 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/XRootDFindLibs.cmake:124 (find_package)
  CMakeLists.txt:31 (include)
```

amadio:
Hi @smuzaffar, can you please try the patch from the pull request above? Then we can add it to 5.5.2 (to be released next month).

smuzaffar:
thanks @amadio , I confirm that  #1805 works

amadio:
Closing as #1805 has been merged.

PR #1805, opened 2022-10-20
[CMake] Make switches for server-only options dependent on XRCL_ONLY=FALSE

amadio:
Fixes: ff67c2a0, Closes: #1804

### DIFF
--- a/cmake/XRootDDefaults.cmake
+++ b/cmake/XRootDDefaults.cmake
@@ -9,6 +9,8 @@ if( "${CMAKE_BUILD_TYPE}" STREQUAL "" )
   endif()
 endif()
 
+include( CMakeDependentOption )
+
 define_default( PLUGIN_VERSION    5 )
 option( ENABLE_FUSE      "Enable the fuse filesystem driver if possible."                 TRUE )
 option( ENABLE_CRYPTO    "Enable the OpenSSL cryprography support."                       TRUE )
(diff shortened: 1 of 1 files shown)


amadio:
@simonmichal I guess at least `ENABLE_VOMS` could be turned into a dependent option as well. Which other options are client-only?

simonmichal:
@amadio : thanks a lot for the PR :-)

I think, it is worth investigating first why did we run into the problem described in #1804 in the first place, if I read the code correctly, if the `XRDCL_ONLY` is set to true Macaroons and SciTokens plug-ins should have been excluded from the build:
https://github.com/xrootd/xrootd/blob/6ef78195d2279a2e462f2aaae831859822e2a6f9/src/CMakeLists.txt#L52-L88

or do I miss something?

amadio:
The lookup for SciTokens and Macaroons happens in [cmake/XRootDFindLibs.cmake](https://github.com/xrootd/xrootd/blob/master/cmake/XRootDFindLibs.cmake#L122-L157) and that's not protected with `if(NOT XRDCL_ONLY)`. That's why I said maybe moving the calls to `find_package(...)` into the subdirectories where they are needed could make sense. It's an alternative solution to what I'm proposing here, but may not be as nice if multiple subdirectories depend on the same packages, as those would be needed at top level anyway.

simonmichal (approved):

simonmichal:
@amadio : thanks for the explanation, it all makes sense now!

Issue #1810, opened 2022-10-26
Python3 bindings segfault on PyXRootD::FileSystem::Prepare()


Einscrew:
I faced this problem when porting python2 code we use on EOS to issue transferes from and into tape. I was also able to reproduce it on a forked repo https://github.com/Einscrew/xrootd/commit/b125c81ce2f99bd3e334075a25b04af16180aae8
```
$ gdb --args python3 -c 'from XRootD import client; from XRootD.client.flags import PrepareFlags; fs=client.FileSystem("root://localhost:1094"); print(fs); fs.prepare(["/eos/cms/opstest/mbatalha/ola"], PrepareFlags.STAGE)'
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-120.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /usr/bin/python3.6...Reading symbols from /usr/bin/python3.6...(no debugging symbols found)...done.
(no debugging symbols found)...done.
Missing separate debuginfos, use: debuginfo-install python3-3.6.8-18.el7.x86_64
(gdb) r
Starting program: /usr/bin/python3 -c from\ XRootD\ import\ client\;\ from\ XRootD.client.flags\ import\ PrepareFlags\;\ fs=client.FileSystem\(\"root://localhost:1094\"\)\;\ print\(fs\)\;\ fs.prepare\(\[\"/eos/cms/opstest/mbatalha/ola\"\],\ PrepareFlags.STAGE\)
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
<XRootD.client.filesystem.FileSystem object at 0x7ffff7e77f98>

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff0469364 in PyXRootD::FileSystem::Prepare (self=0x7ffff7e4abf0, args=0x7fffeb274f98, kwds=0x0) at /usr/src/debug/xrootd/xrootd/bindings/python/src/PyXRootDFileSystem.cc:622
[...]

Issue #1588, opened 2022-01-11
Codify support for Python 2 and Python 3 in tests


matthewfeickert:
Hi. While trying to run tests for PR #1585 I realized that the tests for the Python bindings are written to be Python 2 only (as of 98d83fb4b70159f31c3311ce9d1d217874ccef45). Doing some quick checks with a rough pass of 

```
2to3 --write .
```

also show that the examples and tests all contain Python 2 only code.

* The `bindings/python/examples/` can be easily fixed by just properly applying the `print()` function.
* The `bindings/python/tests/` are mostly also just need `print()` functions but also have some string encoding differences and a breaking `xrange` vs. `range` change

https://github.com/xrootd/xrootd/blob/98d83fb4b70159f31c3311ce9d1d217874ccef45/bindings/python/tests/test_threads.py#L31

* I see that there has been some work and care put into ensuring that the Python 2 only code in `bindings/python/libs/` is properly guarded, so I think that's fine as is. :+1: 

To be able to run tests I'm willing to contribute a PR that updates the example and tests code to Python 3, but is there a desire to continue to run the tests for Python 2? I don't know how long the XRootD team plans to support Python 2 as it is now EOL for over a year and there isn't any clear statements in the docs about LTS given that the supported OS part of the docs

https://github.com/xrootd/xrootd/blob/98d83fb4b70159f31c3311ce9d1d217874ccef45/README#L12-L19

was [added 10 years ago in 2011](https://github.com/xrootd/xrootd/commit/e5a2ad11456b18fa24b04f483bbeef481635488c) and I don't think is accurate anymore.

Also, if I have this correct, is the https://gitlab.cern.ch/dss/xrootd GitLab mirror used to run the CI. Do you have instructions on how to have contributed PRs run the CI _before_ they get merged so that PRs can link a passing test suite?

simonmichal:
@matthewfeickert : sorry for the late reply, there's no need to keep the tests compatible with Python2, if you could submit a PR that ports them to Python3 it would be great. (Unfortunately we are not able at the moment to run the CI in gitlab for PR.)

amadio:
Since we will drop support for Python 2.x in XRootD 6.0.0, I'm marking this with that milestone, but will try to port the tests to Python 3.x and add them to run in our CI earlier than that (and mark the correct milestone fixing this later on).

PR #1818, opened 2022-11-03
[Python] Avoid crash and fix module method name (issue #1810)

smithdh:


bindings/python/src/PyXRootDFileSystem.cc
```diff
@@ -608,15 +608,15 @@ namespace PyXRootD
...
+    uint16_t                   flagval  = 0;
     uint8_t                    priority = 0;
     uint16_t                   timeout  = 0;
     PyObject                  *pyfiles = NULL, *callback = NULL;
     PyObject                  *pyresponse = NULL, *pystatus = NULL;
     XrdCl::XRootDStatus        status;
 
-    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "OK|bHO:prepare",
-         (char**) kwlist, &pyfiles, &flags, &priority, &timeout, &callback ) )
+    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "OH|bHO:prepare",
```
amadio, line 618: This is the only line mentioned as the fix in #1810. Just out of curiosity, why are the other changes necessary?

smithdh: The difference with respect to my suggestion in #1810 was to change the storage type of the variable used to store the flags-value, that PyArg_ParseTupleAndKeywords() fills, to be at least 16bits. Then the value is static_cast to an XrdCl::PrepareFlags::Flags (an enum). ParseTupleAndKeywords() needs storage large enough to be consistent with its conversion string (which was the cause of the crash of #1810) and I thought the storage size of PrepareFlags::Flags wasn't obvious:

The storage used for the enum should be an integer type large enough to hold the enumerated values. Looking at the enum definition the largest value would be 'Evict', which needs 9 bits, so the storage is infact going to be at least 16 bits. Also both gcc and clang (that I tried) use at least 32 bits even for small enums. But I thought this was all quite some reasoning for each person that reads this source, and it would be clearer to make an explicit conversion. However there are other places at PyXRootDFileSystem.cc that use ParseTupleAndKeywords to store directly to an enum and I'm not proposing to change them with no other reason.

So if you'd suggest to change it back to the shorter version, possibly with a comment, I'd be very fine with that! (I'm not going to merge this PR, at least not before next week, that will also let Michal have a look too).

While preparing this I've seen some other issues in the python bindings (python ref count leaks & the file line iteration, which doesn't work in python3 and some things like attribute name have to be given as byte strings despite messages which say they have to be strings); so I would like to open some more PRs for those soon. While testing I've also been using the CI tests with python3, with some conversion, and porting the tests to python3 was also the suggestion of issue #1588, so perhaps I could also address that issue by changing the tests. (But best to keep all these as separate PRs I think).

PR #1822, opened 2022-11-08
Show install command being run by pip and fail when pip install fails

amadio:
My previous debugging of the problem was wrong, staged installations (with `DESTDIR` set don't work with the old patch, it only worked when I was using a branch where the version of XRootD was different than what I had installed, so pip didn't complain).

Issue: #1768

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -83,20 +83,29 @@ if ( NOT ${VALID_PIP_EXIT_CODE} EQUAL 0 )
   install(
     CODE
     "EXECUTE_PROCESS(
+      COMMAND_ECHO STDOUT
+      RESULT_VARIABLE INSTALL_STATUS
       COMMAND /usr/bin/env ${XROOTD_PYBUILD_ENV} ${PYTHON_EXECUTABLE} ${SETUP_PY} install \
                 --verbose \
                 --prefix \$ENV{DESTDIR}/${CMAKE_INSTALL_PREFIX} \
                 ${DEB_INSTALL_ARGS}
-    )"
-  )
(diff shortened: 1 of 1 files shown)


amadio:
The "joy" of using CMake...
```
CMake Error at bindings/python/cmake_install.cmake:41 (EXECUTE_PROCESS):
  EXECUTE_PROCESS given unknown argument "COMMAND_ERROR_IS_FATAL".
```
I will check which version adds that and do it conditionally...

simonmichal:
@matthewfeickert : could you please have a look and share your opinion? ;-)

amadio:
@simonmichal Can you please merge this? I'd like to see the commands in the CI so I can try to debug locally. This only shows the command being issued to install the Python bindings and returns non-zero for a make install that fails to avoid masking the problem that a build failure of the bindings doesn't fail the top build.

matthewfeickert (commented):
While I don't really see the need for this (you could just use debug flags instead) there's nothing wrong with it, so sure if this is helpful then go for it.

Not personally a fan of taking commands that span multiple lines and smushing them into 2 lines instead of clearly breaking them out as I had them before, but that's stylistic stuff that doesn't need to hold up a PR.

amadio:
Adding `--verbose` still does not show the command that was invoked (i.e. `python3 -m pip ...`, which is what I'd like to see to be able to run the same command out of the CI environment to reproduce the problem. The style is not so important, if @simonmichal also prefers the old formatting I can revert that part. However, failing the install if pip fails needs to go in to avoid silent failures.

simonmichal:
@amadio : thanks for the PR, looks good!

PR #1823, opened 2022-11-08
CMake build system updates

amadio:
These are mostly fixes for warnings with newer versions of CMake. I'm also picking the fix for #1804 to apply onto master.


cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-# Check for the TLS support in the OpenSSL version that is available
-# (assume available if use of static libs is detected and the openssl version
-# is at least 1.0.2)
-#-------------------------------------------------------------------------------
-
-if ( OPENSSL_FOUND )
-
-  if( OPENSSL_USE_STATIC AND OPENSSL_GOOD_VERSION )
-
-    add_definitions( -DHAVE_TLS -DHAVE_TLS12 -DHAVE_TLS11 -DHAVE_TLS1 -DHAVE_DH_PADDED -DHAVE_DH_PADDED_FUNC )
```
simonmichal, line 79: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-    set( CMAKE_REQUIRED_LIBRARIES ${OPENSSL_LIBRARIES} )
-    set( CMAKE_REQUIRED_QUIET FALSE)
-
-    check_function_exists(TLS_method HAVE_TLS_FUNC)
-    check_symbol_exists(
-            TLS_method
-            ${OPENSSL_INCLUDE_DIR}/openssl/ssl.h
-            HAVE_TLS_SYMB)
-    if( HAVE_TLS_FUNC AND HAVE_TLS_SYMB )
-        add_definitions( -DHAVE_TLS )
```
simonmichal, line 92: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-    endif()
-
-    check_function_exists(TLSv1_2_method HAVE_TLS12_FUNC)
-    check_symbol_exists(
-            TLSv1_2_method
-            ${OPENSSL_INCLUDE_DIR}/openssl/ssl.h
-            HAVE_TLS12_SYMB)
-    if( HAVE_TLS12_FUNC AND HAVE_TLS12_SYMB )
-        add_definitions( -DHAVE_TLS12 )
```
simonmichal, line 101: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-    endif()
-
-    check_function_exists(TLSv1_1_method HAVE_TLS11_FUNC)
-    check_symbol_exists(
-            TLSv1_1_method
-            ${OPENSSL_INCLUDE_DIR}/openssl/ssl.h
-            HAVE_TLS11_SYMB)
-    if( HAVE_TLS11_FUNC AND HAVE_TLS11_SYMB )
-        add_definitions( -DHAVE_TLS11 )
```
simonmichal, line 110: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-    endif()
-
-    check_function_exists(TLSv1_method HAVE_TLS1_FUNC)
-    check_symbol_exists(
-            TLSv1_method
-            ${OPENSSL_INCLUDE_DIR}/openssl/ssl.h
-            HAVE_TLS1_SYMB)
-    if( HAVE_TLS1_FUNC AND HAVE_TLS1_SYMB )
-        add_definitions( -DHAVE_TLS1 )
```
simonmichal, line 119: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-    endif()
-
-    check_function_exists(DH_compute_key_padded HAVE_DH_PADDED_FUNC)
-    check_symbol_exists(
-            DH_compute_key_padded
-            ${OPENSSL_INCLUDE_DIR}/openssl/dh.h
-        HAVE_DH_PADDED_SYMB)
-    if( HAVE_DH_PADDED_FUNC)
-       if( HAVE_DH_PADDED_SYMB )
-         add_definitions( -DHAVE_DH_PADDED )
```
simonmichal, line 129: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-       else()
-         add_definitions( -DHAVE_DH_PADDED_FUNC )
```
simonmichal, line 131: I'm not sure if removing this file is the right thing to do, those definitions are used in the source code.

cmake/FindOpenSSL.cmake
amadio: I grepped for them, and most are not used. The one that was used I added back (`-DHAVE_DH_PADDED`).

cmake/XRootDFindLibs.cmake
```diff
@@ -39,11 +39,12 @@ find_package( CURL )
...
 if( ENABLE_CRYPTO )
   if( FORCE_ENABLED )
-    find_package( OpenSSL REQUIRED )
+    find_package( OpenSSL 1.0.2 REQUIRED )
   else()
-    find_package( OpenSSL )
+    find_package( OpenSSL 1.0.2 )
   endif()
   if( OPENSSL_FOUND )
+    add_definitions( -DHAVE_DH_PADDED )
```
amadio, line 47: This is where I added back the definitions that was set in the OpenSSL module we had.

cmake/FindOpenSSL.cmake
```diff
@@ -1,136 +0,0 @@
...
-      ${OPENSSL_ROOT_DIR}
-      PATH_SUFFIXES
-      ${LIBRARY_PATH_PREFIX}
-      ${LIB_SEARCH_OPTIONS})
-
-    string(FIND ${OPENSSL_SSL_LIBRARY} ".a" hasstaticext)
-    if (NOT ${hasstaticext} EQUAL -1)
-      if ( ${opensslvers} LESS "0x1000201fL" )
-        set( OPENSSL_GOOD_VERSION FALSE )
-        message(WARNING " >>> Cannot build XRootD crypto modules: static openssl build version is < 1.0.2")
```
amadio, line 36: This check was the reason for me to add the minimum version to `find_package(OpenSSL)`.

cmake/FindOpenSSL.cmake
amadio: I think when CMake provides a find module, it's better to use the upstream one, unless the version of CMake we support doesn't provide it. Usually find modules should not change compile options of the projects using them, but just export which compile options may be needed as variables, etc. If we need to `add_definitions(...)` for things specific to XRootD, then we should do outside of the find module, like was already being done in `cmake/XRootDFindLibs.cmake`. At least for `-DHAVE_TLS`, the definition added by the find module was redundant.

cmake/XRootDFindLibs.cmake
```diff
@@ -39,11 +39,12 @@ find_package( CURL )
 
 if( ENABLE_CRYPTO )
   if( FORCE_ENABLED )
-    find_package( OpenSSL REQUIRED )
+    find_package( OpenSSL 1.0.2 REQUIRED )
```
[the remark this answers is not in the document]
simonmichal: OK, understood, I though it requires that exact version, thanks for clarifying!

cmake/FindOpenSSL.cmake
simonmichal: from what I see `HAVE_TLS` is being used in XrdTls/XrdTlsContext.cc

cmake/XRootDFindLibs.cmake
simonmichal: Could please double check if this works well with statically linked openssl, this is something @gganis added in fea67e7844246e18ecdaa1505c9c57040cf1f29f

amadio: Sure, I will test against a statically-linked openssl.

cmake/FindOpenSSL.cmake
amadio: Ah, indeed, `cmake/XRootDFindLibs.cmake` has `HAVE_SSL`, not `HAVE_TLS`. I mistook one for the other, too many Ss and Ls. I will add also `HAVE_TLS`, I did see that when I grepped.

amadio: This is addressed in my last commit.

cmake/XRootDFindLibs.cmake
amadio: I tested against a statically linked OpenSSL locally on my machine, and also on a server where I built ROOT using the builtin OpenSSL and builtin XRootD (which also uses a statically linked OpenSSL). Everything still works fine.

amadio:
Turned out that removing the `#ifdef` altogether was a bit too optimistic, so I replaced it with a versioned `#ifdef` instead to keep it working with both shared and static libraries of OpenSSL.

@simonmichal We *could* remove the `#ifdef` and keep using only the older `SSLv23_method` name, as that's a macro that defines it as `TLS_method` anyway in newer OpenSSL (and why we didn't see any problems when not adding `-DHAVE_TLS` before. Let me know if you prefer this, I did the `#ifdef` to ensure the current version is future-proof. We can drop the `#ifdef` when we move to require a newer version of OpenSSL (i.e. when CentOS 7 support is dropped).

simonmichal (approved):

simonmichal:
Look's good, again thanks for the PR :-)

PR #1824, opened 2022-11-09
[XrdCl] Ensure URL::GetChannelId returns an Id which can be parsed again as a url

smithdh:
A Channel's ChannelInfo contains an instance of a SIDManager which is a reference counted instance keyed by a channel id. SIDMgrPool::GetSIDMgr() accepts a URL, from which it calculates the channel id and returns the appropriate instance. Currently GetSIDMgr is passed the channel id directly when setting up the ChannelInfo. There is a problem that some URLs produce a channel id which cannot be re-parsed to a URL and back to the channel id again.

e.g. the url

root://machine:1194//big5?xrd.k5ccname=/foo

has channel id:

root://machine:1194?xrd.k5ccname=/foo

but instantiating a URL with the above and requesting its channel id gives

://

(Some channel id can be parsed in this way, but e.g. those with special opaque key-values can not). This patch changes the value passed to GetSIDMgr to the url rather than the derived channel id. That is the same way that the SIDMgr instance held by XRootDMsgHandler handlers obtain it. Possibly this PR could be enhanced to change URL so that it is able to parse the channel id and give back the same channel id.

The effect of the current situation is that for affected URLs the SIDMgr may be destroyed while the channel is open but there are no outstanding requests (i.e. handlers). If there is a single request, which times-out and another request is then sent there is a chance that the stream id of the first request can be re-used. If the sid is reused and the server then responds to the first request, the result is interpreted as belonging to the second request. It is possible that this causes problems, or a failure, or maybe it is interepred as a valid result of the second call, giving bad data but an Ok status.

### DIFF
--- a/src/XrdCl/XrdClURL.cc
+++ b/src/XrdCl/XrdClURL.cc
@@ -493,7 +493,7 @@ namespace XrdCl
   //------------------------------------------------------------------------
   std::string URL::GetChannelId() const
   {
-    std::string ret = pProtocol + "://" + pHostId;
+    std::string ret = pProtocol + "://" + pHostId + "/";
     bool hascgi = false;
 
     std::string keys[] = { "xrdcl.intent",


simonmichal:
@smithdh : thanks a lot for the PR. Your observation is absolutely correct, that said if you use the whole URL as the key it will make the client open a new connection for every single file, which would kill the performance.

Instead of using the full url as the key, I would change the `GetChannelID` method and replace:

https://github.com/xrootd/xrootd/blob/f00acdc7559b5189583945b4c27e06dc2637df3e/src/XrdCl/XrdClURL.cc#L496

with 

```
std::string ret = pProtocol + "://" + pHostId + "/";
```

I think this should work, maybe you could give it a try?

smithdh:
hi @simonmichal: Thanks for reviewing this. The original patch didn't intend to replace the key with the whole URL, but to avoid a double use of GetChannelId(), i.e. effectively the value URL(url.GetChannelId()).GetChannelId(), so I don't believe there should have been any problem introduced by switching to overly specific keys. But for sure another approach could be preferred. I tried the change you suggested which is fixing the parsing, of course that seems fine and I it solves the problem as I understand it. I've updated this PR so that it adds that change; so feel free to either eventually merge this pr, or close this and add your change directly.

abh3:
The fixes will appear in 5.5.2-RC2 this week and the actual release is planned for next week. These are he fixes that @smithdh referred to.

PR #1850, opened 2022-12-05
Pick commits from master and apply to stable-5.5.x

amadio:
Excluded commits 2630fe1 and 8a6d7c0 which are new features.

@simonmichal, @abh3 Please let me know if there's anything else that should be removed (i.e. things that should wait for a 5.6 release). Given the number of commits, it may make sense to go for a feature release instead of a patch release, what do you think? In that case, for 5.6, I would like to try to merge the histories with `git merge` such that the tag for v5.6.0 is actually in the linear history of the master branch, then `stable-5.6.x` branches off of there. This is to make `git describe` give more reasonable descriptions of commits on the `master` branch (try `git describe` on `master` and `stable-5.5.x`). When I rebase master on the branch of this PR, I get the output below (compare with what current `github/master` shows instead).
```
epsftws xrootd $ git describe master
v5.5.1-50-gf23a89ba9
epsftws xrootd $ git describe github/master
v5.2.0-99-osghotfix-TlsErrorHandling-947-g31a7f6c05
```


abh3:
I think the easiest way to review this is to see what commit will *not* be included as there should be much fewer of those. As far as this list goes, nothing stands out as should not be there.

amadio:
@abh3 This is what @simonmichal suggested as well, I listed the two excluded commits in the description of the pull request (and I also excluded later some commits only touching release notes, which I will merge/fix before tagging).

simonmichal:
@amadio : thanks a lot for the PR! Please add following 2 commits and it will be ready to merge:
0a5c36b1b0c0dcae6288f43b91ef546a4945d61d [[CMake] Make switches for server-only options dependent on XRCL_ONLY=](https://github.com/xrootd/xrootd/commit/0a5c36b1b0c0dcae6288f43b91ef546a4945d61d)FALSE
7410d53917c0c822533f352570ad2aad1b9533be [XrdCeph] Re-sync accidentally reverted sub-module.

simonmichal:
all have been clarified offline :-) ready to merge :-)

amadio:
Yes, just for the record, commit 0a5c36b is already in the stable branch (was there first, then cherry-picked into master), and the other commit was only needed on master, as the accidental revert was only committed there.

PR #1854, opened 2022-12-07
[XrdCl] Refactor kXR_status response handling.

simonmichal:
Move kXR_status response handling to `AsyncMsgReader`.

### DIFF
--- a/src/XrdCl/XrdClAsyncMsgReader.hh
+++ b/src/XrdCl/XrdClAsyncMsgReader.hh
@@ -135,6 +135,7 @@ namespace XrdCl
                 readstage = ReadRawData;
                 continue;
               }
+
               //----------------------------------------------------------------
               // The next step is to read the message body
               //----------------------------------------------------------------
@@ -168,6 +169,23 @@ namespace XrdCl
               continue;
             }
             //------------------------------------------------------------------
+            // kXR_status is special as it can have both body and raw data,
+            // handle it separately
+            //------------------------------------------------------------------
+            case ReadMore:
+            {
+              XRootDStatus st = xrdTransport.GetMore( *inmsg, &socket );
+              if( !st.IsOK() || st.code == suRetry )
+                return st;
+              inmsgsize = inmsg->GetCursor();
+
+              //----------------------------------------------------------------
+              // The next step is to finalize the read
+              //----------------------------------------------------------------
+              readstage = ReadDone;
+              continue;
+            }
+            //------------------------------------------------------------------
             // We need to call a raw message handler to get the data from the
             // socket
             //------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


amadio:
Should this be rebased/merged soon? Is this for 5.5.3 or for 5.6?

simonmichal:
@amadio : actually now would be the perfect time to merge it, could you have a look and do a quick review?

src/XrdCl/XrdClAsyncMsgReader.hh
```diff
@@ -168,6 +169,23 @@ namespace XrdCl
...
+            {
+              XRootDStatus st = xrdTransport.GetMore( *inmsg, &socket );
+              if( !st.IsOK() || st.code == suRetry )
+                return st;
+              inmsgsize = inmsg->GetCursor();
+
+              //----------------------------------------------------------------
+              // The next step is to finalize the read
+              //----------------------------------------------------------------
+              readstage = ReadDone;
```
amadio, line 185: Can `ReadMore` be required more than once?

src/XrdCl/XrdClXRootDTransport.cc
```diff
@@ -1334,53 +1376,64 @@ namespace XrdCl
...
-    cse->cseCRC = ntohl( cse->cseCRC );
-    size_t length = rsp->status.bdy.dlen - sizeof( uint32_t );
-    void*  buffer = msg.GetBuffer( sizeof( ServerResponseV2 ) + sizeof( uint32_t ) );
-    uint32_t crcval = XrdOucCRC::Calc32C( buffer, length );
-
-    //--------------------------------------------------------------------------
-    // Do the integrity checks
-    //--------------------------------------------------------------------------
-    if( crcval != cse->cseCRC )
+    switch( reqType )
```
amadio, line 1384: Why a switch with a single case inside? Do you intend this to be expanded with more cases later on?

src/XrdCl/XrdClAsyncMsgReader.hh
simonmichal: yes, certainly

src/XrdCl/XrdClXRootDTransport.cc
```diff
@@ -388,6 +413,23 @@ namespace XrdCl
...
+    Log *log = DefaultEnv::GetLog();
+    XRootDStatus st = XRootDTransport::UnMarchalStatusMore( message );
+    if( !st.IsOK() && st.code == errDataError )
+    {
+      log->Error( XRootDTransportMsg, "[msg: 0x%x] %s", &message,
+                  st.GetErrorMessage().c_str() );
+      return st;
+    }
+
+    if( !st.IsOK() )
```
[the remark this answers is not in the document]
simonmichal: well, the first one is a special case where we have corrupted data, the other one handles all other errors, the only difference is the format of the error message (I wanted to preserve the original log message structure)

src/XrdCl/XrdClXRootDTransport.cc
simonmichal: yes, exactly! right now `kXR_status` is used only for 'pgread'/'pgwrite' but in the future it could be used for other requests hence the switch statement.

src/XrdCl/XrdClAsyncMsgReader.hh
amadio: What I wonder is if setting `readstage = ReadDone` will not prevent `ReadMore` from being called twice in a row if needed.

src/XrdCl/XrdClXRootDTransport.cc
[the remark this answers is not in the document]
amadio: Sure, but you can have an `if ( !st.IsOK() )`, and inside it have another if (or switch) to treat differently the different kinds of errors. But I guess this is not so performance critical, so it's not the end of the world if we check twice.

PR #1859, opened 2022-12-13
Update XrdSciTokensAccess.cc

biozit:


src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -1083,6 +1083,16 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper
                      section.c_str());
                 continue;
             }
+	    // prevent two identical issuers break the config 
+	    for (size_t i = 0; i < issuer.length(); i++)
```
abh3, line 1087: Why didn't you use index() and instead implemented a manual version of it?

abh3 (commented):
A question for you along with a suggested improvement.

src/XrdSciTokens/XrdSciTokensAccess.cc
biozit: How about that?

abh3: Well, I was thinking of something much simpler. But before we get to that, I am wondering what is being fixed here. According to INIReader, the new line character ('\n') is considered a separator so there should never be a new line character in the value string. So, why is it that the code assumes otherwise?

Second, if indeed a new line can be in the value string then there is no easy solution here as the new line character can appear anywhere and multiple times as well. So, consider "\nissuer", "issuer\n", or "\nissuer\n" the fixes are very different in each case. The variations are endless (well actually two basic cases).

OK, simplicity (note we only handle the simple case here):
size_t i = issuer.find('\n');
if (i != std::string::npos) {
    if (i != 0)  issuer = issuer.substr(0, i);
       else /*error* string starts with newline and we don't handle that */
}

Yes, I should have have mentioned index() -- old habits.

abh3 (commented):
So, now I am a bit confused as what we are trying to fix. More comments inside.

biozit:
If we have a duplicated entry, something like this:

[Issuer https://chtc.cs.wisc.edu/icecube]
issuer = https://chtc.cs.wisc.edu/icecube
base_path = /icecube

[Issuer https://chtc.cs.wisc.edu/icecube]
issuer = https://chtc.cs.wisc.edu/icecube
base_path = /icecube

The issuer and the base path became:
221213 04:51:12 596604 scitokens_Reconfig: Configuring issuer
https://chtc.cs.wisc.edu/icecube
https://chtc.cs.wisc.edu/icecube

With that, the token can find the issuer.

Maybe a good approach is to use rindex.


*--Fábio Andrijauskas*

abh3:
Ah, so what you are saying is that since INIReader simply adds new values to the key:value pair by catenating them with a new line character. A rather poor design. Anyway, if you want the last value then use find_last_of(). The biggest problem here is whether the package accepts null values If it does then you will need to deal with multiple new lines. Frankly, I think that consider it an error if a null value is found (i.e. if the '/n' is in the first or last position). Anyway, my simple example would still hold.

djw8605:
Also, it’s not only the base path and issuer values that may be duplicated. All possible values under an issuer could be concatonated in a duplicate entry. I think we need a more general solution.

PR #1868, opened 2022-12-15
Provide a mechanism to override the detected hostname by environment …

### LINKED ISSUE #1855: Explicitly override the hostname used by XRootD
Right now, the hostname that will be used by an XRootD daemon is bootstrapped here:

https://github.com/xrootd/xrootd/blob/master/src/Xrd/XrdConfig.cc#L574-L579

Basically, it selects an IP address and does a reverse DNS resolution to determine the hostname.  This hostname can't be overridden as far as I can tell.

We've run into two cases where an override is necessary:

1. Where the forward DNS resolution (name -> IP address) configured as desired but it's essentially impossible to setup the reverse DNS.  This is because the university that owns the project (and the DNS zone for the project) is different than the university that owns the IPs.
2. When running inside Kubernetes, the visible device IP is distinct from the incoming IP address meaning neither forward nor reverse DNS lookups result in the correct value.  We need to explicitly set the hostname instead of having it autodetected.

bbockelm:
…variable.

Note the `myName` variable needed to be changed to a string as it must own the memory; the memory pointed to by `getenv` my change after future `setenv` calls.

Fixes #1855

### DIFF
--- a/src/Xrd/XrdConfig.hh
+++ b/src/Xrd/XrdConfig.hh
@@ -30,6 +30,7 @@
 /******************************************************************************/
 
 #include <vector>
+#include <string>
 
 #include "Xrd/XrdProtLoad.hh"
 #include "Xrd/XrdProtocol.hh"
@@ -94,7 +95,7 @@ static const char  *TraceID;
 XrdNetSecurity     *Police;
 XrdTcpMonInfo      *tmoInfo;
 const char         *myProg;
-const char         *myName;
+std::string         myName;
 const char         *myDomain;
 const char         *mySitName;
 const char         *myInsName;
--- a/src/Xrd/XrdConfig.cc
+++ b/src/Xrd/XrdConfig.cc
@@ -572,11 +572,19 @@ int XrdConfig::Configure(int argc, char **argv)
 // run in v4 mode and that doesn't get set until after the options are scanned.
 //
    static XrdNetAddr *myIPAddr = new XrdNetAddr((int)0);
-   if (!(myName = myIPAddr->Name(0, &temp))) myName = "";
+   auto envName = getenv("XRDHOST");
+   const char *ipAddrName;
+   if (envName) {
+      myName = envName;
+   } else if ((ipAddrName = myIPAddr->Name(0, &temp))) {
+      myName = ipAddrName;
+   } else {
+      myName = "";
+   }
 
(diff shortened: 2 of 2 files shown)


src/Xrd/XrdConfig.hh
```diff
@@ -94,7 +95,7 @@ static const char  *TraceID;
 XrdNetSecurity     *Police;
 XrdTcpMonInfo      *tmoInfo;
 const char         *myProg;
-const char         *myName;
+std::string         myName;
```
amadio, line 98: It does not seem to make sense to me to replace one `const char*` with `std::string`, while leaving all the rest as `const char*`.

amadio (changes requested):
I wonder if something that uses [`gethostname(2)`](https://man7.org/linux/man-pages/man2/gethostname.2.html) couldn't work in this case, rather than setting an environment variable. You could probably use `hostnamectl` from within your containers to adjust the hostname, no?

amadio:
Should not be a problem on macOS either: [gethostname(3)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/gethostname.3.html).

src/Xrd/XrdConfig.hh
bbockelm: I would be more than happy to clean it all up, I just wasn't sure if a surgical approach was preferred.

bbockelm:
> You could probably use hostnamectl from within your containers to adjust the hostname, no?

Sadly, this is not an option for unprivileged users and in some container runtimes (for example, Docker has a default syscall filter that eliminates the ability to generate a user namespace where the user has capabilities to unshare the UTS namesapce).  Otherwise, I'd already utilize the trick suggested in #1855.

abh3:
Gentleman,

This is not the way I envisioned doing this and definitely not via an 
envar that tends to get forgotten buruied and creates enormous problems. 
We are up against holiday brtek. This is a new feature and won't be 
included until a feature release anyway. So, please hold your horses and 
stop rushing.

Andy

abh3:
Once again, I want to urge you to understand that the server has a very defined way of identifying itself and I will not tolerate hacks here. So, please be patient. Thus, I am closing this pull request as itr does not correspond to the architecture.

Issue #1874, opened 2022-12-19
TLS SessionTicket enabled after certificate refresh, gfal davs writes failing


jthiltges:
After upgrading from v5.5.0 to v5.5.1, we're seeing our EL7 local redirector stop accepting GFAL davs writes after 8 hours. I suspect the issue is associated with the changes in #1796 (or other recent changes to certificate refresh). After the redirector xrootd daemon has run 8 hours, the TLS v1.2 server hello includes "SessionTicket TLS", and it provides session tickets to clients. But if a client attempts to (re)start a TLS session using the ticket, the redirector immediately closes the connection. Before the refresh interval, the redirector did not report TLS session ticket support. A restart of the xrootd service returns it to normal, until the next 8h interval has passed.

Local redirector config comments:
- `xrootd.tls capable all -data`
- `xrd.tlsca noverify`
- `http.tlsreuse` is not set, using the default of `off`
- RPMs
    - `xrootd-5.5.1-1.4.osg36.el7.x86_64`
    - `openssl-1.0.2k-25.el7_9.x86_64`
- No cluster filesystem mounted. `cms.dfs lookup distrib redirect immed`
- Macaroons enabled

davs write to show issue on an affected redirector:
```
gfal-copy source davs://local-redirector.example.edu:1094/target
```

<details>
<summary>gfal-copy log excerpt showing failure</summary>

```
2022-12-18T11:15:56.092990 DEBUG Davix: NEON start internal request
2022-12-18T11:15:56.093083 DEBUG Davix: cached ne_session found ! taken from cache
2022-12-18T11:15:56.093165 DEBUG Davix: configure session...
2022-12-18T11:15:56.103282 DEBUG Davix: define connection timeout to 30
2022-12-18T11:15:56.103361 DEBUG Davix: define operation timeout to 3600
2022-12-18T11:15:56.103415 DEBUG Davix: add CA PATH /etc/grid-security/certificates/
2022-12-18T11:15:56.103464 DEBUG Davix: disable login/password authentication
2022-12-18T11:15:56.103616 DEBUG Davix: enable client cert authentication by callback
2022-12-18T11:15:56.103696 DEBUG Davix: Disable Session recycling
2022-12-18T11:15:56.103750 DEBUG Davix: Running pre_send hooks
2022-12-18T11:15:56.103836 DEBUG Davix: > PUT /store/user/jthiltge/test.1671383724.975305632 HTTP/1.1
[...]

PR #1875, opened 2022-12-21
XrdTlsContext: Recreate session cache in the Clone() method

ccaffy:
This fixes the issue #1874


ccaffy:
Hi @abh3 ,

Here is the fix. I tested it both with HTTP and xroots and it works fine.

Cheers,
Cedric

abh3 (approved):
Sorry for the late review but we were off the day before this was posted -- 2.5 weeks of bliss :-) Anyway, it looks good so please merge it and thanks for doing the fix!

abh3:
OK, I decided to merge it myself as it was a no brainer!

ccaffy:
Thanks @abh3 and happy new year!

PR #1884, opened 2023-01-20
[CI] Move from Fedora 35 to Fedora 37

amadio:
Fedora 35 repos are returning an HTTP 404 code. Making a PR just so CI runs.


amadio:
This change is now in `stable-5.5.x`. I will merge into `master` after 5.5.2 is released.

Issue #1707, opened 2022-05-12
500 Internal Server Error on unknown checksum algorithm


paulmillar:
I can request an adler32 checksum of a file:
```console
paul@sprocket:~$ curl -I -H "Want-Digest: adler32" -H "Authorization: Bearer $TOKEN" https://xrootd-hv1-vanderbilt.sites.opensciencegrid.org:1094/store/test/bad-zero-file.root
HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Length: 0
Digest: adler32=f08fab4f

paul@sprocket:~$ 
```

I can also ask for an MD5 checksum:
```console
paul@sprocket:~$ curl -I -H "Want-Digest: md5" -H "Authorization: Bearer $TOKEN" https://xrootd-hv1-vanderbilt.sites.opensciencegrid.org:1094/store/test/bad-zero-file.root
HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Length: 0
Digest: md5=niK3oCxW39lw3BDVMDfimw==

paul@sprocket:~$ 
```

However, if I ask for a SHA checksum, the server replies with a 500 error:
```console
paul@sprocket:~$ curl -I -H "Want-Digest: sha" -H "Authorization: Bearer $TOKEN" https://xrootd-hv1-vanderbilt.sites.opensciencegrid.org:1094/store/test/bad-zero-file.root
HTTP/1.1 500 Internal Server Error
Connection: Close
Content-Length: 0

paul@sprocket:~$ 
```

The server should ignore checksum algorithms it does not understand.

abh3:
The reason you cannot ask for a SHA checksum (other than it is ambiguous) is that none of the SHA checksums are natively supported. As whether or not the server should ignore the request, it certainly will not do so for xroot protocol as that is the defined behavior. As for http, I see no advantage in ignoring there either.

paulmillar:
Hi Andy,

Thanks for the quick reply.

When opening this issue, I perhaps didn't give sufficient information to fully understand what's happening, so let me provide more details.

First, the label `sha` is not ambiguous.  It's defined in RFC 3230 as the SHA-1 algorithm from FIPS 180.

Second, unlike in the xroot protocol, HTTP 3230 allows the client to specify a list of algorithms, partially ordered by "priority".  Here's a couple of examples:
  * `Want-Digest: adler32, md5` 
    Here, the client would like either an adler32 or MD5 checksum (or both), with no priority given to either algorithm.
[...]

PR #1890, opened 2023-01-30
XrdHttp checksum handling logic change

ccaffy:
Solves Issue #1707


src/XrdHttp/XrdHttpReq.hh
```diff
@@ -116,6 +116,14 @@ private:
   
   // Sanitize the resource from http[s]://[host]/ questionable prefix
   void sanitizeResourcePfx();
+
+   /**
+   * Select the checksum to be computed depending on the userDigest passed in parameter
+   * @param userDigest the digest request from the user (extracted from the Want-Digest header)
+   * @param selectedChecksum the checksum that will be performed
+   */
+  void selectChecksum(const std::string & userDigest, std::string & selectedChecksum);
```
amadio, line 125: Is this a private class in XRootD? I wonder if adding this won't require users to recompile (meaning that it should go into 5.6).

abh3: I think this is OK as the method aded is non-virtual and the whole set of classes is in one library which gets replaced in any release update. So, it should not require recompilation of anything that was created by a user.

docs/PreReleaseNotes.txt
```diff
@@ -31,6 +31,8 @@ Prerelease Notes
   **Commit: 0f79a38
   **[XrdMacaroons] Fix authentication when different tokens are used in the same TCP session**
   **Commit: 4410d56
+  **[XrdHttp] 500 Internal Server Error on unknown checksum algorithm**
+  **Commit: 8c9fbb3
```
abh3, line 35: Out of curiosity, how do you know the final commit hash?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1032,6 +1050,42 @@ void XrdHttpReq::mapXrdErrorToHttpStatus() {
...
 
+/**
+ * Select the checksum to be computed depending on the userDigest passed in parameter
+ * @param userDigest the digest request from the user (extracted from the Want-Digest header)
+ * @param selectedChecksum the checksum that will be performed
+ */
+void XrdHttpReq::selectChecksum(const std::string &userDigest, std::string & selectedChecksum) {
+    char * configChecksumList;
+    selectedChecksum = "unknown";
+    if((configChecksumList = getenv("XRD_CSLIST"))) {
```
abh3, line 1061: Not that there is any significant performance penalty here but this transformation of XRD_CSLIST should really be done only once.

src/XrdCl/XrdClUtils.hh
```diff
@@ -31,6 +31,7 @@
 #include "XrdCl/XrdClPostMaster.hh"
 #include "XrdCl/XrdClXRootDTransport.hh"
 #include "XrdNet/XrdNetUtils.hh"
+#include "XrdUtils/XrdUtils.hh"
```
abh3, line 34: I see an introduction of a whole new utils scheme here. In general, utility functions should go into one of the established locations. In this case XrdOucUttils.hh and just add a static method to the class to avoid a namespace implementation. That, of course, leads to the next issue is that we would (in either case) have to implementation of this utility method -- one in XrdOucUtils (formerly XrdUtils) and one in XrdClUtils. The consensus is that the best solution is to convert usage of XrdClUtils::plitString to use the one in XrdOucUtils -- messy but fortunately it would only involve name changes. There are other alternatives here like adding a method to XrdOucString that does what splitstring does as there is already a similar method there called tokenize().

abh3 (changes requested):

docs/PreReleaseNotes.txt
ccaffy: That's the one from my repo, I did not know whether I should have left it blank or if I should have put mine as a placeholder... I can remove it if you want

src/XrdCl/XrdClUtils.hh
ccaffy: Ah yes, so this I doubted when I did it... I did not know the existence of the XrdOucUtils.
I wanted the code to not be duplicated. I can remove my class and put it into the XrdOucUtils if you want.
For the XrdOucString, is there a wish that one day, this class is replaced by the std::string ? If yes, I believe we can keep the implementation of the `splitString(...)` method in XrdOucUtils :man_shrugging:

docs/PreReleaseNotes.txt
```diff
@@ -31,6 +31,8 @@ Prerelease Notes
   **Commit: 0f79a38
   **[XrdMacaroons] Fix authentication when different tokens are used in the same TCP session**
   **Commit: 4410d56
+  **[XrdHttp] 500 Internal Server Error on unknown checksum algorithm**
```
abh3, line 34: That really is not a meaningful summary of the changes. How about "Follow RFC checksum requirements in XRootD http".

src/XrdCl/XrdClUtils.hh
```diff
@@ -56,25 +57,7 @@ namespace XrdCl
...
-          else
-            length = input.length() - start;
-
-          if( length )
-            result.push_back( input.substr( start, length ) );
-
-          start = end + delimiter.size();
-        }
-        while( end != std::string::npos );
+        XrdOucUtils::splitString(result,input,delimiter);
```
abh3, line 60: Likely add a comment providing some history on this with a mention that at some point it would be good to collapse this change (i.e., eliminate the middle man call).

src/XrdOuc/XrdOucUtils.hh
```diff
@@ -130,7 +131,36 @@ static bool PidFile(XrdSysError &eDest, const char *path);
 
 static int getModificationTime(const char * path, time_t & modificationTime);
 
-       XrdOucUtils() {}
-      ~XrdOucUtils() {}
+    //------------------------------------------------------------------------
```
abh3, line 134: It appears (I may be wrong) that this was inlined in the "hh" file. Could you move the actual implementation to the "cc" file to keep the "hh" file reasonably clean.

abh3 (changes requested):
Looks good, some nit-picking changes will seal the deal!

abh3 (approved):
Looks OK. Thanks!

PR #1897, opened 2023-02-07
core: Update genversion.sh script work as expected in --print-only mo…

esindril:
…de with user given version

The behavior of the genversion.sh script does not work as advertised for the --print-only flag. Namely when passing the USER_VERSION parameter this is not taken into consideration. As it's the case when used from this location:
https://github.com/xrootd/xrootd/blob/master/packaging/makesrpm.sh#L137

This pull request fixes this issue and makes the general behavior more consistent with the expectations.


genversion.sh
```diff
@@ -79,33 +79,34 @@ function getVersionFromLog()
...
-    shift
-  else
-    SOURCEPATH=${1}
+    SOURCEPATH=${3}
   fi
-  shift
-done
+fi
+
+echo "user version: ${USER_VERSION} and sourcepath: ${SOURCEPATH}"
```
amadio, line 109: Is this statement meant to be in the final version?

esindril:
No ... fixed! :)

esindril:
This is already fixed in master. All works fine!

Issue #1901, opened 2023-02-08
WebDav: Invalid server checksum answer (5.5.2)


kreczko:
Dear all,

I just tried version 5.5.2 today, which caused the CMS WebDav tests to fail.

Their test command is
```
gfal-sum -D "HTTP PLUGIN:RETRIEVE_BEARER_TOKEN=false" -t 90 davs://xrootd.phy.bris.ac.uk:1094/xrootd/cms/store/mc/SAM/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v1/00000/CE860B10-5D76-E711-BCA8-FA163EAA761A.root ADLER32
```
which returns 
```
davs://xrootd.phy.bris.ac.uk:1094/xrootd/cms/store/mc/SAM/GenericTTbar/AODSIM/CMSSW_9_2_6_91X_mcRun1_realistic_v2-v1/00000/CE860B10-5D76-E711-BCA8-FA163EAA761A.root 96fb2295
```
for version 5.5.0

and
```
gfal-sum error: 5 (Input/output error) - Invalid server checksum answer
```
for version 5.5.2

### our setup
We use a script to get the checksum: ` xrootd.chksum max 100 adler32 /etc/xrootd/xrdsum.sh`, which content:
```bash
#!/usr/bin/env bash

RESULT=$(xrdsum --verbose --debug -l /var/log/xrootd/clustered/checksum.log get --store-result --read-size 128 "$1")
ECODE=$?

# XRootD expects return on stdout - checksum followed by a new line
printf "%s\n" "$RESULT"
exit "$ECODE"
```

`RESULT` is just the checksum, `96fb2295` in this case.
The script is run successfully in both 5.5.0 and 5.5.2, but the response the client gets is different.

EDIT:
Debug info on gfal-sum show the checksum as unknown
```
DEBUG    Davix: < HTTP/1.1 200 OK
DEBUG    Davix: < Connection: Keep-Alive
DEBUG    Davix: < Server: XrootD/v5.5.2
DEBUG    Davix: < Content-Length: 1293057986
DEBUG    Davix: < Digest: unknown=96fb2295
DEBUG    Davix: <
DEBUG    Davix: End of headers.
```

Relates to #1707

ccaffy:
Hi Luke,

Thanks a lot for reporting this.
Can you please tell me what is the request sent by gfal to ask for the checksum ? (The content of the `Want-Digest:` header)

Thanks again,

[...]

PR #1904, opened 2023-02-10
XrdHttp: Supports non-natively supported checksum in the case a user provides a checksum computation program

ccaffy:
Solves issue #1901


abh3:
Looks good. Presumably this was tested with a script and without.

Issue #1898, opened 2023-02-07
Segfault on a disk server, in XrdXrootdMonFile::GetSlot


smithdh:
Hello,

A segfault with the following stack trace was recently reported on an EOS disk server node at cern (running custom xrootd build, and with EOS plugins, approximately equivalent to xrootd 5.5.2):

```
Core was generated by `/opt/eos/xrootd/bin/xrootd -n fst -c /etc/xrd.cf.fst -l /var/log/eos/xrdlog.fst'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f8a36b66a4b in XrdXrootdMonFile::GetSlot (slotSZ=<optimized out>) at /usr/include/bits/byteswap.h:47
[...]
(gdb) where
#0  0x00007f8a36b66a4b in XrdXrootdMonFile::GetSlot (slotSZ=<optimized out>) at /usr/include/bits/byteswap.h:47
#1  0x00007f8a36b66bd3 in XrdXrootdMonFile::Close (fsP=<optimized out>, isDisc=<optimized out>)
    at /usr/src/debug/xrootd-5.5.5/src/XrdXrootd/XrdXrootdMonFile.cc:169
#2  0x00007f8a36b6409b in XrdXrootdFileTable::Del (this=0x7f82a27db520, monP=<optimized out>, 
    fnum=<optimized out>, fnum@entry=0, dodel=<optimized out>)
    at /usr/src/debug/xrootd-5.5.5/src/XrdXrootd/XrdXrootdFile.cc:309
#3  0x00007f8a36b79e88 in XrdXrootdProtocol::do_Close (this=this@entry=0x7f8792088500)
    at /usr/src/debug/xrootd-5.5.5/src/XrdXrootd/XrdXrootdMonitor.hh:192
#4  0x00007f8a36b6fffe in XrdXrootdProtocol::Process2 (this=0x7f8792088500)
    at /usr/src/debug/xrootd-5.5.5/src/XrdXrootd/XrdXrootdProtocol.cc:492
#5  0x00007f8a368af8c0 in XrdLinkXeq::DoIt (this=<optimized out>)
    at /usr/src/debug/xrootd-5.5.5/src/Xrd/XrdLinkXeq.cc:320
#6  XrdLinkXeq::DoIt (this=0x7f8a0f8d0c50) at /usr/src/debug/xrootd-5.5.5/src/Xrd/XrdLinkXeq.cc:308
[...]

PR #1907, opened 2023-02-14
[Server] Avoid a race condition during deferred file close

smithdh:
This is a change to address #1898; similar to the proposed fix attached in that issue discussion.


abh3 (approved):
I couldn't find any loopholes here, so I'd say we go with it.

PR #1908, opened 2023-02-14
Support for musl libc and uclibc

amadio:
Based on work from and replaces pull request #1632.

### DIFF
--- a/src/XrdSys/XrdSysPthread.hh
+++ b/src/XrdSys/XrdSysPthread.hh
@@ -351,7 +351,7 @@ enum PrefType {prefWR=1};
 
         XrdSysRWLock(PrefType ptype)
                     {
-#ifdef __linux__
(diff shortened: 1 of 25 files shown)


src/XrdSys/XrdSysPthread.hh
```diff
@@ -351,7 +351,7 @@ enum PrefType {prefWR=1};
 
         XrdSysRWLock(PrefType ptype)
                     {
-#ifdef __linux__
+#if defined(__linux__) && !defined(MUSL)
```
amadio, line 354: This may actually be a problem, since it's on a header. I may use an alternative for glibc and uclibc that does not need the value of the `MUSL` macro.

amadio: Fixed by using `__GLIBC__` and `__UCLIBC__` defines instead.

PR #1906, opened 2023-02-14
fix: Use CMake 3.12+ FindPython to get Python components

matthewfeickert:
Resolves #1474

* Require CMake 3.16 at minimum to allow use of modern find_package from CMake 3.12+ as well as fixes in FindPython for future planned work.
   - c.f. https://cmake.org/cmake/help/v3.16/command/find_package.html
* Set range of valid CMake versions to allow for more modern CMakes to be used if available.
   - c.f. https://cliutils.gitlab.io/modern-cmake/chapters/basics.html
* Use FindPython from CMake 3.12+ which relies on find_package advancements to select the components that are required. Select the 'Interpreter' and 'Development' components to find the required executables and files.
   - c.f. https://cmake.org/cmake/help/v3.16/module/FindPython.html
* Modern FindPython is capitalization sensitive and will set the following variables that are used throughout the project.
  * `Python_FOUND`
  * `Python_Interpreter_FOUND`
  * `Python_EXECUTABLE`
  * `Python_Development_FOUND`
   - c.f. https://cmake.org/cmake/help/v3.16/module/FindPython.html#result-variables

<details>
<summary>Demonstration of fixing Issue 1474:</summary>

```console
$ docker run --rm -ti -u root matthewfeickert/pyenv-virtualenv-conda:pyenv-v2.3.12       
root@8f5f7b99e0cf:~/data# pyenv virtualenv 3.11.1 issue-1474 && pyenv activate issue-1474
(issue-1474) root@8f5f7b99e0cf:~/data# apt update && apt install -y cmake python3-dev
(issue-1474) root@8f5f7b99e0cf:~/data# python -m pip --quiet install --upgrade pip setuptools wheel
(issue-1474) root@8f5f7b99e0cf:~/data# git clone --recursive https://github.com/matthewfeickert/xrootd.git && cd xrootd
(issue-1474) root@8f5f7b99e0cf:~/data/xrootd# git checkout origin/fix/use-modern-find-package -b fix/use-modern-find-package
(issue-1474) root@8f5f7b99e0cf:~/data/xrootd# cp packaging/wheel/* .
(issue-1474) root@8f5f7b99e0cf:~/data/xrootd# ./publish.sh
(issue-1474) root@8f5f7b99e0cf:~/data/xrootd# python -m pip install ./dist/xrootd-*.tar.gz
(issue-1474) root@8f5f7b99e0cf:~/data/xrootd# python -m pip list
Package         Version
--------------- -------------
build           0.10.0
packaging       23.0
[...]

PR #1909, opened 2023-02-14
fix: Avoid 'may be used uninitialized' warning

matthewfeickert:
* Initialize all variables to avoid a potential 'may be used uninitialized' warning that will become an error with `-flto=auto` is used.

This is needed to for PR #1906.


matthewfeickert:
@simonmichal @amadio this is ready for review.

PR #1910, opened 2023-02-15
[XrdSciTokens] Add global option 'validation = none' to disable token…

apeters1971:
… validation during ZTN handshakes

### DIFF
--- a/src/XrdSciTokens/README.md
+++ b/src/XrdSciTokens/README.md
@@ -88,6 +88,10 @@ Within the `Global` section, the available attributes are:
      group or issuer information from the token.  The username is only populated if either scope-based mapping or
      the mapfile-based approach is successful.
 
+    - `validation` (optional): when the library is used for ZTN authentication the Validate() function is called as part of the
+      ZTN handshake. To disable validating tokens during the ZTN handshake this variable can be set to the following value:
+        - `none`: Don't validate a token during the ZTN handshake. This is useful when the passed token is not handled by the SciToken library!
+
 Each section name specifying a new issuer *MUST* be prefixed with `Issuer`.  Known attributes
 are:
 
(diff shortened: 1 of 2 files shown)


src/XrdSciTokens/README.md
```diff
@@ -88,6 +88,10 @@ Within the `Global` section, the available attributes are:
      group or issuer information from the token.  The username is only populated if either scope-based mapping or
      the mapfile-based approach is successful.
 
+    - `validation` (optional): when the library is used for ZTN authentication the Validate() function is called as part of the
+      ZTN handshake. To disable validating tokens during the ZTN handshake this variable can be set to the following value:
+        - `none`: Don't validate a token during the ZTN handshake. This is useful when the passed token is not handled by the SciToken library!
+
```
abh3, line 94: I think you will need to fix the indentation here as markdown is not formatting this correctly and it really is not well readable.

abh3 (changes requested):

abh3:
I asked for a change in the markdown but I do also want to bring to your attention that the approach taken is is likely not the best, Turning off validation is specific to ztn usage not to SciToken usage in any sense of the word. This is as much as you said. So, this change does not belong in the SciToken libraries which likely would not have been needed at all if validation is not desired. Additionally, it prevents coming up with say a ytn protocol that does validation and have a ztn protocol without it. Once validation is turned off it is turned off for everyone; which limits future options.

I strongly suggest that the ztn parameter "-tokenlib" which accepts a parameter also accept "-tokenlib=none" which automatically says you do not wish to use the tokenlib and this would turnoff validation. The changes are slightly more complicated but not enough to avoid this approach.

So, I guess what I am saying I would prefer the choice to validate or not to be done in XrdSecProtocolztn.cc as that is the logical place to do it; unless there is another reason that I don't know about.

apeters1971:
Ah yes, you are absolutely right. If only ZTN is calling the Validate function, it is much better to do it in ZTN, I was not sure if that is the case. Thanks, will do another one!

PR #1912, opened 2023-02-17
Allow XRootD to return trailers indicating failure

bbockelm:
HTTP provides a response to include a "trailer" in addition to the better-known "header".  If a user sets the following headers in the request:

```
X-Transfer-Status: true
TE: trailers
Transfer-Encoding: chunked
```

Then the response will used chunked encoding and indicate, on the last returned chunk, whether an error has occurred.

Clients aware of these headers can now receive an error message from XRootD if there's an IO error in the middle of the response.  This is expected to be useful in XCache use cases where failure mid-response is somewhat more common.

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -52,7 +52,7 @@ int XrdHttpExtReq::StartChunkedResp(int code, const char *desc, const char *head
 {
   if (!prot) return -1;
 
-  return prot->StartChunkedResp(code, desc, header_to_add, true);
+  return prot->StartChunkedResp(code, desc, header_to_add, -1, true);
 }
 
 int XrdHttpExtReq::ChunkResp(const char *body, long long bodylen)
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1545,7 +1545,7 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc, const char *hea
 /*                      S t a r t C h u n k e d R e s p                       */
 /******************************************************************************/
   
-int XrdHttpProtocol::StartChunkedResp(int code, const char *desc, const char *header_to_add, bool keepalive) {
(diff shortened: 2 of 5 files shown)


bbockelm:
@djw8605 -- the intent is the `stashcp` / `stash_plugin` client should be able to use this information to build better information for XCache requests.

ccaffy:
Hi @bbockelm ,

Thanks for this :)

I compiled everything and it works in the case the transfer succeeds. Out of curiosity, how do you manage to test that functionality ?

I tried to fail the transfer but I get no information from the body that it failed...

bbockelm:
@ccaffy - I test with this small patch to inject a fault:
```
--- a/src/XrdOss/XrdOssApi.cc
+++ b/src/XrdOss/XrdOssApi.cc
@@ -884,8 +884,13 @@ ssize_t XrdOssFile::Read(void *buff, off_t offset, size_t blen)
            else   retval = cxobj->Read((char *)buff, blen, offset);
         else 
 #endif
+     if (offset < 64*1024) {
              do { retval = pread(fd, buff, blen, offset); }
                 while(retval < 0 && errno == EINTR);
+     } else {
+        errno = EIO;
+        retval = -1;
+     }
 
      return (retval >= 0 ? retval : (ssize_t)-errno);
 }
```

Any reads after 64KB will fail (though, since reads are issued as 1MB chunks, the test file needs to be >1MB).

Note that you'll need to add `--raw -v` to the `curl` command to see the trailer.  Sadly trailers aren't printed by `-v` by default.

bbockelm:
> Is it normal that I get a 200 OK response though ?

Yeah -- this is a typical issue in HTTP where the HTTP status has to be sent before the body.  So, if there's any I/O error, it's too late to change the response status code.

Currently, when an error happens mid-transfer, the server simply closes the TCP connection and the client reports something like "unexpected EOF".  With this patch, if you ask for a status trailer, the client can get a useful error message (you might notice other PRs to improve error messages -- it's all related!).

This has to be opt-in, however, to prevent unexpecting clients from interpreting the early exit as a success as the response no longer generates TCP-level problems.

ccaffy:
Alright, all things are clear now :)
Thanks for your detailed explanations that allowed me to learn ;)
Approving this PR...

ccaffy (approved):

PR #1913, opened 2023-02-18
Add an error object to XrdOss.

bbockelm:
While XrdOss has a simple POSIX-like interface, in reality the underlying implementation can be quite complicated -- XrdPfc, for example.  Accordingly, we should allow for a richer error message interface between the OSS and its callers.  This adds a new virtual function which the SFS interface can invoke to get additional error information.

The intent is this will be used by XrdPfc and allow clients to differentiate between "the cache gave you a permission denied" and "the origin gave the cache a permission denied".

This uses the existing "features" interface so we can prevent invoking the new API for a class that's compiled against the old API.

*A note on ABI compatibility*: Technically, this is an ABI change and hence would need to wait until XRootD 6 to be deployed.  However, this takes advantage of a quirk of GCC: the ordering of the virtual functions are preserved.  Hence, the new functions are placed out of alphabetical order so as to not screw up the virtual function table offsets for existing code.  So, I think this actually _won't_ break ABI and hence could go in for 5.6.0.  Would like to hear other's opinions though!

*Second note on ABI compatibility*: If we decide we like the above logic, in XRootD 6 we could add a few placeholder virtual functions that simply hold an empty slot in the vtable, allowing for adding new ABI calls during the 6 series for GCC-based platforms without worrying about ABI breaks.


abh3:
Unfortunately, other compilers (i.e. non-gcc) are used and we have no 
control over that. Plus, I wouldn't feel confortable on relying on a 
"quirk" as this can change at any time. So, I think it's best to wait 
until 6.0.0.

Andy

bbockelm:
The ABI is compiler-specific so if a different compiler is used between the server and the plugin, then it should already crash.  So we don't need to worry about someone using MSVC with GCC or somesuch.

Anyhow, I just pushed a new approach.  This causes the function to be virtual if the major number is greater than 5 (note: in unreleased builds, the major number gets set to 100).  That keeps the ABI for the remainder of XRootD 5 but allows one to test out the new functionality in unreleased builds.

src/XrdOss/XrdOss.hh
```diff
@@ -447,6 +451,23 @@ const char     *getTID() {return tident;}
...
+//! @return Error object associated with the last error.  Caller does not own
+//! the returned memory.  Lifetime of the returned object is only guaranteed
+//! until the next method call.  May return nullptr even if an error occurred.
+//-----------------------------------------------------------------------------
+
+#if XRDOSS_ERR
+virtual const XrdOucErrInfo *getError() const;
+#else
+const XrdOucErrInfo *getError() const;
+#endif
```
amadio, line 470: I am not a fan of changing interface in this manner. I think it's better to have a clean patch and add it only in the next major version.

bbockelm: Is it clear when the next major version is expected?  The level of ugly hacks would likely be different if it's immediately after 5.6 versus waiting until 2024...

xrootd-dev:
It is likely to come this year.

bbockelm:
So I'm going to close this one out for now and continue to work on the other PR (the one with a few examples) so we can work on how such chained error messages should be formatted.

PR #1915, opened 2023-02-18
Example of proposed XrdOss advanced error interface usage

bbockelm:
This PR builds on #1913 to demonstrate how it might be used.  With this, the PSS returns the following when the cache is unable to authenticate with the origin:

```
[ERROR] Server responded with an error: [3030] Unable to open /user/ligo/test_access/access_ligo; invalid exchange (Failed to open file in cache: invalid exchange)
```

Note that it generates a lousy message ("invalid exchange") because EAUTH maps to a EBADE on Linux.  I'll fix that in a follow-up PR.

With this approach, we could continue digging through the `XrdPosix` layer to generate a better error message there as well.  However, I just wanted to illustrate what the new XrdOss API usage would look like.

@osschar


abh3 (commented):
I think the idea here is good, though somewhat limited as we can only get extended information from file objects not from generic methods. I suppose that's better than not but not completely satisfying. Of course, all of this is historic so at the moment it seems the best that can be done. All that said, in our meeting today we decided that this change really needs to go into R6 as it breaks ABI. Fortunately, that allows simplification because the whole XRDOSS_HASAERR checking can be eliminated. It's expected that the "extended information" is either implemented or returns nothing. So, code should simply check for extended error information. We are also worried on what the composite message would look like. Can you give some examples? If it becomes largely unreadable (think composite OpenSSL messages) this this patch would be of limited value.  I know there is some concern about when R6 will be cut but it looks very likely to be in 2023. In the mean time this does open up the possibility to perhaps address the non-file based methods that return nothing more than an errno type message to make this feature complete.

abh3:
Just an update. We've accepted the whole concept but the particular pull request here does not address the underlying problem. Most of the issues come from the POSIX interface which has practically no smarts when reporting errors. We already have an architectural design in place and will be implementing this as quickly as possible. Suffice it to say, your example was very helpful in spurring this activity as well as saving us much time. I'll close this pull request when we have the complete changes in place.

bbockelm:
Hi Andy!

Just to help clean things up -- I'll go ahead and close out this PR.  Will work to revise it once the new error propagation framework is available.

Brian

PR #1916, opened 2023-02-18
Override error message for EAUTH

bbockelm:
On Linux platforms, the "authentication denied" error is mapped to the EBADE errno (as EAUTH doesn't exist).  That results in authentication errors in XrdPss to generate the "invalid exchange" error message (this is because XrdPss takes the errno from XrdPosix and XrdPosix returns EBADE/EAUTH).

Rather than start tinkering with alternate errno's for this case, simply map EBADE's error message to "authentication denied".

Before:

```
Server responded with an error: [3030] Unable to open /user/ligo/test_access/access_ligo; invalid exchange
```

After:

```
Server responded with an error: [3030] Unable to open /user/ligo/test_access/access_ligo; authentication denied
```

This misleading error message was found when working on #1915

### DIFF
--- a/src/XrdSys/XrdSysE2T.cc
+++ b/src/XrdSys/XrdSysE2T.cc
@@ -63,6 +63,12 @@ int initErrTable()
            lastGood = i;
           }
       }
+   // NOTE: On systems without EAUTH (authentication error; currently all Glibc systems but GNU Hurd),
+   // EAUTH is remapped to EBADE ('invalid exchange').  Given there's no current XRootD use of a
+   // syscall that can return EBADE, we assume EBADE really means authentication denied.
+#if defined(EBADE)
+   Errno2String[EBADE] = "authentication failed - possible invalid exchange";
+#endif
 
 // Supply generic message for missing ones
 //


src/XrdSys/XrdSysE2T.cc
```diff
@@ -55,7 +58,15 @@ int initErrTable()
...
    for(int i = 1; i <errSlots; i++)
-      {eTxt = strerror(ERRNOBASE + i);
+      {// NOTE: On systems without EAUTH (authentication error; currently all systems but GNU Hurd),
+       // EAUTH is remapped to EBADE ('invalid exchange').  Given there's no current XRootD use of a
+       // syscall that can return EBADE, we assume EBADE really means authentication denied.
+#if defined(EBADE)
+       if (i == EBADE)
+          {eTxt = authenticationFailed;
+       } else
+#endif
```
amadio, line 68: Considering that no one uses GNU Hurd, it seems to me that these `#ifdefs` are not desirable. Also, rather than use an `if` inside a loop, it's better to just overwrite the message for `EBADE` after the `for` loop in my opinion, with something like:

```cpp
e2sMap[EBADE] = "Authentication failed";
```

amadio (changes requested):

src/XrdSys/XrdSysE2T.cc
amadio: In any case, remapping the error message is just a cosmetic change. I'll wait for @abh3's opinion before merging anything.

src/XrdSys/XrdSysE2T.cc
```diff
@@ -63,6 +63,12 @@ int initErrTable()
            lastGood = i;
           }
       }
+   // NOTE: On systems without EAUTH (authentication error; currently all Glibc systems but GNU Hurd),
+   // EAUTH is remapped to EBADE ('invalid exchange').  Given there's no current XRootD use of a
+   // syscall that can return EBADE, we assume EBADE really means authentication denied.
+#if defined(EBADE)
+   Errno2String[EBADE] = "authentication failed";
+#endif
```
amadio, line 71: Another minor note. Is the assumption that `EBADE` == authentication failed reasonable? Can something else be mistaken for authentication failure because of this? 

Checking a bit, I also figured out that macOS has `EAUTH` but not `EBADE`, and on Linux musl doesn't seem to have either. I wonder what happens in that case.

src/XrdSys/XrdSysE2T.cc
bbockelm: Hoisted the code, also simplified the patch itself.

src/XrdSys/XrdSysE2T.cc
bbockelm: (Indeed, MacOS lacks `EBADE`, that's why I had to do the `#if` macro)

> Is the assumption that EBADE == authentication failed reasonable? Can something else be mistaken for authentication failure because of this?

I think it's reasonable but not necessarily bulletproof.  The EBADE response is pretty esoteric and I couldn't spot any syscalls used that would trigger it on MacOS.

Can I guarantee that this remains true in the future?  Of course not.  My feeling is this would lead to an improvement in 99.9% of cases, however.

abh3:
In our meeting today we saw that there were cases where EBADE could be returned and cause confusion. The solution was to reword the message. We suggest:
Invalid exchange - possible authentication failure
At least that tell us it was EBADE but more likely it's an auth problem. Sort of addresses both concerns. If you change it we will approve it :-) Of course, feel free to present counter arguments.

abh3:
OK, it's a deal.  "Authentication failure - possible invalid exchange" works as well.

bbockelm:
Ok, cleared up the error message as requested.  To keep the git history simple for such a small patch, I rebased and squashed as well.

PR #1917, opened 2023-02-18
Map authentication failure to HTTP 401

bbockelm:
The authentication failure error message was previously mapped to HTTP 500 (internal server error).  401 Unauthorized (despite its name) is what HTTP servers typically utilize for authentication problems.


bbockelm:
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401 for more information about the 401 status.

ccaffy (approved):
Looks good, thanks !

PR #1919, opened 2023-02-18
Report cache object age for caching proxy mode

bbockelm:
When in caching proxy mode, have the HTTP response populate the standardized `Age` field so the client knows the approximate age of the object in cache.

@osschar

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2288,18 +2288,16 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
               
               // Always try to parse response.  In the case of a caching proxy, the open
               // will have created the file in cache
-              filectime = 0;
               if (iovP[1].iov_len > 1) {
                 TRACEI(REQ, "Stat for GET " << resource.c_str()
                          << " stat=" << (char *) iovP[1].iov_base);
 
                 long dummyl;
-                sscanf((const char *) iovP[1].iov_base, "%ld %lld %ld %ld %ld",
+                sscanf((const char *) iovP[1].iov_base, "%ld %lld %ld %ld",
                       &dummyl,
                       &filesize,
                       &fileflags,
-                      &filemodtime,
-                      &filectime);
+                      &filemodtime);
 
                 // As above: if the client specified a response size, we use that.
                 // Otherwise, utilize the filesize
@@ -2320,7 +2318,7 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
                   if (!responseHeader.empty()) {
                     responseHeader += "\r\n";
                   }
-                  long object_age = time(NULL) - filectime;
+                  long object_age = time(NULL) - filemodtime;
                   responseHeader += std::string("Age: ") + std::to_string(object_age < 0 ? 0 : object_age);
               }
 
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2286,36 +2286,51 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
 
               getfhandle();
               
-              // Now parse the stat info if we still don't have it
-              if (filesize == 0) {
-                if (iovP[1].iov_len > 1) {
-                  TRACEI(REQ, "Stat for GET " << resource.c_str() 
-                           << " stat=" << (char *) iovP[1].iov_base);
-              
-                  long dummyl;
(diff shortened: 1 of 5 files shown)


bbockelm:
@osschar - can you review this one as well?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2272,33 +2272,48 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
...
+              else {
+                TRACEI(ALL, "GET returned no STAT information. Internal error?");
+              }
+
+              std::string responseHeader;
+              if (!m_digest_header.empty()) {
+                responseHeader = m_digest_header;
+              }
+              long one;
+              if (filemodtime && XrdOucEnv::Import("XRDPFC", one)) {
```
osschar, line 2305: Are you sure you mean Import(...), not GetInt("XRDPFC")?

osschar (commented):
Maybe someone else should check the XrdHttpReq stuff.

src/XrdPfc/XrdPfcIOFile.cc
```diff
@@ -115,6 +115,11 @@ int IOFile::initCachedStat(const char* path)
    {
       res = GetInput()->Fstat(tmpStat);
       TRACEIO(Debug, trace_pfx << "got stat from client res = " << res << ", size = " << tmpStat.st_size);
+      // The mtime / atime / ctime for cached responses come from the file on disk in the cache hit case.
+      // To avoid weirdness when two subsequent stat queries can give wildly divergent times (one from the
+      // origin, one from the cache), set the times to "now" so we effectively only report the *time as the
+      // cache service sees it.
+      tmpStat.st_ctime = tmpStat.st_mtime = tmpStat.st_atime = time(NULL);
```
osschar, line 122: I missed this is for the fail case, res != 0 ... git snipped the diff in change view.
This makes my comments above orthogonal to this change -- the change here is ok.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Yeah, I think this is right as it matches the `Export`.  There's no obvious global to pull from.

osschar: Yes, right ... I missed that Export puts stuff into process environment.

osschar (approved):

src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -274,6 +274,9 @@ bool Cache::Config(const char *config_filename, const char *parameters)
    const char *theINS = getenv("XRDINSTANCE");
    m_isClient = (theINS != 0 && strncmp("*client ", theINS, 8) == 0);
 
+   // Tell everyone else we are a caching prooxy
```
amadio, line 277: There's a typo here: prooxy

src/XrdPfc/XrdPfcIOFile.cc
```diff
@@ -83,7 +83,7 @@ int IOFile::initCachedStat(const char* path)
    int res = -1;
    struct stat tmpStat;
 
-   if (m_cache.GetOss()->Stat(path, &tmpStat) == XrdOssOK)
+   if (m_cache.GetOss()->Stat(GetFilename().c_str(), &tmpStat) == XrdOssOK)
```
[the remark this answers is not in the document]
osschar: We don't have the time of the last write to data-file -- it could be added to the cinfo file. Technically one could divine it from the access records in cinfo but those get merged at some point and become less and less precise time-wise.

src/XrdPfc/XrdPfcConfiguration.cc
bbockelm: Sigh.  I am "blessed" with a MacBook butterfly keyboard so it enjoys inserting double or triple characters.

Fixed the typo, rebased to avoid polluting history for a typo.

src/XrdPfc/XrdPfcIOFile.cc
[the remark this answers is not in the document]
bbockelm: > We don't have the time of the last write to data-file -- it could be added to the cinfo file

Nah, I don't think this is necessary.  I'm happy using what the OS itself is reporting as the last write to the file.

abh3:
This pull request has been slowed due to conflicts in previous patches. Since it has been approved you can withdraw it and rebase it on the existing code in src/XrdHttp/XrdHttpReq.cc. and resubmit without a re-review (assuming your honest :-) Your choice but it will be faster if you resolve the conflicts.

bbockelm:
Ok, rebased to clear up conflict.  No functional changes.

PR #1920, opened 2023-02-19
XrdHttp: Fix byte range requests

bbockelm:
This PR (a) advertises the server supports byte range requests and (b) fixes the byte range request when resuming a download.

This allows libraries which can resume interrupted partial downloads to function with the xrootd server.


ccaffy (approved):
Yes, everything works :) Thanks !

PR #1921, opened 2023-02-20
[XrdSecztn] Add security library option to configure token validation…

apeters1971:
Add **ZTN** security library option to configure token validation policy using new configuration entry:
`
-validation {required|optional|ignore}'
`
The default policy is `required` and does not need to be specified.

Example 1: don't validate tokens at all
`sec.protocol ztn -validation ignored`

Example 2: validate token but don't fail if validation failed
`sec.protocol ztn -validation optional`

### DIFF
--- a/src/XrdSecztn/XrdSecProtocolztn.cc
+++ b/src/XrdSecztn/XrdSecProtocolztn.cc
@@ -107,6 +107,7 @@ inline uint64_t monotonic_time() {
 /******************************************************************************/
   
 int expiry = 1;
+bool tokenlib = true;
 }
 
 /******************************************************************************/
@@ -631,8 +632,9 @@ int XrdSecProtocolztn::Authenticate(XrdSecCredentials *cred,
 //
    std::string msgRC;
    long long   eTime;
+   bool validated = false;
    if (Entity.name) {free(Entity.name); Entity.name = 0;}
-   if (sthP->Validate(tResp->tkn, msgRC, (expiry ? &eTime : 0), &Entity))
+   if (tokenlib && sthP->Validate(tResp->tkn, msgRC, (expiry ? &eTime : 0), &Entity))
       {if (expiry)
           {if (eTime < 0 && expiry > 0)
               {Fatal(erp, "'ztn' token expiry missing", EINVAL, false);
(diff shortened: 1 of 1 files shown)


bbockelm:
@apeters1971 - would it be possible to also specify the policy outside of a flag?

```
sec.protocol ztn -validation ignored
```

does not compose across multiple configuration files while

```
sec.protocol ztn
ztn.validation ignored
```

does (as the second line could be part of a separate drop-in config file in an RPM).

apeters1971:
Sure, I probably should than also change the other option in ZTN to support multiline? (-expire) ... 
Will do.

apeters1971:
Hi Brian, I checked. i cannot do this so easily, because the plug-ins are only called with the parameters on the line of the security library. This is  something which has to be changed 'on top' of the library, since the configuration file is not parsed inside security plugins.

apeters1971:
To say "No tokenlib" does not work, because they are needed and stacked. 

SciTokens->[ALICE Tokens]->EosTokens

All works via HTTP, not via ZTN. I don't want to rule out to use SciTokens via ZTN.

I just need this simple switch as suggested in the pull request. Brians comment is about nested config files, but this is already not possible by design for security plugin configuration and I don't need it.

bbockelm:
@apeters1971 - I'll defer to you on this but do note that new XRootD 5 features allow you to pull the configuration from a global.  See: https://github.com/xrootd/xrootd/blob/master/src/XrdSciTokens/XrdSciTokensAccess.cc#L907-L916; it was a similar setup where the plugin doesn't get the config file.

abh3 (commented):
Two questions before this goes in. The first in a comment in the code about "optional". The second is why do you think that one cannot say "-tokenlib none"  as a parameter to ztn. The ztn protocol is not responsible for loading the SciToken plugin for authorization purposes. This is done via the authlib directive. However, it needs that plugin it is in order to validate the token. If no validation is desired there is no reason to load that plugin as it will never be called. Right?

What we have now is an oddity that when you say you don't want validation ztn still requires that the plugin that validates the token be present even though it will never use it. I hope you see the inconsistency here.

apeters1971:
I will go for tokenlib "none". That is the easiest indeed!

apeters1971:
I will create one for option '-tokenlib none' to disable token validation

PR #1922, opened 2023-02-21
Allow building against system installed version of isa-l

amadio:
I noticed that `make install` was rebuilding isa-l and some other things, so I tried to update the external project for isa-l to fix that, but also to allow building XRootD against isa-l which is already installed in `/usr`. I also added an option to enable using the system version, which I left disabled by default so that it will keep the previous behavior of preferring the internal copy of isa-l.


ccaffy (approved):
Looks good to me !

PR #1923, opened 2023-02-21
Add workaround to avoid sporadic crashes of the test runner

amadio:
Sometimes, the call to dlclose() being commented out here gets called while test cases using the loaded library are still running (on another thread), which causes random crashes in the test runner. As a workaround, let the OS do the clean up of loaded libraries.


amadio:
Looks like this workaround actually is not enough...

abh3:
OK, looks like there is nothing to review at the moment then.

PR #1926, opened 2023-02-22
[XrdSecztn] Allow to point to a token file using CGI '?xrd.ztn=tokenfile

apeters1971:
As available in kerberos,gsi and sss I need to have a way to define in a multithreaded application the location of a token using CGI since environment variables don't work here. I have added the same mechanism supported by XrdCl to specify a tokenfile location in the URL, in this case you can say something like 
```
xrdcp foo root://....?xrd.ztn=/tmp/mytokenfile
```

### DIFF
--- a/src/XrdSecztn/XrdSecProtocolztn.cc
+++ b/src/XrdSecztn/XrdSecProtocolztn.cc
@@ -51,6 +51,7 @@
 #include "XrdVersion.hh"
 
 #include "XrdNet/XrdNetAddrInfo.hh"
+#include "XrdOuc/XrdOucEnv.hh"
 #include "XrdOuc/XrdOucErrInfo.hh"
 #include "XrdOuc/XrdOucPinLoader.hh"
 #include "XrdOuc/XrdOucString.hh"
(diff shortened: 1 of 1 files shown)


abh3 (commented):
So, we have a problem here. In general, this violates WLCG search rules. You are allowed to add additional search possibilities *after* all the official places you must look at are exhausted. This patch basically short-circuits that and says if it's on the URL don't look anywhere else. So, either move the code to the end of the search criteria or get WLCG dispensation. Please note the WLCG never envisioned this kind of thing so they will need to decide the policy here, I think all you have to do is work with Oliver on this.

apeters1971:
Ok, I have moved it for the time being to the end of the chain. The environment variables are not active in my case, the only side-effect is that '/tmp/bt_<uid' is used before the CGI one, but in principle this what I want

abh3:
OK, thanks. I know my pointing this out was a pain but the WLCG rules apply across multiple projects so that every implementation acts the same way. Anyway, I will look at this after all the checks complete and when I get some sleep. So, tomorrow but I don't expect any problems.

PR #1929, opened 2023-02-22
Make CMAKE_CXX_STANDARD a CACHE variable

greenc-FNAL:
This allows CMAKE_CXX_STANDARD to be customized from the CMake CLI with (e.g.) `-DCMAKE_CXX_STANDARD=17`.


amadio:
I think this is a good idea, and it seems to show that we need some work before merging this. This exposed some bugs, like using `data` for names, which clashes with [`std::data`](https://en.cppreference.com/w/cpp/iterator/data) from C++17.

greenc-FNAL:
Fair enough: I've only ever used it with `-DXRDCL_ONLY:BOOL=TRUE`.

amadio:
I will take care of fixing the compile errors with C++17 then I can merge this.

amadio:
Actually, it looks like this is not so easy to fix, as the problem is coming from a header from voms. They have a `struct data` and a `struct voms` with a `std::vector<data> std;` as a member. This is not good. We can fix the problem on our side, but the header needs to be fixed upstream, so I filed an issue there for this: https://github.com/italiangrid/voms/issues/114

amadio:
We will need to get rid of this before moving to C++17:

```
$ git grep 'using namespace std' | grep \.hh
src/XrdOuc/XrdOucString.hh:using namespace std;
src/XrdSys/XrdSysHeaders.hh:using namespace std;
```
I will give it a try, but it may require lots of changes.

abh3:
I agree, this was done before it was considered bad form to do so; 
especially in header files. I suppose it really depends on how many cc 
files exploited this. Of course, it's easy to figure out.

amadio:
This depends on #1933, which requires a major release. We will merge it after merging #1933.

amadio:
I think that at this point it should be actually ok to merge this. Could you please rebase your branch on the current master and push it here again? Thank you.

amadio:
I thought with the fixes already merged this might have been ok to merge, but we really need to remove `using namespace std;` from all our headers beforehand.

greenc-FNAL:
> I think that at this point it should be actually ok to merge this. Could you please rebase your branch on the current master and push it here again? Thank you.

Sorry, missed this. Glad you were able to proceed.

amadio:
Rebased and updated PR to apply to `devel` branch instead of `master`. Will merge it for 5.7 after fixing the CI to use C++14 where needed.

amadio:
@abh3 This should be good to go into v5.7. The only caveat being that `-DENABLE_VOMS=1` and `-DCMAKE_CXX_STANDARD=17` cannot be used together until we remove `using namespace std;` from all our headers/sources. I will merge this after you give your ok.

PR #1931, opened 2023-02-23
[XrdSecztn] Allow option '-tokenlib none' to disable token validation

apeters1971:
As proposed an even simpler solution to disable token validation in XrdSecztn by using "-tokenlib none".


abh3 (approved):
Looks OK.

PR #1933, opened 2023-02-24
Remove "using namespace std;" from XRootD (necessary to support C++17)

amadio:
This is necessary to avoid clashes of names from the global namespace with names from the std namespace. It was identified as a problem when trying to compile XRootD with C++17 standard and support for VOMS, which has a 'struct data' in the global namespace that clashes with std::data from C++17.

What I'm looking at for review is if using `std::` everywhere is fine or if there's a preference to add `using namespace std;` in the `.cc` files. I think adding `std::` explicitly is more reliable, but the patch is quite large. Once we merge this, #1929 should be fine to merge as well without problems.

### DIFF
--- a/src/XrdApps/XrdCpConfig.cc
+++ b/src/XrdApps/XrdCpConfig.cc
@@ -47,7 +47,6 @@
 #include "XrdSys/XrdSysHeaders.hh"
 #include "XrdSys/XrdSysLogger.hh"
 
-using namespace std;
 
 /******************************************************************************/
 /*                         D e f i n e   M a c r o s                          */
--- a/src/XrdApps/XrdCrc32c.cc
+++ b/src/XrdApps/XrdCrc32c.cc
@@ -42,7 +42,6 @@
 #include "XrdOuc/XrdOucCRC.hh"
 #include "XrdSys/XrdSysE2T.hh"
 
-using namespace std;
 
 namespace
 {const char *pgm = "xrdcrc32c";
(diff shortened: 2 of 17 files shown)


amadio:
Actually, I think I will split this into two changes. One patch to not need `using namespace std;` in our code, which can be merged into master without causing problems downstream, and another that just removes the `using namespace std;` lines from our headers, which should only be merged when we do the next major release. What do you think, @abh3?

amadio:
Ok, I've pushed the commit that changes the code not to rely on `using namespace std;` to master, so now we only need to add later the change that actually removes the lines with `using namespace std;` from our headers, which is what remains in this pull request and is actually the part that may break downstream code. This will only go in for the next major version.

abh3 (approved):
OK, all I see is using std removals. I also see that implicit use of std was already corrected in a previous pull request. So, it looks OK. I was wondering why did we eschew using declaration notation to fix this?

amadio:
> OK, all I see is using std removals. I also see that implicit use of std was already corrected in a previous pull request. So, it looks OK. I was wondering why did we eschew using declaration notation to fix this?

This is the same pull request, but I separated the parts fixing implicit usage of std and pushed it, so all that's left here is the removal of `using namespace std;` from the code.

As for using declarations, in one of our meetings when we discussed this, we decided to keep using `std::` in code rather than declarations for the sake of consistency. Since we cannot use declarations in our headers, rather than have a separate policy for headers and sources, we decided to go with using `std::` everywhere.

amadio:
This has to wait for 6.0 to be merged, so I will close for now and keep the branch on my fork to avoid premature merging by accident. I will open a new PR when the right time arrives.

Issue #1934, opened 2023-02-24
XrdCl client crash on connection failure


smithdh:
Hi. The crashes below were seen in an application, 'eoscp', inside libXrdCl (eos build of xrootd, approx XRootD version 5.5.2). Believed to be in the situation where an XRootD server (the local machine on which the eoscp was running and where it was connecting to) was immediately closing connections, due to #1928. The server has two addresses (ipv4 and ipv6).

Most frequently at this location:

```
(gdb) bt
#0  XrdCl::XRootDTransport::GetHeader (this=<optimized out>, message=..., socket=0x2c021c7330) at /usr/src/debug/xrootd-5.5.5/src/XrdCl/XrdClXRootDTransport.cc:323
#1  0x00007f90c87f1115 in XrdCl::AsyncHSReader::Read (this=0x7f90b0000960) at /usr/src/debug/xrootd-5.5.5/src/XrdCl/XrdClAsyncHSReader.hh:90
#2  XrdCl::AsyncSocketHandler::OnReadWhileHandshaking (this=0x21c7d00) at /usr/src/debug/xrootd-5.5.5/src/XrdCl/XrdClAsyncSocketHandler.cc:515
#3  0x00007f90c87f226d in XrdCl::AsyncSocketHandler::Event (this=0x21c7d00, type=1 '\001') at /usr/src/debug/xrootd-5.5.5/src/XrdCl/XrdClAsyncSocketHandler.cc:227
#4  0x00007f90c87673d8 in (anonymous namespace)::SocketCallBack::Event (this=0x7f90b0000d70, chP=<optimized out>, cbArg=<optimized out>, evFlags=<optimized out>) at /usr/src/debug/xrootd-5.5.5/src/XrdCl/XrdClPollerBuiltIn.cc:83
#5  0x00007f90c99f07db in XrdSys::IOEvents::Poller::CbkXeq (this=0x21a68c0, cP=0x7f90b0000d90, events=1, eNum=<optimized out>, eTxt=<optimized out>) at /usr/src/debug/xrootd-5.5.5/src/XrdSys/XrdSysIOEvents.cc:721
#6  0x00007f90c99f1b03 in XrdSys::IOEvents::PollE::Dispatch (this=this@entry=0x21a68c0, cP=0x7f90b0000d90, pollEv=<optimized out>) at /usr/src/debug/xrootd-5.5.5/src/XrdSys/XrdSysIOEventsPollE.icc:275
[...]

PR #1935, opened 2023-02-24
[XrdCl] Check for closed connection between handling of read and write events

smithdh:
This change is to fix issue #1934.

### DIFF
--- a/src/XrdCl/XrdClAsyncSocketHandler.cc
+++ b/src/XrdCl/XrdClAsyncSocketHandler.cc
@@ -46,6 +46,7 @@ namespace XrdCl
     pStreamName( ToStreamName( url, subStreamNum ) ),
     pSocket( new Socket() ),
     pHandShakeDone( false ),
+    pCloseCount( 0 ),
     pConnectionStarted( 0 ),
     pConnectionTimeout( 0 ),
     pHSWaitStarted( 0 ),
@@ -190,6 +191,7 @@ namespace XrdCl
 
     pPoller->RemoveSocket( pSocket );
     pSocket->Close();
+    pCloseCount++;
     return XRootDStatus();
   }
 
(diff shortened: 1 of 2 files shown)


simonmichal:
@smithdh : thanks a lot for the PR. OK I see this is handling the case when w have Bothe read and write events, and on the read event there's an error and the socket gets disconnected, in this case we are not suppose to handle the write event, is this correct? 
Is this only a problem in case the stream gets reconnected and assigned to a different event loop, or is this a problem in general because the stream got closed and the write event is invalid?

smithdh:
Hi @simonmichal I'm only aware of a problem when the AsyncSocket is re-opened and reassigned to another poller.

I just re-pushed a change to specify relaxed memory order on the std::atomic introduced by this patch; it took a while to convince myself this was ok (not that this makes any difference in generated code, in this case and on intel). I see we generally use relaxed order with std::atomics, and a comment in the XrdSys XrdSysRAtomic utility class contains a a suggestion to use a mutex to some cases, where a stronger guarantee is needed, explicit.

abh3:
Hi David,

I would suggest using XrdSysRAtomic.hh whenever a relaxed atomic is 
wanted. This is the only way to gaurantee that everyone uses that mutex in 
relaxed mode. People forget as we have seen.

Andy

smithdh:
Hi. Thanks, I've replaced the std::atomic that this patch was introducing with RAtomic.

smithdh:
Hi @simonmichal I've realised that there are some paths where the read event action (e.g. OnReadTimeout) can cause the AsyncSocketHandler to be not only closed but destroyed too. So this patch does nothing to help in those cases (and adds at least one access to invalid memory); so I'd propose to abandon/close this PR. I will look for another solution.

PR #1936, opened 2023-02-25
Adjust build rules to not depend on scitokenscpp to build libXrdSecztn

amadio:
This partially reverts commit 984efbc72bdad86b43923569f4dfa707b7a287a2. Actually, libXrdSecztn does not depend on scitokenscpp, we can build it on the client side even when `-DENABLE_SCITOKENS=FALSE`.

@abh, I believe this is what you meant initially that we should do, right?


amadio:
Looks like some changes are also necessary in the spec file to reflect the changes in CMake. I will take care of that and push a new patch later.

Issue #1928, opened 2023-02-22
server drops some new connections and logs "attempt to reuse active link"


smithdh:
Hello. It was noticed that for a time (about 24 hours) one of the EOS disk servers at cern would drop some new incoming connections and, in apparent correlation log entries like:

```
230217 01:12:17 281970 XrdLink: attempt to reuse active link
```

The situation seemed to return to normal about >24 hours after the first such message. (I've done some investigation and hope to have a potential fix. Will describe below, in the comments)

smithdh:
Trying to summarise some (perhaps) relevant extra information:

Over several hours before the situation where the server was dropping connections, there were a number of log entries mentioning 'bad file descriptor', such as:

e.g. grepping for ":1387@" gave this sequence:

```
[...]
230216 22:36:41 time=1676583401.392132 func=_close                   level=INFO  logid=unknown unit=fst@st-096-gg500f11.cern.ch:1095 tid=00007f14d91fe700 source=XrdFstOfsFile:2251             tident=daemon.75437:3834@st-096-gg500b5c sec=      uid=1 gid=1 name=nobody geo="" msg="done close" rc=0 errc=0
230216 22:36:41 282023 XrootdXeq: daemon.75437:3834@st-096-gg500b5c disc 0:00:02
230216 22:36:48 25875 XrootdXeq: 3.22020:3834@p06636710m46495 pub IP64 login as daemon
230216 22:36:48 25875 XrootdXeq: 3.22020:3834@p06636710m46495 disc 0:00:07
230216 22:36:48 178174 XrdLink: Unable to close 3.22020:3834@p06636710m46495; bad file descriptor
230216 22:36:58 292432 XrootdXeq: daemon.189977:3834@st-096-ff809e4a pub IP64 login as daemon
[...]
```

The first line is EOS specific and relates to closing a file, next line the close of connection by xrootd. Then a new connection on fd 3834 is logged (22:36:48), which is then closed very shortly afterwards (also logged at 22:36:48). The connection time indicates that connection started 7 seconds before; it is possible the machine had some loading or other issue that delayed everything for a few seconds at this point (indeed this may be a contributing factor). However a short time later, still at 22:36:48 there is a close error logged from a different thread than the "disc" message. The final line is a new connection on this descriptor 10 seconds later, which is unremarkable.

About 24 hours after the issue there are examples of a few log messages such as:

```
[...]

PR #1941, opened 2023-03-06
[Xrd] Wait for the current PollE poll to finish after removing a Link

smithdh:
This is a change to avoid issue #1928.

### DIFF
--- a/src/Xrd/XrdPollE.icc
+++ b/src/Xrd/XrdPollE.icc
@@ -33,6 +33,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/epoll.h>
+#include <sys/eventfd.h>
 
 #include "Xrd/XrdPollE.hh"
 #include "Xrd/XrdScheduler.hh"
@@ -43,7 +44,7 @@
   
 XrdPoll *XrdPoll::newPoller(int pollid, int maxfd)
 {
-   int pfd, bytes, alignment, pagsz = getpagesize();
+   int pfd, wfd, bytes, alignment, pagsz = getpagesize();
    struct epoll_event *pp;
 
 // Open the /dev/poll driver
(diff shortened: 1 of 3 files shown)


src/Xrd/XrdPollE.icc
```diff
@@ -79,9 +88,30 @@ XrdPoll *XrdPoll::newPoller(int pollid, int maxfd)
 XrdPollE::~XrdPollE()
 {
    if (PollTab) free(PollTab);
+   if (WaitFd >= 0) close(WaitFd);
    if (PollDfd >= 0) close(PollDfd);
 }
-  
+
+/******************************************************************************/
+/*                              A d d E v e n t F d                           */
```
abh3, line 96: This should be AddWaitFD (looks like you must have changed your mind at some point).

src/Xrd/XrdPollE.icc
```diff
@@ -164,6 +194,44 @@ void XrdPollE::Exclude(XrdPollInfo &pInfo)
...
+// Wait to make sure the poll thread has completed handling the last set of
+// events which may have included this Link. The handing includes examining the
+// Link's PollInfo before deciding if to schedule the Link. After we return,
+// the PollInfo may be reset and the Link could be subsequently reused.
+//
+   WaitPoller();
+}
+
+/******************************************************************************/
+/*                          H a n d l e E v e n t F d                         */
```
abh3, line 207: This should be HandleWaitFD

src/Xrd/XrdPollE.icc
```diff
@@ -272,9 +351,28 @@ void XrdPollE::Start(XrdSysSemaphore *syncsem, int &retcode)
        //
        if (num2sched == 1) Sched.Schedule(jfirst);
           else if (num2sched) Sched.Schedule(num2sched, jfirst, jlast);
+
+       if (haveWaiters) HandleWaitFd(waitFdEvents);
       } while(1);
 }
 
+/******************************************************************************/
+/*                             W a i t E v e n t F d                          */
```
abh3, line 360: This should ne WaitPoller (perhaps a better name would be Wait4Poller)

abh3 (changes requested):
Just comment changes are needed. Presumably the code was tested with the "sleeps" that caused it to fail and now it doesn't fail, right? The only other suggestion is consider repackaging the "eventfd' processing into a separate class so that it can be used elsewhere should the need arise.

smithdh:
Thanks very much for the comments. I've changed those; the banners and renamed Wait4Poller. I've tried it with my "sleeps" it solves what I could provoke. (I don't know how to reproduce the original production-system problem, as that seems to be quite rare).

Repackaging as a separate class is likely a better way to make it. However I'm not sure I can find a nice general interface for it. (It would probably be used with an epoll-poller, as eventfd is linux specific like epoll). I don't think we have the same problem in XrdSysIOEventsPollE because of how that poller works. So I propose to leave it here (not separated into another class) now.

PR #1946, opened 2023-03-07
[XrdCl] Add return codes for error checking in the network Event handler

### LINKED ISSUE #1934: XrdCl client crash on connection failure

smithdh:
This is a (second try) fix for issue #1934. (Closes #1934)

### DIFF
--- a/src/XrdCl/XrdClAsyncSocketHandler.cc
+++ b/src/XrdCl/XrdClAsyncSocketHandler.cc
@@ -136,6 +136,9 @@ namespace XrdCl
     }
 
     pHandShakeDone = false;
+    pTlsHandShakeOngoing = false;
+    pHSWaitStarted = 0;
+    pHSWaitSeconds = 0;
 
     //--------------------------------------------------------------------------
     // Initiate async connection to the address
@@ -214,17 +217,35 @@ namespace XrdCl
     type = pSocket->MapEvent( type );
 
     //--------------------------------------------------------------------------
+    // Handle any read or write events. If any of the handlers indicate an error
+    // we will have been disconnected. A disconnection may cause the current
+    // object to be asynchronously reused or deleted, so we return immediately.
+    //--------------------------------------------------------------------------
+    if( !EventRead( type ) )
+      return;
+
+    if( !EventWrite( type ) )
+      return;
+  }
+
+  //----------------------------------------------------------------------------
+  // Handler for read related socket events
+  //----------------------------------------------------------------------------
+  bool AsyncSocketHandler::EventRead( uint8_t type )
+  {
+    //--------------------------------------------------------------------------
     // Read event
     //--------------------------------------------------------------------------
     if( type & ReadyToRead )
     {
       pLastActivity = time(0);
       if( unlikely( pTlsHandShakeOngoing ) )
-        OnTLSHandShake();
-      else if( likely( pHandShakeDone ) )
-        OnRead();
-      else
-        OnReadWhileHandshaking();
+        return OnTLSHandShake();
+
+      if( likely( pHandShakeDone ) )
+        return OnRead();
+
+      return OnReadWhileHandshaking();
     }
 
     //--------------------------------------------------------------------------
@@ -233,14 +254,25 @@ namespace XrdCl
     else if( type & ReadTimeOut )
     {
       if( pHSWaitSeconds )
-        CheckHSWait();
+      {
+        if( !CheckHSWait() )
+          return false;
+      }
 
       if( likely( pHandShakeDone ) )
-        OnReadTimeout();
-      else
-        OnTimeoutWhileHandshaking();
+        return OnReadTimeout();
+
+      return OnTimeoutWhileHandshaking();
     }
 
+    return true;
+  }
+
+  //----------------------------------------------------------------------------
+  // Handler for write related socket events
+  //----------------------------------------------------------------------------
+  bool AsyncSocketHandler::EventWrite( uint8_t type )
+  {
     //--------------------------------------------------------------------------
     // Write event
     //--------------------------------------------------------------------------
@@ -248,19 +280,21 @@ namespace XrdCl
     {
       pLastActivity = time(0);
       if( unlikely( pSocket->GetStatus() == Socket::Connecting ) )
-        OnConnectionReturn();
+        return OnConnectionReturn();
+
       //------------------------------------------------------------------------
       // Make sure we are not writing anything if we have been told to wait.
       //------------------------------------------------------------------------
-      else if( pHSWaitSeconds == 0 )
-      {
-        if( unlikely( pTlsHandShakeOngoing ) )
-          OnTLSHandShake();
-        else if( likely( pHandShakeDone ) )
-          OnWrite();
-        else
-          OnWriteWhileHandshaking();
-      }
+      if( pHSWaitSeconds != 0 )
+        return true;
+
+      if( unlikely( pTlsHandShakeOngoing ) )
+        return OnTLSHandShake();
+
+      if( likely( pHandShakeDone ) )
+        return OnWrite();
+
+      return OnWriteWhileHandshaking();
     }
 
     //--------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


smithdh:
Sorry, converted to draft. It seems likely this still needs some work.

smithdh:
Reworked to add return codes on to various event handlers called by AsyncSocketHandler::Event. I abandoned the previous approach of enhancing "PollerBuiltIn" to allow for a check on whether a handler has been unregistered, since the return codes seem more direct way of checking.

Return code checks allow Event to stop processing any further actions once one of the handers indicates an error.

Moving the PR out of draft status.

smithdh:
May also help with issue #1979. (I reproduced an apparently similar deadlock, and this patch did stop the deadlock I could trigger).

src/XrdCl/XrdClAsyncSocketHandler.cc
```diff
@@ -220,11 +221,11 @@ namespace XrdCl
     {
       pLastActivity = time(0);
       if( unlikely( pTlsHandShakeOngoing ) )
-        OnTLSHandShake();
+        st = OnTLSHandShake();
```
amadio, line 224: In reviewing this with Andy, we asked ourselves the question why on the parts where writing is concerned below, the status from `OnTLSHandShake()` and other calls is not being checked.

src/XrdCl/XrdClAsyncSocketHandler.hh
```diff
@@ -149,67 +149,67 @@ namespace XrdCl
       //------------------------------------------------------------------------
       // Connect returned
       //------------------------------------------------------------------------
-      virtual void OnConnectionReturn();
+      virtual XRootDStatus OnConnectionReturn();
```
amadio, line 152: Andy suggests to add `XRD_WARN_UNUSED_RESULT` to these to ensure we are not ignoring any of the status being returned in other places.

src/XrdCl/XrdClAsyncSocketHandler.cc
smithdh: For the call to OnTLSHandShake() below (concerning writing) the flow doesn't do anything after the call (i.e. it returns from AsyncSocketHandler::Event(), which is what I was going to do for the error case) so I didn't add the check. But it's a good point this looks quite strange perhaps it's a trap for future modifications, e.g. if something is added it might also get executed during the error condition.

smithdh:
Pushed changes; addressing the two review comments above (i) check return code everywhere, (ii) make compiler warn on unused return code from modified methods. Also added comment(s) in code about the significance of return-ing on error during AsyncSocketHandler::Event.

smithdh:
Hi. I think there was the suggestion during review to change to use a smaller and simpler return type for the modified methods, i.e. bool rather than XRootDStatus. This is ok for me; I need a day or so to prepare and push the changes.

smithdh:
I've pushed  changes:

- changed return type to bool
- refactored previous read/write parts of AsyncSocketHandler::Event() into two methods EventRead() EventWrite() in order to remove the previous large number of 'st = Func();' style assignments.
- Added reset of pTlsHandShakeOngoing, pHSWaitStarted, pHSWaitSeconds=0 in AsyncSocketHandler::Connect(). This is because I made a change of logic in AsyncSocketHandler::CheckHSWait(): e.g. previously on wait-time elapse, pHSWaitSeconds was always set to 0 even if the previous SendHSMsg() caused a failure. The reset was never strictly safe (since perhaps the object is deleted or reused), however simply returning after the failure added another way for pHSWaitSeconds to be non-zero after a re-connect (I believe there were already other ways this can happen too). As there are paths where several members that track connection state are not reset on re-connect I took the opportunity to reset those in Connect(); although I don't believe this is involved in any reported problem that I've look into.

Issue #1772, opened 2022-08-26
Cannot start server on OS X without resolvable hostname


nsmith-:
So I get this cryptic error on running `xrootd` in v5.4.3:
```
220826 14:22:19 259 XrdConfig: Unable to determine host name;  ai_family not supported ; execution terminated.
```
and after compiling in debug mode, I found that the error was because in
https://github.com/xrootd/xrootd/blob/4245d0933d6ac49df3bfb79b5fb0f726a0baa67a/src/Xrd/XrdConfig.cc#L572-L573
we create a new address, and the constructor calls `Set`
https://github.com/xrootd/xrootd/blob/4245d0933d6ac49df3bfb79b5fb0f726a0baa67a/src/XrdNet/XrdNetAddr.cc#L97-L102
using the hostname `mac-132492.local` that was constructed initially in
https://github.com/xrootd/xrootd/blob/4245d0933d6ac49df3bfb79b5fb0f726a0baa67a/src/XrdNet/XrdNetIdentity.cc#L55
but unfortunately for me that hostname is not resolvable:
```bash
$ nslookup $(hostname)
** server can't find mac-132492.local: NXDOMAIN
```
and because of this, the `Set` call returns the error string `nodename nor servname provided, or not know` which is ignored in the constructor and so the variable `myIPAddr` is in an invalid state. The next line `myIPAddr->Name(0, &temp)` then raises the original error.

I tried a hack: putting the line `::1 mac-132492.local` in `/private/etc/hosts` and that did allow the server to start up. But I think probably a better solution exists.

abh3:
I believe this was fixed by ignoring hostname resolution on MacOS during start-up, was it not?

abh3:
This actually was resolved to an insufficient identity scheme. The resolution is documented in #2159 as it is essentially the same problem.

amadio:
I think that the current fix might not be enough: 
https://github.com/amadio/xrootd/actions/runs/10283927101/job/28458862396#step:4:478

PR #1950, opened 2023-03-09
 [XrdHttp] Q-values presence is considered BUT discarded when pattern matching the digest algorithm name to run

ccaffy:


abh3 (commented):
I don't see anything wrong here (famous last words). A couple of comments looking for your reaction and thoughts. Tell me what you think and I'll merge it.

amadio:
I'd like to pick the fix for the patch release, but leave out the addition of testing changes just on master, as I'm still working on it (i.e. my commit should not be merged yet). @ccaffy can you please split the changes? Thanks!

ccaffy:
@amadio done :)

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1062,37 +1062,56 @@ void XrdHttpReq::mapXrdErrorToHttpStatus() {
...
+    //First, extract the user digests algorithms name
+    //Digests are separated by comas
+    std::vector<std::string> userDigestsVecWithWeights;
+    extractChecksumFromList(userDigest, userDigestsVecWithWeights);
+    //A weight can also be passed by the user for each digest with digestName;q=xx. We just discard them
+    for(auto digestWithWeight: userDigestsVecWithWeights) {
+        std::vector<std::string> currentDigest;
+        XrdOucTUtils::splitString(currentDigest, digestWithWeight, ";");
+        // Don't forget to convert the digest name coming from the user to an XRootD checksum name
+        xrootdChecksums.push_back(convert_digest_name(currentDigest[0]).c_str());
```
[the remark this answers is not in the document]
ccaffy: I agree yes! I did not want to modify too much of the existing code. But now that I have some unit tests, I'll refactor the entire thing to extract the logic from this XrdHttpReq class.

I believe there is a need to document this mapping in the XRootD documentation. A user that will configure a sha-512 checksum (with a custom script) will not be able to call the sha-512 checksum in HTTP as the mapping will not match anything. BUT, if they call the checksum sha512 in the config, the mapping will work... I believe this needs to be documented properly to avoid surprises from the user.

xrootd-dev:
Well, we all decided a rewrite would be preferable as the code got pretty 
gnarly after handling this case. The logic is pretty simple it was 
complcated by the existing code that didn't mesh well. I also agree the 
testing change and the checksum change should be separate pr's.

Andy

abh3:
@ccaffy Could you address @amadio concerns. I too would like to avoid these construct and it woud seem hey are  not needed as all of this occurs single-threaded during initialization. Also, I see there have been numerous changes to his pull request including spiting it. Maybe one should close this one and post a brand new clean one so we don't trip over the old history. What do you think?

ccaffy:
> @ccaffy Could you address @amadio concerns. I too would like to avoid these construct and it woud seem hey are not needed as all of this occurs single-threaded during initialization. Also, I see there have been numerous changes to his pull request including spiting it. Maybe one should close this one and post a brand new clean one so we don't trip over the old history. What do you think?

Yep it's done in my latest commit, it uses unique_ptr instead of shared_ptr: only one initialization at server start.

ccaffy:
It looks like macos tests are not happy: https://github.com/xrootd/xrootd/actions/runs/4468692198/jobs/7849770342

Probably caused by https://github.com/xrootd/xrootd/issues/1779 or https://github.com/xrootd/xrootd/issues/1772 ?

amadio:
@abh3 The "override" messages are just warnings, they've been there for a while, I already have a local fix for those. I think the problem here is that the CI mac has no hostname, so we hit #1772. We have two options, fix #1772 or disable this test on macOS while a solution for that is found. What do you think we should do?

abh3:
@amadio I was under the impression that warnings were considered fatal. I guess that's not the case for MacOS. Since no one really runs a server on MacOS for production purposes I suppose the solution is if there is no host name (i.e. keep the check) but this is MacOS, ignore the error.

tests/XrdHttpTests/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+if(NOT BUILD_HTTP)
```
amadio, line 1: @ccaffy I guess for now we can just bail here if we're on macOS by using
```cmake
if(APPLE OR NOT BUILD_HTTP) # test hangs on macOS
...
```

ccaffy:
@abh3 do I have your green light for this functionality to be merged to master? 
Thanks in advance.

src/XrdHttp/XrdHttpChecksumHandler.hh
```diff
@@ -0,0 +1,107 @@
...
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+#ifndef XROOTD_XRDHTTPCHECKSUMHANDLER_HH
+#define XROOTD_XRDHTTPCHECKSUMHANDLER_HH
+
+#include <string>
+#include <map>
+#include <vector>
+#include "XrdHttpChecksum.hh"
+#include <memory>
```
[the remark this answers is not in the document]
abh3: Correct. System headers go first followed by our headers. I know it's arbitrary but that's what we decided on way back.

src/XrdHttp/XrdHttpChecksum.cc
```diff
@@ -0,0 +1,47 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+
+#include "XrdHttpChecksum.hh"
+#include <algorithm>
```
abh3, line 24: System header first than local headers. Stylistically all local headers are always qualified by the directory they are in so we would says:

#include "XrdHttp/XrdHttpChecksum.hh"

However, I noticed that style was never followed for any code in this directory. So, don't bother changing it. We'll address this as some future point.

src/XrdHttp/XrdHttpChecksumHandler.cc
```diff
@@ -0,0 +1,116 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+
+#include "XrdHttpChecksumHandler.hh"
+#include "XrdOuc/XrdOucTUtils.hh"
+#include "XrdOuc/XrdOucUtils.hh"
+#include <exception>
```
abh3, line 26: System headers ahead of local headers.

src/XrdHttp/XrdHttpChecksumHandler.cc
```diff
@@ -0,0 +1,116 @@
...
+
+std::map<std::string,XrdHttpChecksumHandlerImpl::XrdHttpChecksumPtr> XrdHttpChecksumHandlerImpl::XROOTD_DIGEST_NAME_TO_CKSUMS;
+
+void XrdHttpChecksumHandlerImpl::initializeCksumsMaps() {
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("md5","md5",true));
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("adler32","adler32",false));
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("sha1","sha",true));
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("sha256","sha-256",true));
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("sha512","sha-512",true));
+    addChecksumToMaps(std::make_unique<XrdHttpChecksum>("cksum","UNIXcksum",false));
```
abh3, line 37: See my comment about cksum vs crc32 in a previous note.

src/XrdHttp/XrdHttpChecksumHandler.hh
```diff
@@ -0,0 +1,116 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+#ifndef XROOTD_XRDHTTPCHECKSUMHANDLER_HH
+#define XROOTD_XRDHTTPCHECKSUMHANDLER_HH
+
+#include "XrdHttpChecksum.hh"
```
abh3, line 25: Local headers after system headers.

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -46,6 +46,7 @@
 #include "XrdOuc/XrdOucStream.hh"
 #include "Xrd/XrdProtocol.hh"
 #include "XrdOuc/XrdOucHash.hh"
+#include "XrdHttpChecksumHandler.hh"
 
 #include <openssl/ssl.h>
```
abh3, line 51: System headers before local headers.

src/XrdHttp/XrdHttpReq.hh
```diff
@@ -43,6 +43,7 @@
 
 #include "XProtocol/XProtocol.hh"
 #include "XrdXrootd/XrdXrootdBridge.hh"
+#include "XrdHttpChecksumHandler.hh"
 
 #include <vector>
```
abh3, line 48: OK, I get it. the header inversion was there before you started. However, we might as well correct it in the files that git updated. We should make note that it should be corrected everywhere: system header followed by local headers.

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1400,5 +1400,14 @@ int XrdOucUtils::getModificationTime(const char *path, time_t &modificationTime)
     return statRet;
 }
 
+void XrdOucUtils::trim(std::string &str) {
+    // Trim leading non-letters
+    while( str.size() && !isgraph(str[0]) ) str.erase(str.begin());
+
+    // Trim trailing non-letters
+
+    while( str.size() && !isgraph(str[str.size()-1]) )
```
abh3, line 1409: This is a silly optimization but when you think about it, you want to trim the end first as std:string will not move any characters but merely adjust the size and place a zero at the end.  Then when you trim the front there are potentially less characters to move around. I say silly because you likely will never notice this optimization. I couldn't resist though.

tests/XrdHttpTests/XrdHttpTests.cc
```diff
@@ -0,0 +1,167 @@
+#undef NDEBUG
```
abh3, line 1: So, I am confused. Wasn't the test suite spun off into a separate pull request as @amadio requested? Am seeing something wrong here?

tests/XrdHttpTests/XrdHttpTests.cc
```diff
@@ -0,0 +1,167 @@
...
+
+#include "XrdHttp/XrdHttpReq.hh"
+#include "XrdHttp/XrdHttpProtocol.hh"
+#include "XrdHttp/XrdHttpChecksumHandler.hh"
+#include <exception>
```
abh3, line 6: System headers before local headers.

abh3 (changes requested):
I think we still need some changes. Nothing drastic.

src/XrdHttp/README-CKSUM.md
```diff
@@ -0,0 +1,15 @@
...
+Here is a table summarizing the IANA-compliant checksum names:
+
+| Digest algorithm | HTTP digest | Will be base64 encoded |
+|------------------|-------------|------------------------|
+| md5              | md5         | true                   |
+| adler32          | adler32     | false                  |
+| sha1             | sha         | true                   |
+| sha256           | sha-256     | true                   |
+| sha512           | sha-512     | true                   |
+| cksum            | UNIXcksum   | false                  |
```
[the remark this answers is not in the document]
bbockelm: This is the IANA definition:

https://www.iana.org/assignments/http-dig-alg/http-dig-alg.xhtml

> The algorithm computed by the UNIX ``cksum'' command, as defined by the Single UNIX Specification, Version 2

So, it seems fairly natural to call this digest algorithm `cksum`.  At https://reveng.sourceforge.io/crc-catalogue/all.htm, there are 12 different CRC-32 algorithms documented, so it seems reasonable to name the `cksum` variant this.

In terms of compatibility, at least the xrootd-hdfs plugin and xrootd-multiuser plugin utilize the `cksum` name for the digest algorithm name for this particular algorithm.

In the "bad news" department, it appears the built-in algorithms call the cksum algorithm the `crc32` digest whereas xrootd-hdfs/multiuser call the Zlib CRC32 algorithm the `crc32` digest (which the built-in algorithms call `zcrc32`.  That means someone requesting the `crc32` checksum algorithm will get differing results, which is really unfortunate.

abh3:
@amadio So, I had two comments on corrections needed (i.e. sysheaders before local headers and non-directory qualified includes). I suggested correcting the sysheader problem but ignoring the include problem as it was endemic to this whole package. Then I realized that so was the sysheader problem. So, perhaps asking @ccaffy to correct the sysheader problem in files he touched is not a good idea and instead we should do a global change using another pr. What do you think?

tests/XrdHttpTests/XrdHttpTests.cc
amadio: I think it's not necessary to split into another pull request. Just a separate commit within this pull request is fine. I just want to cherry-pick just the fix for the patch release, and leave the tests just on master, since it depends on GoogleTest.

amadio:
Yes, let's go for a global change for the header ordering, it's out of scope for this pull request. However, I'd argue that we should put our headers first, to ensure that we are not missing any includes (e.g. using `std::cout` without including `<iostream>` in the header, but getting it only as side effect of previous includes). Also, if we use `""` instead of `<>` when including, then I think we need to leave the base directory out. So, for example, we do either `#include <XrdHttp/XrdHttpReq.hh>` or `#include "XrdHttpReq.hh"` in `src/XrdHttp/XrdHttpReq.cc`, but we will handle this later in another pull request.

src/XrdHttp/README-CKSUM.md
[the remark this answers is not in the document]
ccaffy: The only things I know about cksum is that we don't support natively that checksum and that it should not be base64 encoded before being returned by the client.

`addChecksumToMaps(std::make_unique<XrdHttpChecksum>("cksum","UNIXcksum",false));`

So it's up to the person writing the script to deal with what the return value should be isn't it?

src/XrdHttp/README-CKSUM.md
```diff
@@ -0,0 +1,15 @@
...
+| crc32            | crc32       | false                  |
```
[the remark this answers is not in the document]
ccaffy: I'll just remove it from that table then :)

src/XrdHttp/README-CKSUM.md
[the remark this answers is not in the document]
ccaffy: @bbockelm thanks for your reply. What you said about the xrootd-hdfs plugins triggers some questioning on my side: I wanted to be nice with server administrator allowing them to set, for example, `sha512` or `sha-512` for configuring a sha-512 algorithm. This would not have worked before my modification, see this function here: https://github.com/xrootd/xrootd/blob/master/src/XrdHttp/XrdHttpReq.cc#L134

If there are other plugins that can be used to compute checksums, could my new implementation cause problems?

tests/XrdHttpTests/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+if(APPLE OR NOT BUILD_HTTP)
```
amadio, line 1: @ccaffy You can now revert this back to just `if(NOT BUILD_HTTP)`, as I have put a workaround for issue #1772 in master that will allow the tests to run even without a proper hostname.

ccaffy: Sure :)

abh3:
> Yes, let's go for a global change for the header ordering, it's out of scope for this pull request. However, I'd argue that we should put our headers first, to ensure that we are not missing any includes (e.g. using `std::cout` without including `<iostream>` in the header, but getting it only as side effect of previous includes). Also, if we use `""` instead of `<>` when including, then I think we need to leave the base directory out. So, for example, we do either `#include <XrdHttp/XrdHttpReq.hh>` or `#include "XrdHttpReq.hh"` in `src/XrdHttp/XrdHttpReq.cc`, but we will handle this later in another pull request.

Changing the ordering would involve a massive number of files. I don't think he "strd" problem is sufficient to warrant a massive change like that, especially since we've been doing it this way for 20+ years.  Yes, I know there are good reasons to do it the other way because then you are forced to include every dependency all the time for system headers. Though you still have the same problem for local headers. So, in the end to avoid busy work, I'd do the minimum change and simply correct the files in http to follow what is done elsewhere.

As for putting on the directory ahead of an include file, again this is stylistic. We don't need to worry about this as long as people follow the naming convention for files as it all works out in the end. Requiring "src/" prefix is silly and it not needed. It works fine without that. The cmake include spec makes things work in a natural way. So, I'm neutral but emphasize there is no reason to change everything that has a directory prefix to not have one or add "src/" everywhere. That's a nonsense change IMO.

src/XrdHttp/README-CKSUM.md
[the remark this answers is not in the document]
abh3: For xrootd checksum plugins the return value is always a hex-string and that implies for certain checksums like SHAxxx (which we don't have implementation for as of yet) the HTTP code would need to convert it to base64. Thus far, all script implementation do the same. Having a common format for all checksums is the "standard" for xroot but not for http. I will make that obvious in the documentation. I think the table already specifies when conversion is necessary. The only odd one is Unixcksum (a.k.a. cksum) which is supposed to be a dec-string. However, since you are removing from the table the conversion issue becomes moot.

amadio:
Ok, we can discuss this during the workshop (header ordering). Unless you have a strong objection, however, I will pick up this pull request (on master), and cherry pick the commit with the fix so I can release 5.5.4. This is the last change we are waiting for, and the EOS team would like to have the release soon.

PR #1954, opened 2023-03-11
Implement `only-if-cached` mode for HTTP

bbockelm:
This implements the HTTP `only-if-cached` mode, allowing the client to prevent triggering a download from the origin.

This is envisioned to be used in the case where pulling from the origin is expensive and the client may want to decide to go to another cache first.

This is based on top of #1953 - to be reviewed after that work is committed.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1041,6 +1041,10 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
        return 1;
       }
 
+// Some headers must always be converted to CGI key=value pairs
+//
+   hdr2cgimap["Cache-Control"] = "cache-control";
+
 // Test if XrdEC is loaded
    if (getenv("XRDCL_EC")) usingEC = true;
 
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -44,6 +44,7 @@
 #include <arpa/inet.h>
 #include <sstream>
 #include "XrdSys/XrdSysPlatform.hh"
+#include "XrdOuc/XrdOucCacheDirective.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 #include "XrdHttpProtocol.hh"
 #include "Xrd/XrdLink.hh"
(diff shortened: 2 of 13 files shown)


ccaffy:
I am OK with the HTTP part. I never worked with the XrdPfc library so I leave it to somebody more expert on that :)

osschar:
For only-if-cached ... isn't a cache supposed to return error when some non-cached data is requested? This could be implemented as a flag in read-request from IO.

I understand this might be a problem for xcache as there might be hole in a file causing the read to be going smoothly for a long time before it hits a hole in cached data.

Again (as for #1953) I think this should not be a state of XrdPfcFile. We can find a way to disable prefetching on such IO.

And, also again, for direct mode proxies we should find a way to coalesce clients with the same cache-control to the same IO. One has to think at this point how various control flags interact ... and which are clearly invalid, e.g., no-cache&only-if-cached.

bbockelm:
> For only-if-cached ... isn't a cache supposed to return error when some non-cached data is requested?

Possibly, but this is fairly impractical.  Remember the HTTP response is sent up front.  So, anything after the first byte and you lose the ability to send an error.

So, for now, calling “any cached byte means the response is cached” is good enough for me.

> And, also again, for direct mode proxies we should find a way to coalesce clients with the same cache-control to the same IO.

that would require some serious surgery of the OFS layer to have multiple file handles per file and it’d break the threading contracts … I think the approximations done here are not perfect but better than the more serious surgery.

osschar:
OK, agreed about the semantics. Let's see how we deal with #1953 and then proceed.

osschar:
And I shamefully admit I don;t know how http file transfer works ... I was thinking about this from the perspective of xroot protocol (as these flags would apply there as well) where every sub read gets a status code.

Or you were only thinking of using http?

bbockelm:
> Or you were only thinking of using http?

I would like the feature to be generally useful for any protocol.  However, since the driver is better HTTP 1.1 compliance, unless there's a need to push the functionality all the way through to the reads themselves, it seems simpler to leave it as a file-level property.  I can see some downsides around code complexity in the read path.

abh3:
To that end. Could the actual flags be fully documented in a "TO-DO" file 
explaining what each one does. If this is applicable to file level only 
then it's easily extended to xroot protocol. At the moment no one (other 
than JT) is asking for this kind of functionality and when it is asked 
it's in the context of HTTP. I also agree that we don't want to start 
changing everthing to provide 100% consistency in terms of cache control. 
From what I understand, these flags are considered "best effort" and might 
be ignored if they cannot be fullfilled even in the HTTP world world.

In the end, it all winds up being a compromise.

abh3 (commented):
I guess this is in the same comment as the max-age pull request. The same comments apply here.

amadio:
Thank you for the idea, @bbockelm, this has been implemented in #2104 and will be in the next feature release.

PR #1957, opened 2023-03-13
Implement `max-age` directive

bbockelm:
This is the third PR in the series (to be reviewed after #1953 and #1954; it'll be rebased on top of whatever those become).

This implements the HTTP `max-age` directive, allowing the client to control how old the data returned can be.  If the cached copy is over the age threshold, either the PFC will delete it (if the file is being opened for the first time) or the HTTP request will (for files that are already opened by other clients).

This will need the most careful feedback from @osschar as it contains a key improvement for the XrdPfc -- an `Unlink` request no longer invalidates all open file handles but rather the next read will cause it to be reopened.

Note the careful cleanup of the HTTP GET state machine -- now each request state represents a unique operation instead of overloading the request state with other pieces of information.  I also tried to add some more careful comments to the state machine to try and make it more understandable by the next developer.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1041,6 +1041,10 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
        return 1;
       }
 
+// Some headers must always be converted to CGI key=value pairs
+//
+   hdr2cgimap["Cache-Control"] = "cache-control";
+
 // Test if XrdEC is loaded
    if (getenv("XRDCL_EC")) usingEC = true;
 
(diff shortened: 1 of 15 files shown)


abh3 (commented):
OK, I have looked at this and I cannot say I am happy. This is not what we have traditionally doe visive the architecture. The approach here is rather invasive as it tries to force the whole http header protocol all the way through the end-point code. That really is not what we want to do. In general, we have always promoted headers to cgi elements. In this case, one would need to promote header sub-elements to cgi elements. We then do not pollute the code with specific syntax requirements, Additionally, the protocol gets to choose what actually gets exposed and it's all done in a way that is transparent. I know you spent a lot of work here (obviously) but the approach is really not the way to go as it really muddles the code.

I also see that you better structured the http state machine. That was a good thing. Unfortunately, it's totally tied up with this feature patch. They really should be sperate changes. Could we start with that (i.e. fixing the http state machine)?

bbockelm:
Closing this one out -- Alja has been picking up where I've left off in separate PRs.

Issue #1955, opened 2023-03-12
XRootD day-light savings handling wrong again


stlammel:
Dear xrootd team,
   a few years ago we opened an issue reporting that out services in the US have authentication failures for a little over an hour after the US switches to day-light savings time. We had an exchange and traced this down to the daemons calculating the timezone offset to UTC at start time and then using it throughout their execution. I thought the issue was fixed and all recent versions would include the fix.
   But aparently not: All CMS xrootd services in the US failed aur tests with authentication error this morning Sunday, March 12th at 2 am UTC for a bit over an hour. This smells like the day-light savings bug. Sites run xrootd versions between 5.3.1 and 5.5.3, i.e. rather recent. The error is XRootDStatus.code=204 "[FATAL] Auth failed: No protocols left to try".
   Could you please take a look if the old fix got lost? or a new issue was introduced?
   Thanks,
 - Stephan

stlammel:
One of my colleagues found the old bug report,
https://github.com/xrootd/xrootd/issues/985

amadio:
Closing, fixed by #1958.

Issue #985, opened 2019-05-09
GSI/Daylight Savings Time authentication issue


brianhlin:
With the US DST transition (March 9/10), CMS saw failures at US-CMS sites using a proxy certificate generated between 2-3am UTC:

```
Authentication with gsi failed: Secgsi: ErrParseBuffer: certificate chain
verification failed: Proxy: certificate expired: kXGC_cert
```

This issue did not appear for non-US sites using the same certificate and connections using proxies generated after 3am UTC worked with US sites. In the original support ticket [2], we proposed reproducing this ticket by setting up an XRootD server with its clock set to March 9/10 with a US time zone and a second host acting as a client set to the same time with a European time zone.

[1] http://wlcg-sam-cms.cern.ch/templates/ember/#/historicalsmry/heatMap?end_time=2019%2F03%2F10%2006%3A00&metrics=org.cms.SE-xrootd%20(connection)%2Corg.cms.SE-xrootd%20(contain)%2Corg.cms.SE-xrootd%20(read)%2Corg.cms.SE-xrootd%20(version)%2Corg.cms.WN-analysis%20(%2Fcms%2FRole_lcgadmin)%2Corg.cms.WN-xrootd-access%20(%2Fcms%2FRole_lcgadmin)&profile=CMS_CRITICAL_FULL&site=T1_US_FNAL%2CT2_BR_SPRACE%2CT2_US_Caltech%2CT2_US_Florida%2CT2_US_MIT%2CT2_US_Nebraska%2CT2_US_Purdue%2CT2_US_UCSD%2CT2_US_Vanderbilt%2CT2_US_Wisconsin&start_time=2019%2F03%2F10%2000%3A00&time=manual&view=Test%20History
[2] https://support.opensciencegrid.org/public/tickets/ae8dc3e90125b8499e6dd7a06ecfe6025ccad8a0e871b6d7dbd4cb34c00393a0

bbockelm:
Bumping this ticket - what's the latest?

abh3:
We have not had time to work on this and, frankly, this is not a high 
priority as there is some time before this issue rears it's ugly head.

bbockelm:
Does it happen at all transitions?  If so, the next occurrence is November 3.  Given the cadence of releases and site upgrades, this likely has to be done in the next month or so, right?

abh3:
From my understanding, this happens when the clocks move forward not 
backward. If actually true, we have 6 more months :-)

abh3:
[...]

PR #1958, opened 2023-03-13
[XrdCrypto] Fix utility function XrdCryptosslASN1toUTC to return in UTC

smithdh:
To fix issue #1955 and #985, concerning authentication failures for certificates/proxies with notbefore/after dates during the changeover hour of daylight saving time.

### DIFF
--- a/src/XrdCrypto/XrdCryptoAux.cc
+++ b/src/XrdCrypto/XrdCryptoAux.cc
@@ -77,6 +77,7 @@ void XrdCryptoSetTrace(kXR_int32 trace)
 time_t XrdCryptoTZCorr()
 {
    // Time Zone correction (wrt UTC)
+   // Assumes no DST, the correction is not expected to change during the year
    
    if (!TZInitialized) {
       time_t now = time(0);
--- a/src/XrdCrypto/XrdCryptoX509.cc
+++ b/src/XrdCrypto/XrdCryptoX509.cc
@@ -104,12 +104,10 @@ int XrdCryptoX509::BitStrength()
 //_____________________________________________________________________________
 bool XrdCryptoX509::IsValid(int when)
 {
-   // Check validity at local time 'when'. Use when =0 (default) to check
+   // Check validity at UTC time 'when'. Use when =0 (default) to check
    // at present time.
 
    int now = (when > 0) ? when : (int)time(0);
-   // Correct for time zone (certificate times are UTC plus, eventually, DST
-   now -= XrdCryptoTZCorr();
    return (now >= (NotBefore()-kAllowedSkew) && now <= NotAfter());
 }
 
@@ -120,8 +118,6 @@ bool XrdCryptoX509::IsExpired(int when)
    // at present time.
 
    int now = (when > 0) ? when : (int)time(0);
-   // Correct for time zone (certificate times are UTC plus, eventually, DST
-   now -= XrdCryptoTZCorr();
    return (now > NotAfter());
 }
 
--- a/src/XrdCrypto/XrdCryptosslAux.cc
+++ b/src/XrdCrypto/XrdCryptosslAux.cc
@@ -735,7 +735,7 @@ time_t XrdCryptosslASN1toUTC(const ASN1_TIME *tsn1)
    // Init also the ones not used by mktime
    ltm.tm_wday  = 0;        // day of the week 
    ltm.tm_yday  = 0;        // day in the year
-   ltm.tm_isdst = -1;       // daylight saving time
+   ltm.tm_isdst = 0;        // we will correct with an offset without dst
    //
    // Renormalize some values: year should be modulo 1900
    if (ltm.tm_year < 90)
@@ -744,10 +744,10 @@ time_t XrdCryptosslASN1toUTC(const ASN1_TIME *tsn1)
    // month should in [0, 11]
    (ltm.tm_mon)--;
    //
-   // Calculate UTC
+   // Calculate as if the UTC stamp was a localtime with no dst
    etime = mktime(&ltm);
-   // Include DST shift; here, because we have the information
-   if (ltm.tm_isdst > 0) etime += XrdCryptoDSTShift;
+   // Correct to UTC
+   etime += XrdCryptoTZCorr();
    // Notify, if requested
 // DEBUG(" UTC: "<<etime<<"  isdst: "<<ltm.tm_isdst);
    //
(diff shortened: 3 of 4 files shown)


smithdh:
Hi. To help in PR review it was suggested some explanation/considerations be added, so I try to elaborate below:

The direct issue that the patch tries to address is to restore the apparent intent of the XrdCryptosslASN1toUTC to convert the ASN1 date (with caveat noted later) to UTC, specifically returning the UTC value in the time_t type, presumably to be understood as the usual unix-utc-time-since-epoch:

I show the inputs and outputs of the existing XrdCryptosslASN1toUTC() function to illustrate how it currently works: I consider some chosen ASN1 dates (e.g. the string content of tsn1->data), for some dates chosen to cross the DST timezone changes for a region, e.g. NY in USA.

(with TZ="America/New_York" set in the environment)

The "Z" in the date string indicates the time is in UTC: In the list below I've evaluated the existing function, and then in parenthesis indicated this value minus the actual epoch-time for this date (as looked up elsewhere), 

(localtime skips forward, DST comes into effect on 12 March):

230312001001Z => 1678597801 (result-real = 18000)
230312011001Z => 1678601401 (result-real = 18000)
230312021001Z => 1678608601 (result-real = 21600)
230312031001Z => 1678608601 (result-real = 18000)
230312041001Z => 1678612201 (result-real = 18000)
230312051001Z => 1678615801 (result-real = 18000)

(localtime skips backwards, DST will end 5 Nov):

231105001001Z => 1699161001 (result-real = 18000)
231105011001Z => 1699164601 (result-real = 18000)
231105021001Z => 1699168201 (result-real = 18000)
231105031001Z => 1699171801 (result-real = 18000)
231105041001Z => 1699175401 (result-real = 18000)
231105051001Z => 1699179001 (result-real = 18000)

the result of the function XrdCryptoTZCorr() for this TZ is -18000 (this value is independent of the date during the year on which it is evaluated). Most of the existing uses of the return value from XrdCryptosslASN1toUTC() involve comparing it to the result from time(0) where the "XrdCryptoTZCorr()" is subtracted from the time(0) before the compare. So in a roundabout way all the comparisons are correct except for those during the hour after the DST shift forward, where the XrdCryptosslASN1toUTC() is returning a date 1 hour later than intended.

The reported problems are understood to be due to the notBefore times (of the proxy part of the chain) that lie in this range being seen as sufficiently in the future to not be valid:

The patch then intends to

(i) return unix-time-since-epoch (i.e. utc, in the time_t type) time from XrdCryptosslASN1toUTC function
(ii) remove additional corrections applied to comparisons that involve the return value.

Other considerations: Not all places where the return value was used were corrected with XrdCryptoTZCorr(). i.e. the use of the CRL NextUpdate() date/time. So this patch will change what is evaluated at those places. However I believe this controls only a DEBUG warning message ("WARNING: CRL is expired: you should download the updated one"), and presumably we already are comparing this with various offsets (across our sites in various timezones).

The ASN1 notation has some types, including UTCTime and GeneralizedTime. (For the former there is a 2 digit year, and the later uses a four digit year starting with '20'). RFC5280 gives some additional restrictions on the ASN1 date used for x509, e.g. no fraction of seconds, always specify the 'Z' and, for dates before 2050 UTCTime must be used and after then GeneralizedTime must be used. Our XrdCryptosslASN1toUTC function assumes and will only parse the UTCTime type. I thought to leave this as is for now. The largest dates in the future I believe are for the expiration dates of CA; e.g. cern is currently using one with validy 2013-2023. So later this year I guess it will use one with a notAfter date 2033.

--

How it works: In the patch I added a comment about the independence of XrdCryptoTZCorr() with time of year, because I think there was some suspecion this calculation would vary over the year (and so not be suitable for the current evaluate-once approach). The function find the difference between two dates with two routes:

(i) time(0) ->localtime_r() -> mktime() => result is a time_t
(ii) time(0) -> gmtime_r() -> mktime() => result is a time_t

In route (i) the result should always be the same of the input, i.e. the current unix-utc-epoch time. The second interprets and struct tm, in UTC (i.e. hour/min/seconds in UTC, isdst=0) and converts with mktime(), which interpres as a localtime. i.e. a localtime with isdst=0. So the result is time difference between localtime (assuming DST=0) and UTC. (So does not change accross DST switch).

The existing XrdCryptosslASN1toUTC() also interprets the UTC string from tsn1->data as a localtime with mktime(). It is noted in the existing code-comment that mktime() is more portable than timegm(). timegm() would otherwise seem to be the natural choice here. However the DST conversion is left to mktime() to detect when to apply (setting tm_isdst=-1 on input), and then the application of this conversion is undone again (by the line "if (ltm.tm_isdst > 0) ...") using the returned value of tm_isdst.

e.g. here is the behaiour over the DST change forward:

mktime( 230312011001Z, tm_isdst=0) => (1678601401, tm_isdst=0)
mktime( 230312011001Z, tm_isdst=1) => (1678597801, tm_isdst=0)
mktime( 230312011001Z, tm_isdst=-1) => (1678601401, tm_isdst=0)

mktime ( 230312021001Z, tm_isdst=0 ) => (1678605001, tm_isdst=1)
mktime ( 230312021001Z, tm_isdst=1) => (1678601401, tm_isdst=0)
mktime ( 230312021001Z, tm_isdst=-1) => (1678605001, tm_isdst=1)

mktime ( 230312031001Z, tm_isdst=0) => (1678608601, tm_isdst=1)
mktime ( 230312031001Z, tm_isdst=1) => (1678605001, tm_isdst=1)
mktime ( 230312031001Z, tm_isdst=-1) => (1678605001, tm_isdst=1)

The tm_isdst=0 on input returns a time which advances by 3600 each hour. The same is true for tm_isdst=1 case (but this is 3600 different from the non-dst case).  However the auto-detecting case of tm_isdst=-1 is more complicated, as it returns the time "02:10:01" with a value indicating it has been interpreted as no-DST, but the return indicates DST is in effect. This is presumably a subtle case with auto-detect on a local-time which does not actually occur, with "DST-in-effect" being different to "This was interpreted with DST-in-effect".

This PR patch changes the input to always specify tm_isdst=0 and to not use the returned value of tm_isdst: It is corrected by the XrdCryptoTZCorr() which always has the non-DST value.

Not shown is the re-evaluation of the output of the patched XrdCryptosslASN1toUTC() across the forwards-and-backwards DST change, but it was seen to be in agreement with the looked-up value of the unix-utc-time-since-epoch value.

As noted in #985 this logic in involved enough to be error prone: Using timegm() would have simplified things, and the suggestion to use openssl functions to compare ASN1 structures might well be preferable. Depending on the number of functions converted swapping to the openssl might change the signatures of several of these functions (e.g. to not pass around time_t, or presumably not to return time_t). However the purpose of the patch and to address the specific problem I kept the time_t approach.

PR #1965, opened 2023-03-15
CMake improvements and addition of Google Test to start converting existing CppUnit tests to it

amadio:
Opening a pull request to give others a chance to review this before merging. Will merge tomorrow if there are no objections. The CI system is adapted to run the tests already, including the one I converted as an example.


ccaffy:
All good for me :) Thanks @amadio for this work :)

simonmichal:
I wondering if we shouldn't guard the following block:
```
%check
 cd xrootd/build
 %if %{use_cmake3}
 ctest3 --output-on-failure
 %else
 ctest --output-on-failure
 %endif
```

with `%if %{?_with_tests:1}%{!?_with_tests:0}`, meaning enforce the check only if we are building with `_with_tests` flag.

Otherwise looks good :-)

amadio:
We could do that, but it's not necessary, the way it is now, ctest will just do nothing. You can see what happens [here](https://github.com/xrootd/xrootd/actions/runs/4428613328/jobs/7767954337?pr=1965#step:5:15688), in the fedora-rpm build.

Issue #1967, opened 2023-03-18
XrdHttp - Wrong handling of SSL_Shutdown() in XrdHttpProtocol::Cleanup()


ccaffy:
We see a lot of messages of SSL_shutdown failed in the logs.

This is due to the following code in `XrdHttpProtocol::Cleanup()`:

```
if (SSL_shutdown(ssl) != 1) {
      TRACE(ALL, " SSL_shutdown failed");
      ERR_print_errors(sslbio_err);
    }
```

The documentation of SSL_shutdown says that it is only when the return code of SSL_shutdown() is <0 that there is a fatal error: https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html

I'll therefore change this behaviour.

ccaffy:
Done and merged.

PR #1968, opened 2023-03-20
[XrdHttp] Proper handling of SSL_Shutdown() return code

ccaffy:
Hi @abh3 , 

This PR to address the issue https://github.com/xrootd/xrootd/issues/1967

It looks to me like the shutdown of the TLS /SSLconnection via `SSL_shutdown()` is not as well protected as the way it is done in the XrdTlsSocket code: https://github.com/xrootd/xrootd/blob/master/src/XrdTls/XrdTlsSocket.cc#L753

For example, according to the documentation:
> Note that SSL_shutdown() must not be called if a previous fatal error has occurred on a connection i.e. if SSL_get_error() has returned SSL_ERROR_SYSCALL or SSL_ERROR_SSL.

Is this a problem that needs to be addressed?

Thanks in advance for your review

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1758,11 +1758,26 @@ void XrdHttpProtocol::Cleanup() {
   }
 
   if (ssl) {
-
-
-    if (SSL_shutdown(ssl) != 1) {
-      TRACE(ALL, " SSL_shutdown failed");
-      ERR_print_errors(sslbio_err);
+    // Shutdown the SSL/TLS connection
+    // https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html
+    // We don't need a bidirectional shutdown as
+    // when we are here, the connection will not be re-used.
+    // In the case SSL_shutdown returns 0,
+    // "the output of SSL_get_error(3) may be misleading, as an erroneous SSL_ERROR_SYSCALL may be flagged even though no error occurred."
+    // we will then just flush the thread's queue.
+    // In the case an error really happened, we print the error that happened
+    int ret = SSL_shutdown(ssl);
+    if (ret != 1) {
+        if(ret == 0) {
+            // Clean this thread's error queue for the old openssl versions
+            #if OPENSSL_VERSION_NUMBER < 0x10100000L
+                ERR_remove_thread_state(nullptr);
+            #endif
+        } else {
+            //ret < 0, an error really happened.
+            TRACE(ALL, " SSL_shutdown failed");
+            ERR_print_errors(sslbio_err);
+        }
     }
 
     if (secxtractor)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1759,10 +1759,17 @@ void XrdHttpProtocol::Cleanup() {
...
-
-    if (SSL_shutdown(ssl) != 1) {
-      TRACE(ALL, " SSL_shutdown failed");
-      ERR_print_errors(sslbio_err);
+    int ret = SSL_shutdown(ssl);
+    if (ret != 1) {
+        if(ret == 0) {
+            //https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html
+            //Call again SSL_shutdown
+            ret = SSL_shutdown(ssl);
```
abh3, line 1767: So, tell me why we need a bidirectional shutdown here. From the same documentation "If a unidirectional shutdown is enough (the underlying connection shall be closed anyway), this first call to SSL_shutdown() is sufficient." I think the actual problem was that the original code insisted that the return value be one when in fact 1 or 0 was good enough and only -1 should have reported an error. However, a zero return should clear the error stack as a misleading error may have been placed on the stack. Agree?

abh3 (commented):
OK, this will work but it does slow the connection turnaround because it tries to do a bidirectional shutdown which I don't think is needed. So, while not technically wrong there is a possible improvement and I leave it up to you to decide which way to go.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1759,10 +1759,17 @@ void XrdHttpProtocol::Cleanup() {
 
   if (ssl) {
 
-
-    if (SSL_shutdown(ssl) != 1) {
-      TRACE(ALL, " SSL_shutdown failed");
-      ERR_print_errors(sslbio_err);
+    int ret = SSL_shutdown(ssl);
```
amadio, line 1762: Don't you maybe need also a call to [`SSL_set_shutdown`](https://www.openssl.org/docs/man1.1.1/man3/SSL_set_shutdown.html) before like seen [here](https://github.com/xrootd/xrootd/blob/master/src/XrdTls/XrdTlsSocket.cc#L750)?

src/XrdHttp/XrdHttpProtocol.cc
amadio, line 1767: Is a single retry always sufficient? It seems that in [XrdTlsSocket it may be called up to 4 times](https://github.com/xrootd/xrootd/blob/master/src/XrdTls/XrdTlsSocket.cc#L752-L769) in order to shutdown SSL.

amadio: > OK, this will work but it does slow the connection turnaround because it tries to do a bidirectional shutdown which I don't think is needed. So, while not technically wrong there is a possible improvement and I leave it up to you to decide which way to go.

The documentation says that bidirectional shutdown is necessary if the connection is going to be reused. If the connection will be closed anyway, then the unidirectional shutdown is enough. Therefore, if we are going to reuse the connection in question here, we should do the complete shutdown, as done by @ccaffy here. However, I think from the above, before calling any shutdown function, we must check that there are no pre-existing errors on the connection.

amadio: Replying to my own comment here, but the documentation states that at most two calls are needed.

theanalyst: Instead of calling `ERR_print_errors(bio*)` can't we just clear the thread local queue, and if we really need just peek on the last error code. Depending on how openssl implements its thread local vs global queues it might give us a slight edge over trying to use ERR_print_errors ?

https://www.openssl.org/docs/man1.0.2/man3/ERR_remove_thread_state.html & 
https://www.openssl.org/docs/man1.0.2/man3/ERR_peek_last_error.html

amadio: I was thinking about calling `SSL_get_error`, as recommended in the docs.

theanalyst: > I was thinking about calling `SSL_get_error`, as recommended in the docs.

The current docs only mention this function as a switch statement for a few functions (not including ssl_shutdown), not sure whether the doc. is reflective of reality tho :/

"SSL_get_error() returns a result code (suitable for the C "switch" statement) for a preceding call to SSL_connect(), SSL_accept(), SSL_do_handshake(), SSL_read(), SSL_peek(), or SSL_write() on ssl."

abh3: What does "reuse connection" really mean? The socket stays open and connected and we start a new session on the existing socket?  When Recycle() gets called the socket will definitely be closed unless it was marked as a shared socket which doesn't appear to ever happen.

src/XrdHttp/XrdHttpProtocol.cc
theanalyst: Shouldn't SSL_shutdown anyway set this flag without need for us to set this. The docs do mention that 
"Whether the operation succeeds or not, the SSL_SENT_SHUTDOWN flag is set and a currently open session is considered closed and good and will be kept in the session cache for further reuse."

src/XrdHttp/XrdHttpProtocol.cc
ccaffy: > I was thinking about calling `SSL_get_error`, as recommended in the docs.

It looks like I need an error code as second argument, which I do not have right now. There's only one place where `SSL_get_error()` is called: https://github.com/xrootd/xrootd/blob/master/src/XrdHttp/XrdHttpProtocol.cc#L531

> What does "reuse connection" really mean? The socket stays open and connected and we start a new session on the existing socket? When Recycle() gets called the socket will definitely be closed unless it was marked as a shared socket which doesn't appear to ever happen.

Alright so thanks to that architecture, there is no need to call this SSL_Shutdown() twice as the connection will never be re-used. Thanks!

abh3: OK. Could you make sure the comment reflects this as I can envision this conversation happening again several years from now.

ccaffy: @abh3 done and documented :)

Issue #1969, opened 2023-03-20
xrdcrypto doesn't support certificate dates past 2049


rcarpa:
https://github.com/xrootd/xrootd/blob/c3afc27d72c9b0460bfac05967d6a7ea8f9fa99a/src/XrdCrypto/XrdCryptosslAux.cc#L718

only supports parsing [UTCTime](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.2.5.1), which is only used for dates from 1950 to 2049. 

If the server certificate expires after 2049, the server will not be able to start. 

If you are curious, and ask yourself why I even got confronted to this problem in the first place, I have to say a huge "thank you" to whatever person randomly selected a nicely looking days=9999 in our [script](https://github.com/rucio/rucio/blob/4ef457c25e4219da64086a52592b82ed570131b4/etc/certs/generate.sh#L41) which generates  the test certificates and made me lose a ton of time searching for non-existing problems after executing this script  :smiling_face_with_tear:

PR #1970, opened 2023-03-20
Possible fix for the time parsing related issue (#1969)

gganis:
Allow for GeneralizedTime (> year 2049) in certificates.
Fix also a bug in renormalizing the Year from UTCTime parsing.

### DIFF
--- a/src/XrdCrypto/XrdCryptosslAux.cc
+++ b/src/XrdCrypto/XrdCryptosslAux.cc
@@ -730,7 +730,14 @@ time_t XrdCryptosslASN1toUTC(const ASN1_TIME *tsn1)
        &(ltm.tm_year), &(ltm.tm_mon), &(ltm.tm_mday),
        &(ltm.tm_hour), &(ltm.tm_min), &(ltm.tm_sec),
                                       &zz) != 7) || (zz != 'Z')) {
-       return -1;
+       // Try GeneralizedTime
+       if ((sscanf((const char *)(tsn1->data),
+           "%04d%02d%02d%02d%02d%02d%c", 
+           &(ltm.tm_year), &(ltm.tm_mon), &(ltm.tm_mday),
+           &(ltm.tm_hour), &(ltm.tm_min), &(ltm.tm_sec),
+                                          &zz) != 7) || (zz != 'Z')) {
+           return -1;
+       }
    }
    // Init also the ones not used by mktime
    ltm.tm_wday  = 0;        // day of the week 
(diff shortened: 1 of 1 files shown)


src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -730,16 +730,26 @@ time_t XrdCryptosslASN1toUTC(const ASN1_TIME *tsn1)
...
                                       &zz) != 7) || (zz != 'Z')) {
-       return -1;
+       // Try GeneralizedTime
+       if ((sscanf((const char *)(tsn1->data),
+           "%04d%02d%02d%02d%02d%02d%c", 
+           &(ltm.tm_year), &(ltm.tm_mon), &(ltm.tm_mday),
+           &(ltm.tm_hour), &(ltm.tm_min), &(ltm.tm_sec),
+                                          &zz) != 7) || (zz != 'Z')) {
+           return -1;
+       }
```
amadio, line 740: Hi @gganis, I wonder if we shouldn't also handle the case with a timezone if we are touching this code (i.e. something like the third example [here](https://en.wikipedia.org/wiki/GeneralizedTime))? What do you think?

src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -730,16 +730,26 @@ time_t XrdCryptosslASN1toUTC(const ASN1_TIME *tsn1)
...
    ltm.tm_yday  = 0;        // day in the year
    ltm.tm_isdst = 0;        // we will correct with an offset without dst
    //
    // Renormalize some values: year should be modulo 1900
-   if (ltm.tm_year < 90)
+   if (ltm.tm_year < 50)
       ltm.tm_year += 100;
+   // In case GeneralizedTime is used
+   if (ltm.tm_year > 2000)
+      ltm.tm_year -= 1900;
```
amadio, line 752: I think we need `ltm.tm_year > 1900` here instead of `ltm.tm_year > 2000`, otherwise I see wrong dates (i.e. with year >3600) being generated.

amadio: In the end, I think this is what would give the most reasonable answers:

```cpp
   if (ltm.tm_year < 50)
      ltm.tm_year += 2000;
   else if (ltm.tm_year < 100)
      ltm.tm_year += 1900;

   ltm.tm_year -= 1900;
```
This works for years 00(2000) 01(2001) 49(2049) 51(1951) 1800 1900 1949 1951 2000, and 2100 without wrapping into 3000+.

amadio:
Thank you very much, @gganis!

PR #1971, opened 2023-03-21
[XrdHttp] Avoid sending empty header in the response as this conflict…

esindril:
…s with the reported content-length

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1528,7 +1528,7 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc, const char *hea
   if ((bodylen >= 0) && (code != 100))
     ss << "Content-Length: " << bodylen << crlf;
 
-  if (header_to_add)
+  if (header_to_add && (header_to_add[0] != '\0'))
     ss << header_to_add << crlf;
 
   ss << crlf;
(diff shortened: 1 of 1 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1528,7 +1528,7 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc, const char *hea
   if ((bodylen >= 0) && (code != 100))
     ss << "Content-Length: " << bodylen << crlf;
 
-  if (header_to_add)
+  if (header_to_add && strlen(header_to_add))
```
amadio, line 1531: A test for non-empty string probably suffices here, so how about `if(header_to_add && *header_to_add)`? It would be cheaper than `strlen`, which has the added problem that I can give you an enormous header and make this function very expensive.

theanalyst: Shouldn't if(header_to_add) already check for nullness? Also does bodylen already account for the `ss << crlf ` line we append in the beginninng or is it just a simple strlen of the body?

amadio: Yes, if you pass `nullptr`, but if you pass `""`, then this test will succeed, which is not what we want. With my suggestion, `""` will also not pass the test.

theanalyst: > Yes, if you pass `nullptr`, but if you pass `""`, then this test will succeed, which is not what we want. With my suggestion, `""` will also not pass the test.

Makes sense

esindril: Actually, now that you mention it probably it would be even better to protect against a huge header size being returned and it looks like the upper limit of the usual HTTP implementations is around 8kB - 16kB. If the header goes beyond this the server should probably reply: `413 Entity Too Large`. But this is outside the scope of this small patch and probably up for discussion with the rest of the devs.

amadio: `XrdHttpProtocol::StartChunkedResp` below (and maybe other functions) probably need the same treatment as this function.

esindril: I update the pull request to cover also the `StartChunkedResp` - I think these two are the only occurrences.

amadio:
@esindril Thanks!

PR #1985, opened 2023-03-31
Add Prometheus HTTP extension plugin

bbockelm:
With this work, there is a new built-in HTTP extension plugin that can be used to export OpenMonitoring / prometheus-based metrics.

An example output on a dev host:
```
$ curl -H "Authorization: Bearer $BEARER_TOKEN" http://localhost:8000/metrics/
# HELP server_bytes Number of bytes read into the server
# TYPE server_bytes counter
server_bytes{direction="tx"} 882
server_bytes{direction="rx"} 3107
# HELP server_metadata XRootD server metadata
# TYPE server_metadata gauge
server_metadata{instance="f4hp7ql65f.local:1095",site="bbockelm_laptop",version="v5.5.4.post     248"} 1680294742
```

I am told by @ivukotic it would be immensely helpful in "collapsing" the monitoring pipeline.  His central Prometheus monitoring could work directly with the caches instead of going through the XRootD/UDP->TCP->Collector->LogStash->database (or in the case of OSG, XRootD/UDP->Shoveler/AMQ->Collector/AMQ->database).

Beyond shortening the chain it also simplifies the authorization scheme -- only the central prometheus instance needs a token.  The caches just need to be configured to trust the central service.

Note this includes a dependency on https://github.com/jupp0r/prometheus-cpp; the build of the module should be disabled if the library is missing.  This only needs the `prometheus-cpp-core` library; it might be quite reasonable to think about vendoring it.

@jthiltges @djw8605 @biozit @brianhlin -- you guys might be interested in this as well.


bbockelm:
@amadio -- I think I am going to close this one for now as I'm trying to clean out some backlog.

We've moved our Prometheus integration out-of-process.  Would have been quite nice to have direct support within Xrootd but I think this is going to have to sit on the backburner until someone has time to pick it up again.

PR #1994, opened 2023-04-06
[XrdApps] Let XrdClProxyPlugin work with PgRead

### LINKED ISSUE #1993: xrdcp fails when redirected for current versions
Hi all,
We currently have problems with xrdcp for our local resources.
We have a XRootD server and redictor on a ceph node to provide local datasets of our users to the (external) worker nodes of our T3 via XRootD.

**Setup**:
Ceph configs: [data server](https://etpwww.etp.kit.edu/~rhofsaess/xrootd-ceph-ds.cfg),  [redirector](https://etpwww.etp.kit.edu/~rhofsaess/xrootd-ceph-mgr.cfg)

On the worker node, a caching proxy and a redirector are running. Configs: [proxy](https://etpwww.etp.kit.edu/~rhofsaess/xrootd-cache-server.cfg), [redirector](https://etpwww.etp.kit.edu/~rhofsaess/xrootd-cache-manager.cfg)

The client is configured to use the redirector. Config: [client-plugin-proxy.conf](https://etpwww.etp.kit.edu/~rhofsaess/client-plugin-proxy.conf)

**What we observe**
xrdcp fails for current versions (tested with: 5.5.1)

What we see is that the final url is build up correctly (`final_url=root://f03-001-110-e.gridka.de:1094//root://ceph-node-j.etp.kit.edu:1094//mschnepf/data_mnist-tar.gz`), but the transfer somewhat fails with the error:

```
Run: [ERROR] Operation is not implemented:  (source)
```
Full log: [transfer_5-5-1.log](https://etpwww.etp.kit.edu/~rhofsaess/transfer_5-5-1.log)

With older versions of xrootd, however, our setup works fine, see [transfer_5-3-1.log](https://etpwww.etp.kit.edu/~rhofsaess/transfer_5-3-1.log).

I tracked it down to the client(config) or the redirector.
If I do not use the redirector but ask the server directly, it works with a current xrootd version (5.5.4). [Log: transfer_5-5-4-without.log](https://etpwww.etp.kit.edu/~rhofsaess/transfer_5-5-4-without.log)

What I also observed was that changing the major version in "lib =libXrdClProxyPlugin-<V>.so" in the client plugin conf
does not have any influence. When changing it to V=1000, the redirect still works (and it still fails :D).
Just wanted to make sure that this behaviour is intended.

Maybe also our configs may aren't compatible with newer versions. If so, thanks for any help in advance to fix it! :)

Thanks and cheers

smithdh:
Hello. To fix #1993, a missing specialisation for PgRead is added to ProxyPrefixFile (in the XrdClProxyPlugin).
This patch also adds a few other missing method overrides. i.e.

(i) Added missing override of PgRead, Write (1 already present, 2 other overloads added), PgWrite, VectorWrite, WriteV.
(ii) Did not add override for Read(6 argument version); present in the base (XrdCl::FilePlugIn) but not called from XrdClFile. Method appears unused, possibly could be removed in release 6. (The XrdClPlugInInterface.hh is public so can't change before then).
(iii) Likely we're missing ReadV, but this is not present in the base. The ReadV() interface was added after release of xrootd5, so maybe ReadV can be added to XrdClPlugInInterface for release 6.
(iv) Added override specifier on declarator of overridden methods in ProxyPrefixFile.hh. (c++11)

### DIFF
--- a/src/XrdApps/XrdClProxyPlugin/ProxyPrefixFile.hh
+++ b/src/XrdApps/XrdClProxyPlugin/ProxyPrefixFile.hh
@@ -26,6 +26,8 @@
 #include "XrdCl/XrdClDefaultEnv.hh"
 #include "XrdCl/XrdClPlugInInterface.hh"
 
+#include <utility>
+
 using namespace XrdCl;
 
 namespace xrdcl_proxy
@@ -45,7 +47,7 @@ public:
   //----------------------------------------------------------------------------
   //! Destructor
   //----------------------------------------------------------------------------
-  virtual ~ProxyPrefixFile();
+  virtual ~ProxyPrefixFile() override;
 
   //----------------------------------------------------------------------------
   //! Open
@@ -54,13 +56,13 @@ public:
                             OpenFlags::Flags flags,
                             Access::Mode mode,
                             ResponseHandler* handler,
-                            uint16_t timeout);
+                            uint16_t timeout) override;
 
   //----------------------------------------------------------------------------
   //! Close
   //----------------------------------------------------------------------------
   virtual XRootDStatus Close(ResponseHandler* handler,
-                             uint16_t         timeout)
+                             uint16_t         timeout) override
   {
     return pFile->Close(handler, timeout);
   }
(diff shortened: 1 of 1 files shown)


amadio:
Looks good to me. @abh, please ack and I'll merge.

abh3:
Looks fine to me. Thanks David -- that was super-fast!

PR #1997, opened 2023-04-13
Fix warnings from Clang compiler

amadio:


src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -51,7 +51,7 @@ XrdCl::URL *XrdPosixAdmin::FanOut(int &num)
    XrdCl::URL                    *uVec;
    XrdNetAddr netLoc;
    const char *hName;
-   unsigned int i;
+   unsigned long i;
```
abh3, line 54: Why does Clang need this to be a long?

abh3 (commented):
OK, I do know Clang wants override but why only this set of methods. We haven't used override (though we should) in most other classes simply because that code predates "override". Also one other question about an oddity.

src/XrdPosix/XrdPosixAdmin.cc
amadio: It's not clang that needs it, it just warns that the condition `i > std::numeric_limits<ptrdiff_t>::max() / sizeof(XrdCl::URL)` in the [`if` below](https://github.com/xrootd/xrootd/pull/1997/files#diff-71e5adaf610226018b09990d066c0888d3f3ed551809561d057b4cecf6214da0R71) is never going to trigger because `std::numeric_limits<ptrdiff_t>::max() / sizeof(XrdCl::URL)` is much larger than the largest `int`. We are checking there that we are not trying to allocate more memory (i.e. `i * sizeof(XrdCl::URL)`) than `std::numeric_limits<ptrdiff_t>::max()` bytes, but we can only do so if `i` can actually hold up to that value itself. The "solution" is to make `i` big enough for that.

amadio:
> OK, I do know Clang wants override but why only this set of methods. We haven't used override (though we should) in most other classes simply because that code predates "override". Also one other question about an oddity.

GCC doesn't warn about it, but adding override informs the compiler about what we want to do and if the overriden/overriding method's signature is wrong or if the method we are overriding is not virtual, it will fail to compile. The commit adds the override specifier in the few places where it was expected but was not there yet. The errors shown in [this example](https://en.cppreference.com/w/cpp/language/override#Example) in cppreference show well the kind of checks we get by adding the override specifier.

abh3 (approved):

PR #2000, opened 2023-04-17
Use consistently SHA-256 for signatures (fixes #1992)

### LINKED ISSUE #1992: xrdgsiproxy doesn't work by default in alma9 due to sha1 signature
The proxies created by `xrdgsiproxy` use a SHA1 signature. As a result, they don't work in alma9, which refuses SHA1 signature by default: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/security_hardening/using-the-system-wide-cryptographic-policies_security-hardening#doc-wrapper

```
# xrdgsiproxy init -cert /tmp/cert.pem -key /tmp/key.pem 
: problems creating proxy
```

But it works after allowing SHA1:

```
# sudo update-crypto-policies --set DEFAULT:SHA1
Setting system policy to DEFAULT:SHA1
Note: System-wide crypto policies are applied on application start-up.
It is recommended to restart the system for the change of policies
to fully take place.
# xrdgsiproxy init -cert /tmp/cert.pem -key /tmp/key.pem 
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
file        : /tmp/x509up_u0
type        : Proxy (RFC)
issuer      : /CN=Rucio User
subject     : /CN=Rucio User/CN=1547110546
path length : 0
bits        : 2048
time left   : 12h:0m:0s
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

gganis:
On systems honouring, by default, the deprecation of SHA1, such as AlmaLinux 9, xrdgsiproxy fails (see issue https://github.com/xrootd/xrootd/issues/1992).
This patch replaces use of SHA1 by SHA-256 for signatures. SHA-256 is supported by all OpenSSL versions still around,
which makes it easier from point of view of portability. Move to SHA3 should be considered for the future.

Fixes #1992.


amadio:
Thank you, Gerri!

PR #2006, opened 2023-05-04
Fix problem creating ZIP archive with lots of entries

### LINKED ISSUE #2004: Trouble creating a ZIP archive with lots of entries
I am chasing an issue that I initially saw when testing XrdEc. I have since come up with a standalone testing case that is not related to using XrdEc. The test code and instruction to compile the code is pasted at the bottom. Note that

1. Once compile, pls run against a xrootd server to created a new ZIP archive
2. The program will never finish. If turning on XRD_LOGEVEL=Error, I see message like the following after a while:
[2023-04-30 21:13:01.694793 -0700][Error  ][AsyncSock         ] [cent7b:1094.0] Socket error encountered: [ERROR] Socket error: connection reset by peer
[2023-04-30 21:13:01.694852 -0700][Error  ][XRootD            ] [cent7b:1094] Impossible to send message kXR_writev (handle: 0x00000000, chunks: 1, total size: 2097211). Trying to recover.
[2023-04-30 21:13:01.694863 -0700][Error  ][XRootD            ] [cent7b:1094] Unable to get the response to request kXR_writev (handle: 0x00000000, chunks: 1, total size: 2097211)

3. it can not be compiled with devel rpm because of "#include <XrdCl/XrdClCtx.hh>". This is what is used in XrdEc.
4. This issue is sensitive to the setting of MAX_FILES_IN_ZIP, size_t fsize and std::string fn in the testing code. If we reduce MAX_FILES_IN_ZIP by 1 (while keeping "fsize" and "fn" unchanged), all will work just fine. 
5. It is possible that the MAX_FILES_IN_ZIP setting is related to my environment. You may need to set it to something much larger to see this issue. 

---------------------------------------

/* How to compile
 scl enable devtoolset-9 sh
 g++ -g -o xrdzip_write.exe xrdzip_write.cc \
     -I/afs/slac/package/xrootd/xrootd-5.5.1/xrootd/src \
     -I/afs/slac/package/xrootd/xrootd-5.5.1/amd64_rhel70/src \
     -Wl,--copy-dt-needed-entries \
     -L/afs/slac/package/xrootd/xrootd-5.5.1/amd64_rhel70/src \
     -L/afs/slac/package/xrootd/xrootd-5.5.1/amd64_rhel70/src/XrdCl \
     -l XrdCl  
*/

#include <stdio.h>
#include <string>
#include <XrdCl/XrdClXRootDResponses.hh>
#include <XrdCl/XrdClURL.hh>
#include <XrdCl/XrdClFile.hh>
#include <XrdCl/XrdClCtx.hh>
#include <XrdCl/XrdClFileSystem.hh>
#include <XrdCl/XrdClZipArchive.hh>
#include <XrdCl/XrdClZipOperations.hh>
#include <XrdOuc/XrdOucCRC32C.hh>

#define MAX_FILES_IN_ZIP 36251

int main(int argc, char* argv[]) {
    XrdCl::URL url(argv[1]);
    XrdCl::XRootDStatus st;
    
    size_t fsize = 128*1024;
    char *buff = (char*)malloc(fsize);
    uint32_t cksm = crc32c(0, buff, fsize);

    XrdCl::ZipArchive zip;
    XrdCl::Ctx<XrdCl::ZipArchive> pzip = zip;

    st = XrdCl::WaitFor(XrdCl::OpenArchive(pzip, url.GetURL(), XrdCl::OpenFlags::New | XrdCl::OpenFlags::Update));
    if (!st.IsOK()) {
        printf("ZipArchive Open fail\n");
        exit(1);
    }
    else
        printf("ZipArchive Open successfully\n");

    for (int i=0; i<MAX_FILES_IN_ZIP; i++) {
        std::string fn = "obj." + std::to_string(i) + ".1";
        printf("Open file %s : ", fn.c_str());
        st = XrdCl::WaitFor(XrdCl::AppendFile(pzip, fn, cksm, fsize, buff));
        if (!st.IsOK()) {
           printf("fail %s\n", st.ToString().c_str());
           exit(1);
        }
        else
           printf("successfully\n");
    }

    printf("closing ZipArchive ...\n");
    st = XrdCl::WaitFor(XrdCl::CloseArchive(pzip));
    if (!st.IsOK()) {
        printf("ZipArchive Close fail\n");
        exit(1);
    }
    else
        printf("ZipArchive Close successfully\n");

    exit(0);
}

smithdh:
Intended to fix #2004. Initially in draft.

### DIFF
--- a/src/XrdCl/XrdClOperations.cc
+++ b/src/XrdCl/XrdClOperations.cc
@@ -210,6 +210,32 @@ namespace XrdCl
     final = std::move( f );
   }
 
+  //------------------------------------------------------------------------
+  // Called by a pipeline on the handler of its first operation before Run
+  //------------------------------------------------------------------------
+  void PipelineHandler::PreparePipelineStart()
+  {
+    // Move any final-function from the handler of the last operaiton to the
+    // first. It will be moved along the pipeline of handlers while the
+    // pipeline is run.
+
+    if( final || !nextOperation ) return;
+    PipelineHandler *last = nextOperation->handler.get();
+    while( last )
+    {
+      Operation<true> *nextop = last->nextOperation.get();
+      if( !nextop ) break;
+      last = nextop->handler.get();
+    }
+    if( last )
+    {
+      // swap-then-move rather than only move as we need to guarantee that
+      // last->final is left without target.
+      std::function<void(const XRootDStatus&)> f;
+      f.swap( last->final );
+      Assign( std::move( f ) );
+    }
+  }
 
   //------------------------------------------------------------------------
   // Stop the current pipeline
--- a/src/XrdCl/XrdClOperations.hh
+++ b/src/XrdCl/XrdClOperations.hh
@@ -123,6 +123,11 @@ namespace XrdCl
       //------------------------------------------------------------------------
       void Assign( std::function<void(const XRootDStatus&)>  final );
 
+      //------------------------------------------------------------------------
+      //! Called by a pipeline on the handler of its first operation before Run
+      //------------------------------------------------------------------------
+      void PreparePipelineStart();
+
     private:
 
       //------------------------------------------------------------------------
@@ -487,6 +492,10 @@ namespace XrdCl
         if( !operation ) std::logic_error( "Empty pipeline!" );
 
         Operation<true> *opr = operation.release();
+        PipelineHandler *h = opr->handler.get();
+        if( h )
+          h->PreparePipelineStart();
+
         opr->Run( timeout, std::move( prms ), std::move( final ) );
       }
 
--- a/src/XrdCl/XrdClUtils.cc
+++ b/src/XrdCl/XrdClUtils.cc
@@ -865,4 +865,54 @@ namespace XrdCl
       if( dst_supported.count( *itr ) ) return *itr;
     return std::string();
   }
+
+  //----------------------------------------------------------------------------
+  //! Split chunks in a ChunkList into one or more ChunkLists
+  //----------------------------------------------------------------------------
+  void Utils::SplitChunks( std::vector<ChunkList> &listsvec,
+                           const ChunkList        &chunks,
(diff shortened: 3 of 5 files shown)


abh3:
Let me know when this is ready for review.

smithdh:
Thanks Andy. I'd been making some last changes; but it's ready, I'll move it out of draft now, so it could be reviewed. I expect depending on the feedback we can consider this, or not, for including in 5.5.5 (since that's targeted for tagging tomorrow).

abh3 (commented):
I don't see anything wrong here. The question is does this use checkpointing? If so, perhaps one needs to enable checkpointing on the server and see if another problem crops up, The checkpointing allows additions to the zip archive in-place.

abh3:
Oh yes. while I didn't specifically approve, I am OK if you feel that my comment is not applicable. Just ask Guilherme to merge it.

smithdh:
(I just made a force-push, only changed a typo in code comment).

Thank you. I just checked that it does appear to work on a server with checkpointing enabled (in the no error case); infact appending against an existing archive fails against a server without checkpointing enabled. But, following on from your comment, we could verify more. e.g. that the checkpointing is giving the protection we want: keeping the archive valid if there's a problem during addition of files (I saw we don't use the kXR_ckpXeq->kXR_write sequenced version of write during CloseArchive and I'm not sure if that will properly preserve integrity if the checkpoint changes are discarded; and I didn't try to change that in this PR).

I'll suggest to Guilherme to merge, so we can address #2004 now.

abh3:
Interesting. If you know what you are doing you don't need to subsequently 
use CkpXeq. The idea here is to protect the index of the zipfile. If we 
overwrite that then it gets checkpointed. After that it's all appends and 
checkpointing isn't explcitly necessary (assuming the whole index was 
overwritten). Upon failure all changes are discarded and the saved data is 
restored. I suppose we should independently verify that this actually is 
what is happening.

PR #2008, opened 2023-05-12
Remove xrootd-ceph git submodule and bring the code back to the main repository

amadio:
This is the last git submodule we have. If there are no objections, I'd like to bring it back to the main repository and handle RPM/packaging in a way that doesn't require the split, as it complicates releases and the last several versions of xrootd-ceph actually have no changes. I am bringing the code back as a subtree so the history is preserved.

@simonmichal Could you please remind us who asked to split this, so we can add them here for review? Thank you!


amadio:
@abh3 What are your thoughts here?

abh3:
So, I've seen actions on this but I figured that you already decided that Ceph was ready to be part of the repo. So, I don't quite understand what my thoughts should be. Historically, we never put it into the repo because Ceph source was quite unstable and the plugin-had to be built with whatever the site was deploying which,, for us,, was unknown. So, could you add more context to the question?

snafus:
Hi,
  To jump in here from the RAL perspective. xroot-ceph package is under active development under the fork https://github.com/stfc/xrootd-ceph. There has always been the intention to push the changes from this repo back to the main repo under xrootd, but we had not yet reached a version to consider as 'fully complete' (for want of a better phrase). 
  
I suspect that if xrootd-ceph submodule was incorporated back into xrootd, it would  make this process more difficult (and for future changes)? Perhaps this needs further discussion?
James

abh3:
I see @amadio issue here that this change would remove some extra steps in the release process. On the other hand, if we don't automatically build Ceph because we don't know what to build against then a site would have to build the complete xrootd package to just get the Ceph part built (unless we provide an option to only build the Ceph part). As for making it difficult in the future not exactly so. For your first update after including Ceph in the repo you will need to generate a patch file and apply it to a clone of the xrootd repo (it should be clean as you are the only one modifying the code) and the create a pull request. After that, your changes would be made in the cloned xrootd repo and that step is pretty straightforward. @amadio could you comment on whether this is all correct?

amadio:
> To jump in here from the RAL perspective. xroot-ceph package is under active development under the fork https://github.com/stfc/xrootd-ceph. There has always been the intention to push the changes from this repo back to the main repo under xrootd, but we had not yet reached a version to consider as 'fully complete' (for want of a better phrase).

I see. If it is actively developed in your fork, I'd like to discuss possible ways for us to be able to avoid our version becoming stale. We release RPMs for xrootd-ceph based on the repository at https://github.com/xrootd/xrootd-ceph. Just as a test, I rebased the master branch from https://github.com/stfc/xrootd-ceph onto the master from our repository and added the changes to this pull request. The differences with your master branch are just a couple of small extra fixes we have on our side.

> I suspect that if xrootd-ceph submodule was incorporated back into xrootd, it would make this process more difficult (and for future changes)? Perhaps this needs further discussion?
James

Rebasing your master branch on top of ours was not a problem, as it's done above. If you have active development on other branches, we can discuss how to incorporate them too. However, I think that it would be interesting to discuss how we can better coordinate things so that we keep the main repository as up to date as possible with external developments. If we use a git subtree as I am doing in this merge, you should be able to keep developing on your fork, then propose changes back to us with normal pull requests later. If you are available, you could join our [development meeting](https://indico.cern.ch/event/1259794/) today. If you cannot join, we can set another date/time to discuss things further.

amadio:
> I see @amadio issue here that this change would remove some extra steps in the release process. On the other hand, if we don't automatically build Ceph because we don't know what to build against then a site would have to build the complete xrootd package to just get the Ceph part built (unless we provide an option to only build the Ceph part). As for making it difficult in the future not exactly so. For your first update after including Ceph in the repo you will need to generate a patch file and apply it to a clone of the xrootd repo (it should be clean as you are the only one modifying the code) and the create a pull request. After that, your changes would be made in the cloned xrootd repo and that step is pretty straightforward. @amadio could you comment on whether this is all correct?

Hi @abh, it's mostly correct, as an exercise, I've added changes from their master branch to this pull request, and it was not difficult, just had a small and easy to resolve conflict. What I'd like to have as a goal is to have a single source of truth for everybody, and make new developments appear in our released RPMs so everybody can benefit. As the main user of xrootd-ceph, RAL may have features under development on their fork, but if they agree to contribute them back as merge requests for each feature release when they are ready for production, I think we can avoid the fork diverging too much from the main repository.

@snafus What do you think?

PR #2012, opened 2023-05-21
Fix build failure with latest glibc

ellert:
The latest updates to glibc headers adds __nonnull to the fclose()

Declaration in old version:

    extern int fclose (FILE *__stream);

Declaration in new version:

    extern int fclose (FILE *__stream) __nonnull ((1));

This change causes a compilation error in xrootd:
~~~
/builddir/build/BUILD/xrootd-5.5.5/src/XrdTls/XrdTlsTempCA.cc:54:48: error: ignoring attributes on template argument 'int (*)(FILE*)' [-Werror=ignored-attributes]
   54 | typedef std::unique_ptr<FILE, decltype(&fclose)> file_smart_ptr;
      |                                                ^
~~~
This commit rewrites the code so the error is not triggered.

### DIFF
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -50,8 +50,8 @@
 #include <atomic>
 
 namespace {
-    
-typedef std::unique_ptr<FILE, decltype(&fclose)> file_smart_ptr;
+
+typedef std::unique_ptr<FILE, int(*)(FILE*)> file_smart_ptr;
 
 
 static uint64_t monotonic_time_s() {


src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -50,8 +50,9 @@
 #include <atomic>
 
 namespace {
-    
-typedef std::unique_ptr<FILE, decltype(&fclose)> file_smart_ptr;
+
+static void FClose(FILE *f) {fclose(f);}
```
amadio, line 54: Is adding this really necessary? I wonder if just adjusting the typedef to use `void(*)(FILE*)` instead of `decltype(&fclose)` is already enough.

ellert: Yes, that works too. I have changed the PR accordingly.

bbockelm:
@amadio - do you know if there are any recent C++ improvements for code patterns like this?  This sort of smart pointer trickery makes me miss the `defer` keyword in golang.

amadio:
I think it may be possible to use something like [`std::decay`](https://en.cppreference.com/w/cpp/types/decay) in a case like this, but using the actual type is also fine, I guess.

PR #2016, opened 2023-05-26
XrdHttp: The deletion of a non-empty directory returns a 405 error code instead of 500

ccaffy:
The mapping between errno error codes and XRoot protocol error code has also been changed. ENOTEMPTY is mapped to kXR_ItExists

### DIFF
--- a/src/XProtocol/XProtocol.hh
+++ b/src/XProtocol/XProtocol.hh
@@ -1370,7 +1370,12 @@ static int mapError(int rc)
            case ENOTBLK:       return kXR_NotFile;
            case ENOTSUP:       return kXR_Unsupported;
            case EISDIR:        return kXR_isDirectory;
-           case EEXIST:        return kXR_ItExists;
+           case ENOTEMPTY: [[fallthrough]];
+           // In the case one tries to delete a non-empty directory
+           // we have decided that until the next major release
+           // the kXR_ItExists flag will be returned
+           case EEXIST:
+                return kXR_ItExists;
            case EBADRQC:       return kXR_InvalidRequest;
            case ETXTBSY:       return kXR_inProgress;
            case ENODEV:        return kXR_FSError;
@@ -1391,10 +1396,6 @@ static int mapError(int rc)
            case ETIMEDOUT:     return kXR_ReqTimedOut;
            case EBADF:         return kXR_FileNotOpen;
            case ECANCELED:     return kXR_Cancelled;
-           // In the case one tries to delete a non-empty directory
-           // we have decided that until the next major release
-           // the kXR_ItExists flag will be returned
-           case ENOTEMPTY:     return kXR_ItExists;
            default:            return kXR_FSError;
           }
       }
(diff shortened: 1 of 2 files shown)


src/XProtocol/XProtocol.hh
```diff
@@ -1391,6 +1391,10 @@ static int mapError(int rc)
            case ETIMEDOUT:     return kXR_ReqTimedOut;
            case EBADF:         return kXR_FileNotOpen;
            case ECANCELED:     return kXR_Cancelled;
+           // In the case one tries to delete a non-empty directory
+           // we have decided that until the next major release
+           // the kXR_ItExists flag will be returned
+           case ENOTEMPTY:     return kXR_ItExists;
```
abh3, line 1397: What should happen here is that this case needs to be put ahead if the kXR_ItExists case as a falthrough.  That is
using [[fallthrough]]; to indicate that the fallthrough is correct.

abh3 (changes requested):

src/XProtocol/XProtocol.hh
ccaffy: I do not understand :thinking: Is it a fallthrough case? I just return kXR_ItExists when the errno code is ENOTEMPTY. Each switch case has a return value, so there's no need for a `[[fallthrough]];` attribute, isn't it?

abh3:
Yes, it's a fallthrough case which makes it much more obvious what is 
happening. Don't forget to specify [[falthrough]]

abh3 (changes requested):
I don't think I really explained myself here.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -926,7 +926,14 @@ void XrdHttpReq::mapXrdErrorToHttpStatus() {
...
         break;
       case kXR_ItExists:
-        httpStatusCode = 409; httpStatusText = "File already exists";
+        if(request != ReqType::rtDELETE) {
+          httpStatusCode = 409; httpStatusText = "File already exists";
+        } else {
+          // In the case the XRootD layer returns a kXR_ItExists after a deletion
+          // was submitted, we return a 405 status code with the error message set by
+          // the XRootD layer
+          httpStatusCode = 405;
```
amadio, line 935: What about `httpStatusText` here?

src/XProtocol/XProtocol.hh
```diff
@@ -1370,7 +1370,11 @@ static int mapError(int rc)
            case ENOTBLK:       return kXR_NotFile;
            case ENOTSUP:       return kXR_Unsupported;
            case EISDIR:        return kXR_isDirectory;
-           case EEXIST:        return kXR_ItExists;
+           case EEXIST:        [[fallthrough]];
+           // In the case one tries to delete a non-empty directory
+           // we have decided that until the next major release
+           // the kXR_ItExists flag will be returned
+           case ENOTEMPTY:     return kXR_ItExists;
```
amadio, line 1377: This is really just nitpicking, so feel free to ignore, but I'd put the `return kXR_ItExists;` on `case EEXIST`, and put `case ENOTEMPTY` as the fall through.

src/XrdHttp/XrdHttpReq.cc
ccaffy: The httpStatusText here is just inherited from the XRootD layer, so we don't modify it

src/XProtocol/XProtocol.hh
ccaffy: Yes, I'll do that :)

ccaffy:
My pleasure :)

Issue #2018, opened 2023-05-26
[XrdHttp] Error 500 response when issuing GET with checksum request on a file that does not exist


ccaffy:
Interesting :)

When I issued a HEAD request with a `Want-Digest: adler32` header on a file that does not exist,  I properly get 404 Not Found Error.

```
$ curl -H 'Want-Digest: sha-512' -I https://something.cern.ch:1096/tmp/non-empty-dir/does_not_exist.txt
HTTP/1.1 404 Not Found
Connection: Close
Server: XrootD/v5.2.0-99-osghotfix...687
Content-Length: 0
```
But if I do a GET with a 'Want-Digest: sha-512' header on a file that does not exist:

```
$ curl -H 'Want-Digest: sha-512' -X GET https://something.cern.ch:1096/tmp/non-empty-dir/does_not_exist.txt
HTTP/1.1 500 Internal Server Error
Connection: Close
Server: XrootD/v5.2.0-99-osghotfix...687
Content-Length: 51

Underlying filesystem failed to calculate checksum.
```

abh3:
Could you try the same test using xroot protocol to see where the problem 
really lies?

ccaffy:
I spotted this "Ugly hack": https://github.com/xrootd/xrootd/blob/master/src/XrdHttp/XrdHttpReq.cc#L2096

If the server is NOT a manager, then the error on stat() is ignored... Will investigate further and look on how to NOT modify this behaviour ;)

abh3:
It may be that Fabrizio knows the history behind that hack.

ccaffy:
Yes, I asked him directly. My fix has been done according to what he said to me ;)

PR #2019, opened 2023-05-30
[XrdHttp] A GET with 'Want-Digest' header request issued on a non-existing file will return a 404 error instead of 500

ccaffy:
Fixes https://github.com/xrootd/xrootd/issues/2018.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -1805,7 +1805,7 @@ XrdHttpReq::PostProcessChecksum(std::string &digest_header) {
     if (convert_to_base64) {free(digest_value);}
     return 0;
   } else {
-    prot->SendSimpleResp(500, NULL, NULL, "Underlying filesystem failed to calculate checksum.", 0, false);
+    prot->SendSimpleResp(httpStatusCode, NULL, NULL, httpStatusText.c_str(), httpStatusText.length(), false);
     return -1;
   }
 }


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1798,7 +1798,16 @@ XrdHttpReq::PostProcessChecksum(std::string &digest_header) {
...
+    if(xrderrcode != kXR_NotFound) {
+      prot->SendSimpleResp(500, NULL, NULL, "Underlying filesystem failed to calculate checksum.", 0, false);
+    } else {
+      // This is a result of the "Ugly" hack put in place to prevent a failing stat() on a
+      // non-manager server to fail a transfer.
+      // In the case a user issues a GET with a 'Want-Digest' header on a non-existing file, the non-failing
+      // stat() will result in a failed checksum query giving a 500 error back to the user.
+      // Here we just return a 404 instead. httpStatusCode and httpStatusText contain the right reason why the request failed.
+      prot->SendSimpleResp(httpStatusCode, NULL, NULL, httpStatusText.c_str(), httpStatusText.length(), false);
+    }
```
amadio, line 1810:
```suggestion
    // This is a result of the "Ugly" hack put in place to prevent a failing stat() on a
    // non-manager server to fail a transfer.
    // In the case a user issues a GET with a 'Want-Digest' header on a non-existing file, the non-failing
    // stat() will result in a failed checksum query giving a 500 error back to the user.
    // Here we just return a 404 instead. httpStatusCode and httpStatusText contain the right reason why the request failed.
    if(xrderrcode != kXR_NotFound) {
      httpStatusCode = 500;
      httpStatusText = "Underlying filesystem failed to calculate checksum.";
    }
        
    prot->SendSimpleResp(httpStatusCode, NULL, NULL, httpStatusText.c_str(), httpStatusText.length(), false);
    }
```

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1798,7 +1798,16 @@ XrdHttpReq::PostProcessChecksum(std::string &digest_header) {
     if (convert_to_base64) {free(digest_value);}
     return 0;
   } else {
-    prot->SendSimpleResp(500, NULL, NULL, "Underlying filesystem failed to calculate checksum.", 0, false);
+    // This is a result of the "Ugly" hack put in place to prevent a failing stat() on a
+    // non-manager server to fail a transfer.
+    // In the case a user issues a GET with a 'Want-Digest' header on a non-existing file, the non-failing
+    // stat() will result in a failed checksum query giving a 500 error back to the user.
+    // Here we just return a 404 instead. httpStatusCode and httpStatusText contain the right reason why the request failed.
+    if(xrderrcode != kXR_NotFound) {
```
abh3, line 1806: I'm OK with the way it is but you could be a bit more discerning here. For instance, if the error is kXR_Unsupported then the status code should really be 501 with the text as returned by the server. Which leads to the question why override the server supplied error text with a generic one? Is the original error message lost somewhere? One would expect that the server's error message will better explain how to correct the problem.

abh3 (commented):
Like I said, I'm OK with this though I do have a question that may or may not lead to another change. So, if  @amadio is OK with it so am I.

src/XrdHttp/XrdHttpReq.cc
amadio: Overriding is what used to be done, so I understand @ccaffy's choice. However, I agree that your suggestion is an improvement. I suppose we could have something like
```cpp
switch(xrderrcode) {
case kXR_NotFound:
    break; // reuse error code and message already set
case kXR_Unsupported:
    httpStatusCode = 501;
    break;
default:
    httpStatusCode = 500;
    httpStatusText = "Underlying filesystem failed to calculate checksum.";
}
prot->SendSimpleResp(httpStatusCode, NULL, NULL, httpStatusText.c_str(), httpStatusText.length(), false);
```

xrootd-dev:
Well, yes, but the question remains why isn't the server's error text 
promoted to the repsonse. Should that not be the actual explanation as 
opposed to substituting a generic response?

ccaffy:
It was just that it was done like that. But I agree, let's then just use the server error text that is set by the XRootD layer, it will simplify the code!

abh3:
Excellent!

ccaffy:
> @ccaffy Very nice, but could you squash the two commits into a single one? Thanks!

Done :)

amadio (approved):

src/XrdHttp/XrdHttpReq.cc
ffurano: uhmmmmm wait please, IMO there may be another corner case, let's talk

ccaffy: After the chat with @ffurano , we are good to go! @amadio you may merge this fix whenever you want :)

Cheers,
Cedric

PR #2025, opened 2023-06-05
Modernize Python bindings build system

amadio:
This is a rewrite of the build system of the Python bindings, to solve some of the open tickets we have.

@henryiii Could you please review these changes? I'd appreciate your input.

@adriansev Could you also have a look to check that it would work for your use cases? Some of the open tickets this solves were filed by you. Thank you!


bindings/python/project.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
+requires = ["setuptools", "wheel"]
```
matthewfeickert, line 2: Shouldn't this be `pyproject.toml` and not `project.toml`? Or are you doing something different that I don't know about?

```suggestion
[build-system]
requires = ["setuptools>=61.0"]
build-backend = "setuptools.build_meta"
```

If you are using a `pyproject.toml` based system in keeping with [PEP 518](https://peps.python.org/pep-0518/) then `build-backend` should be included. In https://scikit-hep.org/developer/pep621#pyprojecttoml-build-system we recommend a modern 

```toml
[build-system]
requires = ["setuptools>=61.0"]
build-backend = "setuptools.build_meta"
```

but maybe there are limitations on this version of `setuptools` for XRootD?

project.toml
```diff
@@ -0,0 +1,2 @@
+[build-system]
```
matthewfeickert, line 1: Same comments as above. Though if it is possible to get rid of the duplication of toml files in this PR that's an even better win.

matthewfeickert (commented):
Awesome @amadio! I just scanned this now, so this isn't really a review, but just some things that hit me right away.

bindings/python/project.toml
henryiii: >=42 is fine if you need Python 2 support.

bindings/python/setup.py
```diff
@@ -0,0 +1,143 @@
+import os
+import platform
+import subprocess
+import sys
+
+from setuptools import setup, Extension
+from setuptools.command.build_ext import build_ext
+from subprocess import check_call, check_output
+from distutils.spawn import find_executable
```
matthewfeickert, line 9: Please don't put `distutils` back in. This is deprecated behavior.

In general it seems like this is a lot of extra work to use `setuptools` when `scikit-build-core` already exists. https://scikit-build-core.readthedocs.io/

bindings/python/project.toml
amadio: Yeah, pyproject.toml, that's why I'm asking for your review :-) It works as is, but the misnaming was an oversight. I will fix and push a new version.

project.toml
amadio: Yes, I thought the same, even using a `setup.cfg` for metadata deduplication as well. I will probably use a symlink from bindings/python and keep just the top one as a regular file. For setup.py that is not possible without some adding some more hacks, so that I'd like to keep separate because the setup of RPATH and CMake flags propagation is different.

bindings/python/setup.py
henryiii: I expect this is here instead of `shutil.which` (and `scikit-build-core` is not used) due to the desire to support Python 2. I'd at least make this conditional on `sys.version_info`, though!

henryiii: Classic scikit-build does support Python 2, though :wink:

amadio: I'd like to avoid a dependency on scikit-build if possible, but I don't mind removing distutils if there's an alternative way, I'm only using `find_executable`, so should be easy to find a good alternative.

henryiii: `shutil.which` replaces `distutils.spawn.find_executable`, Python 3 only.

bindings/python/project.toml
amadio: For 5.x we still need to keep support for Python 2, but when 6.0 comes out, we will drop support for Python 2.x.

bindings/python/setup.py
matthewfeickert: c.f. https://github.com/xrootd/xrootd/commit/80890de66f66208346c3f3e81aca8209fb60a852 from https://github.com/xrootd/xrootd/pull/1585

henryiii: Okay, but you have a dependency on setuptools already. In Python 3.12.0b1 it's not pre-installed anymore, so it's on equal footing with scikit-build. ;)

amadio: Ok, nice, I will use something similar to what's in 80890de to only use distutils if necessary.

henryiii:
If you really don't want to use scikit-build(-core), I'd recommend looking at https://github.com/pybind/cmake_example/blob/master/setup.py, which still handles more situations than this does, like macOS ARM/Universal support. Though no editable installs and other features.

amadio:
> If you really don't want to use scikit-build(-core), I'd recommend looking at https://github.com/pybind/cmake_example/blob/master/setup.py, which still handles more situations than this does, like macOS ARM/Universal support. Though no editable installs and other features.

At least for 5.x, I'd like to not add new dependencies, as @abh is rightly a fan of keeping dependencies to a minimum. For 6.0, we can look into moving to scikit-build, if the platforms we support have it readily available (i.e. via package manager). Indeed, I'm sure scikit-build covers more than what I did here, as I'm kind of new to the Python packaging ecosystem. This just covers the OSs we already support and what I identified and tested on my own machine and macOS.

bindings/python/project.toml
amadio: Fixed in latest version.

bindings/python/setup.py
amadio: Fixed.

PR #2029, opened 2023-06-08
[gsi] Fix some memory leaks when using Secgsi

### LINKED ISSUE #2021: XRootD GSI Authentication memory leaks
Hi @amadio,

As discussed privately, we noticed that one headnode of the EOSLHCb instance had its memory usage increasing.
After enabling jemalloc profiling on the headnode, I managed to produce the following outputs: 

[prof2.pdf](https://github.com/xrootd/xrootd/files/11622290/prof2.pdf)
[prof3.pdf](https://github.com/xrootd/xrootd/files/11622291/prof3.pdf)

I believe the memory leaks come from the GSI authentication and non-freed cryptossl objects?

Thanks in advance :)

smithdh:
To fix issue #2021. (Draft initially).

### DIFF
--- a/src/XrdCrypto/XrdCryptosslgsiAux.cc
+++ b/src/XrdCrypto/XrdCryptosslgsiAux.cc
@@ -42,6 +42,7 @@
 #include <openssl/pem.h>
 #include <openssl/rsa.h>
 #include <openssl/x509v3.h>
+#include <memory>
 
 #include "XrdSut/XrdSutRndm.hh"
 #include "XrdCrypto/XrdCryptogsiX509Chain.hh"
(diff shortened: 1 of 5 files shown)


src/XrdCrypto/XrdCryptosslgsiAux.cc
```diff
@@ -900,9 +950,45 @@ int XrdCryptosslX509SignProxyReq(XrdCryptoX509 *xcpi, XrdCryptoRSA *kcpi,
...
+
+   // These will be assigned dynamically allocated ssl structures later.
+   // An RAII style guard is used to free them on exit from this method.
+   EVP_PKEY       *ekpi = 0;
+   X509           *xpo  = 0;
+   X509_EXTENSION *ext  = 0;
+   PROXY_CERT_INFO_EXTENSION *pci   = 0;
+   STACK_OF(X509_EXTENSION)  *xrisk = 0;
+
+   class sslGuard {
```
abh3, line 962: This definition of sslGaurd is almost identical to the previous version. It should be trivial to define a superset of sslGaurd and only have one definition placed at file level and used where need.

abh3 (commented):
Other than a code reduction comment I don't see anything inherently wrong here assuming all the frees are appropriate. Let's see what @gganis has to say.

src/XrdCrypto/XrdCryptosslgsiAux.cc
```diff
@@ -656,17 +656,61 @@ int XrdCryptosslX509CreateProxyReq(XrdCryptoX509 *xcpi,
...
+     sslGuard(EVP_PKEY *&epk, X509_EXTENSION *&xext, X509_NAME *&xname,
+              X509_REQ *&xreq, PROXY_CERT_INFO_EXTENSION *&pci,
+              STACK_OF(X509_EXTENSION) *&ske) :
+            epk_(epk), xext_(xext), xname_(xname), xreq_(xreq), pci_(pci),
+            ske_(ske) { }
+
+       ~sslGuard() {
+#if OPENSSL_VERSION_NUMBER >= 0x10000000L
+         sk_X509_EXTENSION_pop_free(ske_, X509_EXTENSION_free);
+#else /* OPENSSL */
```
bbockelm, line 680: Can we drop support for pre-OpenSSL 1.0?  That would mean we drop support for RHEL5 (which seems OK).

src/XrdCrypto/XrdCryptosslgsiAux.cc
```diff
@@ -656,17 +656,61 @@ int XrdCryptosslX509CreateProxyReq(XrdCryptoX509 *xcpi,
       PRINT("EEC certificate has expired");
       return -kErrPX_ExpiredEEC;
    }
+
+   // These will be assigned dynamically allocated ssl structures later.
+   // An RAII style guard is used to free them on exit from this method.
+   EVP_PKEY       *ekro  = 0;
```
bbockelm, line 662: If you don't want the bother of an entire class, you could do something like:

```
std::unique_ptr<EVP_PKEY, decltype(EVP_PKEY_free)> ekro(nullptr, &EVP_PKEY_free)
```

Type definition is uglier but maybe clearer to some?

abh3 (commented):
Well, I suppose using unique_ptr removes a class definition but I can't say it makes it easier to understand. Yes, the introduction of type synonyms helps but still feels a bit forced. That's just an individual preference here. Otherwise, I don't see anything wrong. Let's see of @gganis has anything more to say.

PR #2034, opened 2023-06-12
[XrdSecGsi] Use SHA256 as default message digest algorithm

### LINKED ISSUE #2030: Update default message digest algorithm for GSI
@adriansev noted that none of the default message digest algorithms are supported out-of-the-box by OpenSSL 3 / RHEL 9.

The current code (https://github.com/xrootd/xrootd/blob/master/src/XrdSecgsi/XrdSecProtocolgsi.cc#L161) has set `sha1:md5`.

Can we simply update this to `sha256` (remove sha1 and md5)?  Don't know when `sha256` was added to OpenSSL (could have been named SSLeay at the time!) but I can't imagine we support clients old enough that they don't have `sha256`.

amadio:
Fixes #2030.


amadio:
@bbockelm Is this what you intended to do for #2030?

bbockelm:
Almost - I was going to suggest removing the insecure fallbacks and just use `sha256` (since it seems RHEL5 and later have support for SHA256).

amadio:
> Almost - I was going to suggest removing the insecure fallbacks and just use `sha256` (since it seems RHEL5 and later have support for SHA256).

Ok, can do. Let's hear what @abh has to say on this, though.

abh3:
I think in the end it really does not matter. Having SHA1 and, fo that matter, MD5 as fallbacks would be rejected by any modern OpenSSL library. I will agree hat it looks somewhat insecure but on the other hand it looks like it's backwards compatible; which is not particularly useful.

Issue #2033, opened 2023-06-10
Security hardening in systemd unit files


badshah400:
Hi,
The security team at openSUSE recommend strengthening the security of xrootd's systemd services by making certain additions to its service files. These look like the following, for example, and similarly for the other service files:

```patch
Index: xrootd-5.5.0/packaging/common/xrootd@.service
===================================================================
--- xrootd-5.5.0.orig/packaging/common/xrootd@.service
+++ xrootd-5.5.0/packaging/common/xrootd@.service
@@ -6,6 +6,17 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
+# added automatically, for details please see
+# https://en.opensuse.org/openSUSE:Security_Features#Systemd_hardening_effort
+PrivateDevices=true
+ProtectHostname=true
+ProtectClock=true
+ProtectKernelTunables=true
+ProtectKernelModules=true
+ProtectKernelLogs=true
+ProtectControlGroups=true
+RestrictRealtime=true
+# end of automatic additions
 ExecStart=/usr/bin/xrootd -l /var/log/xrootd/xrootd.log -c /etc/xrootd/xrootd-%i.cfg -k fifo -s /run/xrootd/xrootd-%i.pid -n %i
 User=xrootd
 Group=xrootd
```

Please let us know if you find these options or specifically any of these options useful, and I can submit patches. Fwiw, we already have had these patches applied on our openSUSE packages for a while (since Jan 2022) now. More details are [here](https://en.opensuse.org/openSUSE:Security_Features#Systemd_hardening_effort).

openSUSE packages are here: <https://build.opensuse.org/package/show/openSUSE:Factory/xrootd>

Thanks in advance.

amadio:
We discussed this issue this morning. Since we are quite close to the release date (v5.6 should happen next week), we came to  the conclusion that we'd like to merge these changes initially commented out, to minimize the risk that they could break something for users on older operating systems. For the following release, once we have enough time to test everything, then we can uncomment and leave the new settings enabled by default. Would you like to make a pull request with the changes (so we properly attribute credit to you)? Thank you!

badshah400:
That makes a lot of sense. I'll make a PR by tomorrow. Many thanks for the consideration.

badshah400:
Thanks for the quick and helpful responses.

amadio:
I guess we can close this issue, right? I forgot to link the commit with the issue when merging.

amadio:
[...]

PR #2036, opened 2023-06-13
Add sandboxing settings to systemd service files.

badshah400:
This is a first attempt to implement stronger security settings in systemd service files. The sandboxing settings added with this commit are commented out until they can receive more testing, as discussed in issue #2033.

### DIFF
--- a/packaging/common/cmsd@.service
+++ b/packaging/common/cmsd@.service
@@ -6,6 +6,14 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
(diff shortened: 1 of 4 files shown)


amadio:
I will do a quick check with these changes uncommented tomorrow just to check that nothing obvious is wrong, then I will merge. Thanks a lot for the contribution!

amadio:
These looked fine uncommented on Alma8. Thank you for the contribution!

PR #2037, opened 2023-06-14
[XrdClTests] Change PostMaster test to use the postmaster instance in the DefaultEnv

smithdh:
This patch is to fix a failing XrdClTest, the failure noticed after a recent change in the PostMaster component ( eae57 ): But I believe the failure is in the test rather than the change itself: The test used various instances of a 'PostMaster' where as in usual use a single instance is always used. This cased the test to fail in a way not that's not expected in other (usual) uses.


amadio:
Thank you very much, @smithdh! This does fix the test. I'm merging.

PR #2041, opened 2023-06-15
Enable XrdEc tests by default

### LINKED ISSUE #2040: Erasure coding test failures: AlignedWrite2MissingTest{,IsalCrcNoMt}
In order to reproduce, need to build XRootD with XrdEc plugin enabled, and assuming you are in the build directory, run the following:
```
build $ cd tests/XrdEcTests
XrdEcTests $ rm -rf data && ../common/test-runner ./libXrdEcTests.so 'All Tests/MicroTest/MicroTest::AlignedWrite2MissingTest'
```

I get
```
You have selected: 

Selected tests/
  Selected tests/MicroTest::AlignedWrite2MissingTest

Running:

.F

MicroTest.cc:661:Assertion
Test name: MicroTest::AlignedWrite2MissingTest
assertion failed
- Expression: _st.IsOK()
- [*status]: [ERROR] Local error: permission denied

Failures !!!
Run: 1   Failure total: 1   Failures: 1   Errors: 0
```
The test only fails when 2 pieces are missing, the test for 1 piece missing passes, and if I add an extra parity block it passes as well.

I would like to fix and add these tests to the main test target for 5.6 release, that's why I'm investigating this.

amadio:
This is to add XrdEc tests to always run in CI when erasure coding is enabled, but before merging we need to fix #2040 (the two tests mentioned there should fail on this branch). Once all is green, we can merge this to master.

### DIFF
--- a/packaging/common/cmsd@.service
+++ b/packaging/common/cmsd@.service
@@ -6,6 +6,14 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
+#PrivateDevices=true
+#ProtectHostname=true
+#ProtectClock=true
+#ProtectKernelTunables=true
+#ProtectKernelModules=true
+#ProtectKernelLogs=true
+#ProtectControlGroups=true
+#RestrictRealtime=true
 ExecStart=/usr/bin/cmsd -l /var/log/xrootd/cmsd.log -c /etc/xrootd/xrootd-%i.cfg -k fifo -s /run/xrootd/cmsd-%i.pid -n %i
 User=xrootd
 Group=xrootd
--- a/packaging/common/frm_purged@.service
+++ b/packaging/common/frm_purged@.service
@@ -6,6 +6,14 @@ Requires=network-online.target
 After=network-online.target
 
 [Service]
+#PrivateDevices=true
+#ProtectHostname=true
+#ProtectClock=true
+#ProtectKernelTunables=true
+#ProtectKernelModules=true
+#ProtectKernelLogs=true
+#ProtectControlGroups=true
+#RestrictRealtime=true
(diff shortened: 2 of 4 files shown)


smithdh:
Hi @amadio 

I think it looks like an existing off by one problem in XrdCl declarative API utilities, I think here:

diff --git a/src/XrdCl/XrdClParallelOperation.hh b/src/XrdCl/XrdClParallelOperation.hh
index 6736fefba..dd38c56d4 100644
--- a/src/XrdCl/XrdClParallelOperation.hh
+++ b/src/XrdCl/XrdClParallelOperation.hh
@@ -313,7 +313,7 @@ namespace XrdCl
           // although we might have the minimum to succeed we wait for the rest
           if( status.IsOK() ) return ( pending == 0 );
           size_t nb = failed_cnt.fetch_add( 1, std::memory_order_relaxed );
-          if( nb + 1 == failed_threshold ) res = status; // we dropped below the threshold
+          if( nb == failed_threshold ) res = status; // we dropped below the threshold
           // if we still have to wait for pending operations return false,
           // otherwise all is done, return true
           return ( pending == 0 );

just trying to understand why it works in CI.. Do the CI tests run as root? - that might explain it. The stripe 'failure' is simulated by setting the unix permissions to '0' of a local file, i.e. '---------', but if the test is running as root it would still be able to open it, and not actually fail.

amadio:
Nice catch! Yes, that very much looks like it's the culprit, it fixes the test for me. I  also came to the conclusion that it must pass in CI because of running as root within the container. I think I will change the test to truncate some of the files instead of using chmod. That way it should fail also when run as root. Can you please just push the change above to master? I will rebase my branch on it, apply the truncate change and add to master. I've also added some simple tests using xrdfs and xrdcp on a simple server which I plan to add to the standard set as well.

smithdh:
I've pushed that one to master branch.

amadio:
Thanks!

amadio:
Added also a set of simple smoke tests. Will merge once CI finishes.

PR #2047, opened 2023-06-20
Increase default number of parallel event loops to 10

amadio:
This value has been chosen as default based on benchmarks that demonstrated that this is the minimum required number of threads to be able to saturate a link of 100Gbps.

Note: I have a small concern about clients launched by the server starting up way too many threads if enough of them are launched together after this patch. We should ensure this change will not cause problems for EOS, or make sure to set the number of event loops back to the old value in the environment in EOS instances before upgrading.


abh3 (approved):
Looks OK

PR #2053, opened 2023-07-10
[XrdCl] Avoid race in postmaster QueryTransport

smithdh:
This change is avoid the possibility of accessing pChannelMap (a std::map) during PostMaster::QueryTransport, while that map is modified by another thread, e.g. by one executing PostMaster::Send. A crash due to this was seen in a test, but is probably rare.


amadio:
Was this about the Poller test crash on macOS?

Issue #2058, opened 2023-07-13
[Regression] Chunked PUT creates empty files with 5.6.1 XrdHTTP


olifre:
This feels a bit similar to #1411, but the details are different. 

Running the following minimal example (based on the Nagios probe used by EGI ARGO monitoring, i.e. https://gitlab.cern.ch/lcgdm/nagios-plugins-webdav/-/blob/master/src/check_webdav), note you may need to adapt the URL for testing yourself:
```python
#!/usr/bin/env python

import os
import pycurl

curl = pycurl.Curl()
#curl.setopt(pycurl.FOLLOWLOCATION, 1)

curl.setopt(pycurl.CAINFO, os.getenv("X509_USER_PROXY"))
curl.setopt(pycurl.SSLCERT, os.getenv("X509_USER_PROXY"))
curl.setopt(pycurl.CAPATH, "/etc/grid-security/certificates")
curl.setopt(pycurl.URL, "https://myserver..com:1094//cephfs/grid/atlas/user/scratch/freyermu/testfile")
curl.setopt(pycurl.HTTPHEADER, ["Transfer-Encoding: chunked"])

TEST_DATA = "This is some text 0x1d bytes."

curl.setopt(curl.VERBOSE, 1)

def print_traffic(msg_kind, msg):
  if msg_kind < 1 or msg_kind > 4:
    return
  if msg_kind % 2 ==0:
    print("[%2d]> %s" % (msg_kind, str(msg)))
  elif msg_kind % 2 == 1:
    print("[%2d]< %s"  % (msg_kind, str(msg)))

curl.setopt(pycurl.DEBUGFUNCTION, lambda x, y: print_traffic(x, y))

class StringReader:
  def __init__(self, data):
    self.pos = 0
    self.data = data
  def read(self, size):
    ret = self.data[self.pos:self.pos+size]
    self.pos=self.pos+size

    if len(ret) == 0:
      self.pos = 0
    return ret

curl.setopt(pycurl.UPLOAD, 1)
stringreader = StringReader(TEST_DATA)
curl.setopt(pycurl.READFUNCTION, stringreader.read)

curl.perform()

print(curl.getinfo(pycurl.HTTP_CODE))

curl.setopt(pycurl.UPLOAD, 0)
curl.perform()
```
yields the following when run against XRootD 5.5.5:
```
[...]

PR #2059, opened 2023-07-17
[XrdHttp] Add back parsing of Transfer-Encoding header

CohenQU:
This commit reintroduces a header parser for `Transfer-Encoding: chunked` to maintain compatibility with the EGI Nagios probe. This issue was first mentioned in #2058.

Fixes: [d96b2b70487e159b47c24925abcfa00f975ec3d6](https://github.com/xrootd/xrootd/pull/2009/commits/d96b2b70487e159b47c24925abcfa00f975ec3d6), [#2009](https://github.com/xrootd/xrootd/pull/2009)

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -186,6 +186,8 @@ int XrdHttpReq::parseLine(char *line, int len) {
       sendcontinue = true;
     } else if (!strcasecmp(key, "TE") && strstr(val, "trailers")) {
       m_trailer_headers = true;
+    } else if (!strcasecmp(key, "Transfer-Encoding") && strstr(val, "chunked")) {
+      m_transfer_encoding_chunked = true; 
     } else if (!strcasecmp(key, "X-Transfer-Status") && strstr(val, "true")) {
       m_transfer_encoding_chunked = true;
       m_status_trailer = true;


olifre:
Thanks for the patch! As already mentioned in the linked issue, I can confirm that this fixes the EGI probe (and other use cases using `Transfer-Encoding: chunked`) in my testing. 
It still feels a bit surprising that setting the status header also toggles the transfer-encoding, so I would be curious as to which clients require this workaround (maybe it is also possible to fix these in their corresponding upstreams instead).

ccaffy (approved):
All good for me.

CohenQU:
@olifre I concur with your recommendations. I will initiate a dialogue with the developers on the osdf-client side to address the issue. Once we have implemented the necessary changes, I will then proceed to establish an additional PR here. Thank you for your invaluable input.

amadio:
I think the commit message could be improved, but the patch itself looks good to go. I'd use `[XrdHttp] Add back parsing of Transfer-Encoding header` or something like that, and also a reference to the commit that removed it with `Fixes: <commit hash>, <issue#>.` Thanks!

amadio:
Only the title of the pull request has been updated, the commit message remains the same.

CohenQU:
@amadio Sorry, the update on my local branch didn't seem to deploy well, but it should be updated now.

amadio:
Thank you! I am merging, but please note that the issue will only close when I merge the `devel` branch into `master`, you do not need to close that by hand.

PR #2063, opened 2023-07-24
Implement Linux epoll maxfd limit

pllopis:
On Linux systems, the latest xrootd release `abort()`s during initialisation if the `RLIMIT_NOFILE` (ulimit -n) isn't the max, but is still high, and a default isn't set via the new `maxfd strict` directive.

For instance, on a default docker environment on CentOS Stream 9, the default limit is set to `1073741816`, while the epoll limit is way less than that (it's `INT_MAX / sizeof(struct epoll_event)`, which on that system was `178956970`). 
In the Linux Poller implementation, surpassing this limit causes `EINVAL` to be returned in the first call to `epoll_wait()` as the `maxevents` parameter will be too high. The error output will look like below:

```
Config maximum number of connections restricted to 1073741816
230724 12:21:09 453 Xrd_Config: sendfile enabled.
230724 12:21:09 453 Xrd_LinkCtl: Allocating 64 link objects at a time
230724 12:21:13 453 Xrd_Poll: Starting poller 0
230724 12:21:13 458 XrdPoll: Unable to poll for events; invalid argument
230724 12:21:13 453 Xrd_Poll: Starting poller 1
Aborted (core dumped)
```

This PR contributes an additional sanity check when run on Linux systems so that the max can never surpass the maximum allowed by the epoll subsystem. With this patch applied, the above will instead read:
```
Config maximum number of connections restricted to 178956970
230724 17:37:51 001 Xrd_Config: sendfile enabled.
230724 17:37:51 001 Xrd_LinkCtl: Allocating 64 link objects at a time
230724 17:37:52 001 Xrd_Poll: Starting poller 0
230724 17:37:52 001 Xrd_Poll: Starting poller 1
230724 17:37:53 001 Xrd_Poll: Starting poller 2
```

Note that under some circumstances, this bug is sometimes hidden away when it's run under the rpm-provided systemd unit file, which [sets a sane limit](https://github.com/xrootd/xrootd/blob/8f0a3ebf6b14c65c57667d07c0c624f3d4b215a8/packaging/common/xrootd%40.service#L24), but this isn't the case when not running under systemd.

### DIFF
--- a/src/Xrd/XrdConfig.cc
+++ b/src/Xrd/XrdConfig.cc
@@ -93,6 +93,9 @@
 #if defined(__linux__) || defined(__GNU__)
 #include <netinet/tcp.h>
 #endif
+#if defined(__linux__)
+#include <sys/epoll.h>
+#endif
 #ifdef __APPLE__
 #include <AvailabilityMacros.h>
 #endif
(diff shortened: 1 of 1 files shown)


pllopis:
Hi, just saw the failing CI due to the int to unsigned comparison and pushed a fix. (Let me know if you'd like to rebase into a single commit when pushing fixes.)

amadio:
Yes, a single commit with the final version would be better. You can just force push to the same branch. Otherwise, I will do it upon merging.

abh3 (approved):
Looks good to me.

PR #2064, opened 2023-07-25
Fix logic error in user mapping

### LINKED ISSUE #2056: username != username in commit #1978
Hello,
in commit #1978 in the file [src/XrdSciTokens/XrdSciTokensAccess.cc](https://github.com/xrootd/xrootd/pull/1978/files#diff-8cd57d0057f83be94fb75362e5c1c31c8592a2b6ef6403192d00e28837bc001d) line 241 reads
`if (!m_username.empty() && username != username) {return "";}`(https://github.com/xrootd/xrootd/pull/1978/files#diff-8cd57d0057f83be94fb75362e5c1c31c8592a2b6ef6403192d00e28837bc001dR241)
Where `username != username` is always going to return false, right? Is that intended?
I assume this is supposed to be `m_username != username`?
In the current version on the masterbranch, the line is still the same (it's just line 242 in there: https://github.com/xrootd/xrootd/blob/master/src/XrdSciTokens/XrdSciTokensAccess.cc#L242)
Cheers,
Simon

bbockelm:
When setting up mapping rules, if the user feature was utilized, a logic error would cause the rule to be ignored.

Fixes #2056

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -239,7 +239,7 @@ struct MapRule
     {
         if (!m_sub.empty() && sub != m_sub) {return "";}
 
-        if (!m_username.empty() && username != username) {return "";}
+        if (!m_username.empty() && username != m_username) {return "";}
 
         if (!m_path_prefix.empty() &&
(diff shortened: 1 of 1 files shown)


amadio:
Please fix the typo in the commit message. Thanks!

bbockelm:
@amadio - good catch!  Done.

PR #2066, opened 2023-07-28
Add unprivileged Dockerfile for alma9

pllopis:
Hi,
This PR contributes a Dockerfile with the following changes with respect to the other alma9 dockerfiles:
- The main change is that it allows to run unprivileged since it's built without systemd.
- Adds arguments to activate or deactivate certain build options

I mostly wanted to share this to learn whether you would be interested in such an approach. If so, perhaps more effort can be put into it so it runs the tests, works with your existing CI, etc.

### DIFF
--- a/docker/build/Dockerfile.unprivileged.alma9
+++ b/docker/build/Dockerfile.unprivileged.alma9
@@ -1,0 +1,133 @@
+# This dockerfile assumes that you're passing the binary with its corresponding arguments to run it.
+# Example: docker run xrootd:latest xrootd -c /etc/xrootd/xrootd.cfg
+# But you could also run cmsd or other xrootd binaries in the same way.
+#
+# The VERSION argument must correspond to a github tag name.
+#
+ARG VERSION=v5.6.1
+ARG ENABLE_HTTP=ON
+ARG ENABLE_FUSE=OFF
+ARG ENABLE_KRB5=OFF
+ARG ENABLE_SCITOKENS=ON
+ARG ENABLE_MACAROONS=ON
+ARG ENABLE_XRDCEPH=ON
+ARG ENABLE_PYTHON=ON
+
+FROM almalinux:9 AS builder
+# Re-declare the ARGs for use in this stage
+ARG VERSION
+ARG ENABLE_HTTP
+ARG ENABLE_FUSE
+ARG ENABLE_KRB5
+ARG ENABLE_SCITOKENS
+ARG ENABLE_MACAROONS
+ARG ENABLE_XRDCEPH
+ARG ENABLE_PYTHON
+
+# Set the workdir to a source directory and copy the entire project into the Docker image
+WORKDIR /xrootd
+COPY ../../ .
+
+# Enable codeready builder repos
+RUN dnf install -y yum-utils
+RUN dnf config-manager --set-enabled crb
+
+# Install dependencies
+RUN dnf install -y epel-release && \
+    dnf install -y \
+    git \
+    cmake \
+    gcc-c++ \
+    cppunit-devel \
+    curl-devel \
+    davix-devel \
+    diffutils \
+    file \
+    fuse-devel \
+    gcc-c++ \
+    git \
+    gtest-devel \
+    json-c-devel \
+    libmacaroons-devel \
+    libtool \
+    libuuid-devel \
+    libxml2-devel \
+    libcurl-devel \
+    make \
+    openssl-devel \
+    python3-devel \
+    python3-setuptools \
+    readline-devel \
+    scitokens-cpp-devel \
+    tinyxml-devel \
+    yasm \
+    zlib-devel 
+
+# Install Ceph dependencies to enable XRDCEPH
+RUN dnf install -y \
+    centos-release-ceph-quincy
+RUN dnf install -y \
+    librados-devel \
+    libradosstriper-devel
+
+RUN ls -l
+RUN cmake -S . -B xrootd_build -DCMAKE_BUILD_TYPE=RelWithDebInfo \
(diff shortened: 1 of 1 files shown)


amadio:
Dear @pllopis,

Thank you for your pull request. Indeed, having a container that can run unprivileged would be nice, but I think that for packaging/installation, it would be better to stick to the RPM-based solution, as it creates a much simpler Dockerfile which also ensures we keep dependencies needed up to date, etc. Maybe you can update our existing Dockerfile to run unprivileged instead? Note that our tests already run unprivileged when run via the xrootd-docker script, as the servers are started by systemd. This is something we'd like to preserve.

docker/build/Dockerfile.unprivileged.alma9
```diff
@@ -0,0 +1,133 @@
+# This dockerfile assumes that you're passing the binary with its corresponding arguments to run it.
```
amadio, line 1: A container that can be run this way is nice to have, but I think it's better to modify our current container to allow it rather than create a different install procedure, as this one is a lot more complicated than what we currently have, and can become out of date, as dependencies, build options, etc, are hard-coded. For example, if we remove ENABLE_SCITOKENS option in the future, it would require changing this file. I also think that it is already possible to do this with our current container if you pass the full command line with `xrootd ...` when running the container, no?

docker/build/Dockerfile.unprivileged.alma9
```diff
@@ -0,0 +1,133 @@
...
+    tinyxml \
+    davix-libs
+
+RUN if [ "${ENABLE_KRB5}" =  "ON" ]; then dnf install -y krb5-libs; fi
+RUN if [ "${ENABLE_HTTP}" = "ON" ]; then dnf install -y libcurl-minimal; fi
+RUN if [ "${ENABLE_FUSE}" = "ON" ]; then dnf install -y fuse-libs; fi
+RUN if [ "${ENABLE_PYTHON}" = "ON" ]; then dnf install -y python3; fi
+RUN if [ "${ENABLE_SCITOKENS}" = "ON" ]; then dnf install -y scitokens-cpp; fi
+RUN if [ "${ENABLE_MACAROONS}" = "ON" ]; then dnf install -y libmacaroons; fi
+RUN if [ "${ENABLE_XRDCEPH}" = "ON" ]; then dnf install -y librados2 libradosstriper1; fi
```
amadio, line 120: These create many unnecessary layers in the final container and will cause many rebuilds when one of the layers change.

amadio (changes requested):

pllopis:
Hi @amadio thanks for reviewing!

If I understand correctly, you'd like to preserve these two:

1. `xrootd` and `cmsd` are started via systemd running inside the container. I guess you a referring specifically to: https://github.com/xrootd/xrootd/blob/8f0a3ebf6b14c65c57667d07c0c624f3d4b215a8/docker/xrd-docker#L128-L130
2. I understand that you prefer to keep the dependencies and build implementations in one place (the RPM spec).

Regarding 1:
> Note that our tests already run unprivileged when run via the xrootd-docker script, as the servers are started by systemd

Sorry, I'm confused and may have missed something here. I gave it a try on a system and noticed that when launching it via `xrd-docker`, the container was being launched in privileged mode. This seems consistent with the way xrd-docker is launching things: https://github.com/xrootd/xrootd/blob/8f0a3ebf6b14c65c57667d07c0c624f3d4b215a8/docker/xrd-docker#L118
Therefore I don't understand how you can get unprivileged mode when starting things under systemd (unless you meant that you're using podman)?

As far as I can see, the reason the `--privileged` flag is required is entirely due to systemd running inside the container.
Therefore, I don't think we can have both unprivileged containers, and systemd inside the container starting xrootd and cmsd. Podman of course will achieve this, but this isn't a general solution, as for instance you will still require to run as privileged when using anything else other than podman (such as container orchestrators like Kubernetes).

Regarding 2:
 I think this might be doable indeed (provided point 1 above is resolved somehow).

I was hoping to find a general solution that will allow to run xrootd in unprivileged mode that works for any environment such as k8s, docker. But I don't see a technical solution other than to drop the requirement to have systemd start the daemons inside the container. Do you have any ideas about this?

Thanks!

amadio:
> I was hoping to find a general solution that will allow to run xrootd in unprivileged mode that works for any environment such as k8s, docker. But I don't see a technical solution other than to drop the requirement to have systemd start the daemons inside the container. Do you have any ideas about this?

We just want systemd to start the xrootd service for testing and CI, I'm fine with creating a container that can run unprivileged, but the container built by xrd-docker can be run unprivileged like this, for example, so I don't think a lot is needed to allow what you want:

```sh
$ docker run --ulimit nofile=262144:262144 -u xrootd:xrootd --network=host xrootd:alma8 xrootd -c /etc/xrootd/xrootd-srv2.cfg
```
With 5.6.2 you will not need `--ulimit nofile=262144:262144` anymore.

You can test that the server runs ok with `xrdfs localhost:1099 query config version`. So I think that the container the way it is built is fine, you just need to change the default configurations, as the ones shipping inside are not suitable for running as you'd like.

pllopis:
Oh I see! Stupid me, I never tried invoking the xrootd binary directly.
Thanks for pointing this out, I think this might indeed work running as unprivileged. Pretty sure this solution will work, so feel free to close this PR.

amadio:
Not at all! I think the idea you have is quite good, to have a ready-made container to run xrootd as a service. I will close the PR, but in the upgrade we are doing for the CI I will see if we can improve the defaults in the containers to make it easier to run unprivileged and maybe consider also providing a Dockerfile specifically for this case which installs a pre-built version of XRootD using our RPM repositories rather than building it from source.

Issue #2065, opened 2023-07-28
Curl used by XrdHttpTpc fails depending on the CRLfile being used


esindril:
This is happening on CentOS7 running the following package versions:
```
rpm -qa | grep "openssl\|curl"
libcurl-7.29.0-59.el7_9.1.x86_64
curl-7.29.0-59.el7_9.1.x86_64
openssl-libs-1.0.2k-26.el7_9.x86_64
python-pycurl-7.19.0-19.el7.x86_64
openssl-1.0.2k-26.el7_9.x86_64
```

We saw in our EOSATLAS instance failed HTTP TPC pull transfers with the following signature:
```
230727 17:27:23 32353 TPC_PushRequest: event=TRANSFER_FAIL, local=/eos/atlas/<some>/<path>, user=(anonymous), bytes_transferred=0; HTTP library failure: Failed to load CRL file (path? access rights?, format?)
```

The reason for this failure seems to be the fact that the OpenSSL library can not handle CRL files that are larger than 100kB - and truncating the data that it reads. In this particular case the CRLfile constructed by the XRootD daemon [1] and passed as an argument to the curl library is just above the 100kB limit:
```
# ls -lrt /tmp/fst/.xrdtls/
total 1184
-rw-------. 1 daemon daemon  151532 Jul 28 10:31 ca_file.pem
-rw-------. 1 daemon daemon 1058459 Jul 28 10:31 crl_file.pem
```
For all other nodes where this CRLfile is below the 100kB limit the XRootD HTTP TPC transfers are successful. This seems to be a known limitation in that particular version of OpenSSL:
`https://github.com/openssl/openssl/blob/081314d07705aa58912845c213a48414d8f616a9/crypto/ocsp/ocsp_ht.c#L87C9-L87C29`

As a (dev) workaround XRootD could append CRL to this file taking into account this limitation. Operationally we'll try to limit the number of CRLs that go into this file for the time being.
There are other variations of a similar problem reported on the OpenSSL bug tracker. [2]

[1] https://github.com/xrootd/xrootd/blob/master/src/XrdTls/XrdTlsTempCA.cc#L178
[2] https://github.com/openssl/openssl/issues/8581

esindril:
Ok, actually things are more twisted than that. The error above is thrown when the following CRL is the first in the bundle:
```
 openssl crl -in /etc/grid-security/certificates/8dd53007.r0 -text 
Certificate Revocation List (CRL):
        Version 2 (0x1)
[...]

PR #2073, opened 2023-08-21
XrdTls: XrdTlsTempCA - CRLs containing critical extensions are inserted at the end of the bundled CRL file

ccaffy:
Solves issue #2065

### DIFF
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -161,6 +161,13 @@ public:
      * processFile(...) method, false otherwise
      */
     bool atLeastOneValidCRLFound() const;
+    /**
+     * https://github.com/xrootd/xrootd/issues/2065
+     * To mitigate that issue, we need to defer the insertion of the CRLs that contain
+     * critical extensions at the end of the bundled CRL file
+     * @return true on success.
+     */
+    bool processCRLWithCriticalExt();
 
 private:
     XrdSysError &m_log;
@@ -171,6 +178,9 @@ private:
     std::unordered_set<std::string> m_known_crls;
     const int m_output_fd;
     std::atomic<bool> m_atLeastOneValidCRLFound;
+    //Store the CRLs containing critical extensions to defer their insertion
+    //at the end of the bundled CRL file. Issue https://github.com/xrootd/xrootd/issues/2065
+    std::vector<std::unique_ptr<XrdCryptosslX509Crl>> m_crls_critical_extension;
 };
 
 
@@ -179,7 +189,7 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
 {
     file_smart_ptr outputfp(fdopen(dup(m_output_fd), "w"), &fclose);
     if (!outputfp.get()) {
-        m_log.Emsg("CAset", "Failed to reopen file for output", fname.c_str());
+        m_log.Emsg("CRLSet", "Failed to reopen file for output", fname.c_str());
         return false;
     }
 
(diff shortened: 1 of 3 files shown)


src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -202,10 +212,17 @@ CRLSet::processFile(file_smart_ptr &fp, const std::string &fname)
...
+          // This CRL will be put at the end of the bundled file
+          m_crls_critical_extension.emplace_back(std::move(xrd_crl));
+        } else {
+          // No critical extension found on that CRL, just insert it on the CRL bundled file
+          if (!xrd_crl->ToFile(outputfp.get())) {
             m_log.Emsg("CRLset", "Failed to write out CRL", fname.c_str());
             fflush(outputfp.get());
             return false;
+          }
         }
```
amadio, line 226: It seems to me that you could declare the vector `m_crls_critical_extension` before the `for` loop, then process it here, no? You can then pass a reference to it to `processCRLWithCriticalExt` and avoid adding a member to the class.

ccaffy: Yes, I did not want to change the signature of the member function, but sure, if you prefer it that way, it can be done :)

amadio: Yes, I realize now that my suggestion doesn't make much sense as the CRLs with extensions would be processed at the end of each file, which doesn't help much. Let's leave it like this and wait for Andy's comments. Then I will merge this into devel for 5.6.2.

src/XrdTls/XrdTlsTempCA.cc
```diff
@@ -217,6 +234,26 @@ bool CRLSet::atLeastOneValidCRLFound() const {
     return m_atLeastOneValidCRLFound;
 }
 
+bool CRLSet::processCRLWithCriticalExt() {
+  // Don't open the output file if not necessary
+  if(!m_crls_critical_extension.empty()) {
+    file_smart_ptr outputfp(fdopen(dup(m_output_fd), "w"), &fclose);
```
abh3, line 240: Please use XrdSysFD_Dup() instead of dup() to avoid FD leakage (see line 104 in the original file). This should also be corrected in line 180 that somehow missed using XrdSysFD_Dup() in CRLSet::processFile(). I know it's not technically part of this patch but it should be corrected either via this patch or with an addition pull request (technically cleaner) so that we don't forget.

abh3 (changes requested):
Somewhat minor issue and a suggestion to carry through the needed change to another part of the code. I was wondering why we are playing so many games with fdopen(), Would a single open stream suffice for this? I'm not asking to change it just curious why this is necessary.

src/XrdTls/XrdTlsTempCA.cc
ccaffy: Hi @abh3,

Thanks for your very pertinent comment, I'll correct that and ship it within the same PR, but in a different commit :)
Then we are good to go!

Cheers,
Cedric

ccaffy: About your question, I don't know why, but I believe it can be changed...

ccaffy: If you want me to change this, I can do it :) But probably in a different PR then ;)

ccaffy: I actually did it as it did not take a lot of time. I will submit a different PR once this one has been approved and merged to master

abh3:
Thanks Cedric!

abh3 (approved):

amadio:
@ccaffy I guess this is ready to merge, right?

ccaffy:
> @ccaffy I guess this is ready to merge, right?

Yes let's gooo !!!

xrootd-dev:
Hi Cedric,

Ah, don't forget with the follow-up pull request!

Andy

PR #2075, opened 2023-08-23
XrdTls: XrdTlsTempCA - Refactored CASet and CRLSet to open the output file only once before the processing

ccaffy:
This PR is coming from the PR https://github.com/xrootd/xrootd/pull/2073.

The Maintenance thread creates bundle CA and CRL files. Before this modification, the output files created by the maintenance were open and closed for each CA and CRL read. It is not the case anymore. There is only one open at the beginning of the processing and one close at the end.

### DIFF
--- a/src/XrdTls/XrdTlsTempCA.cc
+++ b/src/XrdTls/XrdTlsTempCA.cc
@@ -60,13 +60,28 @@ static uint64_t monotonic_time_s() {
   return tp.tv_sec + (tp.tv_nsec >= 500000000);
 }
 
-
-class CASet {
+/**
+ * Class managing the CRL or CA output file pointer. It is a RAII-style class that opens the output
+ * file in the constructor and close the file when the instance is destroyed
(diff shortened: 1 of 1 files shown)


abh3 (commented):
Well, this is rather an extensive charge *though for the better). I can inly assume it has been tested ad nauseum. So, if you confirm that is the case I will approve.

ccaffy:
> Well, this is rather an extensive charge *though for the better). I can inly assume it has been tested ad nauseum. So, if you confirm that is the case I will approve.

It's just using RAII to open the file in the constructor and close it when the object is destroyed. I tested it and got no issue. Used `lsof -t` on the concatenated CRL and CA files, with no output --> everything looks good to me ;)

abh3:
Thanks for confirming that :-)

abh3 (approved):
Can't see anything wrong here. This is a good improvement!

Issue #2074, opened 2023-08-22
Scitoken.conf


biozit:
Hi All,

   Using this config for scitokens, the Xrootd says the latest issuer is not listed:

`[Global]
audience = IceCube, NRP

[Issuer https://chtc.cs.wisc.edu/icecube]
issuer = https://chtc.cs.wisc.edu/icecube
base_path = /icecube
map_subject = False

[Issuer https://token.nationalresearchplatform.org/modis]
issuer = https://token.nationalresearchplatform.org/modis
base_path = /nrp/protected/sio/MODIS_Aqua_microphysics_images/
map_subject = False

[Issuer https://token.nationalresearchplatform.org/ucsdphysics]
issuer = https://token.nationalresearchplatform.org/ucsdphysics
base_path = /ucsd/physics
map_subject = False`

However, XrootD is ok with this:

`[Global]
audience = IceCube, NRP

[Issuer https://chtc.cs.wisc.edu/icecube]
issuer = https://chtc.cs.wisc.edu/icecube
base_path = /icecube
map_subject = False

[Issuer https://token.nationalresearchplatform.org/modis]
issuer = https://token.nationalresearchplatform.org/modis
base_path = /nrp/protected/sio/MODIS_Aqua_microphysics_images/
map_subject = False

[Issuer https://**token.nationalresearchplatform.ort**/ucsdphysics]
issuer = https://**token.nationalresearchplatform.ort**/ucsdphysics
base_path = /ucsd/physics
map_subject = False
`

abh3:
Hmm, which issues would be the latest? Can you turn on some debugging to 
see what xrootd (well actually the SciTokens plugin) thinks is going on?

biozit:
`stash-origin-auth]# xrootd -version
v5.5.5`

This is the case with 2 issuers:

`[root@node-1-5 stash-origin-auth]# cat scitokens.conf 
[Global]
audience = IceCube, NRP

[Issuer https://chtc.cs.wisc.edu/icecube]
issuer = https://chtc.cs.wisc.edu/icecube
base_path = /icecube
map_subject = False

[Issuer https://token.nationalresearchplatform.org/modis]
issuer = https://token.nationalresearchplatform.org/modis
base_path = /nrp/protected/sio/MODIS_Aqua_microphysics_images/
map_subject = False


[Issuer https://token.nationalresearchplatform.org/ucsdphysics]
issuer = https://token.nationalresearchplatform.org/ucsdphysics
[...]

PR #2079, opened 2023-08-31
Updating maximum header size and maximum line length for any line in …

biozit:
This update is to fix this issue: https://github.com/xrootd/xrootd/issues/2074


abh3 (approved):
This looks fine, though usually the line length is longer than the typical tokens it contains. Note that the original used a 4:1 ratio (50 chars for tokens in a 200 char line). So, a line length of 2K or even 4K would make more logical sense though I don't think it will be materially different the way it is now.

Issue #2084, opened 2023-09-12
Many threads waiting around ERR_clear_error


smithdh:
Hi,

On some busy EOS instances (xrootd severs ~v5.5.5, on centos 7) it has sometimes been seen that many,  e.g. 1800 threads, can be waiting inside ERR_clear_error at any moment. The process was not deadlocked, it is assumed there is just too much contention and/or lock held for too long. For example:

```
TID 234963:
#0  0x00007f32ed9b554d __lll_lock_wait
#1  0x00007f32ed9b0e9b _L_lock_883
#2  0x00007f32ed9b0d68 pthread_mutex_lock
#3  0x00007f32ed02cf83 int_thread_get
#4  0x00007f32ed02d48d int_thread_get_item
#5  0x00007f32ed02de20 ERR_get_state
#6  0x00007f32ed02e065 ERR_clear_error
#7  0x00007f32e35af3f1 XrdCl::XRootDTransport::DoAuthentication(XrdCl::HandShakeData*, XrdCl::XRootDChannelInfo*)
#8  0x00007f32e35b2c0b XrdCl::XRootDTransport::HandShakeMain(XrdCl::HandShakeData*, XrdCl::AnyObject&)
#9  0x00007f32e35b2ec3 XrdCl::XRootDTransport::HandShake(XrdCl::HandShakeData*, XrdCl::AnyObject&)
#10 0x00007f32e36244fa XrdCl::AsyncSocketHandler::HandleHandShake(std::unique_ptr<XrdCl::Message, std::default_delete<XrdCl::Message> >)
#11 0x00007f32e3624a91 XrdCl::AsyncSocketHandler::OnReadWhileHandshaking()
#12 0x00007f32e3625df5 XrdCl::AsyncSocketHandler::Event(unsigned char, XrdCl::Socket*)
#13 0x00007f32e359aba8 (anonymous namespace)::SocketCallBack::Event(XrdSys::IOEvents::Channel*, void*, int)
#14 0x00007f32ee843bbb XrdSys::IOEvents::Poller::CbkXeq(XrdSys::IOEvents::Channel*, int, int, char const*)
#15 0x00007f32ee844ee3 XrdSys::IOEvents::PollE::Dispatch(XrdSys::IOEvents::Channel*, unsigned int)
#16 0x00007f32ee8450c0 XrdSys::IOEvents::PollE::Begin(XrdSysSemaphore*, int&, char const**)
[...]

PR #2085, opened 2023-09-12
[XrdTls] Make the thread-id returned to libcrypto appear more uniform for openssl 1.0

smithdh:
Connected to issue #2084


src/XrdTls/XrdTlsContext.cc
```diff
@@ -366,9 +378,11 @@ const char *sslCiphers = "ALL:!LOW:!EXP:!MD5:!MD2";
 
 XrdSysMutex            ctxMutex;
 #if __cplusplus >= 201103L
-std::atomic<bool>      initDone( false );
+std::atomic<int>      initTlsDone( 0 );
```
abh3, line 381: So, what is wrong with using an atomic bool?

src/XrdTls/XrdTlsContext.hh
```diff
@@ -19,7 +19,7 @@
 //------------------------------------------------------------------------------
 
 #include <cstdint>
-//#include <string>
+#include <string>
```
abh3, line 22: What make <string> be required now? Mostly out of curiosity.

abh3 (commented):
Just a couple of questions here.

src/XrdTls/XrdTlsContext.cc
smithdh: Ah, that was because I was changing that area because to not lose the debug initialisation even if Init() had already been called from elsewhere, and I noticed that our AtomicGet may be defined "__sync_fetch_and_or" and on the gcc docs here: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fsync-Builtins.html it mentioned "The object pointed to by the first argument must be of integer or pointer type. It must not be a boolean type."; so I changed it.

src/XrdTls/XrdTlsContext.hh
smithdh: there's a compiler error when including XrdTlsContext.hh in XrdCryptoLite_bf32.cc without it. There are std::string used in XrdTlsContext.hh, so it seemed to make sense it was required, I suppose it's not previously been needed because elsewhere when XrdTlsContext.hh was included something else had previously included string. e.g. this is the error:

```
In file included from /data/dhsmith/xrootd-smithdh-repo/src/XrdCrypto/XrdCryptoLite_bf32.cc:46:
/data/dhsmith/xrootd-smithdh-repo/src/XrdTls/XrdTlsContext.hh:75:13: error: ‘string’ in namespace ‘std’ does not name a type
       {std::string cert;   //!< -> certificate path.
```
etc.

abh3: I think the underlying issue here is that the legacy line
#if __cplusplus >= 201103L
is still there. We require the use of at least gcc 7 and there really is no need to use gcc homebrew atomics.

src/XrdTls/XrdTlsContext.cc
abh3: See my comment on the underlying issue here.

abh3 (commented):
So, all of this looks fine. Last question (promise) as noted in the comment. I ask simply because I haven't looked at how OpenSSL handles its hash table so you can easily say that isn't a good solution; which is fine but at least we will have considered all the options.

smithdh:
Thanks; I like that idea. I looked, and the actual hash that the openssl table is using is h=key*13 (where key is e.g. XrdSysThread::Num()); but indeed the Num() based id should be much better, and then no need for 'mixer'. I'll revise the PR to use the Num().

smithdh:
gettid (used by Num()) is a system call, whereas the ID() (using pthread_self()) is a singe instruction [on linux]. so possibly there's a bit more overhead.

smithdh:
Hi @abh3 I did a quick test just now: On a single thread test program on an centos7 machine, single threaded repeated syscall with gettid gave about 2MHz. Instead repeated pthread_self() + mix was about 100 the rate. My best guess at the frequency we're calling the sslTLS_id_callback() function in production was about ~1kHz, that was just based on the number of retrieves recorded in the openssl hash-table statistics and assuming there was 1 call to the callback for each. But 100 times is quite a bit slower:

I think for now I'll leave the PR with the XrdSysThread::ID() + mix. I believe Guilherme was thinking of tagging 5.6.2 tomorrow, what do you think about including the PR still with ID()+mix, but possibly subject to later re-visit (and perhaps we anyway want to revisit to include ERR_remove_thread_state() in a future change, anyway)? Or would you favor holding this now, to give more time to consider, and maybe measure more carefully?

abh3:
Hi @smithdh good to know. I was not aware that it was that much slower. Then we are all set.

abh3 (approved):

Issue #2092, opened 2023-09-25
cmsd blacklist behaviour when host cannot be found


Jo-stfc:
Hi,
I've recently found a bug or intended behaviour that is not well documented:
If a host in the blacklist cannot be found, the cmsd managers completely ignore the blacklist after a restart.

```
Config processing blacklist file /etc/xrootd/cms.blacklist
Config Unable to blacklist '<hostname>'; Name or service not known
230925 10:16:25 375365 GetBL: Error(s) encountered in blacklist file!
```
this stops the in-memory blacklist from updating (as documented), but a service restart wipes the existing blacklist and the other valid hosts are not blacklisted.

Could this either be clarified in the cmsd documentation or caught in  XrdCmsBlackList::AddBL ?

amadio:
Could you please test this patch?

```diff
diff --git a/src/XrdCms/XrdCmsBlackList.cc b/src/XrdCms/XrdCmsBlackList.cc
index ed5eda173..3fb81ce99 100644
--- a/src/XrdCms/XrdCmsBlackList.cc
+++ b/src/XrdCms/XrdCmsBlackList.cc
@@ -455,7 +455,7 @@ bool XrdCmsBlackList::GetBL(XrdOucTList  *&bList,
 // Return ending status
 //
    blFile.Close();
-   bList = (aOK ? bAnchor.Export() : 0);
+   bList = bAnchor.Export();
    rList = rAnchor[1].Array(rcnt);
    return aOK;
 }
```

This will try to export the list even if some entries were not added correctly.

Jo-stfc:
Thanks Guilherme,
it now flags the specific host as not found and raises the error but then proceeds to blacklist the other hosts.

```
Config processing blacklist file /etc/xrootd/cms.blacklist
Config Unable to blacklist '<nonexistent host>'; Name or service not known
230925 14:28:52 3087232 GetBL: Error(s) encountered in blacklist file!
Config Blacklisting <host in list before>
Config Blacklisting <host in list after>
```

Jo-stfc:
yes, that's the behaviour I wanted. Thanks!

abh3:
As was correctly surmised, the issue stems from this documentation note:
12)  If the blacklist file is present but contains a syntax error or cannot be read, the current black is not changed.

So, yes, on a restart there is no blacklist so nothing winds up being backlisted upon startup.

[...]

PR #2093, opened 2023-09-26
Try to load blacklist even if some entries are invalid

amadio:
I suppose this is what you meant, right @abh3?

Fixes: #2092


abh3 (approved):

abh3:
Thanks for the fix @amadio

PR #2095, opened 2023-10-01
Pass through currently-configured CA file to the SciTokens library

bbockelm:
I'm working on setting up some end-to-end integration tests including the SciTokens library.

When hosting the issuer's public keys on the XRootD server itself, we need to have a validated HTTPS connection from the libcurl handle used internally by the `scitokens-cpp` library.

The `scitokens-cpp` library currently can only use the system CAs, which makes it impossible to utilize the integration test's generated CA.  We need to pass along the CA file used by xrootd.

Pairs with https://github.com/scitokens/scitokens-cpp/pull/132

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -5,6 +5,7 @@
 #include "XrdSec/XrdSecEntity.hh"
 #include "XrdSec/XrdSecEntityAttr.hh"
 #include "XrdSys/XrdSysLogger.hh"
+#include "XrdTls/XrdTlsContext.hh"
 #include "XrdVersion.hh"
 
 #include <map>
@@ -432,7 +433,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
     };
 
 public:
-    XrdAccSciTokens(XrdSysLogger *lp, const char *parms, XrdAccAuthorize* chain) :
+    XrdAccSciTokens(XrdSysLogger *lp, const char *parms, XrdAccAuthorize* chain, XrdOucEnv *envP) :
         m_chain(chain),
         m_parms(parms ? parms : ""),
         m_next_clean(monotonic_time() + m_expiry_secs),
@@ -441,7 +442,7 @@ public:
         pthread_rwlock_init(&m_config_lock, nullptr);
         m_config_lock_initialized = true;
         m_log.Say("++++++ XrdAccSciTokens: Initialized SciTokens-based authorization.");
-        if (!Config()) {
+        if (!Config(envP)) {
             throw std::runtime_error("Failed to configure SciTokens authorization.");
         }
     }
@@ -926,7 +927,7 @@ private:
     }
 
 
-    bool Config() {
+    bool Config(XrdOucEnv *envP) {
         // Set default mask for logging.
         m_log.setMsgMask(LogMask::Error | LogMask::Warning);
 
@@ -961,6 +962,15 @@ private:
             } while ((val = scitokens_conf.GetToken()));
         }
         m_log.Emsg("Config", "Logging levels enabled -", LogMaskToString(m_log.getMsgMask()).c_str());
+
+        auto xrdEnv = static_cast<XrdOucEnv*>(envP ? envP->GetPtr("xrdEnv*") : nullptr);
+        auto tlsCtx = static_cast<XrdTlsContext*>(xrdEnv ? xrdEnv->GetPtr("XrdTlsContext*") : nullptr);
+        if (tlsCtx) {
+            auto params = tlsCtx->GetParams();
+            if (params && !params->cafile.empty()) {
+                scitoken_config_set_str("tls.ca_file", params->cafile.c_str(), nullptr);
+            }
+        }
 
         return Reconfig();
     }
@@ -1259,10 +1269,10 @@ private:
 };
 
 void InitAccSciTokens(XrdSysLogger *lp, const char *cfn, const char *parm,
-                      XrdAccAuthorize *accP)
+                      XrdAccAuthorize *accP, XrdOucEnv *envP)
 {
     try {
-        accSciTokens = new XrdAccSciTokens(lp, parm, accP);
+        accSciTokens = new XrdAccSciTokens(lp, parm, accP, envP);
         SciTokensHelper = accSciTokens;
     } catch (std::exception &) {
     }
@@ -1273,7 +1283,7 @@ extern "C" {
 XrdAccAuthorize *XrdAccAuthorizeObjAdd(XrdSysLogger *lp,
                                           const char   *cfn,
                                           const char   *parm,
-                                       XrdOucEnv    * /*not used*/,
+                                       XrdOucEnv       *envP,
                                        XrdAccAuthorize *accP)
 {
     // Record the parent authorization plugin. There is no need to use
@@ -1283,7 +1293,7 @@ XrdAccAuthorize *XrdAccAuthorizeObjAdd(XrdSysLogger *lp,
     // If we have been initialized by a previous load, them return that result.
     // Otherwise, it's the first time through, get a new SciTokens authorizer.
     //
-    if (!accSciTokens) InitAccSciTokens(lp, cfn, parm, accP);
+    if (!accSciTokens) InitAccSciTokens(lp, cfn, parm, accP, envP);
     return accSciTokens;
 }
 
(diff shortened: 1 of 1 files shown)


src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -31,6 +32,13 @@
 XrdVERSIONINFO(XrdAccAuthorizeObject, XrdAccSciTokens);
 XrdVERSIONINFO(XrdAccAuthorizeObjAdd, XrdAccSciTokens);
 
+namespace XrdGlobal {
+// The XrdOucEnv object passed to the plugin has a pointer called `XrdTLSContext*`;
+// however, that's a cloned pointer without any of the verification data (CA file or
+// directory), which is precisely what we need.  So, we use the XrdGlobal version.
```
amadio, line 38: Sounds like we should rather solve this issue by passing `XrdOucEnv` with a `XrdTLSContext*` that has the verification data, no? This works around the intended interfaces and creates a hidden connection between `XrdSciTokens` and `XrdTlsContext` which we should try to avoid. Let's hear what @abh3 thinks.

amadio:
> I'm working on setting up some end-to-end integration tests including the SciTokens library.

I like the idea. We've been working on making all our tests run with `make test`. In XRootD 5.7 I will merge @AngeloGalav's work which makes great progress towards this. I had in mind to add also tests for the GSI, Kerberos, and other authentication methods, so if you could integrate your tests with the existing ones, that would be great.

abh3:
I agree with Guilherme, we try to use existing mechanisms. In the past,, 
we tried using global systems to pass information and it turned into a 
dependency mess. So, now we use the env mechanism which makes life much 
easier. It's easy to do and the pointer can be placed in the XrdEnv object 
at the same time it is inserted into the XrdProtocol object in 
XrdConfig.cc. I should point out that we do have a problem that the env 
pointer is only passed to XrdAccAuthorizeObjAdd() not if the SciTokens 
library is setup as the only or the base authorization plugin. This is a 
historical problem and we correct that during major releases in any plugin 
interface where the need to access the environment has been pointed out, 
like in this case. Maybe we should change it for all plugin interfaces at 
this point.

Andy

  In fact, we use this mechanism to provide HTTP wih the 
standard context. The Sciokens library can also use that context during 
initialization. I think he context is created and placed in
  get access to the 
standard context yo
I would just put the "XrdTlsContext*" pointer in the Xrd created 
ENV. In facyt, tht is alreay done so no need to

abh3:
One other think I should point out, we don't need to wait for a major release if wedo what we have done in the past and version the XrdAccAuthorizeObject() API as we have done for other plugns in the same situation. We simly add XrdAccAuthorizeObject2() which is used to pass the additional env parameter. We might as well as bite the bullet for 4.7.0. Any other plugin initialization API's fall into he same hole at this point?

abh3:
The particular patch that solves this problem will be to export into the env the uncloned version of TLS context which is what you want. Note that the cloned context is specific to the xroot protocol created env and what you need is the xrd version of that pointer.. The way you would get is is:
(XrdTlsContext*)(((XrdOucEnv*)(xrootdenv->GetPtr("xrdEnv*)))->GetPtr("XrdTlsContext*");
Where: xrootdenv is the env passed to the SciTokens library via the params.

Then you can use the getParams() method on that pointer. Additionally you can use the passed env either via the stacking API or if SciTokens is not stacked then you would need to get it via the V2 initialization API (see the XrdAccAuthorize.hh file).

Push to devel should be there shortly.

abh3:
@bbockelm The required patches are in the devel branch. You should be able to get the full fledged XrdTlsContext pointer to get the context parameters. Could you please tell us if there is something else you need before we close this pull request?

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: Switched over to the new interface... resolving conversation.

bbockelm:
Just rebased on @abh3's latest commit to the devel branch and switched over to the new object with verification parameters... seems to work like a charm!

amadio:
If I understand correctly, this should work fine with current released versions of scitokens-cpp, right? I mean, it compiles and loads, but the CA setup will only work with a future version including your latest changes. Looks good to me now. Will merge into `devel` for v5.7 after @abh3's approval.

abh3:
Ah, we seem to have a poblem here. The current pull request still has changes to code that no longer needs changes. Brian's commit should only affect the SciTokens code not anything else. So, we could close this pull request and ask for a clean pull request that uses the current code. As it is I am confused on what git tells me will change.

amadio:
Brian just needs to rebase again on devel, because I added David's recent bug fix to master and rebased devel on top. After rebasing, only the change to scitokens will remain visibke here.

amadio:
Actually, it was just a click away in GitHub's interface, so I did it already.

amadio:
I am seeing this when trying to compile:
```
[ 90%] Building CXX object src/CMakeFiles/XrdPosixPreload.dir/XrdPosix/XrdPosix.cc.o
/home/amadio/src/xrootd/src/XrdSciTokens/XrdSciTokensAccess.cc: In member function ‘bool XrdAccSciTokens::Config(XrdOucEnv*)’:
/home/amadio/src/xrootd/src/XrdSciTokens/XrdSciTokensAccess.cc:971:17: error: ‘scitoken_config_set_str’ was not declared in this scope
  971 |                 scitoken_config_set_str("tls.ca_file", params->cafile.c_str(), nullptr);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~
make[2]: *** [src/CMakeFiles/XrdAccSciTokens-5.dir/build.make:76: src/CMakeFiles/XrdAccSciTokens-5.dir/XrdSciTokens/XrdSciTokensAccess.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:2422: src/CMakeFiles/XrdAccSciTokens-5.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
I wonder how it compiled in the CI.

bbockelm:
@amadio - you might have an older version of the scitokens library.  That API function was added in [v1.0.2](https://github.com/scitokens/scitokens-cpp/releases/tag/v1.0.2).

amadio:
Yes, it was a problem on my machine, sorry for the noise.

amadio:
I just noticed that Debian 11, Debian12, and Ubuntu 22.04 all have `libscitokens-dev` that is too old for the changes in this PR. The newest one is 0.7.3 in Debian 12, and 0.7.0 in Ubuntu 22.04. We will need to file version bump bugs on those distros before we can put this change in an XRootD release. See https://repology.org/project/scitokens-cpp/versions.

amadio:
@ellert Is it possible to update scitokens-cpp on Debian/Ubuntu? Where should I file these bugs? I see in EPEL it's all updated to 1.0.2 already, so we are ok there.

ellert:
Debian testing and Ubuntu 23.10 have scitokens-cpp 1.0.2.
Debian is different from Fedora in that it is not easy to update a package in a stable release.
Several criteria must be fulfilled in order to propose an update to a stable Debian release. A proposed update must fix a bug and "The bug should be of severity 'important' or higher". which means "a bug which has a major effect on the usability of a package, without rendering it completely unusable to everyone". Just proposing an update to a newer version will not work.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -962,6 +963,15 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
         }
         m_log.Emsg("Config", "Logging levels enabled -", LogMaskToString(m_log.getMsgMask()).c_str());
 
+        auto xrdEnv = static_cast<XrdOucEnv*>(envP ? envP->GetPtr("xrdEnv*") : nullptr);
+        auto tlsCtx = static_cast<XrdTlsContext*>(xrdEnv ? xrdEnv->GetPtr("XrdTlsContext*") : nullptr);
+        if (tlsCtx) {
+            auto params = tlsCtx->GetParams();
+            if (params && !params->cafile.empty()) {
+                scitoken_config_set_str("tls.ca_file", params->cafile.c_str(), nullptr);
```
amadio, line 971: @bbockelm Given that we support these older Debian/Ubuntu distributions, and scitokens cannot be updated there, I think this part of the patch will need to be `#ifdef`ed out depending on the version of scitokens found. Can you please make a patch for that? Thanks!

amadio:
@ellert Thanks for the explanation. It's a pity that it cannot be updated. However, we cannot drop support for Debian 12 and Ubuntu 22.04, which are recent, so the solution will have to be to adapt to the available scitokens version. Luckily in EPEL scitokens 1.0.2 is available everywhere.

bbockelm:
@amadio - would you prefer to make this `#ifdef`'d out or make this something determined at runtime with `dlopen` or similar?

amadio:
Since the version we build against and the runtime version will likely always be the same, I think that `#ifdef` is probably the better way, but if you'd rather check at runtime, I guess that could work too. Just please test on Debian 11 or Ubuntu 22.04 to make sure it works.

PR #2098, opened 2023-10-06
[XrdCl] Give error return code if xrdfs rm fails to delete any file. Closes #2097

### LINKED ISSUE #2097: xrdfs rm - not exiting with a !=0 exit code in case of errors
Hi,

Since I can't reopen https://github.com/xrootd/xrootd/issues/1885 I'll copy-paste the request here.

This bug seems to have reappeared at least from 5.6.0 and beyond. Even for single files, regardless of the success or failure of the operation, the exit code is always 0:

```
$ v5.6.0-6/bin/xrdfs eosalice.cern.ch rm '/33/33333/33333'
rm /33/33333/33333 : [ERROR] Error response: permission denied
$ echo $?
0

$ v5.6.0-6/bin/xrdfs eosalice.cern.ch rm '/33/33333/33333' '/44/44444/444444'
rm /33/33333/33333 : [ERROR] Error response: permission denied
rm /44/44444/444444 : [ERROR] Error response: permission denied
$ echo $?
0
```

and the same in 5.6.2:

```
$ xrdcp -version
v5.6.2
$ xrdfs eosalice.cern.ch rm '/33/33333/33333'
rm /33/33333/33333 : [ERROR] Error response: permission denied
$ echo $?
0
```

Could you please re-apply the fix?

Cheers,

.costin

smithdh:
Fixing #2097

### DIFF
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -701,9 +701,12 @@ XRootDStatus DoRm( FileSystem                      *fs,
   }
 
   //----------------------------------------------------------------------------
-  // Run the query
+  // Run the query:
+  // Parallel() will take the vector of Pipeline by reference and empty the
+  // vector, so rms.size() will change after the call.
   //----------------------------------------------------------------------------
-  XRootDStatus st = WaitFor( Parallel( rms ).AtLeast( rms.size() ) );
+  const size_t rs = rms.size();
+  XRootDStatus st = WaitFor( Parallel( rms ).AtLeast( rs ) );
   if( !st.IsOK() )
     return st;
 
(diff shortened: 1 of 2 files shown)


abh3 (approved):
Looks OK to me though, frankly, I don't quite understand why it  should be better than what was there before unless rms.size() decreases while the rm is actually executing.

smithdh:
Thanks @abh3 , I've added a code comment since indeed it doesn't seem clear what is happening there.

amadio:
Looks good, just let me know when it's ready to merge. Also, please remember to use "Fixes #2097" or "Closes #2097" instead of "Fixing #2097" in commit messages or PR title/description. GitHub is not smart enough to link this as closing the issue otherwise.

PR #2102, opened 2023-10-10
[XrdHttp] Fix parsing of chunked PUT lengths

djw8605:
The bugfix change was from using `prot->BufAvailable()` to `prot->BufUsed()`.  I also added sanity checks around the size of the chunk and a few trace statement to help with any future problems.

Chunked encoded PUT are formatted with a length before each chunk. Fix the parsing to only parse the buffer that has been filled. Additionally, add a sanity check on the size of a chunk in order to protect against malformed of malicious sized chunks.

Also, added additional trace statements to help when the chunk is rejected.


ccaffy (approved):
Looks good to me. Thanks for your contribution!

ccaffy:
@amadio , do you want to merge this in devel or in master?

abh3:
Thanks @djw8605! While we can redirect pull requests from master to devel it won't necessarily work every time. In the future, please use the devel branch as your base code. Thanks again!

PR #2103, opened 2023-10-11
[XrdHttp] Fix max to min when reading chunk length

djw8605:
Small change to using the minimum of either the used buffer size or the first 13 characters of a chunk to read the chunk length.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -1334,7 +1334,7 @@ int XrdHttpReq::ProcessHTTPReq() {
             // This is to prevent a malicious client from sending a very large chunk size
             // or a malformed chunk request.
             // 1TB in base-16 is 0x40000000000, so only allow 11 characters, plus the CRLF
-            long long max_chunk_size_chars = std::max(static_cast<long long>(prot->BuffUsed()), static_cast<long long>(13));
+            long long max_chunk_size_chars = std::min(static_cast<long long>(prot->BuffUsed()), static_cast<long long>(13));
             for (; idx < max_chunk_size_chars; idx++) {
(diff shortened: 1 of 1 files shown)


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1334,7 +1334,7 @@ int XrdHttpReq::ProcessHTTPReq() {
             // This is to prevent a malicious client from sending a very large chunk size
             // or a malformed chunk request.
             // 1TB in base-16 is 0x40000000000, so only allow 11 characters, plus the CRLF
-            long long max_chunk_size_chars = std::max(static_cast<long long>(prot->BuffUsed()), static_cast<long long>(13));
+            long long max_chunk_size_chars = std::min(static_cast<long long>(prot->BuffUsed()), static_cast<long long>(13));
```
abh3, line 1337: I am a bit skeptical this corresponds to what the variable names say the contents are. Looking at BuffUsed() it either returns the number of bytes available (.e. not used) or some ridiculous value which is greater than the actual buffer size. So, for sanuty sake could you review BuffUsed and verify that what you think you are ggeting is actally what you are getting?

abh3 (commented):
Pretty big change from max to min. There is a potetial issue, please see comment and respond.

src/XrdHttp/XrdHttpReq.cc
djw8605: At least in my tests, `BuffUsed` returns the number of bytes in the buffer that is filled with data from the client.  The goal of this line is to read at most 13 characters for the size of the chunk.  But it can be less.  With using max, reading more than the size of the buffer will result in reading data not from the client.

PR #2104, opened 2023-10-13
Implement only-if-cached cache control using XrdPfcFsctl

alja:
This is a new implementation of  #1954 using XrdPfcFsctl object in XrdPssFile::Open()

Note: 
A stderr code ETIME is returned when cache does not exist. The ETIMEDOUT code would be interpreted as delay error in XrdOfs::Emsg().

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -968,13 +968,26 @@ int Cache::ConsiderCached(const char *curl)
                {
                   is_cached = true;
                }
+               else if (info.GetFileSize() == 0)
+               {
+                  is_cached = true;
+               }
                else
                {
                   long long fileSize = info.GetFileSize();
                   long long bytesRead = info.GetNDownloadedBytes();
-                  if (bytesRead > m_configuration.m_onlyIfCachedMinSize &&
-                      (float)bytesRead/fileSize > m_configuration.m_onlyIfCachedMinFrac)
+
+                  if (fileSize < m_configuration.m_onlyIfCachedMinSize)
+                  {
+                     if ((float)bytesRead / fileSize > m_configuration.m_onlyIfCachedMinFrac)
                         is_cached = true;
+                  }
+                  else
+                  {
+                     if (bytesRead > m_configuration.m_onlyIfCachedMinSize &&
+                         (float)bytesRead / fileSize > m_configuration.m_onlyIfCachedMinFrac)
+                        is_cached = true;
+                  }
                }
             }
             infoFile->Close();
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -896,6 +896,104 @@ int Cache::LocalFilePath(const char *curl, char *buff, int blen,
    return -ENOENT;
 }
 
+//______________________________________________________________________________
+// Check if the file is cached including m_onlyIfCachedMinSize and m_onlyIfCachedMinFrac
+// pfc configuration parameters. The logic of accessing the Info file is the same
+// as in Cache::LocalFilePath.
+//! @return 0      - the file is complete and the local path to the file is in
+//!                  the buffer, if it has been supllied.
+//!
+//! @return <0     - the request could not be fulfilled. The return value is
+//!                  -errno describing why. If a buffer was supplied and a
+//!                  path could be generated it is returned only if "why" is
+//!                  ForCheck or ForInfo. Otherwise, a null path is returned.
+//!
+//! @return >0     - Reserved for future use.
+//------------------------------------------------------------------------------
+int Cache::ConsiderCached(const char *curl)
+{
+   TRACE(Debug, "ConsiderFileCached '" << curl << "'" );
+
+   XrdCl::URL url(curl);
+   std::string f_name = url.GetPath();
+   std::string i_name = f_name + Info::s_infoExtension;
+
+   {
+      XrdSysCondVarHelper lock(&m_active_cond);
+      m_purge_delay_set.insert(f_name);
+   }
+
+   struct stat sbuff, sbuff2;
+   if (m_oss->Stat(f_name.c_str(), &sbuff) == XrdOssOK &&
+       m_oss->Stat(i_name.c_str(), &sbuff2) == XrdOssOK)
+   {
+      if (S_ISDIR(sbuff.st_mode))
+      {
+         TRACE(Info, "ConsiderCached '" << curl << ", why=ForInfo" << " -> EISDIR");
+         return -EISDIR;
+      }
+      else
+      {
+         bool read_ok = false;
+         bool is_cached = false;
+
+         // Lock and check if the file is active. If NOT, keep the lock
+         // and add dummy access after successful reading of info file.
+         // If it IS active, just release the lock, this ongoing access will
+         // assure the file continues to exist.
+
+         // XXXX How can I just loop over the cinfo file when active?
+         // Can I not get is_complete from the existing file?
+         // Do I still want to inject access record?
+         // Oh, it writes only if not active .... still let's try to use existing File.
+
+         m_active_cond.Lock();
+
+         bool is_active = m_active.find(f_name) != m_active.end();
+
+         if (is_active)
+            m_active_cond.UnLock();
+
+         XrdOssDF *infoFile = m_oss->newFile(m_configuration.m_username.c_str());
+         XrdOucEnv myEnv;
+         int res = infoFile->Open(i_name.c_str(), O_RDWR, 0600, myEnv);
+         if (res >= 0)
+         {
+            Info info(m_trace, 0);
+            if (info.Read(infoFile, i_name.c_str()))
+            {
+               read_ok = true;
+
+               if (info.IsComplete())
+               {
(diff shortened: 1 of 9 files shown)


alja:
## Test http header with curl

Trying to access file with the only-if-cached directive where file does not exist in the cache. The 504 Gateway Timeout error is in the response
```
alja@green ~> rm xrootd/cached-data/alja/xrootd/data/A_FILE*
alja@green ~> curl -D respTxt -H "Cache-Control: only-if-cached"  http://localhost:7777/alja/xrootd/data/A_FILE
Unable to open /alja/xrootd/data/A_FILE; timer expired
alja@green ~> cat respTxt
HTTP/1.1 504 Gateway Timeout
Connection: Close
Server: XrootD/v5.6.2-12-gf39cf279b
Content-Length: 55
```

Cache file removing any cache control directive and then access it with the only-if-cached condition:
```
alja@green ~> curl -D respTxt -O http://localhost:7777/alja/xrootd/data/A_FILE
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  8211  100  8211    0     0   5173      0  0:00:01  0:00:01 --:--:--  5173
alja@green ~>  curl -O -D respTxt -H "Cache-Control: only-if-cached"  http://localhost:7777/alja/xrootd/data/A_FILE
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  8211  100  8211    0     0   6022      0  0:00:01  0:00:01 --:--:--  6019
```

alja:
## Test  only-if-cached mode with xrdcp adding CGI options
Get an ETIME error if file is not cached with only-if-cached option
```
alja@green ~> xrdcp -f root://localhost:7777//alja/xrootd/data/nano.root?cache-control=only-if-cached /tmp
Run: [ERROR] Server responded with an error: [3035] Unable to open /alja/xrootd/data/nano.root; timer expired (source)
```
Cache file and then check file is copied using the only-if-cached directive
```
alja@green ~> xrdcp -f root://localhost:7777//alja/xrootd/data/nano.root /tmp
[39.72MB/39.72MB][100%][==================================================][39.72MB/s]
alja@green ~> xrdcp -f root://localhost:7777//alja/xrootd/data/nano.root?cache-control=only-if-cached /tmp
[39.72MB/39.72MB][100%][==================================================][39.72MB/s]

```

alja:
@abh
The only message that is printed in log is on the error with the info severity level
https://github.com/alja/xrootd/blob/only-if-cached-master/src/XrdPfc/XrdPfcFSctl.cc#L148

Let me know if this should be more verbose.

abh3:
Hi Alja,

I don't think we need more info. The cache control is an option and is 
expected to occasionally return an error as the correct response.

Andy

abh3:
Could you also post wwhat happens when you use curl?

alja:
@abh
If cache does not exist, the curl command prints the timer expired error:

```
# curl -D respTxt -H "Cache-Control: only-if-cached"  http://localhost:7777/alja/xrootd/data/A_FILE
Unable to open /alja/xrootd/data/A_FILE; timer expired 
```
With curl option -D the server response lines are written in respTxt text file. The content in the case on non existing cache is:
```
# cat respTxt
HTTP/1.1 504 Gateway Timeout
Connection: Close
Server: XrootD/v5.6.2-12-gf39cf279b
Content-Length: 55

```

src/XrdPfc/XrdPfcFSctl.cc
```diff
@@ -131,6 +132,23 @@ int XrdPfcFSctl::FSctl(const int               cmd,
    rc = SFS_ERROR;
   }
 
+  if (!strcmp(xeq, "cached"))
+  {
+     const char* path = args.ArgP[0];
+     int rval = myCache.LocalFilePath(path, nullptr, 0, XrdOucCache::LFP_Reason::ForInfo);
```
abh3, line 138: So, if I understand this correctly, if at least one block of the file is cached we will get a zero return code and say it's OK.  Is that correct or am I missing something here?

I say that because a previous comment said only if the file is "fully cached". That's the opposite extreme and I do wonder about tat choice as well.

abh3 (commented):
I must say this became a much more concise patch that any of us expected. Out of curiosity, what happened to the original "XrdOuc/XrdOucCacheDirective.hh"? Is it no longer relevant or just not in this case.

BTW I'll merge this once I get a couple of answers, mostly on what "if-cached" really means here.

amadio:
I'd just like to note that, as this is a new feature, it should go into the devel branch instead of master. Once it's ready for merge, I will change the base, rebase, then merge.

abh3:
Absolutely correct. Thank you for pointing that out.

src/XrdPfc/XrdPfcFSctl.cc
alja: At first @bbockelm  was thinking that a file is considered cached if any part of the file is cached. Last week at the Pelican meeting the decision was that the file must be fully cached.

I can’t tell which of the two options would work best.
Perhaps this can be a configuration option.

alja:
@amadio Thanks  Guilherme! I will keep that in mind for the future PRs.

src/XrdPfc/XrdPfcFSctl.cc
abh3: Indeed, this is going to go back and forth until someone figures out what really works in practice. Yes, a configuration option would be best here. I'd go with something like pfc.ifcache <num>% where <num> is a positive integer represent9ing that you return true should the file be at least that percentage in the cache. The question is how to actually get that information to make that decision.

alja:
@abh3 @bbockelm @osschar 

I have added new configuration options to tune the decision if the file is being cached.
There pfc.onlyifcached option has two named parameters: minsize and minfrac. The min size can be given in numeric or human readable form (using a2sz utility)

Here is the example:
```pfc.onlyifcached minsize 500m minfrac 0.5```

The default minsize is 1024*1024 bytes. The default minfrac is 1.0. 

There is one more detail:
For now the minsize **and** minfrac has to be both true if file is considered cached.  In case the file is complete the minsize and minfrac values are not checked.

Please review and comment on the changes.

File: src/XrdPfc/XrdPfc.cc [hunk 2]
```diff
@@ -896,6 +896,104 @@ int Cache::LocalFilePath(const char *curl, char *buff, int blen,
...
+               {
+                  is_cached = true;
+               }
+               else
+               {
+                  long long fileSize = info.GetFileSize();
+                  long long bytesRead = info.GetNDownloadedBytes();
+                  if (bytesRead > m_configuration.m_onlyIfCachedMinSize &&
+                      (float)bytesRead/fileSize > m_configuration.m_onlyIfCachedMinFrac)
+                        is_cached = true;
```
amadio, line 977: What happens to small files? It seems to me that files smaller than the minimum required cached size will always be considered uncached with the condition above and the default of `1024*1024` bytes for minimum size.

amadio: Also, this will cause division by zero with files of zero size, no?

alja: > What happens to small files? It seems to me that files smaller than the minimum required cached size will always be considered uncached with the condition above and the default of `1024*1024` bytes for minimum size.

I did not think about this case. 
So for the files smaller that pfc.onlyifcached minSize, we check only the fraction of file downloaded?
Note, there will be cases also for large files where the only-if-cached directive will let the cache pass by. As I understand, other clients which don't have the onlyifcached condition will take care that cache is being filled.

alja: > Also, this will cause division by zero with files of zero size, no?

Thank you for pointing this out!
I will add that check.

I wonder if there is a cache Info file for empty files.

alja: @amadio Please check.

amadio: It looks good now, thank you!

amadio:
@alja, @abh3 From my point of view, this looks ready for merging. Please let me know if I should wait, otherwise, I will merge this tomorrow. Cheers,

abh3:
@amadio I think it's OK to merge this. We still need to get documentation 
for the new config directive.

amadio:
@abh3 Ok, I will give it a go with the new CI and merge once it completes then.

alja:
Where are the sources for the Pss html documentation? E.g https://xrootd.slac.stanford.edu/doc/dev56/pss_config.htm
I could not see it in this repository.

abh3:
The sources are not in the repo as the documentation is binary (i.e. MS 
Word). So, simply type up some text showing the syntax of the directive 
along with an explanation for each specifiable value and notes on any side 
effects with certain combinations or situations. Use a roughly similar 
format as whown in the current documenatation for the pfc. Send it to me 
or post it. I can then put it into MS word.

alja:
@abh3 
I have documented the 'onlyifcached' option in my copy of html documentation:
http://xrd-cache-1.t2.ucsd.edu/alja/xrd-dev/XRootD%20Proxy%20Configuration.html

PR #2109, opened 2023-10-19
Implemented HTTP TPC Packet Marking

ccaffy:
Hi @abh3 ,

I think this functionality is progressing well.

I've managed to have some packet marking emitted with HTTP TPC transfers PUSH and PULL single and multi (PULL) streams.

Now, by looking at the JSON of the firefly, I'm not sure this is working 100%.

If I do a PULL single stream `curl -v --capath /etc/grid-security/certificates -L -X COPY  -H "Source: $DST" -H "SciTag: 144" "$SRC"` for example, here are the JSON fireflies that is sent by the pulling server at the beginning and the end of the transfer:

```json
{"version":1,"flow-lifecycle":{"state":"start","current-time":"2023-10-19T16:06:25.224773+00:00","start-time":"2023-10-19T16:06:25.224773+00:00"},"usage":{"received":0,"sent":136},"netlink":{"rtt":0.061},"context":{"experiment-id":2,"activity-id":16,"application":"http-tpc"},"flow-id":{"afi":"ipv4","src-ip":"xxx.xxx.xxx.xx","dst-ip":"xxx.xxx.xxx.xx","protocol":"tcp","src-port":52326,"dst-port":2001}}
```

```json
{"version":1,"flow-lifecycle":{"state":"end","current-time":"2023-10-19T16:07:49.906540+00:00","start-time":"2023-10-19T16:06:25.224773+00:00","end-time":"2023-10-19T16:07:49.906540+00:00"},"usage":{"received":0,"sent":0},"netlink":{"rtt":0.000},"context":{"experiment-id":2,"activity-id":16,"application":"http-tpc"},"flow-id":{"afi":"ipv4","src-ip":"xxx.xxx.xxx.xx","dst-ip":"xxx.xxx.xxx.xx","protocol":"tcp","src-port":52326,"dst-port":2001}}
```

What puzzles me is the value of `received` and `sent` that are equal to 0. Are these normal? Otherwise, the other fields look correct to me :)

Thanks.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -225,22 +225,24 @@ int XrdHttpReq::parseHost(char *line) {
 }
 
 void XrdHttpReq::parseScitag(const std::string & val) {
-  int scitag = 0;
+  // The scitag header has been populated and the packet marking was configured, the scitag will either be equal to 0
+  // or to the value passed by the client
+  mScitag = 0;
   std::string scitagS = val;
   trim(scitagS);
   if(scitagS.size()) {
     if(scitagS[0] != '-') {
       try {
-        scitag = std::stoi(scitagS.c_str(), nullptr, 10);
-        if (scitag > XrdNetPMark::maxTotID) {
-          scitag = 0;
+        mScitag = std::stoi(scitagS.c_str(), nullptr, 10);
+        if (mScitag > XrdNetPMark::maxTotID || mScitag < 0) {
+          mScitag = 0;
         }
       } catch (...) {
         //Nothing to do, scitag = 0 by default
       }
     }
   }
-  addCgi("scitag.flow", std::to_string(scitag));
+  addCgi("scitag.flow", std::to_string(mScitag));
 }
 
 int XrdHttpReq::parseFirstLine(char *line, int len) {
--- a/src/XrdTpc.cmake
+++ b/src/XrdTpc.cmake
@@ -38,7 +38,8 @@ if( BUILD_TPC )
     XrdTpc/XrdTpcCurlMulti.cc     XrdTpc/XrdTpcCurlMulti.hh
     XrdTpc/XrdTpcState.cc         XrdTpc/XrdTpcState.hh
     XrdTpc/XrdTpcStream.cc        XrdTpc/XrdTpcStream.hh
-    XrdTpc/XrdTpcTPC.cc           XrdTpc/XrdTpcTPC.hh)
+    XrdTpc/XrdTpcTPC.cc           XrdTpc/XrdTpcTPC.hh
+    XrdTpc/PMarkManager.cc        XrdTpc/PMarkManager.hh)
 
   target_link_libraries(
     ${LIB_XRD_TPC}
--- a/src/XrdTpc/PMarkManager.cc
+++ b/src/XrdTpc/PMarkManager.cc
@@ -1,0 +1,72 @@
+//------------------------------------------------------------------------------
+// This file is part of XrdTpcTPC
+//
+// Copyright (c) 2023 by European Organization for Nuclear Research (CERN)
+// Author: Cedric Caffy <ccaffy@cern.ch>
+// File Date: Oct 2023
+//------------------------------------------------------------------------------
+// XRootD is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// XRootD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+
+
+#include <sstream>
+#include "PMarkManager.hh"
+
+PMarkManager::SocketInfo::SocketInfo(int fd, const struct sockaddr * sockP) {
+  netAddr.Set(sockP,fd);
+  client.addrInfo = static_cast<XrdNetAddrInfo*>(&netAddr);
+}
+
+PMarkManager::PMarkManager(XrdNetPMark *pmark) : mPmark(pmark), mTransferWillStart(false) {}
+
+void PMarkManager::addFd(int fd, const struct sockaddr * sockP) {
+  if(mPmark && mTransferWillStart && mReq->mSciTag >= 0) {
+    // The transfer will start and the packet marking has been configured, this socket must be registered for future packet marking
+    mSocketInfos.emplace(fd, sockP);
+  }
+}
+
+void PMarkManager::startTransfer(XrdHttpExtReq * req) {
+  mReq = req;
+  mTransferWillStart = true;
+}
+
+void PMarkManager::beginPMarks() {
+  if(!mSocketInfos.empty() && mPmarkHandles.empty()) {
+    // Create the first pmark handle that will be used as a basis for the other handles
+    // if that handle cannot be created (mPmark->Begin() would return nullptr), then the packet marking will not work
+    // This base pmark handle will be placed at the beginning of the vector of pmark handles
+    std::stringstream ss;
+    ss << "scitag.flow=" << mReq->mSciTag;
+    auto sockInfo = mSocketInfos.front();
+    mInitialFD = sockInfo.client.addrInfo->SockFD();
+    mPmarkHandles.emplace(mInitialFD,mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc"));
+    mSocketInfos.pop();
+  } else {
+    // The first pmark handle was created, or not. Create the other pmark handles from the other connected sockets
+    while(!mSocketInfos.empty()) {
+      auto & sockInfo = mSocketInfos.front();
+      if(mPmarkHandles[mInitialFD]){
+        mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),mPmark->Begin(*sockInfo.client.addrInfo, *mPmarkHandles[mInitialFD], nullptr));
+      }
+      mSocketInfos.pop();
+    }
+  }
+}
+
+void PMarkManager::endPmark(int fd) {
+  // We need to delete the PMark handle associated to the fd passed in parameter
+  // we just look for it and reset the unique_ptr to nullptr to trigger the PMark handle deletion
+  mPmarkHandles.erase(fd);
+}
--- a/src/XrdTpc/PMarkManager.hh
+++ b/src/XrdTpc/PMarkManager.hh
@@ -1,0 +1,113 @@
+//------------------------------------------------------------------------------
+// This file is part of XrdTpcTPC
+//
+// Copyright (c) 2023 by European Organization for Nuclear Research (CERN)
+// Author: Cedric Caffy <ccaffy@cern.ch>
+// File Date: Oct 2023
+//------------------------------------------------------------------------------
+// XRootD is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Lesser General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// XRootD is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+#ifndef XROOTD_PMARKMANAGER_HH
+#define XROOTD_PMARKMANAGER_HH
+
+#include "XrdNet/XrdNetPMark.hh"
+#include "XrdSec/XrdSecEntity.hh"
+#include "XrdNet/XrdNetAddrInfo.hh"
+#include "XrdNet/XrdNetAddr.hh"
+#include "XrdHttp/XrdHttpExtHandler.hh"
+
+#include <map>
+#include <memory>
+#include <queue>
+
+/**
+ * This class will manage packet marking handles for TPC transfers
+ *
+ * Each time a socket will be opened by curl (via the opensocket_callback), the manager
+ * will register the related information to the socket.
+ *
+ * Once the transfer will start we will start the packet marking by creating XrdNetPMark::Handle
+ * objects from the socket information previously registered.
+ *
+ * In the case of multi-stream HTTP TPC transfers, a packet marking handle will be created for each stream.
+ * The first one will be created as a basic one. The other will be created using the first packet marking handle as a basis.
+ */
+class PMarkManager {
+public:
+
+  /**
+   * This class allows to create and keep a XrdSecEntity object
+   * from the socket file descriptor and address
+   * Everything is done on the constructor
+   *
+   * These infos will be used later on when we create new PMark handles
+   */
+  class SocketInfo {
+  public:
+    SocketInfo(int fd, const struct sockaddr * sockP);
+    XrdNetAddr netAddr;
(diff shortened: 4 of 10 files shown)


ccaffy:
I think I know why we get 0 and 0 for bytes received and sent. It's because when the destructor `XrdNetPMarkFF` happens, the socket has already been closed by curl... I need to find a way out of it :)

ccaffy:
OK now we have what we want :) Destroying the socket' PMark handle before it is closed makes everything working fine:

```
{"version":1,"flow-lifecycle":{"state":"end","current-time":"2023-10-20T13:09:21.190627+00:00","start-time":"2023-10-20T13:08:51.055092+00:00","end-time":"2023-10-20T13:09:21.190627+00:00"},"usage":{"received":1795180572,"sent":18247},"netlink":{"rtt":1.365},"context":{"experiment-id":2,"activity-id":16,"application":"http-tpc"},"flow-id":{"afi":"ipv4","src-ip":"xxx.xxx.xxx.xx","dst-ip":"xxx.xxx.xxx.xx","protocol":"tcp","src-port":52428,"dst-port":2001}}
```

abh3 (commented):
First pass, mostly comments on simplification. Generally, I avoid make_unique when I can make the compiler control the lifetime of an object. Doing it that way is more straightforward than wrapping things in a "unique" object. Here, it seems that this is quite possible based on how the objects are defined, used, and disposed of. Of course, maybe I misunderstood how the PmakManager object is actually used. Also, note my concern about declaring pointers that shadow other objects that are referenced by pointers themselves. It's not something I would suggest doing at all and doesn't seem necessary here as well.

I leave it all as comments to be resolved before we move on.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -225,22 +225,24 @@ int XrdHttpReq::parseHost(char *line) {
...
   std::string scitagS = val;
   trim(scitagS);
   if(scitagS.size()) {
     if(scitagS[0] != '-') {
       try {
-        scitag = std::stoi(scitagS.c_str(), nullptr, 10);
-        if (scitag > XrdNetPMark::maxTotID) {
-          scitag = 0;
+        mScitag = std::stoi(scitagS.c_str(), nullptr, 10);
+        if (mScitag > XrdNetPMark::maxTotID) {
```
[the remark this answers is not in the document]
ccaffy: Thanks, fixed!

src/XrdTpc.cmake
```diff
@@ -38,7 +38,8 @@ if( BUILD_TPC )
     XrdTpc/XrdTpcCurlMulti.cc     XrdTpc/XrdTpcCurlMulti.hh
     XrdTpc/XrdTpcState.cc         XrdTpc/XrdTpcState.hh
     XrdTpc/XrdTpcStream.cc        XrdTpc/XrdTpcStream.hh
-    XrdTpc/XrdTpcTPC.cc           XrdTpc/XrdTpcTPC.hh)
+    XrdTpc/XrdTpcTPC.cc           XrdTpc/XrdTpcTPC.hh
+    XrdTpc/PMarkManager.cc XrdTpc/PMarkManager.hh)
```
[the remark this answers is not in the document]
ccaffy: Fixed :)

src/XrdTpc/PMarkManager.cc
```diff
@@ -0,0 +1,80 @@
...
+
+
+#include <sstream>
+#include <algorithm>
+#include "PMarkManager.hh"
+
+PMarkManager::SocketInfo::SocketInfo(int fd, const struct sockaddr * sockP) {
+  netAddr = std::make_unique<XrdNetAddr>();
+  netAddr->Set(sockP,fd);
+  netAddrInfo = std::make_unique<XrdNetAddrInfo>(*netAddr);
```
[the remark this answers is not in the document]
ccaffy: True, I simplified this now :)

src/XrdTpc/PMarkManager.cc
```diff
@@ -0,0 +1,80 @@
...
+  client = std::make_unique<XrdSecEntity>();
```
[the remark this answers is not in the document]
ccaffy: Correct, changed to normal declarations :)

src/XrdTpc/PMarkManager.cc
```diff
@@ -0,0 +1,80 @@
...
+  client->addrInfo = netAddrInfo.get();
```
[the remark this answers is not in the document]
ccaffy: Yep, I had to go that way to make things working :)

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -119,18 +119,33 @@ int TPCHandler::opensocket_callback(void *clientp,
...
 {
-// See what kind of address will be used to connect
-//
-if (purpose == CURLSOCKTYPE_IPCXN && clientp)
-   {XrdNetAddr thePeer(&(aInfo->addr));
-    ((TPCLogRecord *)clientp)->isIPv6 =  (thePeer.isIPType(XrdNetAddrInfo::IPv6)
-                                      && !thePeer.isMapped());
-   }
+  //Return a socket file descriptor (note the clo_exec flag will be set).
+  int fd = XrdSysFD_Socket(aInfo->family, aInfo->socktype, aInfo->protocol);
```
amadio, line 123: It seems to me that there's no point in continuing if you get `fd < 0` here, so maybe add an if statement like
```cpp
if (fd < 0)
    return CURL_SOCKET_BAD;
```
and below just `return fd;`

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -119,18 +119,33 @@ int TPCHandler::opensocket_callback(void *clientp,
...
+  return (fd >= 0 ? fd : CURL_SOCKET_BAD);
+}
+
+int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
+  TPCLogRecord * rec = (TPCLogRecord *)clientp;
+  if(rec->pmarkManager) {
+    // Destroy the PMark handle associated to the file descriptor before closing it.
+    // Otherwise, we would lose the socket usage information if the socket is closed before
+    // the PMark handle is closed.
+    rec->pmarkManager->endPmark(fd);
```
amadio, line 146: These indirections in combination with the many `if (rec->pmarkManager)` also tell me that maybe these are better moved into `TPCLogRecord`.

src/XrdTpc/XrdTpcMultistream.cc
```diff
@@ -281,6 +281,11 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
         curl_handles.emplace_back(handles.back()->GetHandle());
     }
 
+    // Notify the packet marking manager that the transfer will start after this point
+    if(rec.pmarkManager) {
+      rec.pmarkManager->setTransferWillStart(&req);
```
amadio, line 286: This feels a bit awkward here, maybe a `startTransfer` in `TPCLogRecord` makes more sense, then it'd look like just `rec.startTransfer(&req)` here, and you hide some of the internals of `TPCLogRecord` for external callers. Same for other places where such indirections appear, and `if(rec.pmarkManager)` is needed.

src/XrdTpc/XrdTpcTPC.hh
```diff
@@ -76,6 +79,8 @@ private:
         int tpc_status;
         unsigned int streams;
         bool isIPv6;
+        // The packet manager lifetime is the one of the TPCLogRecord
+        std::unique_ptr<PMarkManager> pmarkManager;
```
amadio, line 83: I think using
```suggestion
        MarkManager pmarkManager;
```
here would allow you to remove a lot of the `if(rec.pmarkManager)` statements you have. You'd just then internally have to deal with null pointers in the `PMarkManager` class itself.

src/XrdTpc/PMarkManager.hh
```diff
@@ -0,0 +1,113 @@
...
+  void endPmark(int fd);
+
+  virtual ~PMarkManager() = default;
+private:
+  using SocketInfoPtr = std::unique_ptr<SocketInfo>;
+  using FdPMarkHandlePtrPair = std::pair<int,std::unique_ptr<XrdNetPMark::Handle>>;
+  // The list of socket information from which we will create the packet marking handles
+  std::list<SocketInfoPtr> mSocketInfos;
+  // The vector of the packet marking handles
+  std::vector<FdPMarkHandlePtrPair> mPmarkHandles;
```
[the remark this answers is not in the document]
ccaffy: I prefer using a vector, unordered_map does not guarantee the order of the elements after insertion :)

src/XrdTpc/XrdTpcTPC.cc
ccaffy: Yes correct, done :)

src/XrdTpc/XrdTpcMultistream.cc
ccaffy: Yes, renaming setTransferWillStart :)

src/XrdTpc/PMarkManager.hh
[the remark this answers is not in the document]
amadio: You could use a normal `std::map` in that case, I think a map or set is more ergonomic than a vector of pairs, but this is just a suggestion...

src/XrdTpc/PMarkManager.hh
```diff
@@ -0,0 +1,113 @@
...
+   * @param fd the fd of the socket to be closed
+   */
+  void endPmark(int fd);
+
+  virtual ~PMarkManager() = default;
+private:
+  using SocketInfoPtr = std::unique_ptr<SocketInfo>;
+  using FdPMarkHandlePtrPair = std::pair<int,std::unique_ptr<XrdNetPMark::Handle>>;
+  // The list of socket information from which we will create the packet marking handles
+  std::list<SocketInfoPtr> mSocketInfos;
```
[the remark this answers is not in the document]
ccaffy: Changed to a queue, which is more meaningful about what it does :)

src/XrdTpc/PMarkManager.cc
```diff
@@ -0,0 +1,77 @@
...
+
+void PMarkManager::endPmark(int fd) {
+  // We need to delete the PMark handle associated to the fd passed in parameter
+  // we just look for it and reset the unique_ptr to nullptr to trigger the PMark handle deletion
+  auto it = std::find_if(mPmarkHandles.begin(), mPmarkHandles.end(),[fd](const auto & pair){
+    return pair.first == fd;
+  });
+  if(it != mPmarkHandles.end()) {
+    it->second.reset(nullptr);
+  }
```
amadio, line 76: With a `std::map`, this would be a lot simpler... :-)

```cpp
mPmarkHandled.erase(fd);
```

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -119,18 +119,36 @@ int TPCHandler::opensocket_callback(void *clientp,
...
+  if(fd < 0) {
+    return CURL_SOCKET_BAD;
+  }
+  TPCLogRecord * rec = (TPCLogRecord *)clientp;
+  if (purpose == CURLSOCKTYPE_IPCXN && clientp)
+  {XrdNetAddr thePeer(&(aInfo->addr));
+    rec->isIPv6 =  (thePeer.isIPType(XrdNetAddrInfo::IPv6)
+                    && !thePeer.isMapped());
+    // Register the socket to the packet marking manager
+    if(fd >= 0) {
```
amadio, line 135: If you arrive here, it's because `fd >= 0` already, so this `if` is not necessary.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -119,18 +119,36 @@ int TPCHandler::opensocket_callback(void *clientp,
...
+
+int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
+  TPCLogRecord * rec = (TPCLogRecord *)clientp;
+
+  // Destroy the PMark handle associated to the file descriptor before closing it.
+  // Otherwise, we would lose the socket usage information if the socket is closed before
+  // the PMark handle is closed.
+  rec->pmarkManager.endPmark(fd);
+
+  return close(fd);
```
amadio, line 151: You could close the socket in `endPmark(fd)` and return the result. Then here you just
```cpp
  return rec->pmarkManager.endPmark(fd);
```

src/XrdTpc/PMarkManager.cc
ccaffy: Is it guaranteed that file descriptors created by connections will have an ascending order?

src/XrdTpc/XrdTpcTPC.cc
ccaffy: Oh yes, I went too fast........

src/XrdTpc/PMarkManager.cc
ccaffy: I can find a way around it though. I can store the first file descriptor integer as a member class, and use the map ;) Which will simplify things...

src/XrdTpc/XrdTpcTPC.cc
amadio: Yeah, you are right, forget about this one!

src/XrdTpc/PMarkManager.cc
ccaffy: No but I don't want sorted socket, I want the **FIRST** file descriptors that got opened by curl when the first transfer connection occurs :)

src/XrdTpc/PMarkManager.hh
```diff
@@ -0,0 +1,113 @@
...
+   * (e.g sent bytes or received bytes) will have values equal to 0.
+   * @param fd the fd of the socket to be closed
+   */
+  void endPmark(int fd);
+
+  virtual ~PMarkManager() = default;
+private:
+  // The list of socket information from which we will create the packet marking handles
+  std::queue<SocketInfo> mSocketInfos;
+  // The vector of the packet marking handles
```
amadio, line 100: You may want to update also this comment.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -849,6 +869,8 @@ int TPCHandler::ProcessPushReq(const std::string & resource, XrdHttpExtReq &req)
 //  curl_easy_setopt(curl, CURLOPT_SOCKOPTFUNCTION, sockopt_setcloexec_callback);
     curl_easy_setopt(curl, CURLOPT_OPENSOCKETFUNCTION, opensocket_callback);
     curl_easy_setopt(curl, CURLOPT_OPENSOCKETDATA, &rec);
+    curl_easy_setopt(curl, CURLOPT_CLOSESOCKETFUNCTION, closesocket_callback);
+    curl_easy_setopt(curl,CURLOPT_CLOSESOCKETDATA,&rec);
```
amadio, line 873: Just noticed the missing space after the comma, different than the ones above it. Not a critical thing, though.

src/XrdTpc/PMarkManager.hh
ccaffy: Yes haha, done :)

src/XrdTpc/XrdTpcTPC.cc
ccaffy: Yep, now it's cleaner :+1:

src/XrdTpc/PMarkManager.cc
ccaffy: As discussed, I want to keep track of the first FD in order to get its associated PMarkHandle to create the other PMarkHandle from it :)

src/XrdTpc/PMarkManager.cc
```diff
@@ -0,0 +1,72 @@
+//------------------------------------------------------------------------------
+// This file is part of XrdTpcTPC
+//
+// Copyright (c) 2013 by European Organization for Nuclear Research (CERN)
```
amadio, line 4: Another quite minor thing, but this is off by 10 years :-)

ccaffy: Actually, what's the policy for the headers of those files? It's quite inconsistent throughout the entire project :D

amadio: Yes, it was likely just copied without change, sometimes with the right year, sometimes with the same as before...

abh3:
We have no established policy. However, I have begun the habit of updating 
te copyright date to the current dates in the files that I change.

src/XrdTpc/PMarkManager.cc
ccaffy: Alright then, done for the PMarkManager.cc :)

PR #2110, opened 2023-10-20
[XrdCrypto] Fix ODR violation reported by address sanitizer

amadio:
Several source files are added both into XrdCrypto and XrdUtils, including XrdCryptoAux.cc. This causes some global symbols to violate ODR because libXrdCrypto.so is linked to libXrdUtils.so.

### DIFF
--- a/src/XrdCrypto/XrdCryptoAux.cc
+++ b/src/XrdCrypto/XrdCryptoAux.cc
@@ -36,11 +36,13 @@
 // For error logging and tracing
 static XrdSysLogger Logger;
 static XrdSysError eDest(0,"crypto_");
-XrdOucTrace *cryptoTrace = 0;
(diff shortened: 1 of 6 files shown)


amadio:
@abh3 Please let me know if you are OK with this going into 5.6.3. Otherwise I will leave it for 5.7. Thanks!

abh3:
Let's leave it to 5.7.0, I won't have time to review this in the next 12 waking hours. Anyway, as per our discussion I though we were looking for a more long lasting fix. Frankly, after I though about this, the correct fix is to take XrdCryptoAux.cc and all its other crypto dependencies and moving them to a separate shared library. Then you can have linUtils and libsecgsi link against the same code without creating ODR problems. So, it really is a packaging issue.

amadio:
As usual, you are right 👍 The extra shared library makes the most sense.

PR #2112, opened 2023-10-24
Conditionally compile support for loading CA files.

bbockelm:
The `scitoken_config_set_str` function was only recently added to the SciTokens library; some supported platforms do not plan on taking the updated version.

Accordingly, add a CMake-level test to see if the symbol exists and only then make the call to the library.

@amadio - it's been a _long_ time since I've done anything non-trivial with cmake.  Please double-check you're happy with my approach.


amadio:
Looks good. I thought scitokens would export a version which we could use to decide, but this works too. I will check tomorrow with my new CI and merge if all goes well. Thank you!

amadio:
Thanks! This works well.

PR #2117, opened 2023-10-31
[XrdCrypto,XrdSecgsi] Update min/default RSA bits to 2048

amadio:


src/XrdSecgsi/XrdSecProtocolgsi.cc
```diff
@@ -149,7 +149,7 @@ String XrdSecProtocolgsi::UsrCert  = "/.globus/usercert.pem";
 String XrdSecProtocolgsi::UsrKey   = "/.globus/userkey.pem";
 String XrdSecProtocolgsi::PxyValid = "12:00";
 int    XrdSecProtocolgsi::DepLength= 0;
-int    XrdSecProtocolgsi::DefBits  = 512;
+int    XrdSecProtocolgsi::DefBits  = 2048;
```
abh3, line 152: This really should come from the default value define in CryptoAux which is already included in this cc file.

src/XrdSecgsi/XrdSecProtocolgsi.hh
```diff
@@ -188,7 +188,7 @@ public:
    char  *proxy;  // [c] user proxy  [/tmp/x509up_u<uid>]
    char  *valid;  // [c] proxy validity  [12:00]
    int    deplen; // [c] depth of signature path for proxies [0] 
-   int    bits;   // [c] bits in PKI for proxies [512] 
+   int    bits;   // [c] bits in PKI for proxies [2048]
```
abh3, line 191: Same here, refer to the CryptoAux definition so that when it becomes 4096 we know where to look.

abh3 (changes requested):
Just two changes that I saw and one is a comment.  Nice job though!

amadio:
@abh3 I hope the current version is in line with what you suggested.

abh3 (approved):

abh3:
Thanks! Now we can simply change the define symbol in the future. I guess we need to wait for @esindril to complete the review.

esindril (approved):
Looks goog!

amadio:
Thanks, @esindril, I'm merging it then. Cheers,

PR #2119, opened 2023-11-03
Fix header dependencies and missing includes/declarations

amadio:
@apeters1971 Reported to me that some headers that he needed to develop Zip support on EOS were not installed by XRootD, so this pull request adds a script that checks that each installed public and private XRootD header can be included individually without errors. The subsequent commits are fixes for the errors reported by this script. The intention is to identify which headers may have missing includes, missing forward declarations, or missing header dependencies, that is, headers from XRootD which it includes, but were not installed by the install target.

### DIFF
--- a/src/XrdHeaders.cmake
+++ b/src/XrdHeaders.cmake
@@ -185,10 +185,15 @@ if( NOT XRDCL_ONLY )
     XrdSut/XrdSutAux.hh
     XrdSut/XrdSutBucket.hh
 
-    XrdVoms/XrdVoms.hh
-
     XrdOuc/XrdOucPgrwUtils.hh
 )
+
+  if ( BUILD_VOMS )
+    set( XROOTD_PRIVATE_HEADERS
+      ${XROOTD_PRIVATE_HEADERS}
+      XrdVoms/XrdVoms.hh
+    )
+  endif()
 endif()
 
 install_headers(
--- a/src/XrdHeaders.cmake
+++ b/src/XrdHeaders.cmake
@@ -128,12 +128,16 @@ set( XROOTD_PRIVATE_HEADERS
   XrdNet/XrdNetIF.hh
   XrdSecsss/XrdSecsssID.hh
   XrdSys/XrdSysPriv.hh
+  XrdOuc/XrdOucCRC32C.hh
   XrdOuc/XrdOucExport.hh
   XrdOuc/XrdOucGatherConf.hh
   XrdOuc/XrdOucPList.hh
   XrdOuc/XrdOucN2NLoader.hh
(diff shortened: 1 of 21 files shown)


amadio:
Looks like `clang++` which is in `PATH` on the macOS images is not properly setup, since it complains about things being C++11 extensions in the error messages. I will leave macOS out of the header checks for now while I figure out how to setup clang with the right C++ standard and include directories on the macOS image.

src/XrdHeaders.cmake
```diff
@@ -39,6 +39,7 @@ set( XROOTD_PUBLIC_HEADERS
   XrdOuc/XrdOucIOVec.hh
   XrdOuc/XrdOucLock.hh
   XrdOuc/XrdOucName2Name.hh
+  XrdOuc/XrdOucPinLoader.hh
```
abh3, line 42: Why does this need to be public?

src/XrdOuc/XrdOucPList.hh
```diff
@@ -31,7 +31,7 @@
 /******************************************************************************/
 
 #include <cstdio>
-#include <strings.h>
+#include <cstring>
```
abh3, line 34: Out of curiosity, why the selective replacement of "string.h" with ctring? There are still a lot of file that have "string.h" in them.

abh3 (commented):
Honestly, I don't quite understand  this pull request other than it makes clang happy.  Nonetheless, I made a couple of commments.

src/XrdOuc/XrdOucPList.hh
amadio: Note that it's `strings.h`, not `string.h`. The first header is deprecated and does not contain the definitions we want. In any case, if you run the scritpt `check-headers.sh` after installing all our `*-devel` RPMs, you will see the errors these commits are fixing.

src/XrdHeaders.cmake
amadio: `XrdOucPinLoader.hh` is needed by `XrdClPlugInManager.hh`. We could, instead of adding this header, stop installing or installing privately both of them. I will check later if the header dependencies can be solved by installing less rather than more for this one.

amadio: I made `XrdClPlugInManager.hh` private (no one could have used it before, because including it would cause a compilation error), and I install both `XrdClPlugInManager.hh` and `XrdOucPinLoader.hh` into `include/xrootd/private`. Later we can review which headers are currently exposed that don't make sense and simply remove them from the install target in a later version.

abh3 (approved):
It looks OK to me.

PR #2122, opened 2023-11-13
Allow configurable SciTokens authorization strategy

### LINKED ISSUE #2121: XrdSciTokens config to refuse tokens without scopes
Having enabled tokens for both xroot and http for the EOS instance in Vienna, a couple of tests in the CMS SAM/ETF testsuite were failing due to the assumption that the scitokens library would refuse tokens without any scopes. This apparently is not the case. Therefore, could this be added as an option to the scitokens configuration. According to @bbockelm, there might already be some option in place to achieve this. If so, please share it with us so that we can pass this info to service manages to properly configure their instances. 

Please keep in mind that such instances also use the scitokens_mapfile.json option to configure certain mappings for paths - a thing which might influence the scitokens ability to refuse tokens without any scopes.

Thanks!

bbockelm:
This is first draft of a proposal to fix #2121.

On a per-issuer basis, decide if the mapping information is only to be done *after* authorization passes or if the mapping information should be kept and passed to the subsequent plugin.


amadio:
Hi @bbockelm I tried this branch, but the mapping is applied even when I set the authorization strategy to just "capability", like this:

```
[Global]
audience = https://demo.scitokens.org

[Issuer DEMO]
issuer = https://demo.scitokens.org
base_path = /
map_subject = False
name_mapfile = /etc/xrootd/scitokens_mapfile.json
default_user = xrootd
authorization_strategy = capability
```

The file `scitokens_mapfile.json` contains

```
[
  {"path": "/data/", "result": "mymappeduser", "comment": "Owner of the data"}
]
```
and I see this in the server log
```
231207 16:03:44 161155 scitokens_Access: Trying token-based access control
231207 16:03:44 161155 scitokens_Access: Token not found in recent cache; parsing.
231207 16:03:45 161155 scitokens_Access: New valid token mapped_username=xrootd, subject=amadio, issuer=https://demo.scitokens.org
231207 16:03:45 161155 scitokens_Access: Request username mymappeduser
```
with a token that contains this JSON (scope is `openid` on purpose so it doesn't have any storage capability):
```
{
  "ver": "scitoken:2.0",
  "sub": "amadio",
  "aud": "https://demo.scitokens.org",
  "iss": "https://demo.scitokens.org",
  "scope": "openid",
  "exp": 2701960770,
  "iat": 1701960867,
  "nbf": 1701960867,
  "jti": "aaf1797e-99c0-4d38-a809-01b0812bbe7c"
}
```
I would expect to see
```
231207 16:03:45 161155 scitokens_Access: Request username amadio
```
since that's the username that comes with the token (i.e. without the mapping to `mymappeduser`).

My xrootd configuration file is also shown below, so you can try this too.
```
all.trace    all
http.trace   all
xrd.trace    all
xrootd.trace all
scitokens.trace all

all.export /
all.sitename  XRootD
all.pidpath   /var/run/xrootd
oss.localroot /srv/xrootd
xrootd.chksum chkcgi adler32 crc32c

xrootd.seclib libXrdSec.so

sec.protocol ztn
sec.protbind * only ztn

xrd.tlsca certdir /etc/grid-security/certificates
xrd.tls /etc/xrootd/hostcert.pem /etc/xrootd/hostkey.pem

xrd.protocol XrdHttp:8080 libXrdHttp.so
http.header2cgi Authorization authz
ofs.authlib ++ libXrdAccSciTokens.so config=/etc/xrootd/scitokens.cfg
ofs.authlib ++ libXrdMacaroons.so
ofs.authorize 1

acc.authdb /etc/xrootd/authdb
```
For completeness, the authdb file contains
```
u amadio /data rw
```
to allow my user to read/write into the directory. Without this, I get permission denied with the token.

amadio:
@bbockelm I guess this one should probably be closed? As far as I understand, tokens without scopes are already rejected, and the problem with existing scopes/paths is addressed by #2152. Please correct me if I'm wrong.

amadio:
Ok, I think we can close this. Please open a new pull request if you'd like to move forward with the approach of having multiple authorization strategies later.

PR #2127, opened 2023-11-20
Switch from using a cert file to a cert chain file

### LINKED ISSUE #2126: XRootD TLS missing certificate chain, causing verification failures
When configuring TLS via `xrd.tls`, we should export the full chain in the specified cert file during the TLS handshake.  Instead, only the first entry in the file is exported.

For example, suppose you set the following line:

```
xrd.tls /etc/xrootd/certificates/tls.crt /etc/xrootd/certificates/tls.key
```

Then, I'd expect the full contents of `/etc/xrootd/certificates/tls.crt` to be sent during the TLS handshake, not just the first entry.

Without the full chain, connections with alternate verification paths -- or with requiring intermediate certificates -- fail.

To test:

1.  Configure a XRootD on the public internet.
2. Use `certbot` to get a Let's Encrypt certificate for the host.
3. Set the host cert file to be the `fullchain.pem` from Let's Encrypt.
4. Run `curl` against the XRootD instance; note that it fails.
5. Run `openssl s_client -connect <hostname>:<port>` and look at the reported `Certificate chain`.

For the last item, when XRootD is used, you see a chain like this:

```
Certificate chain
 0 s:/CN=hcc-briantest7.unl.edu
   i:/C=US/O=Let's Encrypt/CN=R3
```

That can't be verified as `/C=US/O=Let's Encrypt/CN=R3` is not a trust root.

When you use the same set of files with Apache, you'll see:

```
Certificate chain
 0 s:/CN=hcc-briantest7.unl.edu
   i:/C=US/O=Let's Encrypt/CN=R3
 1 s:/C=US/O=Let's Encrypt/CN=R3
   i:/C=US/O=Internet Security Research Group/CN=ISRG Root X1
 2 s:/C=US/O=Internet Security Research Group/CN=ISRG Root X1
   i:/O=Digital Signature Trust Co./CN=DST Root CA X3
```

and verification works.

In general, this prevents use outside the "grid" context: for browser authentication (and most grid CAs - except the CERN CA - are cross-signed so they are valid in the browser as well), intermediate CAs are used and only roots are distributed to clients.

bbockelm:
Outside most grid contexts, we need to present the certificate chain in the TLS handshake for the verification path through the intermediate CA to the root.

This commit switches over to the corresponding OpenSSL function call.

Fixes the ability to use most CAB CAs with XRootD.

If, for some reason, the prior behavior is desired, the host certificate file can be limited to contain only the host certificate.  I don't expected this to cause any backward compatibility issues as certificates issued by CAs needing this functionality were already broken.

Fixes #2126

### DIFF
--- a/src/XrdTls/XrdTlsContext.cc
+++ b/src/XrdTls/XrdTlsContext.cc
@@ -719,7 +719,7 @@ XrdTlsContext::XrdTlsContext(const char *cert,  const char *key,
 
 // Load certificate
 //
-   if (SSL_CTX_use_certificate_file(pImpl->ctx, cert, SSL_FILETYPE_PEM) != 1)
+   if (SSL_CTX_use_certificate_chain_file(pImpl->ctx, cert) != 1)
       FATAL_SSL("Unable to create TLS context; invalid certificate.");
 
 // Load the private key


amadio:
Looks ok for me, but let's hear also from Andy.

abh3 (approved):
Looks acceptable.

PR #2134, opened 2023-11-25
Use a monotonically-increasing counter for the link ID

### LINKED ISSUE #2133: Non-unique user IDs in monitoring
In the [documentation](https://xrootd.slac.stanford.edu/doc/dev56/xrd_monitoring.htm), the user ID is listed as having this interpretation:

```
userid:     prot/user.pid:sid@host
```

Where the fields are as follows:
- `prot`: "Communication protocol being used by the client"
- `user`: "Unix username of the user as reported by the client (i.e. unverified) or the plug-in identifier. You can distinguish between the two as plug-ins always report a pid of zero. The user name is the plug-in component name (e.g. pfc)."
- `pid`: "User’s process number that issued the request."
- `sid`:  "Server’s fingerprint for the connection to user:pid at host."

Now, the `sid` documentation here is incorrect (or, at least, confusing).  It is actually the server instance fingerprint and static until the server restarts.

For the `https` protocol, `pid` is set to a monotonically increasing integer, meaning the `userid` string is always unique for a connection (in fact, the tuple of (pid, sid) is unique).

Now, for the `xroot` protocol, `pid` is set to the client-identified PID, meaning there's no guaranteed of uniqueness of `userid`.

Hence, for the `xroot` protocol, you can't associate a file open (`d` record type) record with a user login (`U`) or a token access (`T`).  This means the token-based monitoring done during #1987 only will work with the f-stream (as those records include the user dictID as part of the f-stream's open record).

I see two approaches here:

1.  The `userid` is not meant to be unique.  That's fine, though I'd request re-wording `sid` to say it's simply the server's unique fingerprint, not the "fingerprint for the connection".
2. The `userid` should be made unique.  For the xrootd protocol, compute the `pid` the same way with `getSID` as is done for the HTTP protocol.

bbockelm:
Instead of using the client-provided PID (which may not be unique), use the session ID counter.  This allows the tuple of the (PID, server ID) to be unique in the produced monitoring packets (until the 32-bit unsigned integer rolls over).  This brings the xrootd protocol in alignment with what's done with the HTTP protocol.

Adds a comment noting which `SID` is being used as there are three different definitions for the `S` (server, session, stream) in the codebase.

Suggested changes to the monitoring documentation (all in section 3.3.1, "Message Info Field"):

- `pid`: Session ID associated with the connection.  Will be unique until 32-bit rollover or server restart.
- `sid`: Unique fingerprint for the server instance.  Will change when the server is restarted.

Even though `pid` is no longer the "process ID", I'd like to keep the acronym as `sid` is already used.  I believe the explanation of `user` can be kept unchanged.

Fixes #2133


abh3 (approved):
Looks fine.

amadio:
This change actually broke TPC transfers where the server calls the client (i.e. `xrcp`) to perform the actual transfer:
```
xrootd/tests/XrdCl/XrdClFileCopyTest.cc:598: Failure
Value of: _st.IsOK()
  Actual: false
Expected: true
[process11.Run(0)]: [ERROR] Server responded with an error: [3005] [ERROR] Server responded with an error: [3010] tpc authorization expired\0
```
This can be reproduced by rebasing the devel branch on top of master and running the tests by running
```
xrootd $ ctest -VV -S test.cmake
```
from the top of the repository, as shown above. This will need to be either reverted or fixed before I can make a new release (planned for this week, but may need to wait until next week).

abh3:
The alternative fix here is simply to change the "tis" in the monitoring 
code when the "tid" gets registered. That way the change is restricted to 
the reporting side of things. However, do you have any idea of why this 
would break TPC?

Andy

amadio:
I'm reverting the change introduced here (we do not do rebases on `master`), so I'm reopening #2133 after that. There is clearly a problem with how the client and the server coordinate for a TPC transfer after the change was introduced, even if the reason is not so clear yet, at least not to me. We will need to be extra careful when reintroducing this change later on. I am attaching here two full logs of the output of the third-party copy test on the whole testing cluster before and after the change. The interesting part is shown as a diff below (mixing all logs together, red is before the change, and green is after the change, and also the one where the TPC transfer fails):

```diff
 Xrd_ProtLoad: matched port 10943 protocol xroot
 anon.0:27@localhost Xrd_Poll: FD 27 attached to poller 0; num=2
-XrootdXeq: amadio.761885:27@localhost pvt IP64 login
+XrootdXeq: amadio.4:27@localhost pvt IP64 login
 Xrd_Inet: Accepted connection on port 10943 from 28@localhost
 Xrd_ProtLoad: matched port 10943 protocol xroot
-anon.0:28@localhost Xrd_Poll: FD 28 attached to poller 1; num=2
 Xrd_Sched: Now have 6 workers
 Xrd_Sched: running main accept inq=0
-XrootdXeq: FD 28#1 bound amadio.761885:27@localhost
-amadio.761885:27@localhost Xrd_Link: Setting FD 27 ref to 1+1 post=0
-Xrd_Sched: running amadio.761885:28@localhost inq=0
-amadio.761885:27@localhost Xrd_Link: Setting FD 27 ref to 2+-1 post=0
-XrootdXeq: amadio.761885:27@localhost disc 0:00:00
-XrootdXeq: amadio.761885:28@localhost unbind 0:00:00
-amadio.761885:28@localhost Xrd_Poll: Poller 1 removing FD 28
-amadio.761885:28@localhost Xrd_Poll: FD 28 detached from poller 1; num=1
-amadio.761885:27@localhost Xrd_Poll: Poller 0 removing FD 27
-amadio.761885:27@localhost Xrd_Poll: FD 27 detached from poller 0; num=1
+anon.0:28@localhost Xrd_Poll: FD 28 attached to poller 1; num=2
+XrootdXeq: FD 28#1 bound amadio.4:27@localhost
+Xrd_Poll: Poller 0 enabled amadio.4:27@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 1 enabled amadio.2:25@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 1 enabled amadio.2:25@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 2 enabled amadio.3:26@localhost
+Xrd_Poll: Poller 2 enabled amadio.3:26@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 0 enabled amadio.1:8@localhost
+Xrd_Poll: Poller 2 enabled amadio.3:26@localhost
 Xrd_Sched: running async response inq=0
-amadio.760743:25@localhost Xrd_Link: Setting FD 25 ref to 1+1 post=0
-amadio.760743:25@localhost Xrd_Link: Setting FD 25 ref to 2+-1 post=0
-Xrd_Sched: running amadio.760743:25@localhost inq=0
-Xrd_Sched: Now have 5 workers
-Xrd_Sched: running chksum inq=0
-amadio.760743:25@localhost Xrd_Link: Setting FD 25 ref to 1+1 post=0
-amadio.760743:25@localhost Xrd_Link: Setting FD 25 ref to 2+-1 post=0
-redirector.760506:8@localhost Xrd_Link: Setting FD 8 ref to 1+1 post=0
-Xrd_Sched: running redirector inq=0
-redirector.760506:8@localhost Xrd_Link: Setting FD 8 ref to 2+-1 post=0
-redirector.760488:8@localhost Xrd_Link: Setting FD 8 ref to 1+1 post=0
+amadio.4:27@localhost Xrd_Link: Setting FD 27 ref to 1+1 post=0
+amadio.4:27@localhost Xrd_Link: Setting FD 27 ref to 2+-1 post=0
+ofs_TPC: localhost tpc grant by amadio.2@localhost expired for /data/cb4aacf1-6f28-42f2-b68a-90a73460f424.dat
+ofs_TPC: localhost tpc grant by amadio.3@localhost expired for /data/cb4aacf1-6f28-42f2-b68a-90a73460f424.dat
+amadio.4:27@localhost Xrd_Poll: Link 27 terminating: hangup
+Xrd_Sched: running amadio.4:27@localhost inq=0
+XrootdXeq: amadio.4:27@localhost disc 0:01:00
+XrootdXeq: amadio.4:28@localhost unbind 0:01:00
+amadio.4:28@localhost Xrd_Poll: Poller 1 removing FD 28
+amadio.4:28@localhost Xrd_Poll: FD 28 detached from poller 1; num=1
+amadio.4:27@localhost Xrd_Poll: Poller 0 removing FD 27
+amadio.4:27@localhost Xrd_Poll: FD 27 detached from poller 0; num=1
+ofs_Run: /home/amadio/src/xrootd/build/src/XrdCl/xrdcp ended with status 54
+ofs_TPC: amadio.2:25@localhost /data/tpcFile [ERROR] Server responded with an error: [3010] tpc authorization expired (source)
+Xrd_Sched: running async response inq=0
+amadio.2:25@localhost Xrd_Link: Setting FD 25 ref to 1+1 post=0
+amadio.2:25@localhost Xrd_Link: Setting FD 25 ref to 2+-1 post=0
+Xrd_Sched: running amadio.3:26@localhost inq=0
+Xrd_Sched: running amadio.2:25@localhost inq=0
+Xrd_Sched: running amadio.1:8@localhost inq=0
+redirector.742454:8@localhost Xrd_Link: Setting FD 8 ref to 1+1 post=0
+Xrd_Sched: running redirector inq=0
+redirector.742454:8@localhost Xrd_Link: Setting FD 8 ref to 2+-1 post=0
+Xrd_Sched: running amadio.1:8@localhost inq=0
+Xrd_Sched: running amadio.3:26@localhost inq=0
+Xrd_Sched: running amadio.3:26@localhost inq=0
+redirector.742436:8@localhost Xrd_Link: Setting FD 8 ref to 1+1 post=0
 Xrd_Sched: running redirector inq=0
```

The full logs are here: [working.txt](https://github.com/xrootd/xrootd/files/13512588/working.txt) and [broken.txt](https://github.com/xrootd/xrootd/files/13512590/broken.txt) (i.e. with #2134 included). Inspecting the logs I founds some places where the id used in different places were the same before this patch, but different afterwards. This could be one of the reasons why the patch breaks the transfers.

abh3:
I just pushed commit d25051cd0 that essentially corrects the pull 
request mentioned below (2134).

The issue was that the trace ID was being changed not just the monitoring 
ID. The commit d25051cd0 only affects the monitoring reporting and thus 
should have no effect on TPC. However, please check it out and also 
completely back out the pull request below.

PR #2137, opened 2023-11-28
Get more useful diagnostics in case of test failures

ellert:
EXPECT_EQ( A, B ) provides more useful diagnostics than EXPECT_TRUE ( A == B ).

EXPECT_NE( A, B ) provides more useful diagnostics than EXPECT_TRUE ( A != B ).


amadio:
Very nice, thank you! The plan is to gradually convert all CppUnit based tests to GTest, then remove them. CppUnit has problems with lto among other things. But these changes are good in any case (talking about the XrdCeph tests here).

Issue #1501, opened 2021-09-01
non-ASCII char in xroot error message


belforte:
looks like recent xroot (4.12.3-llifpc2) insert (at least at times) a non ASCII character at the end of an error message, see screenshot below.
This breaks current code which reports and parse those messages.
Can it be avoided ?
While I only have an example from using CMS code , @makortel  says that in this case it just prints xrootd message "as is"
![Screenshot from 2021-08-31 17-48-12](https://user-images.githubusercontent.com/1837785/131671605-0d4d1ba3-6ff0-463c-87f6-2ce8ce6d19fd.png)
Of course we can filter for non-ASCII on our side, but  @klannon suggested to report this to you as well for a proper resolution. In this I am the current maintainer of a tool which relies on CMS framework reports, so am 2 software layers (CMSSW and CMS/WMCore) away from xroot, simply acting as messenger here, but I doubt I can be of further help. 
Thanks

simonmichal:
@belforte : thank you for reporting this issue and sorry for replaying with such a delay!

Could you you test with the latest release, I believe this should have been fixed in: https://github.com/xrootd/xrootd/commit/22d308a3dd689607ee0396860d377145836ef55b

belforte:
thanks @simonmichal . Unfortunately I am stuck with what version of xroot is built with  CMSSW. But this should be easy to check on your side, just try to open a non-existing file. In any case I believe you can close.

amadio:
Commit 22d308a likely fixes this, but it looks like commit a20fe39385 could have introduced a potential off by one error. The function where the message comes from is below.

https://github.com/xrootd/xrootd/blob/v5.5.1/src/XrdCms/XrdCmsCluster.cc#L2031-L2045

If `n` could be off by one here, it could cause an extra `'\0'` character to be shown in the output due to `Sel.Resp.Data` containing two `NUL` characters at the end instead of just one.

PR #2138, opened 2023-11-28
Avoid null bytes in error message strings

ellert:
If I change a test that uses the GTEST_ASSERT_XRDST_NOTOK macro to use the wrong expected error code so that a failure is triggered, the printed error message ends with a null byte:
```
46: Value of: !_st.IsOK() && _st.code == XrdCl::errNotFound
46:   Actual: false
46: Expected: true
46: [fs.Stat( targetPath, info )]: [ERROR] Server responded with an error: [3011] No servers are available to read the file.\0
```
Note the extra \0 at the end of the string.

The reason for this is that the code that converts an error response from the server to a std::string includes the terminting null byte in the length of the string. This commit reduces the length by 1 so that the terminating null byte is not included in the string. With this change the error message is printed correctly without a null byte at the end:
```
46: Value of: !_st.IsOK() && _st.code == XrdCl::errNotFound
46:   Actual: false
46: Expected: true
46: [fs.Stat( targetPath, info )]: [ERROR] Server responded with an error: [3011] No servers are available to read the file.
```

### DIFF
--- a/src/XrdCl/XrdClXRootDMsgHandler.cc
+++ b/src/XrdCl/XrdClXRootDMsgHandler.cc
@@ -1202,7 +1202,7 @@ namespace XrdCl
       if( pStatus.code == errErrorResponse )
       {
         st->errNo = rsp->body.error.errnum;
-        std::string errmsg( rsp->body.error.errmsg, rsp->hdr.dlen-4 );
+        std::string errmsg( rsp->body.error.errmsg, rsp->hdr.dlen-5 );
         if( st->errNo == kXR_noReplicas && !pLastError.IsOK() )
           errmsg += " Last seen error: " + pLastError.ToString();
         st->SetErrorMessage( errmsg );


amadio:
Please take a look at my [comment](https://github.com/xrootd/xrootd/issues/1501#issuecomment-1333983315) in #1501. I suspect we may have another case of an error message with an embedded `\0` in it, so we should probably try to fix that at the source, otherwise we will cut off other messages.

amadio:
Actually, having a second look at this, the change is actually ok. In the other cases an explicit memcpy is being used on a normal `char` array, in this particular case it's a `std::string` being created, and `dlen` is such that the terminating null byte is included. I think I'd rather just revert https://github.com/xrootd/xrootd/commit/5a832596a59d1055f59620d43fbb740955ce6787 to keep the link between the changes, noting that even though https://github.com/xrootd/xrootd/commit/9987b4b41990df69ae611a523acb76c69eeccbbb did fix a problem, reverting https://github.com/xrootd/xrootd/commit/22d308a3dd689607ee0396860d377145836ef55b was actually a mistake. In the end, I'd actually want to replace all these `dlen-4` with `dlen - offsetof(ServerResponse, errmsg)` in the relevant places.

PR #2142, opened 2023-12-04
[CMake] Fix include path in XRootDConfig.cmake

hahnjo:
Headers are installed into the subdirectory `xrootd/`. This was broken with commit https://github.com/xrootd/xrootd/commit/c6e0e598ce37b258fff8922b08bf5c45515a8e57 ("[CMake] Find only XRootD matching XRootDConfig.cmake installation path") released with version 5.6.3.


amadio:
Actually, no change needed on the other side, as it uses the variable without `/xrootd`, we still need to add the suffix there. It's all good!

Issue #2131, opened 2023-11-22
All XrdEc tests fail on Linux SPARC systems due to alignment issues (bus error)


glaubitz:
Running the testsuite on Linux SPARC fails with the following tests failing:

```
The following tests FAILED:
         20 - XrdEc::AlignedWriteTest (Bus error)
         21 - XrdEc::SmallWriteTest (Bus error)
         22 - XrdEc::BigWriteTest (Bus error)
         23 - XrdEc::VectorReadTest (Bus error)
         24 - XrdEc::IllegalVectorReadTest (Bus error)
         25 - XrdEc::AlignedWrite1MissingTest (Bus error)
         26 - XrdEc::AlignedWrite2MissingTest (Bus error)
         27 - XrdEc::AlignedWriteTestIsalCrcNoMt (Bus error)
         28 - XrdEc::SmallWriteTestIsalCrcNoMt (Bus error)
         29 - XrdEc::BigWriteTestIsalCrcNoMt (Bus error)
         30 - XrdEc::AlignedWrite1MissingTestIsalCrcNoMt (Bus error)
         31 - XrdEc::AlignedWrite2MissingTestIsalCrcNoMt (Bus error)
```

Obtaining a backtrace with gdb is not possible as the stack gets corrupted.

SSH access to SPARC machines for testing can be obtained through the GCC Compile Farm: https://gcc.gnu.org/wiki/CompileFarm

amadio:
Hi, are you using XRootD on a SPARC machine? We unfortunately do not have enough resources to support these less used platforms, but we will gladly review and merge patches fixing these issues if they come in a pull request.

amadio:
Fixed by #2144.

amadio:
@glaubitz Thanks for the tip about the GCC compile farm. In the end, however, I could not install xrootd dependencies there to test, so what I did was to add a QEMU-based workflow that only runs on demand to test alternative architectures. You can see an example run here (and can run on your fork if you have one as well): https://github.com/xrootd/xrootd/actions/runs/7171402996/job/19526316751

PR #2144, opened 2023-12-04
Avoid bus errors

ellert:
Do not dereference unaligned pointers.

Fixes: #2131

### DIFF
--- a/src/XProtocol/XProtocol.cc
+++ b/src/XProtocol/XProtocol.cc
@@ -204,7 +204,7 @@ char* ClientFattrRequest::VVecInsert( const char *value, char *buffer )
 //
 char* ClientFattrRequest::NVecRead( char* buffer, kXR_unt16 &rc )
  {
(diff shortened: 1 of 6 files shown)


amadio:
Thanks, @ellert! At a glance this looks alright.

@glaubitz Could you please try this on SPARC and let us know if it fixes the problems you reported? I tried on Aarch64, which is also somewhat unforgiving on alignment errors, and it seems ok. However, everything was also ok before the changes as well there.

glaubitz:
> Thanks, @ellert! At a glance this looks alright.
> 
> @glaubitz Could you please try this on SPARC and let us know if it fixes the problems you reported? I tried on Aarch64, which is also somewhat unforgiving on alignment errors, and it seems ok. However, everything was also ok before the changes as well there.

Yes, this fixes the problems I am seeing on sparc64 as reported in #2131. Thanks!

amadio:
Ok, great, thank you very much for the changes!

PR #2145, opened 2023-12-04
[XrdZip] Support Big Endian

ellert:
The newly added tests for XrdZip revealed endian problems.
This PR addresses these issues.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -26,6 +26,7 @@ Build-Depends:
  libjson-c-dev,
  libmacaroons-dev,
  uuid-dev,
+ uuid-runtime,
  voms-dev,
  libscitokens-dev,
  davix-dev,
--- a/src/XrdOssCsi/XrdOssCsiTagstoreFile.hh
+++ b/src/XrdOssCsi/XrdOssCsiTagstoreFile.hh
@@ -34,22 +34,10 @@
 #include "XrdOss/XrdOss.hh"
 #include "XrdOssCsiTagstore.hh"
 #include "XrdOuc/XrdOucCRC.hh"
+#include "XrdSys/XrdSysPlatform.hh"
 
 #include <memory>
 #include <mutex>
-
-#if defined(__APPLE__)
-// Make sure that byte swap functions are not already defined.
-#if !defined(bswap_16)
-// Mac OS X / Darwin features
-#include <libkern/OSByteOrder.h>
-#define bswap_16(x) OSSwapInt16(x)
-#define bswap_32(x) OSSwapInt32(x)
-#define bswap_64(x) OSSwapInt64(x)
-#endif
-#else
-#include <byteswap.h>
-#endif
 
 class XrdOssCsiTagstoreFile : public XrdOssCsiTagstore
 {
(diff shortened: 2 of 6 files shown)


ellert:
Scratch builds of current master:

Without this patch: https://koji.fedoraproject.org/koji/taskinfo?taskID=109950752
With this patch: https://koji.fedoraproject.org/koji/taskinfo?taskID=109950773

Without the patch the s390x build fails:
```
The following tests FAILED:
	 10 - XrdCl::ZipTest.ExtractTest (Failed)
	 11 - XrdCl::ZipTest.OpenFileTest (Failed)
	 12 - XrdCl::ZipTest.ListFileTest (Failed)
	 13 - XrdCl::ZipTest.GetterTests (Failed)
	 34 - XrdCl::FileTest.ReadTest (Failed)
	 40 - XrdCl::FileTest.PlugInTest (Failed)
	 46 - XrdCl::FileCopyTest.NormalCopyTest (Failed)
```
[ERROR] Received corrupted data: End-of-central-directory signature not found

With the patch the tests succeed on s390x.

amadio:
The reason I didn't merge this one is that I'm not a big fan of the `#ifdef`s in headers. I was thinking about ways to not need them, but worst case we should probably switch to using plain `#if`s to avoid bit rot should one of the branches of the `#ifdef` become always false. Otherwise the changes look alright. I will try to propose some more concrete review suggestions tomorrow, as I unfortunately didn't have the time for that today.

ellert:
> Hi @ellert, @abh3 pointed out to me that in `XrdSysPlatform.hh` we already have things that should be usable here instead of adding the same or similar `#ifdef`s. Could you recreate the patch using that? Thank you!

That was actually how I had done it originally. I changed it when I realized that the XrdZip code was independent of the rest of the xrootd code and doesn't #include any xrootd headers outside XrdZip/* and including XrdSys/XrdSysPlatform.hh would break this. But that was possibly not that important. Anyway, I have now changed it to include XrdSys/XrdSysPlatform.hh.

amadio:
Thank you. I verified that this patch fixes the tests on s390x by running under emulation with docker+qemu. I may add a build to the CI in the future for this, as it can be done with a minimal change to one of our existing builds. However, I think that the `BSwap` functions belong in `XrdSysPlatform.hh` rather than `XrdClZipUtils.hh`, as they are more general and may be needed elsewhere in the future. I will change the patch to address that and force-push here to then merge into master, so it goes into the next release, 5.6.4. Unfortunately, even though a `std::byteswap` function exists, as usual for useful stuff, it's only avaiable with C++23 and later.

amadio:
@ellert Please ack my modified version of your commit, I will merge it to master along with my other changes for 5.6.4. I added a couple of minor fixes to fix the CI after Alpine release a new version which broke XRootD as well.

ellert:
> @ellert Please ack my modified version of your commit, I will merge it to master along with my other changes for 5.6.4. I added a couple of minor fixes to fix the CI after Alpine release a new version which broke XRootD as well.

I agree with your modifications.
I have added an additional commit to avoid duplication for your consideration.

amadio:
Great! Thanks for the extra change, I missed that one. I will merge this once the CI is OK. I am also adding new CI builds for other architectures on Fedora using QEMU, as that was pretty easy to do.

PR #2149, opened 2023-12-13
Add #include <byteswap.h> for BSD and GNU/Hurd

ellert:
Here are some additional includes for Hurd and BSD.
Is it supposed to be compilable on windows? I have not tried to do anything inside the "#ifdef WIN32".


amadio:
Looks good, thanks!

PR #2151, opened 2023-12-14
avoid duplicating basepath

Jo-stfc:
if the basepath is already present in the scope, do not append it again.
Currently, if a token comes in with a scope already including the basepath (e.g. /store/MC/samtests). If a basepath is set this will be appended to the scope and therefore fail authentication.


abh3:
Waiting for @bbockelm review of this pull request.

bbockelm:
@abh3 - I think this can be closed in favor of the approach in #2152.  While Guilherme has some good follow-up questions there, I think the other PR is the correct approach.

PR #2152, opened 2023-12-15
Attempt to use restricted paths even when scope is more generic

### LINKED ISSUE #2132: Scitoken bug with scope including basepath or /
Hi,
I found a weird interaction with CMS tokens. 
RAL uses lfn2pfn mapping to map cms's /store root folder to the pfn cms:/store.
Testing with a user token, I've confirmed the correct permissions are given (can only access files in the cms pool)
This follows the usual flow of request comes in, lfn2pfn maps /store to cms:/store and permission is granted.
This however caused CMS SAM token tests (external site functional tests) to fail, and from server logs the flow seems to be: request comes in, there is no lfn2pfn mapping and then the request gets denied.
Turning on scitokens.trace all fixed this behaviour and made the SAM tests pass again

bbockelm:
If a token has a broad scope and all the listed restricted paths are more specific, then try to generate ACLs that include the restricted paths.

Example: suppose we have a token with scope `storage.read:/` but restricted paths `/foo` and `/bar`.  Currently, we will reject the scope and give the token no permissions.  With this commit, we will generate acls of `read:/foo` and `read:/bar`.  So, as long as the access being attempted is within the restricted path, we can permit it.

Fixes #2132

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -350,7 +350,16 @@ public:
 
     bool apply(Access_Operation oper, std::string path) {
         for (const auto & rule : m_rules) {
-            if ((oper == rule.first) && !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size())) {
+            // The rule permits if both conditions are met:
+            // - The operation type matches the requested operation,
+            // - The requested path is a substring of the ACL's permitted path, AND
+            // - Either the requested path and ACL path is the same OR the requested path is a subdir of the ACL path.
+            //
+            // The third rule implies if the rule permits read:/foo, we should NOT authorize read:/foobar.
+            if ((oper == rule.first) &&
+                !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size()) &&
+                (rule.second.size() == path.length() || path[rule.second.size()]=='/'))
+            {
                 return true;
             }
         }
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -898,8 +898,12 @@ private:
                     // more possible authorizations.
                     if (!strncmp(acl_path, restricted_path.c_str(), acl_path_size)) {
                         // Only do prefix checking on full path components.  If acl_path=/foo and
-                        // restricted_path=/foobar, then we shouldn't authorize access to /foobar.
-                        if (restricted_path.size() > acl_path_size && restricted_path[acl_path_size] != '/') {
+                        // restricted_path=/foobar, then we shouldn't authorize access to /foobar. Note:
+                        // - The scitokens-cpp library guaranteees that acl_path is normalized and not
+                        //   of the form `/foo/`.
+                        // - Hence, the only time that the acl_path can end in a '/' is when it is
+                        //   set to `/`.
+                        if ((restricted_path.size() > acl_path_size && restricted_path[acl_path_size] != '/') && (acl_path_size != 1)) {
                             continue;
                         }
                         acl_paths.push_back(restricted_path);
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -862,20 +862,43 @@ private:
         int idx = 0;
         std::set<std::string> paths_write_seen;
         std::set<std::string> paths_create_or_modify_seen;
+        std::vector<std::string> acl_paths;
+        acl_paths.reserve(config.m_restricted_paths.size() + 1);
         while (acls[idx].resource && acls[idx++].authz) {
+            acl_paths.clear();
             const auto &acl_path = acls[idx-1].resource;
             const auto &acl_authz = acls[idx-1].authz;
-            if (!config.m_restricted_paths.empty()) {
-                bool found_path = false;
+            if (config.m_restricted_paths.empty()) {
+                acl_paths.push_back(acl_path);
+            } else {
+                auto acl_path_size = strlen(acl_path);
                 for (const auto &restricted_path : config.m_restricted_paths) {
+                    // See if the acl_path is more specific than the restricted path; if so, accept it
+                    // and move on to applying paths.
                     if (!strncmp(acl_path, restricted_path.c_str(), restricted_path.size())) {
-                        found_path = true;
+                        // Only do prefix checking on full path components.  If acl_path=/foobar and
+                        // restricted_path=/foo, then we shouldn't authorize access to /foobar.
+                        if (acl_path_size > restricted_path.size() && acl_path[restricted_path.size()] != '/') {
+                            continue;
+                        }
+                        acl_paths.push_back(acl_path);
                         break;
                     }
+                    // See if the restricted_path is more specific than the acl_path; if so, accept the
+                    // restricted path as the ACL.  Keep looping to see if other restricted paths add
+                    // more possible authorizations.
+                    if (!strncmp(acl_path, restricted_path.c_str(), acl_path_size)) {
+                        // Only do prefix checking on full path components.  If acl_path=/foo and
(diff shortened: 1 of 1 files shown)


abh3:
@bbockelm I'll merge it once @snafus gives the thumbs up after testing.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -862,20 +862,43 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
...
+                        acl_paths.push_back(acl_path);
                         break;
                     }
+                    // See if the restricted_path is more specific than the acl_path; if so, accept the
+                    // restricted path as the ACL.  Keep looping to see if other restricted paths add
+                    // more possible authorizations.
+                    if (!strncmp(acl_path, restricted_path.c_str(), acl_path_size)) {
+                        // Only do prefix checking on full path components.  If acl_path=/foo and
+                        // restricted_path=/foobar, then we shouldn't authorize access to /foobar.
+                        if (restricted_path.size() > acl_path_size && restricted_path[acl_path_size] != '/') {
```
amadio, line 893: pplSame here, but for `acl_path` = `/store` and `restricted_path` = `/store/`. I think there should be a function factored out that checks if a directory is a subdirectory of another, and use it in both places. You may also want to think about symlinks which may let a user bypass the restrictions (i.e. a symlink in the user directory under the restricted path which points to `/` and may let the user read more than they should).

amadio (commented):
I think this probably needs some changes before it can be merged, but since the next release will be 5.7, merging to `master` as it's set now is fine. Just let's iterate first until we are sure the problem is really fixed. Cheers,

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -862,20 +862,43 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
...
+            } else {
+                auto acl_path_size = strlen(acl_path);
                 for (const auto &restricted_path : config.m_restricted_paths) {
+                    // See if the acl_path is more specific than the restricted path; if so, accept it
+                    // and move on to applying paths.
                     if (!strncmp(acl_path, restricted_path.c_str(), restricted_path.size())) {
-                        found_path = true;
+                        // Only do prefix checking on full path components.  If acl_path=/foobar and
+                        // restricted_path=/foo, then we shouldn't authorize access to /foobar.
+                        if (acl_path_size > restricted_path.size() && acl_path[restricted_path.size()] != '/') {
```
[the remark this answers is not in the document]
Jo-stfc: the second condition should be acl_path[restricted_path.size()-1] != '/'

src/XrdSciTokens/XrdSciTokensAccess.cc
Jo-stfc: same here restricted_path[acl_path_size-1] != '/'

abh3:
Hey people, could we resolve the issues here? The problem does exist and we need an good solution.

src/XrdSciTokens/XrdSciTokensAccess.cc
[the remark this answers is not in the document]
Jo-stfc: everything works as intended with restricted_path = /store (the acl_path /store/ gets added if it's given in scope)
with restricted_path = /store/ it would reject /store acl_path

Jo-stfc:
an additional check was needed in XrdAccRules::apply (line 351) to address /foo /foobar access in scope (token scoped /foo could access /foobar).
My Implementation is as follows:
    bool apply(Access_Operation oper, std::string path) {
        for (const auto & rule : m_rules) {
            if ((oper == rule.first) && !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size()) && ( rule.second.size() == path.length() || path[rule.second.size()]=='/') ) {
                return true;
            }
        }
        return false;
    }

src/XrdSciTokens/XrdSciTokensAccess.cc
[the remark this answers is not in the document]
Jo-stfc: a MakeCanonical() pass on restricted_path before this should be sufficient

src/XrdSciTokens/XrdSciTokensAccess.cc
amadio: Actually, I think the check is correct, since it checks that for e.g. `acl_path = /store` and `restricted_path = /store/cms` it works and for `restricted_path = /storecms` it doesn't. That is, the `/` character is at position one bigger than the size of `acl_path`, which is correct.

Jo-stfc: the patch I put in would still reject `/store` if restricted_path is `/storecms`. it would pass the stringcompare and then correctly detect the last character as `e`, rejecting it at the  `!= '/'` condition.
The case I had to put in the -1 for was when the acl_path is `/` and restricted path is `/store`, which would still fail even if you change the strcmp condition

Jo-stfc: restricted_path is already 0 indexed, so the last character's position would be at `acl_path_size-1`
in fact, with the current condition, an `acl_path = /store/` would fail a check with `restricted_path= /store/cms` for the same reason, while with the -1 it would pass and add the restricted_path

src/XrdSciTokens/XrdSciTokensAccess.cc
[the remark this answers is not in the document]
bbockelm: > a MakeCanonical() pass on restricted_path before this should be sufficient to account for the /store/ restricted path case

This is already done [here](https://github.com/xrootd/xrootd/blob/master/src/XrdSciTokens/XrdSciTokensAccess.cc#L213) when parsing the configuration.

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: > You may also want to think about symlinks which may let a user bypass the restrictions (i.e. a symlink in the user directory under the restricted path which points to / and may let the user read more than they should).

Symlinks are a whole other set of (valid) concerns. For example, think about how the `all.export` setting works if you include user-controlled symlinks.  I do think it's a conversation worth having but at a higher level -- I don't want to derail this PR with a larger orthogonal issue.

bbockelm:
@Jo-stfc - I took your basic idea and simply reformatted it to be easier to read and added some comments about the logic.  Since the `restricted_path` setting is already canonicalized, I also resolved that issue.

Did spend some time with a test issuer going through the various cases and it looks good to me.  I would appreciate a second (or third) set of eyes, however!

The only "downside" here is this now _also_ fixes a long-standing issue (going from string prefix matching to more path matching) that's different from the original PR title.  That means the release notes are a bit more complicated.  I would suggest wording along the following lines:

- SciTokens-based authorization now permits requests as long as the request is allowed by both the `restricted_path` configuration and the token's ACLs.  Previously, a request was denied if the token's ACL was more permissive than the restricted path setting.
- SciTokens-based authorization is now based on path-prefix matching instead of string-based matching.  An authorization for the path `/foo` no longer permits operations on `/foobar`.

Jo-stfc:
Hi @bbockelm, I've compiled and ran my tests. The case with scopes broader than the restricted path (e.g. read:/ or read:/dteam:wlcg with restricted_path as /dteam:wlcg/test) are failing with permission denied. It works as intended with the -1 correction in the section I commented before (line 893)

Jo-stfc:
@bbockelm I've tested that PR and it works as intended. (accepts scope / and restricts it to restricted path).
with restricted path /foo/foo:
it will accept scopes /, /foo, /foo/ and restrict access to the equivalent scope of /foo/foo 
it will reject scopes /fo, /foobar and /foo/foobar

bbockelm:
@abh3 / @amadio -- OK, I think this one is finally ready to roll!

Note that it replaces #2151; that PR can be closed as well.

PR #2154, opened 2023-12-16
Send User-Agent as the monitor info

bbockelm:
After the initial login to the xrootd bridge (and after parsing the first request's headers), perform a `kXR_set` for the monitoring info.  The `User-Agent` header is sent as the info.

This is only done for the first request of a connection.  It is assumed that all subsequent connections belong to the same user agent.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -582,9 +582,29 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
 
     } else
       CurrentReq.reqstate++;
+  } else if (!DoneSetInfo && !CurrentReq.userAgent().empty()) { // DoingLogin is true, meaning the login finished.
+    std::string mon_info = "monitor info " + CurrentReq.userAgent();
+    DoneSetInfo = true;
+    if (mon_info.size() >= 1024) {
+      TRACEI(ALL, "User agent string too long");
+    } else if (!Bridge) {
+      TRACEI(ALL, "Internal logic error: Bridge is null after login");
+    } else {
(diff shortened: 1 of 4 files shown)


abh3 (approved):
Looks fine to me, Cedric do you have issues with this?

ccaffy (approved):
Looks good to me too! Thanks @bbockelm !

amadio:
Looks like this breaks HTTP transfers. I will need to revert and make another release.

abh3:
@bbockelm Could you look at this again and see why this is happening?

amadio:
@abh3 Cedric already created a pull request with a fix, which I confirmed and merged into `devel`.

abh3:
Then whose problem was it and what was the problem?

ccaffy:
Hi @abh3,

Brian used the Bridge to set the user-agent for the transfer and did not handle that case in the callback that is ran once the bridge executed the command. This prevented the file `open` to happen. I've added that handling (very simple handling) in the `PostProcessHttpReq()` member function so it does not mess-up the rest of the operations :)

amadio:
The problem was introduced by this pull request, and after I found out it broke http transfers, Cedric looked at it and fixed it. Today I'm going to make another release (5.6.6) including the fix.

abh3:
Hi Guilherme,

So, clearly, he never tested the change, sigh. OK, negative points here which means he gets lower priority unless we are assured the submitted patch has been tested. Perhaps even requiring a description on how to test the change so we can verify (that's the extreme case).

Andy

amadio:
My plan is to expand as much as possible our test suite to cover the most common operations such that we catch problems with patches when the pull request is opened. In #2134, the problem was caught by the new CI, but not the old. I am working on adding tests with http(s) transfers and a few authentication methods for 5.7.

PR #2156, opened 2023-12-17
Allow HTTP request to provide a hint about what file will be read

### LINKED ISSUE #2155: Xrootd significantly slower than `cat`
When downloading a 6.8GB file via HTTP from an underlying CephFS, I noticed that XRootD is significantly slower than `cat`:

- Time to download via HTTPS: 20s
- Time to download via HTTPS (hot cache): 6s.  [I include this case to verify that we're not being limited by HTTPS or the overhead of xrootd itself]
- Time to copy file to /dev/null via `cat`: 11s

Now why would `cat` be twice the speed of XRootD?  I verified that XRootD was issuing a nice, big sequence of 2MB reads (for reference, `cat` reads in 256KB chunks).  Since we've eliminated that xrootd or the HTTPS protocol is not the bottleneck, why would two processes issuing sequential reads have a nearly 2x performance difference?

After much experimentation, it appears the difference is `cat` issues the following system call after opening the file:

```
posix_fadvise(fd, 0, 0, POSIX_FADV_SEQUENTIAL);
```

CephFS stripes data across objects in a manner similar to RAID0.  Suppose you set the object size to 4MB, stripe size to 256KB, and stripe count to 4.  In this case, if you read 1MB, Linux will fetch from 4 objects concurrently (1MB / 256KB stripe size = 4 stripes).  If you have a stripe count of 1, then the data is written contiguously in the object; if you read 1MB, Linux will fetch the from 1 object.  Typically, the more parallelism the better -- you're likely pulling data from different storage devices.

The Ceph client's readahead setting will control the amount of data read and, hence, the number of objects read in parallel.  So, the impact of this `posix_fadvise` invocation is significantly increasing the number of ongoing parallel operations for `cat`, much beyond `xrootd`'s parallelism.  Decreasing the readahead down to 8MB (test above is at 128MB), `cat` slows to about the same speed as XRootD/HTTP.

I did a quick patch that invokes the same `posix_fadvise` call immediately after open within the xrootd framework; the time to download via HTTPS shrank to 13s (consistently slower than `cat` but much closer).

Note that the HTTP protocol could have issued a read that triggers `posix_fadvise(fd, 0, 0, POSIX_FADV_WILLNEED);`.  I tried switching the `posix_fadvise` invocation from `POSIX_FADV_SEQUENTIAL` to `POSIX_FADV_WILLNEED`; transfers took about 18s (better than the baseline 20s, but not by much).

So: is it possible to add a protocol extension (or even something hacked through the CGI parameters? or the bridge?) to allow the HTTP framework to provide this apparently-essential hint to the kernel?

@juztas, @ccaffy, and @smithdh - you might be interested in this one.

bbockelm:
Some filesystems (notably, CephFS) see remarkable performance improvements when you provide hints about upcoming sequential access.  Since we know within the HTTP protocol that sequential access is upcoming, we add a simple command that will notify XRootD of this via the bridge.

Uses the existing query command and adds new internal behavior.  The query command is already listed as "implementation-defined" and shouldn't need documentation in the protocol.

If the fsctl isn't implemented, then the XrdHttpProtocol will never again try to send it.

Fixes #2155

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.hh
+++ b/src/XrdHttp/XrdHttpProtocol.hh
@@ -105,7 +105,10 @@ public:
   int Stats(char *buff, int blen, int do_sync = 0);
 
 
+  /// Returns true if the filesystem supports fadvise
+  bool GetSupportsFadvise() const {return SupportsFadvise;}
 
+  void SetSupportsFadvise(bool val) {SupportsFadvise = val;}
 
   /// Perform a Stat request
   int doStat(char *fname);
@@ -298,10 +301,16 @@ private:
 
   /// Tells that we are just logging in
   bool DoingLogin;
-  
+
+  /// Indicates whether we've attempted to send app info.
+  bool DoneSetInfo;
+
   /// Tells that we are just waiting to have N bytes in the buffer
   long ResumeBytes;
-  
+
+  /// Tells that we will perform a fadvise hint.
+  bool SupportsFadvise{true};
+
   /// Private SSL context
   SSL *ssl;
 
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -500,7 +500,7 @@ bool XrdHttpReq::Error(XrdXrootd::Bridge::Context &info, //!< the result context
         const char *etext_ //!< associated error message
         ) {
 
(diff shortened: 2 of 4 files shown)


bbockelm:
Closing this PR in favor of the `kXR_seqio`-based approach suggested by @abh3.

PR #2160, opened 2024-01-09
force Destination address on HTTP-TPC

ddavila0:
Committer: Diego Davila <davila.foyo@gmail.com>
This was though to work on the Rucio/SENSE context where a single host gets N IPv6 addresses assigned.
When a host has more than 1 IP address assigned, the Destination address used by the HTTP-TPC is selected randomly. We want the address to be selected based on the hostname that was used to contact the server.

### DIFF
--- a/src/XrdTpc/XrdTpcTPC.cc
+++ b/src/XrdTpc/XrdTpcTPC.cc
@@ -1,5 +1,6 @@
 #include "XrdHttp/XrdHttpExtHandler.hh"
 #include "XrdNet/XrdNetAddr.hh"
+#include "XrdNet/XrdNetUtils.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 #include "XrdSec/XrdSecEntity.hh"
 #include "XrdSfs/XrdSfsInterface.hh"
@@ -321,6 +322,7 @@ TPCHandler::~TPCHandler() {
   
 TPCHandler::TPCHandler(XrdSysError *log, const char *config, XrdOucEnv *myEnv) :
         m_desthttps(false),
+        m_sense(false),
         m_timeout(60),
         m_first_timeout(120),
         m_log(log->logger(), "TPC_"),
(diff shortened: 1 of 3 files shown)


abh3 (approved):
This looks OK to me, though I would have preferred more comments on why the code is there to begin with. Not a deal breaker but it would be nice to add! Cedric?

abh3 (changes requested):
Actually, after some discussion, we would prefer that the added cod would be under the control of an option. This should be easy to do because nothing here replaces code it simply adds code. Let's call the option "http.sense" as this is only relevant to sense configurations. If not, then maybe a more descriptive option like "http.tpcdest force".

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -954,40 +954,44 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
...
-
-    // Get all the IP addresses associated with the above hostname
-    XrdNetUtils::GetAddrs(host_used.c_str(), &nP, numIP, XrdNetUtils::prefAuto, 0);
+    // The following change was required by the Rucio/SENSE project where
+    // multiple IP addresses, each from a different subnet, are assigned to a
+    // single server and routed differently by SENSE.
+    // The above requires the server to utilize the same IP, that was used to
+    // start the TPC, for the resolution of the given TPC instead of
+    // using any of the IPs available.
+    char *sense = getenv( "XRD_SENSE" );
```
abh3, line 963: I apologize for note getting to you earlier about this. An alternative is to add a config directive "tpc.sense" no need for an argument as it's implied to be true. This can trivially be done in TPCHandler::Configure @ XrdTpcConfigure.cc.  You'll see that you will be able to set a bool in XrdTpcTpc.hh which can be tested here. The place to do it will be obvious (e.g. before/after the test for tpc.timeout). It's a trivial addition and avoids proliferating global envars. OK?

abh3 (changes requested):

abh3 (approved):
Thank you Diego. It all looks good.

abh3:
Oh, one more thing could you change this to update the devel branch not the master branch? We've changed the flow so that updates first go into the devel branch before being committed to master.

ddavila0:
I've created a new PR: https://github.com/xrootd/xrootd/pull/2172

amadio:
Thanks! Next time, you can just force-push to your branch and keep the same pull request. It's also possible to change the target branch at the top near the title of the pull request. The way you did it is fine as well, though!

Issue #2161, opened 2024-01-11
[XrdTPC::PmarkManager] TSAN failure read after freed variable


ccaffy:
I saw a segmentation fault when trying to do HTTP TPC transfers via EOS with Source server = Destination Server.

By enabling ASAN, I've managed to see what the problem is...

```cpp
==8260==ERROR: AddressSanitizer: heap-use-after-free on address 0x60c00002dc40 at pc 0x7fe94d518ca7 bp 0x7fe9468db180 sp 0x7fe9468db170
240111 15:00:54 8287 anon.0:36@xrootd-server http_Protocol:  Entering SSL_accept...
READ of size 8 at 0x60c00002dc40 thread T3
    #0 0x7fe94d518ca6 in XrdSecEntity::~XrdSecEntity() /root/xrootd/src/XrdSec/XrdSecEntity.cc:51
    #1 0x7fe944574627 in XrdTpc::PMarkManager::SocketInfo::~SocketInfo() /root/xrootd/src/XrdTpc/XrdTpcPMarkManager.hh:58
    #2 0x7fe9445750d2 in XrdTpc::PMarkManager::beginPMarks() /root/xrootd/src/XrdTpc/XrdTpcPMarkManager.cc:54
    #3 0x7fe944563fa9 in TPC::TPCHandler::RunCurlWithUpdates(void*, XrdHttpExtReq&, TPC::State&, TPC::TPCHandler::TPCLogRecord&) /root/xrootd/src/XrdTpc/XrdTpcTPC.cc:681
    #4 0x7fe94456aab7 in TPC::TPCHandler::ProcessPullReq(std::string const&, XrdHttpExtReq&) /root/xrootd/src/XrdTpc/XrdTpcTPC.cc:1044
    #5 0x7fe94455dff6 in TPC::TPCHandler::ProcessReq(XrdHttpExtReq&) /root/xrootd/src/XrdTpc/XrdTpcTPC.cc:300
    #6 0x7fe945a6902c in XrdHttpReq::ProcessHTTPReq() /root/xrootd/src/XrdHttp/XrdHttpReq.cc:950
    #7 0x7fe945a41e5a in XrdHttpProtocol::Process(XrdLink*) /root/xrootd/src/XrdHttp/XrdHttpProtocol.cc:873
    #8 0x421e5b in XrdProtLoad::Process(XrdLink*) /root/xrootd/src/Xrd/XrdProtLoad.cc:236
[...]

PR #2163, opened 2024-01-11
[XrdTpcTPC] PMarkManager - corrected a segfault that may happen in the creation of the first packet marking handle

ccaffy:
Fixes the issue https://github.com/xrootd/xrootd/issues/2161

### DIFF
--- a/src/XrdTpc/XrdTpcPMarkManager.cc
+++ b/src/XrdTpc/XrdTpcPMarkManager.cc
@@ -51,14 +51,14 @@ void PMarkManager::beginPMarks() {
     // This base pmark handle will be placed at the beginning of the vector of pmark handles
     std::stringstream ss;
     ss << "scitag.flow=" << mReq->mSciTag;
-    auto sockInfo = mSocketInfos.front();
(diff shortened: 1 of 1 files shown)


abh3 (commented):
Seems like auto does not deduce reference types in templates. See,
https://stackoverflow.com/questions/56429977/auto-return-type-not-deducing-reference

The suggested fix is to use decltype as the post points out. Alternatively, explicitly specify the correct type. What is your opinion?

amadio:
Looks good to me, feel free to merge, or I will do it tomorrow. Thanks!

ccaffy:
> Seems like auto does not deduce reference types in templates. See, https://stackoverflow.com/questions/56429977/auto-return-type-not-deducing-reference
> 
> The suggested fix is to use decltype as the post points out. Alternatively, explicitly specify the correct type. What is your opinion?

Yes, will use the correct type instead of auto ;) Will test a home-made compiled version of XRootD against EOS before pushing this fix for the merge ;)

ccaffy:
All good, I've tried this fix on EOS and everything work fine now.
Thanks for the merge !

Cheers,
Cedric

PR #2165, opened 2024-01-17
Fix compilation error reported with gcc 14

ellert:
```
/builddir/build/BUILD/xrootd-5.6.4/src/XrdSecgsi/XrdSecgsitest.cc: In function ‘pdots(char const*, bool)’:
/builddir/build/BUILD/xrootd-5.6.4/src/XrdSecgsi/XrdSecgsitest.cc:90:15: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
   90 |    printf("|| %s ", t);
      |               ^~
```

### DIFF
--- a/src/XrdSecgsi/XrdSecgsitest.cc
+++ b/src/XrdSecgsi/XrdSecgsitest.cc
@@ -87,7 +87,11 @@ static void pdots(const char *t, bool ok = 1)
    unsigned int i = 0;
    unsigned int l = (t) ? strlen (t) : 0;
    unsigned int np = PRTWIDTH - l - 8;
-   printf("|| %s ", t);
+   if (l > 0) {
+      printf("|| %s ", t);
+   } else {
+      printf("||  ");
+   }
    for (; i < np ; i++) { printf("."); }
    printf("  %s\n", (ok ? "PASSED" : "FAILED"));
 }


src/XrdSecgsi/XrdSecgsitest.cc
```diff
@@ -87,7 +87,11 @@ static void pdots(const char *t, bool ok = 1)
    unsigned int i = 0;
    unsigned int l = (t) ? strlen (t) : 0;
    unsigned int np = PRTWIDTH - l - 8;
-   printf("|| %s ", t);
+   if (l > 0) {
+      printf("|| %s ", t);
+   } else {
+      printf("||  ");
+   }
```
amadio, line 94: How about
```suggestion
   printf("|| %s ", t ? t : "");
```

ellert: I based the proposed change to the style used a few lines further down in the file to be consistent:

https://github.com/xrootd/xrootd/blob/333a5dd8187a897bc2c3c5e8097be0aef11b8e7e/src/XrdSecgsi/XrdSecgsitest.cc#L99-L103

amadio: Ok, let's keep it this way then.

Issue #2162, opened 2024-01-11
Xrdcp high CPU consuption on AlmaLinux9


joaopblopes:
An increase in CPU consumption was observed on FTS transfers using xrootd on AlmaLinux 9 machines.

After an initial investigation with perf, it seems that most of the CPU time is spent on the openssl `EVP_PKEY_param_check` call.

```
$ perf record -g --call-graph dwarf xrdcp root://eospps.cern.ch:1094//eos/opstest/dteam/batistal/file.1mb root://eospublic.cern.ch:1094//eos/opstest/dteam/batistal/file.1mb_`uuidgen`
```

```
$ perf report --call-graph --stdio
.
.
.
#
# Children      Self  Command  Shared Object          Symbol                                                         
# ........  ........  .......  .....................  ...............................................................
#
    99.34%     0.00%  xrdcp    libXrdSecgsi-5.so      [.] XrdSecProtocolgsi::getCredentials
            |
            ---XrdSecProtocolgsi::getCredentials
               |          
                --98.97%--XrdSecProtocolgsi::ParseClientInput
                          |          
                          |--93.35%--XrdSecProtocolgsi::ClientDoCert
                          |          |          
                          |           --93.13%--XrdCryptosslFactory::Cipher
                          |                     XrdCryptosslCipher::XrdCryptosslCipher
                          |                     |          
                          |                      --92.30%--EVP_PKEY_param_check
                          |                                evp_pkey_param_check_combined
                          |                                try_provided_check
                          |                                evp_keymgmt_validate
                          |                                dh_validate
                          |                                DH_check_ex
                          |                                DH_check
                          |                                BN_check_prime
                          |                                ossl_bn_check_prime
                          |                                bn_is_prime_int
                          |                                |          
                          |                                 --92.17%--ossl_bn_miller_rabin_is_prime
                          |                                           |          
                          |                                            --91.94%--BN_mod_exp_mont
                          |                                                      |          
                          |                                                       --90.59%--bn_mul_mont_fixed_top
                          |                                                                 |          

```

The xrootd and openssl versions installed in the machine are:
```
[root@fts-daq-005 ~]# rpm -qa | grep -E '^(openssl|xrootd)'
openssl-libs-3.0.7-24.el9.x86_64
openssl-3.0.7-24.el9.x86_64
xrootd-libs-5.6.3-3.el9.x86_64
xrootd-client-libs-5.6.3-3.el9.x86_64
[...]

PR #2166, opened 2024-01-17
[GSI] Skip check of our standard DH parameters

smithdh:
This patch is to reduce some of the increased CPU time reported in issue #2162.

Increased CPU usage was seen on alma9 (which uses openssl 3.0.7), when using xrootd and gsi. The extra cpu usage is seen by the client using gsi authentication when setting up a xrootd connection to a server. For example (see #2162), 0.15s with openssl 1.0.2k compared with 1.9s with openssl 3.0.7.

This patch reduces the time increase by conditionally skipping the Diffie-Hellman parameter check. The timings reported in the ticket indicate this reduces runtime from ~1.9s to ~0.24s. (This is still longer than the reference runtime with openssl 1.0.2k). The remaining extra time is not thought to be due to DH checks, so for the sake of keeping this PR compact and help review & discussion, this patch only changes the DH check.

### DIFF
--- a/src/XrdCrypto/XrdCryptosslCipher.cc
+++ b/src/XrdCrypto/XrdCryptosslCipher.cc
@@ -166,7 +166,33 @@ static int DSA_set0_key(DSA *d, BIGNUM *pub_key, BIGNUM *priv_key)
 }
 #endif
 
+static EVP_PKEY *getFixedDHParams() {
+    static EVP_PKEY *dhparms = [] {
+        EVP_PKEY *dhParam = 0;
+
+        BIO *biop = BIO_new(BIO_s_mem());
+        BIO_write(biop, dh_param_enc, strlen(dh_param_enc));
+        PEM_read_bio_Parameters(biop, &dhParam);
+        BIO_free(biop);
+        return dhParam;
+    }();
+
+    assert(dhparms);
+    return dhparms;
+}
+
 static int XrdCheckDH (EVP_PKEY *pkey) {
+   // If the DH parameters we received are our fixed set we know they
+   // are acceptable. The parameter check requires computation and more
+   // with openssl 3 than previously. So skip if DH params are known.
+   const EVP_PKEY *dhparms = getFixedDHParams();
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+   const bool skipcheck = EVP_PKEY_parameters_eq(pkey, dhparms);
+#else
+   const bool skipcheck = EVP_PKEY_cmp_parameters(pkey, dhparms);
+#endif
+   if (skipcheck) return 1;
+
    int rc;
 #if OPENSSL_VERSION_NUMBER < 0x10101000L
    DH *dh = EVP_PKEY_get0_DH(pkey);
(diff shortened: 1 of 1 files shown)


abh3 (commented):
Looks good. Only question is whether it would be possible to devise a consistent hash match? If so, the code would be much shorter.

smithdh:
Thanks. I see there's a build error on some platforms (possibly for openssl 1.1.1 systems), will check & fix.

smithdh:
I've made the patch shorter, there is an openssl function to compare parameters, so I used that instead of writing a custom one. I didn't see a way to introduce a hash of the parameters that would give shorter code, so I left with the compare.
I changed the skipcheck test to apply with all versions of openssl; while the large time saving is seen with openssl 3 there appears to be small advantage even with openssl 1.0.2 (e.g. ~50ms). The patch passes the tests (i.e. on systems with openssl 1.0, 1.1 or 3.).

src/XrdCrypto/XrdCryptosslCipher.cc
```diff
@@ -166,7 +166,33 @@ static int DSA_set0_key(DSA *d, BIGNUM *pub_key, BIGNUM *priv_key)
...
+static EVP_PKEY *getFixedDHParams() {
+    static EVP_PKEY *dhparms = [] {
+        EVP_PKEY *dhParam = 0;
+
+        BIO *biop = BIO_new(BIO_s_mem());
+        BIO_write(biop, dh_param_enc, strlen(dh_param_enc));
+        PEM_read_bio_Parameters(biop, &dhParam);
+        BIO_free(biop);
+        return dhParam;
+    }();
```
amadio, line 178: I there a particular reason to wrap this in a lambda? It's fine as it is, I'm just wondering.

amadio (approved):
Looks good to me, maybe just consider adding a comment (or editing the description of the pull request) showing the tests that were run and the difference observed on Alma 9.

amadio:
I am merging this to proceed with preparation for the next release. Thanks!

amadio:
For the record, this is what I observe on my machine:
```
$ perf diff -s dso ~/perf.data perf.data
# Event 'cycles'
#
# Baseline  Delta Abs  Shared Object         
# ........  .........  ......................
#
    84.87%    -37.46%  libcrypto.so.3        
    13.51%    +32.65%  [kernel.vmlinux]      
     0.68%     +2.47%  libc.so.6             
     0.14%     +0.76%  libssl.so.3           
     0.33%     +0.61%  ld-linux-x86-64.so.2  
     0.10%     +0.30%  libXrdUtils.so.3.0.0  
     0.22%     +0.28%  libXrdCl.so.3.0.0     
     0.04%     +0.13%  xrdcp                 
               +0.07%  libssh2.so.1.0.1      
               +0.07%  [vdso]                
               +0.07%  libXrdSecgsi-5.so     
     0.01%     +0.07%  libnss_myhostname.so.2
     0.06%     +0.00%  libstdc++.so.6.0.33   
     0.03%             libgcc_s.so.1         
     0.02%             libXrdCryptossl-5.so
$ perf diff -c ratio -o0 -s comm ~/perf.data perf.data
# Event 'cycles'
#
# Baseline           Ratio  Command
# ........  ..............  .......
#
    99.95%        0.204778  xrdcp
```
So about 80% reduction in CPU usage to copy a 100MB file from a local server using GSI authentication. Thanks, @smithdh!

PR #2172, opened 2024-01-24
add config val tpc.sense to force the Destination IP address on a HTTP-TPC

ddavila0:
Committer: Diego Davila <davila.foyo@gmail.com>

### DIFF
--- a/src/XrdTpc/XrdTpcConfigure.cc
+++ b/src/XrdTpc/XrdTpcConfigure.cc
@@ -56,7 +56,22 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
                 Config.Close();
                 return false;
             }
-        } else if (!strcmp("tpc.timeout", val)) {
+        } else if (!strcmp("tpc.fixed_route", val)) {
+		if (!(val = Config.GetWord())) {
+                Config.Close();
+                m_log.Emsg("Config", "tpc.fixed_route value not specified");
+                return false;
+            }
+            if (!strcmp("1", val) || !strcasecmp("yes", val) || !strcasecmp("true", val)) {
+                m_fixed_route= true;
+            } else if (!strcmp("0", val) || !strcasecmp("no", val) || !strcasecmp("false", val)) {
+                m_fixed_route= false;
+	    } else{
+                Config.Close();
+                m_log.Emsg("Config", "tpc.fixed_route value is invalid", val);
+                return false;
+	    }
+	} else if (!strcmp("tpc.timeout", val)) {
             if (!(val = Config.GetWord())) {
                 m_log.Emsg("Config","tpc.timeout value not specified.");  return false;
             }
--- a/src/XrdTpc/XrdTpcTPC.cc
+++ b/src/XrdTpc/XrdTpcTPC.cc
@@ -1,5 +1,6 @@
 #include "XrdHttp/XrdHttpExtHandler.hh"
 #include "XrdNet/XrdNetAddr.hh"
+#include "XrdNet/XrdNetUtils.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 #include "XrdSec/XrdSecEntity.hh"
 #include "XrdSfs/XrdSfsInterface.hh"
@@ -321,6 +322,7 @@ TPCHandler::~TPCHandler() {
   
 TPCHandler::TPCHandler(XrdSysError *log, const char *config, XrdOucEnv *myEnv) :
         m_desthttps(false),
+        m_fixed_route(false),
         m_timeout(60),
         m_first_timeout(120),
         m_log(log->logger(), "TPC_"),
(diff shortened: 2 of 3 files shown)


abh3 (approved):
Thanks for retargeting!

abh3:
@ccaffy Cedric could you take a quick look and see if anything is amiss?

ccaffy:
Yes, will compile and try few HTTP-TPC transfers :)

amadio:
@ddavila0 Could you please remove the line `Committer: Diego Davila davila.foyo@gmail.com` from the commit message? Committer will actually be whoever merges this into `devel` for you (git marks this in the commit), and you should be anyway marked correctly as the author of the commit. If it's too much of a hassle, you can leave it and I may remove it upon merging. Thank you!

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -952,6 +954,38 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
...
             return req.SendSimpleResp(rec.status, NULL, NULL, msg, 0);
     }
+    // ddavila 2023-01-05:
+    // The following change was required by the Rucio/SENSE project where
+    // multiple IP addresses, each from a different subnet, are assigned to a
+    // single server and routed differently by SENSE.
+    // The above requires the server to utilize the same IP, that was used to
+    // start the TPC, for the resolution of the given TPC instead of
+    // using any of the IPs available.
+    if (m_sense  == true ){
```
amadio, line 964: I wonder if a more generic name isn't more suitable here, than using the name of the software that needs this (as it may also be useful in other domains, like within containerized setups). I thought maybe `tpc.route` with options `fixed` and `dynamic` (default), or something similar to that. For example, within Kubernetes, that may turn out to be needed as well.

src/XrdTpc/XrdTpcConfigure.cc
```diff
@@ -56,7 +56,22 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
...
                 return false;
             }
-        } else if (!strcmp("tpc.timeout", val)) {
+        } else if (!strcmp("tpc.route", val)) {
+		if (!(val = Config.GetWord())) {
+                Config.Close();
+                m_log.Emsg("Config", "tpc.route value not specified");
+                return false;
+            }
+            if (!strcasecmp("fixed", val)){
```
abh3, line 65: This is really overkill. Can you just use a flag? The route is either fixed or dynamic. So, m_route_fixed can be a bool and when true it's fixed and when false it's dynamic. You can do it the other way around (i.e. define m_route_dynamic).

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -952,6 +954,38 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
...
             return req.SendSimpleResp(rec.status, NULL, NULL, msg, 0);
     }
+    // ddavila 2023-01-05:
+    // The following change was required by the Rucio/SENSE project where
+    // multiple IP addresses, each from a different subnet, are assigned to a
+    // single server and routed differently by SENSE.
+    // The above requires the server to utilize the same IP, that was used to
+    // start the TPC, for the resolution of the given TPC instead of
+    // using any of the IPs available.
+    if (!strcmp("fixed", m_route)){
```
abh3, line 964: Hmmm, let's simplify this. See my comment above where you parse the value to make this a simple bool test.

src/XrdTpc/XrdTpcTPC.hh
```diff
@@ -144,6 +144,8 @@ private:
     static size_t m_block_size;
     static size_t m_small_block_size;
     bool m_desthttps;
+    char const * m_route;  // Defines how the Destination IP is selected in an HTTP-TPC, "fixed" picks the same
```
abh3, line 147: This really can be a bool.

abh3 (changes requested):
Sorry, I though this would be the last iteration but not so, sigh.

src/XrdTpc/XrdTpcConfigure.cc
ddavila0: I'd prefer to have a config variable that has a value assigned, so I'd expect the config file to look like this:
`tpc.fixed_route true`
but that's pretty much the same (or very similar ) code to what it's currently done. Are you asking me to change this to assume 'true' as long as the variable exists ?

abh3: No, I was simply asking not to do constant string compares for something that can be decided by a unitary value. You can use bools and logically use the variable name to indicate what the bool value means (that was my example) or you can use an enum if you find that a variable name/value paradigm is not to your liking (though that would also be overkill given that the enum would have two values of which only one would be tested)

src/XrdTpc/XrdTpcTPC.hh
ddavila0: Done

abh3 (approved):
Tank you Diego. It looks fine to me though @amadio would have preferred something like "tpc.route {fixed|dynamic}" where "fixed would set m_fixed_route to true and dynamic would set it false, from what I gather. If he's OK with this version then we go with it.

amadio:
I'm fine with merging this version in. If we change our minds about the specific naming, I can adjust before the next release.

PR #2173, opened 2024-01-24
[XrdHttp] The PostProcessHttpReq callback now takes into account the setting of the user agent.

ccaffy:


amadio:
I confirm this fixes the problem, thank you @ccaffy!

PR #2176, opened 2024-01-25
Implement ability to have the token groups as a separate claim

s-thiele:
This implements the ability to have the groups claim in the token as a separate variable `groups_claim` to be set in the config, analogous to the username claim in #1978 . If `groups_claim` is not set in the config, it defaults back to the previously hard coded `"wlcg.groups"`.

### DIFF
--- a/src/XrdSciTokens/README.md
+++ b/src/XrdSciTokens/README.md
@@ -115,7 +115,9 @@ are:
    - `username_claim` (optional): Not all issuers put the desired username in the `sub` claim (sometimes the subject is
       set to a de-identified value).  To use an alternate claim as the username, such as `uid`, set this to the desired
       claim name.  If set, it overrides `map_subject` and `default_user`.
-   -  `name_mapfile` (options): If set, then the referenced file is parsed as a JSON object and the specified mappings
+   - `groups_claim` (optional): Not all issuers put the desired groups in the `wlcg.groups` claim. To use an alternate claim
+      as the groups, set this to the desired claim name. If not set, the default is `wlcg.groups`.
+   - `name_mapfile` (options): If set, then the referenced file is parsed as a JSON object and the specified mappings
       are applied to the username inside the XRootD framework.  See below for more information on the mapfile.
    -  `authorization_strategy` (optional): One or more authorizations to use from the token.  Multiple (space separated)
       items may be specified from the following valid values:
@@ -168,7 +170,8 @@ The enumerated keys are:
      For example, if the issuer's base path is `/home`, the operation is accessing `/home/bbockelm/foo`, and the path in
      the rule is `/bbockelm`, then this attribute evaluates to `true`.  Note the path value and the requested path must
      be normalized; if presented with `/home//bbockelm/`, then this is treated as if `/home/bbockelm` was given.
-   - `group`: Case-sensitive match against one of the groups in the token.
+   - `group`: Case-sensitive match against one of the groups in the token (the claim specifying the groups is configurable, controlled by the
+     `groups_claim` variable in the issuer config; default is `wlcg.groups`).
    - `ignore`: If present (regardless of the value), the rule is ignored.
    - `comment`: Ignored; reserved for adding comments from the administrator.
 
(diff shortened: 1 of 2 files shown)


abh3:
@bbockelm This is news to me, Brian can you look at this and approve it, if this is what you want?

olifre:
To further elaborate, this is meant for use with non-WLCG JWTs, i.e. from other issuers — those of course do not use `wlcg.groups`, but other claims to store the group information (e.g. `eduperson_entitlement` is a common claim to use). 
Since the default is unchanged, it does not break WLCG workflows, but of course it's best if @bbockelm also confirms it :wink: .

bbockelm (changes requested):
Happy with the change overall.

Just a small requested change to fix what appears to be a pre-existing bug that could cause a crash.

bbockelm:
@s-thiele (or @olifre) -- would you have the time to do the small refactor I suggested in the code review?  There's just the one minor item and I think this will be good to go.

s-thiele:
> @s-thiele (or @olifre) -- would you have the time to do the small refactor I suggested in the code review? There's just the one minor item and I think this will be good to go.

Yes, sorry for the delay. Next week I'm on a conference, but then I'll take a look. At first glance I'm not actually sure this can actually produce a crash, but it's certainly not the nicest to read :) Either way, I'll make the changes the week after next.

s-thiele:
> @s-thiele (or @olifre) -- would you have the time to do the small refactor I suggested in the code review? There's just the one minor item and I think this will be good to go.

@bbockelm done :) added the new commit to the pull request.

amadio:
@s-thiele, @bbockelm This pull request needs to be updated to take into consideration the new autorization strategy configuration option (#2205), so that it also affects the groups option. So it needs to be rebased on top of the `devel` branch, which already has the new feature merged. Once that is done, I can test and merge it into `devel` to be released in [5.7.0](https://github.com/xrootd/xrootd/milestone/36).

amadio:
Actually, I just rebased this locally. I will test and push here the changes after rebasing and merge, no need for you to do anything. Thanks!

amadio:
@bbockelm After testing this locally, my conclusion is that the defaults for both `username_claim` and `groups_claim` should be handled internally by the scitokens library based on the token profile, that is `groups` for scitokens profile, and `wlcg.groups` for WLCG profile, in the same way that I (painfully) learned that the scope is `storage.read:/foo` in WLCG profile and `read:/` in scitokens profile. The profile JSON should be parsed by the scitokens library and the right thing should be done for each profile. What we are doing here are just workarounds to let things work.

amadio:
The failure in Alpine is due to not being able to checkout the previous commit, likely due to my force-push followed by change of merge base. It can be ignored.

olifre:
> @bbockelm After testing this locally, my conclusion is that the defaults for both `username_claim` and `groups_claim` should be handled internally by the scitokens library based on the token profile, that is `groups` for scitokens profile, and `wlcg.groups` for WLCG profile, in the same way that I (painfully) learned that the scope is `storage.read:/foo` in WLCG profile and `read:/` in scitokens profile.

One problem with that would be the use of more generic tokens without a fixed profile on where groups are stored — actually, the tokens which @s-thiele (mainly) developed this for are `at+jwt`-tokens issued by the [Unity IdM](https://unity-idm.eu/) which is used in the Helmholtz and Research Data Infrastructure communities in Europe, for example. These tokens don't have a specific profile they follow, but adhere to `at+jwt`, and it's up to the IdM implementaiton (in this case Unity) to decide which field to use for the group information (this can for example be `eduperson_entitlement`). 

So using defaults from a profile would of course be a good idea (I wholeheartedly agree), but somewhere a knob is needed to allow for the use of generic tokens which do not have a profile defining a `group` claim which must be used.

amadio:
@olifre Yes, of course. I agree with what you say above. The profile is just so that the defaults are correct, since they can change. The `groups_claim` option is there now, so if you want you can use it and let us know if you encounter any problems.

PR #2177, opened 2024-01-29
Add support for pelican:// protocol

bbockelm:
In https://github.com/PelicanPlatform/xrdcl-pelican, we are developing a XrdCl plugin that can talk to the infrastructure for a new project, christening the URL scheme `pelican://`.

This commit adds the new schema so it can be utilized from both xrdcp (primarily for testing) and XCache.


abh3 (approved):
Looks fine to me. I presume this has been tested to the extent that it passes the URL downstream.

PR #2183, opened 2024-02-03
Fix ordering of the debug levels

bbockelm:
The documentation says "increasing values produce more detail".  However, `DebugLevel 1` was mapped to "Info" and `DebugLevel 3` was mapped to "Error".

With this, `pss.setopt DebugLevel N` has the following meanings:

- 0: None
- 1: Error
- 2: Warning
- 3: Info
- 4: Debug
- 5: Dump

Previously, 1 and 3 were swapped.


amadio:
Looks good, thanks.

PR #2184, opened 2024-02-08
add stat permissions to create, modify and write

Jo-stfc:
stat permissions are implicitly required for storage systems for create, modify  and delete operations. Currently tokens with purely create and modify permissions would fail with permission denied on stat on attempts to perform those operations

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -353,6 +353,12 @@ public:
             if ((oper == rule.first) && !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size())) {
                 return true;
             }
+            // according to WLCG token specs, allow creation of required superfolders for a new file if requested
+            if ((oper == AOP_Stat || oper == AOP_Mkdir) && 
+              !rule.second.compare(0, path.size(), path, 0, path.size()) &&  
+              rule.second.size() >= path.length()  ) {
+                return true;
+            }
         }
         return false;
     }
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -889,6 +889,7 @@ private:
                     xrd_rules.emplace_back(AOP_Mkdir, path);
                     xrd_rules.emplace_back(AOP_Rename, path);
                     xrd_rules.emplace_back(AOP_Excl_Insert, path);
+                    xrd_rules.emplace_back(AOP_Stat, path);
                 } else if (!strcmp(acl_authz, "modify")) {
                     paths_create_or_modify_seen.insert(path);
                     xrd_rules.emplace_back(AOP_Create, path);
@@ -897,6 +898,7 @@ private:
                     xrd_rules.emplace_back(AOP_Insert, path);
                     xrd_rules.emplace_back(AOP_Update, path);
                     xrd_rules.emplace_back(AOP_Chmod, path);
+                    xrd_rules.emplace_back(AOP_Stat, path);
                     xrd_rules.emplace_back(AOP_Delete, path);
                 } else if (!strcmp(acl_authz, "write")) {
                     paths_write_seen.insert(path);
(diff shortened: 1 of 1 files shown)


src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -889,6 +889,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
                     xrd_rules.emplace_back(AOP_Mkdir, path);
                     xrd_rules.emplace_back(AOP_Rename, path);
                     xrd_rules.emplace_back(AOP_Excl_Insert, path);
+		    xrd_rules.emplace_back(AOP_Stat, path);
```
amadio, line 892: The formatting seems odd here. Can you please check that there was no mixing of tabs/spaces? Thank you.

Jo-stfc: thanks, there was a mixing of tabs when I edited it. amended

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -889,14 +889,16 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
...
                 } else if (!strcmp(acl_authz, "modify")) {
                     paths_create_or_modify_seen.insert(path);
                     xrd_rules.emplace_back(AOP_Create, path);
                     xrd_rules.emplace_back(AOP_Mkdir, path);
                     xrd_rules.emplace_back(AOP_Rename, path);
                     xrd_rules.emplace_back(AOP_Insert, path);
                     xrd_rules.emplace_back(AOP_Update, path);
-                    xrd_rules.emplace_back(AOP_Chmod, path);
+                    xrd_rules.emplace_back(AOP_Chmod, path);xrd_rules.emplace_back(AOP_Stat, path);
+                    xrd_rules.emplace_back(AOP_Stat, path);
```
amadio, line 901: You have the same statement twice here.

Jo-stfc: thanks, fixed

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -353,6 +353,10 @@ class XrdAccRules
             if ((oper == rule.first) && !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size())) {
                 return true;
             }
+            // according to WLCG token specs, allow creation of required superfolders for a new file if requested
+            if ((oper == AOP_Stat || oper == AOP_Mkdir) && !rule.second.compare(0, path.size(), path, 0, path.size()) &&  rule.second.size() >= path.length()  ) {
```
bbockelm, line 357: This would allow a `mkdir` operation using a read-only token, correct?  How about something like this:
```suggestion
            if ((oper == rule.first) && (oper == AOP_Stat || oper == AOP_Mkdir) && rule.second.size() >= path.length() && !rule.second.compare(0, path.size(), path, 0, path.size()) && (rule.second.size() == path.length() || rule.second[path.length()] == '/') ) {
```

The last part of the conditional:

```
(rule.second.size() == path.length() || rule.second[path.length()] == '/')
```

mandates either the two strings are the same _xor_, as the ACL path is longer, the next character in the ACL path is '/'.  That is:

- requested: mkdir /foo, acl: mkdir /foobar/baz => FAIL
- requested: mkdir /foobar, acl: mkdir /foobar => OK
- requested: mkdir /foobar/baz, acl: mkdir /foobar/baz => OK
- requested: mkdir /, acl: mkdir /foobar/baz => FAIL

Jo-stfc: thanks Brian, I've updated the PR with your suggestion

bbockelm (approved):
Looks good to me and am satisfied with @Jo-stfc's testing reports.

amadio (changes requested):
Should be good to go, could you please just break the line as suggested? Thank you!

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -353,6 +353,10 @@ class XrdAccRules
             if ((oper == rule.first) && !path.compare(0, rule.second.size(), rule.second, 0, rule.second.size())) {
                 return true;
             }
+            // according to WLCG token specs, allow creation of required superfolders for a new file if requested
+            if ((oper == rule.first) && (oper == AOP_Stat || oper == AOP_Mkdir) && rule.second.size() >= path.length() && !rule.second.compare(0, path.size(), path, 0, path.size()) && (rule.second.size() == path.length() || rule.second[path.length()] == '/') ) {
```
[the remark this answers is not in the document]
Jo-stfc: thanks, I've broken the line by condition

Issue #2190, opened 2024-02-12
Consolidate ENABLE_TESTS and BUILD_TESTS


veprbl:
Neither `ENABLE_TESTS` nor `BUILD_TESTS` seems to be self sufficient. Would it be possible to have just one? Preferably named `BUILD_TESTING` instead, because that's the canonical cmake name for this.

amadio:
I am working towards this. Indeed, my plan is to use the option `BUILD_TESTING` provided by `include(CTest)`, but I need to wait for a feature release (i.e. 5.7, which is likely going to be the next release) to be able to change these things so users are not surprised by scripts which stop working. I will also need to adjust the CI for it as well.

PR #2191, opened 2024-02-12
[Tests] Use `find_package(GTest REQUIRED)`

veprbl:
This enables building against non-cmake gtest

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,4 +1,5 @@
-include(GoogleTest)
+find_package(GTest REQUIRED)
+
 add_subdirectory( XrdCl )
 add_subdirectory(XrdHttpTests)
 


amadio:
The calls to `find_package(GTest)` are here:

https://github.com/xrootd/xrootd/blob/5b5a1f6957def2816b77ec773c7e1bfb3f1cfc5b/cmake/XRootDFindLibs.cmake#L73-L86

What you likely want is just `-DFORCE_ENABLED=1` when configuring. I am already working towards what is described in #2190, so please wait until next release and everything will be there. I will close this one as it's not really necessary, but please see my latest pull request at #2189 where part of this work is being done. When 5.7 is released, everything will be there. Thanks!

veprbl:
Thanks for the pointers. Glad to hear you are already on it. Unfortunately, `-DFORCE_ENABLED=TRUE` is not usable, as it requires "macaroons" (I assume that's not relevant for the tests). Probably, one could have `-DCMAKE_REQUIRE_FIND_PACKAGE_CppUnit=TRUE` and `-DCMAKE_REQUIRE_FIND_PACKAGE_GTest=TRUE` to at least fail early when tests won't be built.

amadio:
`FORCE_ENABLED` enforces that enabled options get built, otherwise they are disabled if dependencies are not found. If you want, you can disable macaroons and other options that get enabled by default when using `FORCE_ENABLED`. Tr to configure without options first, then use `ccmake` to change the configuration and reconfigure. The available options are here:

https://github.com/xrootd/xrootd/blob/5b5a1f6957def2816b77ec773c7e1bfb3f1cfc5b/cmake/XRootDDefaults.cmake#L15-L32

To disable macaroons, use `-DENABLE_MACAROONS=OFF`.

veprbl:
That works!

PR #2193, opened 2024-02-15
Fix BASE_REF/HEAD_REF for pull requests when publishing on CDash

amadio:
This pull request is for testing this functionality.


abh3 (approved):
I tried the test on snprintf() so I am surprised this fix works but maybe I was using an older compiler. Otherwise, OK.

PR #2194, opened 2024-02-15
[XrdTpcTPC] Packet marking - avoid infinite loop in the case the first packet marking handle created got removed after the corresponding socket got closed

### LINKED ISSUE #2192: [XrdHttpTPC] Packet marking beginPmarks infinite loop
It looks like curl does not guarantee that the sockets will be closed in the reverse order that they are opened.

Therefore, the following code can go in an infinite loop in the case the Packet marker handle initialized with the first connected socket is closed before the other sockets.

```cpp
// The first pmark handle was created, or not. Create the other pmark handles from the other connected sockets
    while(!mSocketInfos.empty()) {
      SocketInfo & sockInfo = mSocketInfos.front();
      if(mPmarkHandles[mInitialFD]){
        std::unique_ptr<XrdNetPMark::Handle> pmark(mPmark->Begin(*sockInfo.client.addrInfo, *mPmarkHandles[mInitialFD], nullptr));
        if(pmark) {
          mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),std::move(pmark));
          mSocketInfos.pop();
        } else {
          // We could not create the pmark handle from the socket, we break the loop, we will retry later on when
          // this function will be called again.
          break;
        }
      }
    }
```

I need to change the ` if(mPmarkHandles[mInitialFD]){` with a `map.find(...)`...

ccaffy:
Fixes #2192

### DIFF
--- a/src/XrdTpc/XrdTpcPMarkManager.cc
+++ b/src/XrdTpc/XrdTpcPMarkManager.cc
@@ -45,37 +45,35 @@ void PMarkManager::startTransfer(XrdHttpExtReq * req) {
 }
 
 void PMarkManager::beginPMarks() {
-  if(!mSocketInfos.empty() && mPmarkHandles.empty()) {
-    // Create the first pmark handle that will be used as a basis for the other handles
-    // if that handle cannot be created (mPmark->Begin() would return nullptr), then the packet marking will not work
-    // This base pmark handle will be placed at the beginning of the vector of pmark handles
+  if(mSocketInfos.empty()) {
+    return;
+  }
+
+  if(mPmarkHandles.empty()) {
+    // Create the first pmark handle
     std::stringstream ss;
     ss << "scitag.flow=" << mReq->mSciTag;
     SocketInfo & sockInfo = mSocketInfos.front();
-    mInitialFD = sockInfo.client.addrInfo->SockFD();
-    std::unique_ptr<XrdNetPMark::Handle> initialPmark(mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc"));
-    if(initialPmark) {
-      // It may happen that the socket attached to the file descriptor is not connected yet. If this is the case the initial
-      // Pmark will be nullptr...
-      mPmarkHandles.emplace(mInitialFD,std::move(initialPmark));
-      mSocketInfos.pop();
+    auto pmark = mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc");
+    if(!pmark) {
+      return;
     }
-  } else {
-    // The first pmark handle was created, or not. Create the other pmark handles from the other connected sockets
-    while(!mSocketInfos.empty()) {
-      SocketInfo & sockInfo = mSocketInfos.front();
-      if(mPmarkHandles[mInitialFD]){
-        std::unique_ptr<XrdNetPMark::Handle> pmark(mPmark->Begin(*sockInfo.client.addrInfo, *mPmarkHandles[mInitialFD], nullptr));
-        if(pmark) {
-          mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),std::move(pmark));
-          mSocketInfos.pop();
-        } else {
-          // We could not create the pmark handle from the socket, we break the loop, we will retry later on when
-          // this function will be called again.
-          break;
-        }
-      }
+    mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),std::unique_ptr<XrdNetPMark::Handle>(pmark));
+    mSocketInfos.pop();
+  }
+
+  auto pmarkHandleItor = mPmarkHandles.begin();
+  while(!mSocketInfos.empty()) {
+    SocketInfo & sockInfo = mSocketInfos.front();
+    auto pmark = mPmark->Begin(*sockInfo.client.addrInfo, *(pmarkHandleItor->second), nullptr);
+    if (!pmark) {
+      // The packet marking handle could not be created from the first handle, let's retry next time
+      break;
     }
+
+    int fd = sockInfo.client.addrInfo->SockFD();
+    mPmarkHandles.emplace(fd, std::move(std::unique_ptr<XrdNetPMark::Handle>(pmark)));
+    mSocketInfos.pop();
   }
 }
 
--- a/src/XrdTpc/XrdTpcPMarkManager.hh
+++ b/src/XrdTpc/XrdTpcPMarkManager.hh
@@ -107,8 +107,6 @@ private:
   bool mTransferWillStart;
   // The XrdHttpTPC request information
   XrdHttpExtReq * mReq;
-  // The file descriptor used to create the first packet marking handle
-  int mInitialFD = -1;
 };
 } // namespace XrdTpc
 


src/XrdTpc/XrdTpcPMarkManager.cc
```diff
@@ -46,34 +46,38 @@ void PMarkManager::startTransfer(XrdHttpExtReq * req) {
...
           mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),std::move(pmark));
           mSocketInfos.pop();
         } else {
-          // We could not create the pmark handle from the socket, we break the loop, we will retry later on when
+          // We could not create the pmark handle from the socket because it probably did not connect yet. We break the loop, we will retry later on when
           // this function will be called again.
           break;
         }
+      } else {
+        // There is currently no packet marking on which we can rely on to create the another one. Break this loop. We will try to create a new first packet marking handle
```
amadio, line 78: `s/the another/the other/`, otherwise, looks ok.

ccaffy: correct, will force-push

abh3 (commented):
I'm OK with this though there one may be able to simplify this by simply doing a manual connect in the CURL callback. That avoid encountering unconnected sockets.

ccaffy:
All good, please note that the connect function needs to take the `aInfo->addrlen` parameter instead of `sizeof(aInfo->addr)`. That explains the errno 22 I got previously...
```
 // Connect the socket and register it to the packet marking manager
    if(connect(fd, (struct sockaddr*)&(aInfo->addr), aInfo->addrlen) ==
       -1) {
      return CURL_SOCKET_BAD;
    }
```

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -95,16 +95,20 @@ void CurlDeleter::operator()(CURL *curl)
...
+    /*int oldFlags = fcntl(curlfd,F_GETFD,0);
     if(oldFlags < 0) {
         return CURL_SOCKOPT_ERROR;
     }
     oldFlags |= FD_CLOEXEC;
     if(!fcntl(curlfd,F_SETFD,oldFlags)) {
         return CURL_SOCKOPT_OK;
     }
-    return CURL_SOCKOPT_ERROR;
+    return CURL_SOCKOPT_ERROR;*/
```
amadio, line 111: Why leave commented out code in? I think it's ok to remove and refer to the git log if necessary.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -132,7 +136,11 @@ int TPCHandler::opensocket_callback(void *clientp,
   {XrdNetAddr thePeer(&(aInfo->addr));
     rec->isIPv6 =  (thePeer.isIPType(XrdNetAddrInfo::IPv6)
                     && !thePeer.isMapped());
-    // Register the socket to the packet marking manager
+    // Connect the socket and register it to the packet marking manager
+    if(connect(fd, (struct sockaddr*)&(aInfo->addr), aInfo->addrlen) ==
+       -1) {
```
amadio, line 141: The `-1) {` could go on the same line above, not worth breaking up the line for this, but this is fairly minor, I'm fine if it stays like this.

src/XrdTpc/XrdTpcTPC.cc
abh3: I would say please remove it, it's confusing the way it is.

src/XrdTpc/XrdTpcPMarkManager.cc
```diff
@@ -46,34 +46,38 @@ void PMarkManager::startTransfer(XrdHttpExtReq * req) {
...
-    // if that handle cannot be created (mPmark->Begin() would return nullptr), then the packet marking will not work
-    // This base pmark handle will be placed at the beginning of the vector of pmark handles
+    // Create the first pmark handle
     std::stringstream ss;
     ss << "scitag.flow=" << mReq->mSciTag;
     SocketInfo & sockInfo = mSocketInfos.front();
-    mInitialFD = sockInfo.client.addrInfo->SockFD();
     std::unique_ptr<XrdNetPMark::Handle> initialPmark(mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc"));
     if(initialPmark) {
       // It may happen that the socket attached to the file descriptor is not connected yet. If this is the case the initial
```
abh3, line 55: I don't see how this could happen as the socket is either connected in the call back or it's flagged as dead. When flagged as dead this code path shouldn't be activated should it? If it is, then the socket will never be connected and I suppose you can just simply ignore packet marking for it. In other words, this code seems superfluous with the change to connect the socket.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -95,16 +95,20 @@ void CurlDeleter::operator()(CURL *curl)
  *       was needed for monitoring to report what IP protocol was being used.
  *       It has been kept in case we will need this callback in the future.
  */
-int TPCHandler::sockopt_setcloexec_callback(void *clientp, curl_socket_t curlfd, curlsocktype purpose) {
-    int oldFlags = fcntl(curlfd,F_GETFD,0);
+int TPCHandler::sockopt_callback(void *clientp, curl_socket_t curlfd, curlsocktype purpose) {
```
abh3, line 98: It would seem t me that the safest route here is that the original purpose of the code (i.e. to flag the socket with CLO_EXEC) should be maintained  and done whether or not the socket is connected. The addition of the forced connect and subsequent registration should only occur if packet marking if enabled.

abh3 (changes requested):

src/XrdTpc/XrdTpcPMarkManager.cc
```diff
@@ -46,34 +46,38 @@ void PMarkManager::startTransfer(XrdHttpExtReq * req) {
...
           // this function will be called again.
           break;
         }
+      } else {
+        // There is currently no packet marking on which we can rely on to create the other one. Break this loop. We will try to create a new first packet marking handle
+        // from an available fd in the next round when this function will be called again.
+        break;
       }
     }
   }
```
amadio, line 83: My suggestion for this function is something like this:
```cpp
void PMarkManager::beginPMarks() {
  if (mSocketInfos.empty())
    return;

  if (mPmarkHandles.empty()) {
    std::stringstream ss;
    ss << "scitag.flow=" << mReq->mSciTag;
    SocketInfo & sockInfo = mSocketInfos.front();
    auto pmark = mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc");

    if (!pmark)
      return; /* could not create initial pmark handle */

    int fd = sockInfo.client.addrInfo->SockFD();
    mPmarkHandles.emplace(fd, std::move(std::make_unique<XrdNetPMark::Handle>(pmark)));
    mSocketInfos.pop();
  } 

  while (!mSocketInfos.empty()) {
      SocketInfo & sockInfo = mSocketInfos.front();
      auto pmark = mPmark->Begin(*sockInfo.client.addrInfo, *(pmarkHandleItor->second), nullptr);

      if (!pmark)
        break;

      int fd = sockInfo.client.addrInfo->SockFD();
      mPmarkHandles.emplace(fd, std::move(std::make_unique<XrdNetPMark::Handle>(pmark)));
      mSocketInfos.pop();
    } 
  } 
}
```
What do you think?

src/XrdTpc/XrdTpcTPC.cc
bbockelm: Note this turns the callback from non-blocking to blocking without a timeout.

Blocking behavior is probably not so bad -- a separate event loop is run for each transfer so it's only blocking itself -- but the lack of timeout is problematic.

My $0.02: instead of making this work completely non-blocking (I've done this before in a different project: you basically have to fail the operation, create the connection in the event loop outside curl, then retry and pull from the curl cache.  It's ugly for a use case where not needed), simply add a timeout to the connection here.

src/XrdTpc/XrdTpcTPC.cc
ccaffy: The CLO_EXEC flag is set properly in the `opensocket_callback()` when one calls:
```cpp
//Return a socket file descriptor (note the clo_exec flag will be set).
  int fd = XrdSysFD_Socket(aInfo->family, aInfo->socktype, aInfo->protocol);
```

This sockopt_callback just need to tell curl that the socket got already connected.

src/XrdTpc/XrdTpcPMarkManager.cc
```diff
@@ -45,37 +45,34 @@ void PMarkManager::startTransfer(XrdHttpExtReq * req) {
...
-    std::unique_ptr<XrdNetPMark::Handle> initialPmark(mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc"));
-    if(initialPmark) {
-      // It may happen that the socket attached to the file descriptor is not connected yet. If this is the case the initial
-      // Pmark will be nullptr...
-      mPmarkHandles.emplace(mInitialFD,std::move(initialPmark));
+    auto pmark = mPmark->Begin(sockInfo.client, mReq->resource.c_str(), ss.str().c_str(), "http-tpc");
+    if(pmark) {
+      mPmarkHandles.emplace(sockInfo.client.addrInfo->SockFD(),std::unique_ptr<XrdNetPMark::Handle>(pmark));
       mSocketInfos.pop();
     }
```
amadio, line 61: You are missing a `return` here in case `pmark` creation fails for whatever reason, no?

PR #2205, opened 2024-02-29
Flexible SciTokens authorization strategy (v2)

bbockelm:
This is the follow-up to #2122.

On a per-issuer basis, decide if the mapping information is only to be done after authorization passes or if the mapping information should be kept and passed to the subsequent plugin.

Tests performed:
1.  Scitoken containing correct scopes is authorized in default setting.
2. Scitoken containing group information passes group information to the next plugin in default setting.
3. Scitoken that is mapped in mapfile passes in default setting to next plugin.
4. Scitoken that is mapped in mapfile but has no scopes is ignored when authorization_strategy is set to `capability`
5. Scitoken containing group information and incorrect scopes (no capability-based mapping) is passed to next plugin.

Note there's no config changes in the main xrootd config file but an addition of a configuration parameter in scitokens.cfg.  The default setting if nothing present is the current behavior.  To have the behavior requested by @esindril in #2121 (_only_ capabilities can provide authorization), the following issuer block would be used:

```
[Issuer DEMO]
issuer = https://demo.scitokens.org
base_path = /
restricted_path = /protected
default_user = brian
name_mapfile = /Users/bbockelm/.config/xrootd/mapfile
authorization_strategy = capability
```

Relevant documentation is in the README.md for the module.

Note that when it's "authorized" by either mapping or group, it's not really authorizing -- it's populating the mapped username / group information into the `XrdSecEntity` and passing the information along to the next plugin in the chain.  So, if you're looking to authorize via AuthDB (or, for EOS, in the OSS object), you'll be consuming the output of the plugin.

Fixes: #2121

### DIFF
--- a/src/XrdSciTokens/README.md
+++ b/src/XrdSciTokens/README.md
@@ -117,6 +117,17 @@ are:
       claim name.  If set, it overrides `map_subject` and `default_user`.
    -  `name_mapfile` (options): If set, then the referenced file is parsed as a JSON object and the specified mappings
       are applied to the username inside the XRootD framework.  See below for more information on the mapfile.
+   -  `authorization_strategy` (optional): One or more authorizations to use from the token.  Multiple (space separated)
+      items may be specified from the following valid values:
+
+         - `capability`: Authorize based on capabilities (e.g., `storage.read:/foo`) from the token.
+         - `group`: Pass through the request if there's any group present in the token.
+         - `mapping`: Pass through the request if the user mapping was successful.
+
+      For the `group` and `mapping` cases, the username and group are set in the internal XRootD request credential,
+      but the final authorization must be done by a subsequent plugin.  The default value is `capability group mapping`.
+      *Note*: if `mapping` is present, then a token without a capability may still have authorized actions.
+
 
 Group- and Scope-based authorization
 ------------------------------------
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -42,6 +42,13 @@ enum LogMask {
     All = 0xff
 };
 
+enum IssuerAuthz {
+    Capability = 0x01,
+    Group = 0x02,
+    Mapping = 0x04,
+    Default = 0x07
+};
+
 std::string LogMaskToString(int mask) {
     if (mask == LogMask::All) {return "all";}
 
@@ -272,10 +279,12 @@ struct IssuerConfig
                  const std::vector<std::string> &base_paths,
                  const std::vector<std::string> &restricted_paths,
                  bool map_subject,
+                 uint32_t authz_strategy,
                  const std::string &default_user,
                  const std::string &username_claim,
                  const std::vector<MapRule> rules)
         : m_map_subject(map_subject || !username_claim.empty()),
+          m_authz_strategy(authz_strategy),
           m_name(issuer_name),
           m_url(issuer_url),
           m_default_user(default_user),
@@ -286,6 +295,7 @@ struct IssuerConfig
     {}
 
     const bool m_map_subject;
+    const uint32_t m_authz_strategy;
     const std::string m_name;
     const std::string m_url;
     const std::string m_default_user;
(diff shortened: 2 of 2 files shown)


esindril:
Hi @bbockelm ,

Thank you reviving this topic and providing the above patch for addressing the CMS issue disccused in #2121.
However, I think this will still not fix the initial problem, at least when it comes to EOS since we rely on the mapping (the mapped username) for internal authz inside EOS. Therefore, if I read this code right https://github.com/xrootd/xrootd/pull/2205/files#diff-8cd57d0057f83be94fb75362e5c1c31c8592a2b6ef6403192d00e28837bc001dR618 it might happen that `Capability` authz strategy fails (`scope_sucess=false` i.e what CMS wants) but the `Mapping` authz strategy suceeds (`mapping_success=true` i.e. what EOS requires) and in the end the entire request will succeed - not what CMS wants!

Could you add another config option to enforce both strategies at the same time if specified rather then just one of them? I guess that would solve this particular item.

Thanks,
Elvin

bbockelm:
@esindril - I think these lines earlier in the function prevent what you're talking about:

```
        mapping_success = (access_rules->get_authz_strategy() & IssuerAuthz::Mapping) && !username.empty();
        scope_success = (access_rules->get_authz_strategy() & IssuerAuthz::Capability) && access_rules->apply(oper, path);
```

So, if you set the CMS issuer to _only_ honor capabilities, then `mapping_success` is always going to be false.  So, when the capability-based authorization fails, it should still have its default "on missing" behavior but not inject the results of the mapping (which will cause EOS to fail the request, right?).

Or am I not understanding your question?

esindril:
In EOS we always rely on the mapping as we need an identity when authorizing any type of transfer. Therefore, the `mapping_success` is a prerequisite for any successful transfer. This means that if `scope_success` failed then the transfer would still work - breaking the CMS expectations.

bbockelm:
> I was under the impression that mapping will not be performed at all if the authz strategy does not request it.

Nope -- what this controls is whether "mapping success" should be treated as an "authorization" to continue with the mapped name (as is done now and what CMS doesn't want).

> Then things should indeed be fine. I will test this out in our pre-prod instance. Thanks!

I believe your use case is precisely what I tested in item (4) above.  However, it's always possible I missed something important to EOS so am looking forward to the results of your testing.

esindril:
Hi Brian,

I finally got to test this patch and everything works as expected. Just for reference I put below the config that I used and the results:
```
cat /etc/xrootd/scitokens.cfg
[Global]
audience = https://wlcg.cern.ch/jwt/v1/any,https://elvin-dev01.cern.ch

[Issuer WLCG]
issuer = https://wlcg.cloud.cnaf.infn.it/
base_path = /,/eos/dev
map_subject = False
default_user = nobody
name_mapfile = /etc/xrootd/scitokens_mapfile.json
authorization_strategy = capability
```

Then first test with a token with the correct capabilities:
```
oidc-token -s storage.read:/eos/dev/ WLCG-esindril

echo $TOKEN | cut -d. -f2 | base64 --decode | jq . | grep scope
  "scope": "storage.read:/eos/dev/",

curl -L -H "Authorization: Bearer $TOKEN" https://elvin-dev01.cern.ch:9000/eos/dev/replica/file1.dat -o /var/tmp/dump
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  1919  100  1919    0     0   2244      0 --:--:-- --:--:-- --:--:--  2244
[esindril@esdss000 eos]$ echo $?
0
```

And with a token without any `storage` capability:
```
oidc-token -s email WLCG-esindril

 echo $TOKEN | cut -d. -f2 | base64 --decode | jq . | grep scope
  "scope": "email",

curl -v -L -H "Authorization: Bearer $TOKEN" https://elvin-dev01.cern.ch:9000/eos/dev/replica/file1.dat -o /var/tmp/dump

< HTTP/1.1 403 FORBIDDEN
< Connection: Keep-Alive
< Server: XrootD/v5.7-rc20240319
< Content-Length: 3846
< Content-Type: text/html
< Date: Tue, 19 Mar 2024 16:54:51 GMT
< 
```

I get the same result also when the capability does not match the type of operation, for example `storage.create:/eos/dev/` and we try a read operation.

With this new functionality, I think we can easily accommodate the CMS use case.
Thanks a lot!

Cheers,
Elvin

amadio:
I will rebase this and merge, no need to do anything on your side, @bbockelm. Thanks a lot for the contribution!

PR #2206, opened 2024-03-01
Add g-stream monitoring for IO time, based on the throttle plugin

bbockelm:
Given the throttle plugin is collecting timing statistics for I/O anyway, have it begin to report these monitoring statistics out through the g-stream.

This allows the monitoring programs to determine the overall time spent in the I/O subsystem (which is not available through other monitoring streams).  From the first derivative of the aggregate (most monitoring systems will do this for you), one can calculate an approximation of the "I/O load".


src/XrdXrootd/XrdXrootdConfigMon.cc
```diff
@@ -478,7 +482,7 @@ int XrdXrootdProtocol::xmongs(XrdOucStream &Config)
 
    int numgs = sizeof(gsObj)/sizeof(struct XrdXrootdGSReal::GSParms);
    int selAll = XROOTD_MON_CCM | XROOTD_MON_PFC | XROOTD_MON_TCPMO
-              | XROOTD_MON_TPC;
+              | XROOTD_MON_THROT | XROOTD_MON_TPC;
```
abh3, line 485: I see you nicely added throttle to the comments for XrdXrootdProtocol::xmon, could you do the same for XrdXrootdProtocol::xmongs? It's a one liner.

abh3 (changes requested):
Looks good but a minor nit pick about a missing comment. I know it's a trivial change but I'd like to be current with comments as much as possible.

src/XrdXrootd/XrdXrootdConfigMon.cc
bbockelm: Done in the latest rebase.

amadio:
@bbockelm You have to pull the git tags and push at least one of them to your fork, or the CI will keep failing in your repository.

abh3:
Thanks Brian!

PR #2208, opened 2024-03-03
Fix argument parsing in pyxrootd's File::Stat

chrisburr:
On macOS I've noticed that using `File.stat` results in SIGABORT. Linux doesn't seem to be affected.

After debugging I found that the arugment parsing seems to be broken, with an `int` being placed into a `bool`.

Reproducer:

```python
from XRootD import client

f = client.File()
status, _ = f.open(
    "root://eospublic.cern.ch//eos/opendata/lhcb/Collision12/CHARM/LHCb_2012_Beam4000GeV_VeloClosed_MagDown_RealData_Reco14_Stripping21_CHARM_MDST/00041836/0000/00041836_00009718_1.charm.mdst")
assert status.ok
print(f.stat())
```

Using 5.6.6:

```bash
$ python test.py
fish: Job 1, 'python test.py' terminated by signal SIGABRT (Abort)
```

Patching 5.6.6 with this PR:
```bash
$ python test.py
(<status: 0, code: 0, errno: 0, message: '[SUCCESS] ', shellcode: 0, error: False, fatal: False, ok: True>, <id: '7071570313412045725', size: 4038643315, flags: 48, modtime: 1698682469, modtimestr: '2023-10-30 16:14:29'>)
```

### DIFF
--- a/bindings/python/src/PyXRootDFile.cc
+++ b/bindings/python/src/PyXRootDFile.cc
@@ -106,14 +106,14 @@ namespace PyXRootD
   PyObject* File::Stat( File *self, PyObject *args, PyObject *kwds )
   {
     static const char  *kwlist[] = { "force", "timeout", "callback", NULL };
-    bool                force    = false;
+    int                 force    = 0;
     uint16_t            timeout  = 0;
     PyObject           *callback = NULL, *pyresponse = NULL, *pystatus = NULL;
     XrdCl::XRootDStatus status;
 
     if ( !self->file->IsOpen() ) return FileClosedError();
 
-    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "|iHO:stat", (char**) kwlist,
+    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "|pHO:stat", (char**) kwlist,
         &force, &timeout, &callback ) ) return NULL;
 
     if ( callback && callback != Py_None ) {


amadio:
Thank you, good timing, I'm working on getting the python bindings tests converted to Python 3 and adding them to the CI. I will test this and merge it for 5.6.9.

bindings/python/src/PyXRootDFile.cc
```diff
@@ -106,14 +106,14 @@ namespace PyXRootD
...
-    bool                force    = false;
+    int                 force    = 0;
     uint16_t            timeout  = 0;
     PyObject           *callback = NULL, *pyresponse = NULL, *pystatus = NULL;
     XrdCl::XRootDStatus status;
 
     if ( !self->file->IsOpen() ) return FileClosedError();
 
-    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "|iHO:stat", (char**) kwlist,
+    if ( !PyArg_ParseTupleAndKeywords( args, kwds, "|pHO:stat", (char**) kwlist,
```
chrisburr, line 116: I think p here is python 3.3+, if you still need python 2 then it probably needs to be kept as i and checked for truthiness on the python side.

amadio: Ok, thanks, I will check it when testing.

amadio:
I tried with Python 3.12 and XRootD 5.6.8 on an old Mac, and I get this:
```
>>> f.stat()
(<status: 0, code: 0, errno: 0, message: '[SUCCESS] ', shellcode: 0, error: False, fatal: False, ok: True>, <id: '7071570312732487632', size: 4038643315, flags: 48, modtime: 1698682470, modtimestr: '2023-10-30 16:14:30'>)
>>> print(XRootD.client.__version__)
5.6.8
```
So no `SIGABRT`, what macOS version (and Python version) are you using?

chrisburr:
Python 3.12, XRootD 5.6.6, compiled with clang 16 from conda-forge. I'll look at updating to 5.8.8: https://github.com/conda-forge/xrootd-feedstock/pull/87

```
cburr@ChristohersMBP8 ~> ipython
Python 3.12.2 | packaged by conda-forge | (main, Feb 16 2024, 20:54:21) [Clang 16.0.6 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.22.2 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from XRootD import client
   ...:
   ...: f = client.File()
   ...: status, _ = f.open(
   ...:     "root://eospublic.cern.ch//eos/opendata/lhcb/Collision12/CHARM/LHCb_2012_Beam4000GeV_VeloClosed_MagDown_RealData_Reco14_Stripping21_CHARM_MDST/00041836/0000/00041836_00009718_1.charm.mdst")
   ...: assert status.ok
   ...: f.stat()
fish: Job 1, 'ipython' terminated by signal SIGABRT (Abort)
cburr@ChristohersMBP8 ~ [SIGABRT]> sw_vers
ProductName:		macOS
ProductVersion:		14.3.1
BuildVersion:		23D60
cburr@ChristohersMBP8 ~> python --version                                                                                                                                                                                      (fsspec-xrootd-dev-2)
Python 3.12.2
cburr@ChristohersMBP8 ~> cat ~/mambaforge/pkgs/xrootd-5.6.6-py312hbeacd42_0/info/recipe/meta.yaml | rg clang
    - clang 16.0.6 hac416ee_4
    - clang-16 16.0.6 default_h6b1ee41_4
    - clang_impl_osx-arm64 16.0.6 ha81d56d_8
    - clang_osx-arm64 16.0.6 h40be2b1_8
    - clangxx 16.0.6 default_h6b1ee41_4
    - clangxx_impl_osx-arm64 16.0.6 h4a79b72_8
    - clangxx_osx-arm64 16.0.6 h40be2b1_8
    - libclang-cpp16 16.0.6 default_h6b1ee41_4
```

chrisburr:
Actually maybe the compiler flags are the difference:
```
2024-02-06T11:03:07.0158200Z +CXXFLAGS=-ftree-vectorize -fPIC -fstack-protector-strong -O2 -pipe -stdlib=libc++ -fvisibility-inlines-hidden -fmessage-length=0 -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/xrootd-5.6.7 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix
```

amadio:
There's nothing in recent versions that would change this, my guess is that `-fstack-protector-strong` may be onto something. I will try with that. I'd be interested in seeing what happens if you just change `bool` to `int`, but keep `iHO` in the call to `PyArg_ParseTupleAndKeywords`. We need to keep support for Python 2.7 around at least while  CentOS 7 is still around, unfortunately.

amadio:
Indeed, with the new compile flags, I can reproduce the issue on the Mac.

bindings/python/src/PyXRootDFile.cc
amadio: I think we should only change `bool` to `int` above, while leaving `iHO` as is to not lose compatibility with earlier versions of Python. I tested it and it works. Once this change is made we can merge it. Thanks!

PR #2213, opened 2024-03-16
Fixes for 64 bit time_t on 32 bit systems

ellert:
Debian has enabled 64 bit time_t for most of their 32 bit architectures. This PR fixes some issues to support this.


amadio:
Thank you, @ellert! The changes look alright, I'm merging.

PR #2215, opened 2024-03-18
Fix rsscmd support

cal-jlab:
Here are the two small changes I needed to make to get rsscmd support working.  I tested this in v5.6.8.

### DIFF
--- a/src/XrdOss/XrdOssStat.cc
+++ b/src/XrdOss/XrdOssStat.cc
@@ -123,7 +123,6 @@ int XrdOssSys::Stat(const char *path, struct stat *buff, int opts,
 //
    if ((retc = MSS_Stat(remote_path, buff))) return retc;
    if (popts & XRDEXP_NOTRW) buff->st_mode &= ro_Mode;
-   buff->st_mode |= S_IFBLK;
    return XrdOssOK;
 }
 
--- a/src/XrdOss/XrdOssApi.cc
+++ b/src/XrdOss/XrdOssApi.cc
@@ -609,7 +609,7 @@ int XrdOssDir::Readdir(char *buff, int blen)
 
 // Simulate the read operation, if need be.
 //
-   if (noDread)
+   if (dOpts & noDread)
       {if (ateof) *buff = '\0';
           else   {*buff = '.'; ateof = true;}
        return XrdOssOK;


amadio:
Could you please describe what problem exactly is being fixed here (in the commit message), and how/why the change you made fixes it? This is important for git archeology later. Thank you! For example, see commits 06470b01e24601186db374a7e79f9c1115831d64, 3a0e2ac4da8a77a98c3255ebbe755f7eb8fe6058, 3a2fd5350b1e648df687cd5120f9a6987322df5a.

cal-jlab:
I'm sorry: I'm not particularly good with git.  Can I just type it here?

[XrdOSS] Fix rsscmd support
Fixed two logic errors that were preventing the oss.rsscmd directive from
working properly.  (1) In XrdOssDir::Readdir() a test was made against the constant
noDread, but it needed to be made against a logical and of that constant with
the dOpts variable. (2) In XrdOssSys::Stat() the mode S_IFBLK was being set
indiscriminately for every path, causing directories to appear as files.

amadio:
Thanks, I can pick up from above, but if you want to do it, it's quite simple. Just run `git commit --amend` when you have the commit already checked out. This will allow you to rewrite the commit message. Then you just have to save and do a `git push -f` to your repository after the change and the pull request will be updated automatically.

cal-jlab:
I thought I had done the "commit --amend -F <my message>", and pushed to my repo, but I don't think it worked quite right.  If you'd just take the bit I typed above, that'd be great.  Thanks :-)

amadio:
No problem, I can make the changes.

src/XrdOss/XrdOssStat.cc
```diff
@@ -123,7 +123,7 @@ int XrdOssSys::Stat(const char *path, struct stat *buff, int opts,
 //
    if ((retc = MSS_Stat(remote_path, buff))) return retc;
    if (popts & XRDEXP_NOTRW) buff->st_mode &= ro_Mode;
-   buff->st_mode |= S_IFBLK;
+   // buff->st_mode |= S_IFBLK;
```
abh3, line 126: Frankly, this should be removed. If you want to leave it in, a comment would be necessary on why this is commented out. I agree that the flag should not have been set and full removal avoids provoking questions.

abh3 (changes requested):
Thanks a lot for these changes! The small change I am requesting is to avoid questions later on and the commented out line should be fully removed.

amadio:
@cal-jlab To make things easier, I will fixup the commit and force-push to update the pull request. If it doesn't work, I will just merge the change locally. You don't have to do anything else. Thanks for the contribution!

amadio:
I've split the changes in two commits to be clearer. I kept @cal-jlab as the author. Once the CI passes, I will merge this. Thank you!

src/XrdOss/XrdOssStat.cc
amadio: I've removed it.

amadio:
I'm merging this into the devel branch. Thank you!

cal-jlab:
Thanks :-)

PR #2216, opened 2024-03-19
[XrdTpc] Force HTTP 1.1 for the TPC transfers otherwise transfers fai…

esindril:
…l for Alma 9

In Alma 9.3 the default curl version is 7.76.1 and according to the documentation [1], starting with 7.62.0 the default HTTP protocol used is HTTP 2 TLS. This leads to the following error whenever a HTTP TPC trasfers is attempted: HTTP library failure: Stream error in the HTTP/2 framing layer

This patch will force curl to use HTTP 1.1 without even attempting HTTP 2. This is a workaround for the currently available curl version since normaly curl should be able to gracefully downgrade from HTTP 2 to 1.1 by itself.

[1] https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html


esindril:
Done.

esindril:
Sorry, I merged by mistake. I've updated the commit message and I let you decide when to add it. Thanks!

amadio:
The new version looks fine, I guess you meant you closed by mistake. I will merge this into `devel`.

PR #2223, opened 2024-03-24
Add ability to generate a new token on redirect

### LINKED ISSUE #2170: Generate new token on redirect
Based on email discussions with @Jo-stfc and others, there was a suggestion to add a mode where a limited token is generated on redirect, replacing any original authorization.

To some extent, we already do that, here:

https://github.com/xrootd/xrootd/blob/master/src/XrdMacaroons/XrdMacaroonsHandler.cc#L466

That function takes the existing permissions from the XrdSecEntity and the authorization subsystem (appending any additional limitations requested by the client) and generates a corresponding token.

Now, it would be straightforward - but somewhat unpleasant - to lift this code directly into the HTTP codebase and add in various #ifdef's if XRootD is built without libmacaroons.  That's effectively what's done with the "https2http" code, [here](https://github.com/xrootd/xrootd/blob/master/src/XrdHttp/XrdHttpUtils.cc#L219).  The "https2http" code, however, doesn't handle capability-based authorization so isn't really applicable to the modern configuration (I think it also has some potential versioning problems - it would take a bit to revive it).  By generating a macaroon, it's quite likely the approach would "just work" when enabled because macaroon-based verification is already needed for FTS transfers to work.

Another option would be to build a plugin mechanism for this functionality.  However, this plugin would be fairly narrow in scope and perhaps single-use: it's hard to see others trying to customize "internal token generation on redirects".

@abh3 - thoughts?  Writing it all out, I guess I'm leaning toward just embedding the macaroon logic into XrdHttp.

bbockelm:
If enabled with the `http.redirtoken` option, this PR will create a macaroon containing equivalent authorizations to the existing token understood by the SciToken library and then substitute the SciToken for the macaroon upon redirect.  The macaroon has the same authorizations and mapping _plus_ it's limited to 10 seconds in duration and only for the path specified in the URL.

If there's a token in the URL - but the SciToken library doesn't authorize its access to the redirected URL - then no change to the redirected URL.

Some of the interactions to find the new redirection object provided by the SciTokens library were taken from what was done previously for ZTN.

Fixes #2170


amadio:
@bbockelm Please fetch the git tags from the main repository and push them to your fork. Then you can see the result of the CI there as well. The build failures can also be seen in the "Pull Requests" section in CDash here: https://my.cdash.org/index.php?project=XRootD

abh3:
@bbockelm Where are we on this. It seems this is stalled by a repo synchronization issue and needs to be trivially resolved. See comment from @amadio above.

bbockelm:
Ok -- rebased, squashed some bugs and other commit noise together.  I think this is ready to go / re-review.

@abh3 asked for some notes to add to the documentation.  Here's what I'd mention:

> `http.redirtoken` 
> 
> When set (and Macaroons are otherwise configured), `http.redirtoken` will examine the redirected URL for the
> presence of SciToken (or any other token supported by the SciTokens plugin, such as a WLCG token) and,
> if found, replace it with a macaroon authorizing the redirected request.
> 
> The generated macaroon will only be useful at the same site as the redirector, be short-lived (about a minute),
> and only be valid for the path being requested.  Hence, it is typically far less powerful than the original token
> in the URL.

src/XrdSciTokens/XrdSciTokensRedir.hh
```diff
@@ -0,0 +1,43 @@
+#pragma once
```
amadio, line 1: @abh3 @bbockelm There are only 4 headers currently using this. For consistency, I think we should convert them to normal header guards (as well as this one). What do you think?

src/XrdSciTokens.cmake
```diff
@@ -31,6 +33,22 @@ target_include_directories(
...
+
+   target_compile_definitions(${LIB_XRD_SCITOKENS} PRIVATE HAVE_MACAROONS)
+
+   target_link_libraries(
+       ${LIB_XRD_SCITOKENS}
+       PRIVATE
+       uuid::uuid
+       OpenSSL::Crypto
+       ${MACAROONS_LIB} )
+endif()
```
amadio, line 50: I'm not a big fan of the mixing up between scitokens and macaroons introduced here. Is there no way to do this in a way that doesn't require such invasive changes?

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -463,9 +497,10 @@ class XrdAccSciTokens;
 
 XrdAccSciTokens *accSciTokens = nullptr;
 XrdSciTokensHelper *SciTokensHelper = nullptr;
+XrdSciTokensRedir *SciTokensRedir = nullptr;
 
 class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
-                        public XrdSciTokensMon
+                        public XrdSciTokensMon, public XrdSciTokensRedir
```
amadio, line 503: This changes the class layout, so as it is it can only go into a major version. @abh3 What do you think?

amadio (changes requested):
As it is this patchset seems a bit too dangerous to merge. I think it would be better to introduce integration tests for the functionality being touched here (scitokens and macaroons) before adding such a big change, as it probably won't be much less work to test these things by hand to make sure everything still works as expected. My suggestion is to wait and add this into 6.0 to avoid introducing any regressions in a minor/patch release.

amadio:
You may also look in [CDash](https://my.cdash.org/index.php?project=XRootD&date=2024-05-31), where some new warnings are reported for this pull request.

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: This is the layout of a class in a plugin.  It doesn't affect the ABI of the plugin interface itself.

(This same trick was done with `XrdSciTokensHelper` during the middle of the 5.x series)

src/XrdSciTokens.cmake
abh3: Maybe I can come up with a more directed approach for this. I agree that given the scale of change here we really need to run regression tests before we let this out.

src/XrdSciTokens/XrdSciTokensAccess.cc
abh3: I believe that technically that is true as this relies on GCC behavior, I believe. So, it may not work for other compilers. However, I would check on that. It may not make much of a difference though as it seems were now running the clock and likely won't be ready for releasing this until 6.0 anyway.

src/XrdSciTokens/XrdSciTokensRedir.hh
abh3: I would agree.

src/XrdSciTokens/XrdSciTokensAccess.cc
bbockelm: I don't understand the comment.

This is a plugin, not an interface.

How could a change in a plugin's implementation cause an ABI/API compatibility break in the interface it implements?

abh3:
I think the issue here is that technically changing inheritance (here, adding a new base class) *may* break ABI because it *may* change the derived class layout and vtable. Of course, that depends on the compiler but the recommendation is to not depend on what a particular compiler does. So, yes, this is a plugin but there are users of this plugin and if the memory layout changes then users of the plugin may no longer be compatible with the plugin without a recompilation. Of course, the claim may be that the only user is XRootD server code and that would be recompiled in any case. and that also appears to be the case. So, in that respect while this *may* break ABI it does not matter in this case. So, I think the problem stemmed from the fact that this is a plugin and it was not clear who was using the plugin.

bbockelm:
> may break ABI because it may change the derived class layout and vtable

The SciTokens plugin is not part of any exported header: in fact, this class is not in _any_ header (it's part of a compilation target).  How can a internal class, not part of any public ABI/API and not part of any header, cause ABI/API breaks?

amadio:
> How can a internal class, not part of any public ABI/API and not part of any header, cause ABI/API breaks?

Indeed, I was being too conservative, it should be ok. However, I think if we can create less entanglement between the two plugins, it would be better.

abh3:
I think my answer aligns with @amadio

bbockelm:
@amadio - you wrote:

> However, I think if we can create less entanglement between the two plugins, it would be better

To be clear, there isn't any entanglement between the two plugins.

The CMake snippet you originally commented on was having `libXrdSciTokens-5.so` link against `libmacaroons.so`.

I don't think there's really any practical way to have the SciToken library generate a macaroon without linking against `libmacaroons.so`.

xrootd-dev:
We did think of one other possibility and that is to move the Macaroon generating code to a plugin that depends on LibMacaroons.so (obviously). Then if the server is configured to return a macaroon the plugin is loaded, and its address is placed in the env passed to the libSciTokens plugin. Yes, somewhat complicated, especially w.r.t. the timeline which we have not worked out yet. However, we already have at least one such instance to prevent undue entanglement. There may be better sequences here surrounding the special macaroon generating plugin.

bbockelm:
@abh3 - any further information on your last comment?  My understanding is that this is waiting for your proposed plugin infrastructure that I can utilize to refactor the PR, right?

bbockelm:
Summary of in-person discussion:
- Brian will refactor the PR to move the token generation into libXrdMacaroons with a simple function call interface, directly linking to libXrdMacaroons
- @abh3 will take a look at the proposed interface and, after any refinements, will provide a plugin interface equivalent to the function call.
- Once the plugin interface is available, then Brian will refactor the PR again to use the plugin interface instead of direct linking.

bbockelm:
@abh3 -- I did the first step we discussed, refactoring out the token generation into a single interface.  It's now in `XrdMacaroonsGenerate.hh` and both the SciTokens and Macaroons plugin utilize it (this also allowed me to combine the configuration code as well).

Could you take a look at it and do the next step, which is to provide a plugin equivalent that the SciTokens library would eventually use?

abh3:
Thanks @bbockelm I will look at this soon.

abh3 (commented):
I'll leve this as a comment because we both have more work to so here. Please red the summary comment on the pull request along with the comment in the review.

abh3:
@bbockelm So, I am trying to avoid he classic chicken-egg problem. From what I see, if you have a pointer to the Macaroon plugin then you implicitly have access to the Macaroon generator via XrdMacaroonsHandler::m_generator, which is public. Is that correct? If so, it's a matter of loading the plugin when needed and passing the pointer to the XrdSciTokensRedir plugin via an added init() method. There is no reason for SciTokens to load the Macaroons plugin or use the Macaroons generator function unless http.redirtoken was specified (note there is no corresponding need in xroot protocol). I added a comment in the code for the place where that would happen. So, here is what I propose:

I will add a method to http protocol to load the macaroons plugin if it hasn't been loaded before. This may  happen because the exthandler requested that macaroons be loaded. This may be tricky because ext handlers are not explicitly identified by being any particular type. So, I may need to modify this once I get into it.

My only other question is what triggers a SciTokens initialization if SciTokens itself is not explicitly initialized. For instance, let's say I didn't specify SciTokens as an authlib but did specify http.redirtoken. This can happen if all I am doing is creating a redirector which does not necessarily need authorization but would like to substitute scitokens for macaroons. It doesn't look like his case has been covered.

bbockelm:
@abh3 - @amadio and I discussed this at length (I'm at CERN this week and we could draw on the board).  Here's what I think we converged on:

1.  Instead of doing this at the HTTP protocol level, we load the macaroons plugin (libAccMacaroons.so) inside libAccSciTokens.so when the SciTokensRedir object is created.  This way, macaroons is only loaded if used and we don't have any code shared between the two plugins.
2.  If you have a redirector that only wants to use scitokens translation - but not the authorization plugin for authorization - I don't think this is possible.  SciTokensRedir uses the ACC framework to determine the authorizations to include on the macaroon.  So, if the redirector permits the request then the macaroon will have the equivalent read capabilities.   To make this independent, we'd need a way to load a separate ACC independent of the server-wide authorization.

Jo-stfc:
Thanks Brian. We plan to test these changes soon at RAL, as the token leaks keeps popping up occasionally. 
I've tried to cobble up different approaches, but the best way forward keeps coming back to encoding the token into a macaroon.
I'll try to test this in preprod over next week and let you know how it goes

amadio:
I've pushed my suggested way to rebase this (squashing some history since they are bugfixes) to my fork. You can view it [here](https://github.com/xrootd/xrootd/compare/master...amadio:xrootd:refs/heads/redirtoken). That is mostly a plain rebase on top of the many CMake changes that I've done. @bbockelm please check that everything is ok, then we can move on with adapting it to what we have discussed with @abh3, which is to not add XrdMacaroons source files to XrdSciTokens, but to load the plugin at runtime and enable/use the functionality only if available.

Jo-stfc:
the xrootd-server package is failing to install. the error message looks like this:
```
  Running scriptlet: xrootd-server-1:5.8.2-1.el8.x86_64                                                             5/9
/var/tmp/rpm-tmp.ENPkcR: line 1: fg: no job control
error: %prein(xrootd-server-1:5.8.2-1.el8.x86_64) scriptlet failed, exit status 1

Error in PREIN scriptlet in rpm package xrootd-server
  Cleanup          : xrootd-server-libs-1:5.8.1-99.el8.x86_64                                                       6/9
error: xrootd-server-1:5.8.2-1.el8.x86_64: install failed
```

amadio:
> the xrootd-server package is failing to install. the error message looks like this:
> 
> ```
>   Running scriptlet: xrootd-server-1:5.8.2-1.el8.x86_64                                                             5/9
> /var/tmp/rpm-tmp.ENPkcR: line 1: fg: no job control
> error: %prein(xrootd-server-1:5.8.2-1.el8.x86_64) scriptlet failed, exit status 1
> 
> Error in PREIN scriptlet in rpm package xrootd-server
>   Cleanup          : xrootd-server-libs-1:5.8.1-99.el8.x86_64                                                       6/9
> error: xrootd-server-1:5.8.2-1.el8.x86_64: install failed
> ```
> 
> (this was from @amadio 's version)

Well, that doesn't have much to do with this pull request. I already know what the problem is ("fix" for #2433), you can work around by installing `epel-rpm-macros`, but since the new packages require that new dependency, I have already locally reverted the change to push with XRootD 5.8.2.

Jo-stfc:
Thanks. I've amended the build process and deployed it to a test gw, which passes the functional tests. I'll test it in prod after we upgrade to 5.8.1 first as there are more changes (such as python version) which I'd like to ensure are ok as well.

abh3:
@bbockelm @amadio OK, I get it. The authorization plugin is really not capable of adding/replacing CGI information. Also, we never send the authorization information to the cmsd so it can't do a substitution either unless you munge it up with HTTP. I may have a good solution for you but could you first rebase this pull request to get rid of the 100 extraneous file changes?

abh3:
Gees, this is such a long thread perhaps we should just have a zoom chat. However, I did come up with a solution that does not require any new plugins just some additional code in SciTokens. Basically, we distill the problem to the simple question "I want to substitute a scotoken with a macaroon, can you please do that SciTokenLib?" But how do we ask and how do we get the answer back? Well, it works out to be rather trivial.

1) Http adds the cgi token to the url: sci2mac=<key> (we may want to put the key into a namespace)
The key is an arbitrary key and can be fixed since HTTP is single threaded when it comes to requests relative to the socket. It also means that any particular http session will also be tied to a particular SecEntity object and that is the key here. Note this also accommodates multithreaded session use of sci2mac so no future issues here.

2) When the SciTokenLib gets control and verifies the token as valid and allows access, it checks if there is a sci2mac element in the cgi env passed to it. If there is, it generates a macaroon and adds it to the SceEntity attributes using the provided key.

3) Upon return from the request, if there is a redirect then HttpReq normally calls Redir() which would check if it should sub a macaroon for the scitoken in this request. If it was asked to do that, it gets the macaroon from the SecEntity object using the known attribute key and physically substitutes that for the scitoken authz CGI element. If it's nice, it deletes the key-value pair.

This is already doable and would work without any new plugins. I would suggest some performance improvements but does can be done later. 

Shall we zoom about this?

abh3:
Oh, here is another idea that limits damage to SciTokens library. Simply create a small sci2mac plugin that is stacked on top of the SciToken plugin. If the SciToken library says it's OK to access the resource the plugin, stacked on top of the SciToken plugin, can do the steps to convert the SciToken to a macaroon and set the SecEntity attribute. This localized the changes to a particular library and does no harm to SciTokens. In fact, the new plugin avoids the problem of  contaminating SciTokens with Macaroons because, well the new plugin is really a Macaroon plugin stacked on top of a pre-exiting SciTokens library. This means a much cleaner separation of code. All isci2mac really needs to know is that the token granted access using a public API. I sort of prefer this localized approach because it doesn't burden the SciTokens library with a feature that is only used in specific cases.

Jo-stfc:
Thanks @abh3. 
I've tested the change as it currently stands but the tokens were still being sent without being converted.
(I did add the http.redirtoken line in the config)

amadio:
@Jo-stfc Did you remove `http.header2cgi Authorization authz` from the configuration?

Jo-stfc:
removing it results in a permission denied error from the manager

Jo-stfc:
[lognoauthz.txt](https://github.com/user-attachments/files/20225503/lognoauthz.txt)
I've attached the resulting logfile at debug level

abh3:
So, I'm confused. How are you testing this as it should not have been merged anywhere as far as I can tell. The log does not provide enough information to tell me what is actually happening. Where is the redirector? Where is the server? What does the config look like? Anyway, I think this pull request will likely go through simplification before we merge it.

Jo-stfc:
I've merged @amadio s redirtoken branch into our fork https://github.com/stfc/xrootd/tree/v5.8.2patched and compiled from that.
ceph-dev-gw4 was used as a oidc client to transfer a local file (t6) to the test redirector (echo-internal-manager01) ceph-svc16 is the server behind the redirector.

Jo-stfc:
[manager-xrootd-unified.txt.txt](https://github.com/user-attachments/files/20240620/manager-xrootd-unified.txt.txt)
this is the xrootd config we use

amadio:
I am closing this due to all the merge conflicts with the current `master` branch. With XRootD 6.0 now out, the option `strip-on-redirect` for `http.header2cgi` is the preferred way of keeping things secure. A client like curl will then need to be used as `curl --location-trusted ...` with an XRootD cluster to re-send the credentials after a redirect, which then no longer requires a new token to replace the original one. If this is still needed, please rebase the changes on the current `master` branch and open a new pull request. Thank you.

abh3:
Actually, before creating a new pull requ8st based on this implementation, we did come up with a new design that fully separates the Macaroon library from the SciToken library and can be implemented using a simple stacked plugin. This is the implementation that we would accept as it's simple and tangle-free. If you need more details (which I think you will need) please simply ask and I'll provide them.

Issue #2228, opened 2024-03-28
[XrdTpcTPC] Add new option to allow  for unrestricted redirection


ccaffy:
As discussed in issue [2221](https://github.com/xrootd/xrootd/issues/2221) , we need to add a configuration allowing administrators to allow that the clients credentials are passed to the hosts towards which they are redirected.

We will add a new option `http.tpcallowunrestrictedredirect on/off` (the names are subject to change) that by default will be disabled.

If the option is enabled, we will set the option https://curl.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html to curl for HTTP TPC transfers: In `XrdTpcState.cc:75` for example.

ccaffy:
> > We will add a new option `http.tpcallowunrestrictedredirect on/off` (the names are subject to change) that by default will be disabled.
> 
> I hope we can find a shorter name for this :-) Let me suggest `http.tpc fcreds all/trusted`, similar to `ofs.tpc fcreds`. See https://xrootd.slac.stanford.edu/doc/dev56/ofs_config.htm#_Toc136617306

I like this option name, thanks @amadio !

ccaffy:
I'll do `http.tpc fcreds on/off` as it is clear at the first sight that if that option is `on`, the credentials will be forwarded :)

amadio:
You could also do it the same way as `ofs.tpc`, where `fcreds` already means `fcreds on`, and the default is off.

ccaffy:
Sure, yes, it will allow to simplify the parsing of this command...

amadio:
To make things clear, we normally use
```
ofs.tpc fcreds ?gsi =X509_USER_PROXY
```
to forward delegated credentials using gsi authentication. We could have
```
http.tpc fcreds ?ztn =BEARER_TOKEN
```
for forwarding tokens while reusing the same code for `ofs.tpc`.

ccaffy:
But that's not the goal of the issue right? we just want an option to pass the credentials to the redirected host whatever the authentication is, right?

amadio:
Yes, but if the code for parsing is reused, it creates more consistency between `ofs.tpc` and `http.tpc`, while also allowing to put the contents of the token in a file using 4ef0f2f388f877d0715ffc345afe1b988398749b. What do you think?

ccaffy:
The code is not reused at all, I just use the same parameters name...

PR #2232, opened 2024-04-02
[XrdTpcTPC] Add new option to allow for unrestricted redirection

ccaffy:
Solves issue #2228

Initial simple and basic PR. I don't know if we want to implement the same thing as in https://github.com/xrootd/xrootd/commit/4ef0f2f388f877d0715ffc345afe1b988398749b .

### DIFF
--- a/src/XrdTpc/XrdTpcConfigure.cc
+++ b/src/XrdTpc/XrdTpcConfigure.cc
@@ -73,6 +73,7 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
 	    }
 	} else if (!strcmp("tpc.timeout", val)) {
             if (!(val = Config.GetWord())) {
+                Config.Close();
                 m_log.Emsg("Config","tpc.timeout value not specified.");  return false;
             }
             if (XrdOuca2x::a2tm(m_log, "timeout value", val, &m_timeout, 0)) return false;
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -117,5 +117,7 @@ verb(req->requestverb), headers(req->allheaders) {
   pmark = prot->pmarkHandle;
   mSciTag = req->mScitag;
 
+  tpcForwardCreds = prot->tpcForwardCreds;
+
   length = req->length;
 }
--- a/src/XrdHttp/XrdHttpExtHandler.hh
+++ b/src/XrdHttp/XrdHttpExtHandler.hh
@@ -59,6 +59,8 @@ public:
 
   XrdNetPMark * pmark;
 
+  bool tpcForwardCreds = false;
+
   int mSciTag;
 
   // Get full client identifier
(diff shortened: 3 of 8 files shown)


src/XrdTpc/XrdTpcConfigure.cc
```diff
@@ -82,6 +83,14 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
             } else {
                 m_first_timeout = 2*m_timeout;
             }
+        } else if (!strcmp("http.tpc", val)) {
```
abh3, line 86: This really creates a sore thumb here. As we always try to enforce that each component deal with only one of two prefixes: the one that is exclusive to the component and "all.". Yes, I know, there are exceptions but let's try to squash them when we can. Since this is the TPC component the exclusive prefix is "tpc." the "http." prefix is exclusive to the http component parsed in XrdHttp. So, I would strongly suggest that this really be "tpc.http" at best (you might think of some other keyword). The addition of fccreds is fine. Now I could argue that this is not specific to TPC but any component that would use curl and hence http. So, I could also argue that this really belongs in XrdHttpProtocol::Config() and it becomes a global setting for all uses of curl. That said, I'd like to see @amadio opinion on that.

abh3 (changes requested):
A consistency comment that I think should be corrrcted.

src/XrdTpc/XrdTpcConfigure.cc
amadio: We have `ofs.tpc`, so I thought that `http.tpc` was fine, also because we already have `http.desthttps` being used in this same file. However, I agree with your arguments above. If we have `tpc.` as prefix, having many `*.tpc` options can become confusing. We could merge all authentication related options for TPC into a new `tpc.auth` or similar option that would work for both `http.tpc` and `ofs.tpc` (keeping the old style for backward compatibility). Or we could as you suggest create an `http.auth` or similar where we'd set this as `http.auth fcreds ztn`, for example. This option could then be extended to enable Kerberos authentication via [negotiate](https://everything.curl.dev/libcurl-http/auth.html#negotiate) method with something like `http.auth negotiate` if we don't already try that.

abh3:
This all comes down to consistency and long term vision. If we are trying to only solve the curl issue for TPC and be done with it, then "tpc.http' or more likely 'tpc.auth' would be the correct oath forward. If you want to have a general framework for addressing similar issues in the future and use that to address the TPC problem then something like 'http.auth' would be the correct path forward. To solve a specific component issue  like TPC we could do 'http.auth tpc fcreds'. So, it's really what you want to accomplish. Simply choose one or the other. However, we certainly don't want both (ie. tpc.auth and http.auth).

abh3:
Any progress on making a choice here?

ccaffy:
Hi @abh3 ,

Yes let's go for `http.auth tpc fcreds` to go for future issue adressing ;)
Will do it after pmark socket polling is done :)

Cheers,
Cedric

ccaffy:
Done in the commit https://github.com/xrootd/xrootd/pull/2232/commits/7d102759a5235528e232cce93a20be0f0941e13e 
Please do not merge before I squash

abh3:
Where are we here to move the http.auth to the section that processes http directives as opposed to stranding it in the tpc directive section?

ccaffy:
all good :)

PR #2234, opened 2024-04-05
[XrdSciTokens] Add fix for tokens with a path scope of "/"

jthiltges:
Tokens may include a storage scope for the root path
```
{
  "wlcg.ver": "1.0",
  "scope": "storage.read:/ openid email profile offline_access",
  "iss": "https://cms-auth.web.cern.ch/",
```

This is a corner that slips by the existing token validation logic
> Either the requested path and ACL path is the same OR the requested path is a subdir of the ACL path.

When the ACL path is `/` and is not an exact match for the request, the second character of the request path is compared to `/`, which will usually fail. Add an exception for the ACL path case of `/`.


amadio:
Thank you, but this is already fixed in `devel` by this commit: https://github.com/xrootd/xrootd/commit/4298cfa54847b1c57d94980ba1c056dd7e670f5f

Issue #2236, opened 2024-04-12
Client uses `uint16_t` for timeouts, which allows only up to 65535s (≈18h20min), while the server uses `int` for timeouts


amadio:
It's probably a good idea to use either `int` to match the server side or `uint32_t` to allow timeouts longer than 65535s. For example, setting a timeout to 24h (86400s) would wrap on the client side and give 86400 % 65536 == 20864s (≈5h48min) instead of 24h for the timeout. That means that if someone tries to copy a large file and sets XRD_CPTIMEOUT to what they may think is 24h, this wrapping may make it fail.

PR #2237, opened 2024-04-15
Use `time_t` for times and timeouts in the client

amadio:
Going via pull request to allow review, due to the many changes to the interfaces. Requires recompilation and maybe `s/uint16_t/time_t/` in code using the XRootD client, but should be very straightforward to migrate. I have a similar patch using `int` instead of `time_t` for timeouts (that's what the XRootD server uses). Decided to use `time_t` as it's more future proof. See also [`man time_t(3type)`](https://man7.org/linux/man-pages/man3/time_t.3type.html). 

Fixes: #2236

### DIFF
--- a/bindings/python/src/PyXRootDFile.cc
+++ b/bindings/python/src/PyXRootDFile.cc
@@ -43,7 +43,7 @@ namespace PyXRootD
     const  char            *url;
     XrdCl::OpenFlags::Flags flags    = XrdCl::OpenFlags::None;
     XrdCl::Access::Mode     mode     = XrdCl::Access::None;
-    uint16_t                timeout  = 0;
+    time_t                  timeout  = 0;
     PyObject               *callback = NULL, *pystatus = NULL;
     XrdCl::XRootDStatus     status;
 
@@ -75,7 +75,7 @@ namespace PyXRootD
   PyObject* File::Close( File *self, PyObject *args, PyObject *kwds )
   {
     static const char  *kwlist[] = { "timeout", "callback", NULL };
-    uint16_t            timeout  = 0;
+    time_t              timeout  = 0;
     PyObject           *callback = NULL, *pystatus = NULL;
     XrdCl::XRootDStatus status;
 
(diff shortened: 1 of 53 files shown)


src/XrdCl/XrdClCopyProcess.hh
```diff
@@ -144,9 +144,9 @@ namespace XrdCl
       //! chunkSize      [uint32_t] - size of a copy chunks in bytes
       //! parallelChunks [uint8_t]  - number of chunks that should be requested
       //!                             in parallel
-      //! initTimeout    [uint16_t] - time limit for successfull initialization
+      //! initTimeout    [int]      - time limit for successfull initialization
       //!                             of the copy job
-      //! tpcTimeout     [uint16_t] - time limit for the actual copy to finish
+      //! tpcTimeout     [int]      - time limit for the actual copy to finish
```
amadio, line 149: These are wrong (doing a self-review...). Need to update to `time_t`.

bindings/python/src/PyXRootDFile.cc
```diff
@@ -580,7 +580,7 @@ namespace PyXRootD
     if ( py_timeout && PyObjToUshrt( py_timeout, &tmp_timeout, "timeout" ) )
```
amadio, line 580: Need to also convert `PyObjToUshrt` to more appropriate function (reminder to self).

src/XrdCl/XrdClCopyProcess.hh
amadio: Done.

bindings/python/src/PyXRootDFile.cc
amadio: Done.

amadio:
@abh3 @apeters1971 @esindril Please let me know what you think. If there are no objections, I will push this into devel later to be part of XRootD 5.7.0. ROOT compiles just fine against XRootD with this patch. However, EOS will need changes. I will work on the migration on the EOS side so it's ready when XRootD is released.

abh3 (approved):
Technically, this is fine. Let's hope the recompilation of all user code will not be a real headache. Please run this by the root team.

amadio:
For ROOT it should be fine, I compiled the master branch against this, no changes are needed. I also used the ROOT compiled against this to run the [analysis grand challenge](https://github.com/root-project/analysis-grand-challenge) via XRootD. I noticed that ROOT 6.30.06 has a small issue with our CMake files (not related to this), but in the master branch of ROOT everything is fine. In any case, I will coordinate with Danilo on this.

PR #2242, opened 2024-04-19
[XrdTpcTPC] PMark - Connect curl socket at socket creation

ccaffy:
This allows to improve HTTP-TPC transfers packet marking precision. If the packet marking is enabled and the client submitted a request with a SciTag header, then the connection of the socket will be done within a timeout of 5 seconds. The TPC transfer will fail if the socket could not be connected.

Currently, the timeout is 5 seconds, hardcoded. I can put that in a variable for sure if needed. I have no clue of what a reasonable timeout for a connection could be. I'm open to suggestions :)

### DIFF
--- a/src/XrdNet/XrdNetUtils.cc
+++ b/src/XrdNet/XrdNetUtils.cc
@@ -37,6 +37,8 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <fcntl.h>
+#include <poll.h>
 
 #include "XrdNet/XrdNetAddr.hh"
 #include "XrdNet/XrdNetIdentity.hh"
@@ -937,4 +939,82 @@ bool XrdNetUtils::Singleton(const char  *hSpec, const char **eText)
 // All done
 //
    return isSingle;
+}
+
+bool XrdNetUtils::ConnectWithTimeout(int sockfd, const struct sockaddr* clientAddr,size_t clientAddrLen, uint32_t timeout_sec, std::stringstream &errMsg) {
+
+  if (!SetSockBlocking(sockfd, false, errMsg)) {  // Set to non-blocking
+    close(sockfd);
+    return false;
+  }
+
+  int result = connect(sockfd, clientAddr, clientAddrLen);
+  if (result == 0) {
+    //We've managed to connect immediately
+    // Set the socket back to blocking
+    if(!SetSockBlocking(sockfd, true, errMsg)) {
+      ::close(sockfd);
+      return false;
+    }
+    return true;
+  } else if (errno != EINPROGRESS) {
+    errMsg << "Connection failed: " << strerror(errno);
+    ::close(sockfd);
+    return false;
+  }
+
+  struct pollfd fds;
+  fds.fd = sockfd;
+  fds.events = POLLOUT;  // Wait for the socket to be ready to write
+
+  result = poll(&fds, 1, timeout_sec * 1000); //Timeout in milliseconds
+
+  if (result < 0) {
+    errMsg << "Poll error: " << strerror(errno);
+    ::close(sockfd);
+    return false;
+  } else if (result == 0) {
+    errMsg << "Connection timed out";
+    ::close(sockfd);
+    return false;
+  }
+  // Polling succeeded
+  if (!(fds.revents & POLLOUT)) {
+    // We should normally not reach this code
+    errMsg << "Poll returned without error but the corresponding socket (" << sockfd << ") is not ready to write";
+    ::close(sockfd);
+    return false;
+  }
+  // Check for potential errors on the socket after polling
+  int so_error;
+  socklen_t len = sizeof(so_error);
+  getsockopt(sockfd, SOL_SOCKET, SO_ERROR, &so_error, &len);
+  if (so_error != 0) {
+    errMsg << "Connection failed after poll: " << strerror(so_error);
+    ::close(sockfd);
+    return false;
+  }
+  // Everything succeeded, set the socket back to blocking
+  if(!SetSockBlocking(sockfd, true, errMsg)) {
+    ::close(sockfd);
+    return false;
+  }
+  return true;
+}
+
+bool XrdNetUtils::SetSockBlocking(int sockfd, bool blocking, std::stringstream & errMsg) {
+  int flags = fcntl(sockfd, F_GETFL, 0);
+  if (flags == -1) {
+    errMsg << "Failed to get socket flags " << strerror(errno);
+    return false;
+  }
+
+  flags = blocking ? (flags & ~O_NONBLOCK) : (flags | O_NONBLOCK);
+
+  if (fcntl(sockfd, F_SETFL, flags) == -1) {
+    errMsg << "Failed to set socket blocking/non-blocking " << strerror(errno);
+    return false;
+  }
+
+  return true;
 }
--- a/src/XrdTpc/XrdTpcPMarkManager.cc
+++ b/src/XrdTpc/XrdTpcPMarkManager.cc
@@ -22,6 +22,7 @@
 
 #include <sstream>
 #include "XrdTpcPMarkManager.hh"
+#include "XrdNet/XrdNetUtils.hh"
 
 namespace XrdTpc
 {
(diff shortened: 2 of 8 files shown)


src/XrdTpc/XrdTpcPMarkManager.hh
```diff
@@ -97,16 +109,26 @@ public:
...
+  void addFd(int fd, const struct sockaddr * sockP);
+
   // The queue of socket information from which we will create the packet marking handles
   std::queue<SocketInfo> mSocketInfos;
   // The map of socket FD and packet marking handles
   std::map<int,std::unique_ptr<XrdNetPMark::Handle>> mPmarkHandles;
   // The instance of the packet marking functionality
   XrdNetPMark * mPmark;
+  // The XrdHttpTPC request information
+  XrdHttpExtReq & mReq;
```
amadio, line 129: There is a potential for a crash here if the reference becomes stale. Is it guaranteed that it will outlive the `PMarkManager`?

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -95,16 +95,23 @@ void CurlDeleter::operator()(CURL *curl)
...
+    /*int oldFlags = fcntl(curlfd,F_GETFD,0);
     if(oldFlags < 0) {
         return CURL_SOCKOPT_ERROR;
     }
     oldFlags |= FD_CLOEXEC;
     if(!fcntl(curlfd,F_SETFD,oldFlags)) {
         return CURL_SOCKOPT_OK;
     }
-    return CURL_SOCKOPT_ERROR;
+    return CURL_SOCKOPT_ERROR;*/
```
amadio, line 114: Dead code should be removed.

src/XrdNet/XrdNetUtils.cc
```diff
@@ -938,3 +939,71 @@ bool XrdNetUtils::Singleton(const char  *hSpec, const char **eText)
...
+  }
+
+  fd_set set;
+  struct timeval timeout;
+  FD_ZERO(&set);
+  FD_SET(sockfd, &set);
+  timeout.tv_sec = timeout_sec;
+  timeout.tv_usec = 0;
+
+  result = select(sockfd + 1, NULL, &set, NULL, &timeout);
```
abh3, line 968: It's much better to use poll() for this and avoid all the bit vector mucking for a single fd. See XrdLinkXeq::Peek() for an example how that is done.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -132,8 +139,12 @@ int TPCHandler::opensocket_callback(void *clientp,
   {XrdNetAddr thePeer(&(aInfo->addr));
     rec->isIPv6 =  (thePeer.isIPType(XrdNetAddrInfo::IPv6)
                     && !thePeer.isMapped());
-    // Register the socket to the packet marking manager
-    rec->pmarkManager.addFd(fd,&aInfo->addr);
+    std::stringstream connectErrMsg;
+
+    if(!rec->pmarkManager.connect(fd,&(aInfo->addr),aInfo->addrlen,5,connectErrMsg)) {
```
abh3, line 144: You are correct, hardcoding this may become problematic. The Linux default is 60 seconds which some people claim is too long. It all depends on what you are trying to accomplish. So, yes, I would make this a configurable number. Note that as far as I can tell, curl will use the Linux default. So, it is somewhat strange that when scitag is not enabled it's a long timeout and when it is enabled it is a very short timeout. That will be hard to explain. I would make sure both values are the same.

abh3 (changes requested):

src/XrdTpc/XrdTpcPMarkManager.hh
ccaffy: Yes, because the PMarkManager is a member of the TPCLogRecord which will be destroyed at the end of the transfer, (which will destroy the PMarkManager. But the XrdHttpExtReq will still be alive after the TPCLogRecord instance is destroyed. See: https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L300, https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L304, https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L938 and https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L856

src/XrdTpc/XrdTpcTPC.cc
ccaffy: It's a comment no? Was left here for history... But actually I believe it can be removed

src/XrdNet/XrdNetUtils.cc
ccaffy: OK very nice, did not know about `poll()`, will do :)

src/XrdTpc/XrdTpcTPC.cc
ccaffy: OK, will try to see how I can extract that default value from curl and assign the same value for it...

src/XrdNet/XrdNetUtils.cc
ccaffy: Done in https://github.com/xrootd/xrootd/pull/2242/commits/d283708019b8d3a101979443da377f7b81b62f71
Will squash everything once everything is OK :)

src/XrdTpc/XrdTpcTPC.cc
ccaffy: OK, there's no way I can extract this value. The default is 300 seconds: https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html

 Though I can add a static variable and set the timeout value of 300 seconds with a reference to that documentation

src/XrdTpc/XrdTpcTPC.cc
ccaffy: Done here https://github.com/xrootd/xrootd/pull/2242/commits/7703a2516b2e36bfa489347535726980ed1e7212

src/XrdTpc/XrdTpcTPC.cc
abh3: I don't understand. The Linux default is 60 seconds the curl default is 300 second? I don't think so. I mean 5 minutes is totally excessive. Can't we just set the curl timeout irrespective of whether pmark is enabled or not to the Linux default of 60 seconds?

ccaffy: I am indeed more than happy to set the value for 60 seconds for curl, even if pmark is not enabled.

ccaffy: Done, if everything is good then I can squash and force push before merging.

@abh3 out of curiosity, where did you get the information about the Linux default 60 seconds connect timeout value?

abh3:
I looked it up on the web. It's also been our experience when trying to connect but you can't get through. The conect timeout is relatively long. You can try it yourself.

ccaffy:
Just squashed all the commits, ready to merge now that the timeout is set to 60 seconds ;)

Cheers,
Cedric

src/XrdNet/XrdNetUtils.cc
```diff
@@ -938,3 +940,72 @@ bool XrdNetUtils::Singleton(const char  *hSpec, const char **eText)
...
+        // We managed to connect
+        SetSockBlocking(sockfd, true, errMsg);  // Set back to blocking
+        return true;
+      } else {
+        errMsg << "Connection failed after poll: " << strerror(so_error);
+        ::close(sockfd);
+        return false;
+      }
+    }
+  }
```
amadio, line 991: It's probably fine to merge this as is, but I'd consider refactoring this part to avoid nested `if`s, and such that the success case is when we arrive at the end with a `return true;`.

src/XrdNet/XrdNetUtils.cc
```diff
@@ -938,3 +940,72 @@ bool XrdNetUtils::Singleton(const char  *hSpec, const char **eText)
...
+  int flags = fcntl(sockfd, F_GETFL, 0);
+  if (flags == -1) {
+    errMsg << "Failed to get socket flags " << strerror(errno);
+    return false;
+  }
+
+  flags = blocking ? (flags & ~O_NONBLOCK) : (flags | O_NONBLOCK);
+
+  if (fcntl(sockfd, F_SETFL, flags) == -1) {
+    errMsg << "Failed to set socket blocking/non-blocking";
```
amadio, line 1006: It's probably a good idea to add `<< strerror(errno)` on the set side as well, no?

ccaffy: Done with 773d2469aa4a34e05d6d2ba61b656a8b35303db1

src/XrdNet/XrdNetUtils.cc
ccaffy: Done with 773d2469aa4a34e05d6d2ba61b656a8b35303db1

src/XrdNet/XrdNetUtils.cc
```diff
@@ -938,3 +940,72 @@ bool XrdNetUtils::Singleton(const char  *hSpec, const char **eText)
...
+  if (result < 0) {
+    errMsg << "Poll error: " << strerror(errno);
+    ::close(sockfd);
+    return false;
+  } else if (result == 0) {
+    errMsg << "Connection timed out";
+    ::close(sockfd);
+    return false;
+  } else {
+    if (fds.revents & POLLOUT) {
```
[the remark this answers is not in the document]
ccaffy: Done with 773d2469aa4a34e05d6d2ba61b656a8b35303db1

src/XrdTpc/XrdTpcPMarkManager.cc
```diff
@@ -30,17 +31,38 @@ PMarkManager::SocketInfo::SocketInfo(int fd, const struct sockaddr * sockP) {
...
+  // If pmark is not enabled, we leave libcurl doing the connection
+  return true;
+}
+
+bool PMarkManager::isEnabled() const {
+  if(mPmark && mReq.mSciTag >= 0) {
+    return true;
+  }
+  // If packet marking is not enabled, the connection is done by libcurl
+  return false;
```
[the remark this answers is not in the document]
ccaffy: Absolutely! Done with fe6bb187f4f9218bccc956e1e70cbb8b77d8ace5

amadio (approved):

PR #2244, opened 2024-04-22
[XrdHttp] Reset HTTP request scitag during reset

### LINKED ISSUE #2243: [XrdTpcTPC] SciTag provided by user is kept between requests with and without SciTag header
In the case a user submits a HTTP TPC request with a SciTag, immediately followed by a HTTP TPC request without a SciTag header, the server will still emit a firefly for the second request.

ccaffy:
Fixes #2243


abh3 (approved):
Looks fine to me.

Issue #14, opened 2013-05-08
[XrdCl] does not always try full list of offered authentication methods 


apeters1971:
If the XRootD server offers several authentication methods and the authentication fails during the handshake, the new XrdCl does not continue to try alternative authentication methods.

```
[2013-05-08 22:28:19 +0200][Debug  ][File              ] [0xdc41610@root://localhost//proc/user/?mgm.cmd=cd&mgm.path=/eos/dev/raiddp/&mgm.option=s&eos.ruid=0&eos.rgid=0] Sending an open command
[2013-05-08 22:28:19 +0200][Debug  ][PostMaster        ] [localhost:1094] Found 1 address(es): [::127.0.0.1]:1094
[2013-05-08 22:28:19 +0200][Debug  ][AsyncSock         ] [localhost:1094 #0.0] Attempting connection to [::127.0.0.1]:1094
[2013-05-08 22:28:19 +0200][Debug  ][Poller            ] Adding socket 0xcbbd330 to the poller
[2013-05-08 22:28:19 +0200][Debug  ][AsyncSock         ] [localhost:1094 #0.0] Async connection call returned
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Sending out the initial hand shake + kXR_protocol
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Got the server hand shake response (type: manager [], protocol version 297)
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] kXR_protocol successful (type: manager [], protocol version 297)
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Sending out kXR_login request, username: root
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Logged in
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Authentication is required: &P=krb5,host/eosdevsrv1.cern.ch@CERN.CH,fwd&P=gsi,v:10300,c:ssl,ca:1d879c6c.0&P=unix&P=sss,0.13:/etc/eos.keytab
[2013-05-08 22:28:19 +0200][Debug  ][XRootDTransport   ] [localhost:1094 #0.0] Sending authentication data
[...]

Issue #13, opened 2013-05-08
xrdcp: -DI option segfaults


jmuf:
(Presumably similar/same as https://savannah.cern.ch/bugs/?100772 ?)

<pre>
[root@lxbst2320 tmp]# gdb xrdcp
GNU gdb (GDB) Red Hat Enterprise Linux (7.0.1-45.el5)
Copyright (C) 2009 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /usr/bin/xrdcp...Reading symbols from /usr/lib/debug/usr/bin/xrdcp.debug...done.
done.
(gdb) run -d 1 -v -f -DIReadCacheSize 0 -DIConnectTimeout 30 -DITransactionTimeout 60 -DIRequestTimeout 60 -DIRedirCntTimeout 60   /etc/group root://128.142.188.27//eos/ppsscratch/test/slstest-eospps/foo
Starting program: /usr/bin/xrdcp -d 1 -v -f -DIReadCacheSize 0 -DIConnectTimeout 30 -DITransactionTimeout 60 -DIRequestTimeout 60 -DIRedirCntTimeout 60   /etc/group root://128.142.188.27//eos/ppsscratch/test/slstest-eospps/foo
warning: no loadable sections found in added symbol-file system-supplied DSO at 0x2aaaaaaab000
[Thread debugging using libthread_db enabled]

Program received signal SIGSEGV, Segmentation fault.
0x000000352b478d70 in strlen () from /lib64/libc.so.6
(gdb) bt
#0  0x000000352b478d70 in strlen () from /lib64/libc.so.6
#1  0x00002aaaaad54899 in XrdOucHashVal (KeyVal=0x1d861 <Address 0x1d861 out of bounds>)
    at /usr/src/debug/xrootd/xrootd/src/XrdOuc/XrdOucHashVal.cc:37
#2  0x000000000040cf1c in XrdOucHash<char>::Add (this=0x6194e0, KeyVal=0x1d861 <Address 0x1d861 out of bounds>, 
[...]

Issue #12, opened 2013-05-01
repository numbering differences between osg and xrootd causing problems


dougbenjamin:
Hi,

  Here is an issue that I have seen due to the repository miss match xrootd - osg - epel.

I am including the text of the OSG ticket that I filed.

https://ticket.grid.iu.edu/goc/14976

Hello,

Today we saw these yum errors  due to repository incompatibilities -  osg vs xrootd  for the xrootd software -

---

  YUM - security

---

```
1:xrootd-client-libs-3.3.1-1.2.osg.el5.x86_64 from osg has depsolving problems   --> Missing Dependency: xrootd-libs = 1:3.3.1-1.2.osg.el5 is needed by
package 1:xrootd-client-libs-3.3.1-1.2.osg.el5.x86_64 (osg)

1:xrootd-server-libs-3.3.1-1.2.osg.el5.x86_64 from osg has depsolving problems   --> Missing Dependency: xrootd-libs = 1:3.3.1-1.2.osg.el5 is needed by
package 1:xrootd-server-libs-3.3.1-1.2.osg.el5.x86_64 (osg)

1:xrootd-3.3.1-1.2.osg.el5.x86_64 from osg has depsolving problems   --> Missing Dependency: xrootd-libs = 1:3.3.1-1.2.osg.el5 is needed by
package 1:xrootd-3.3.1-1.2.osg.el5.x86_64 (osg)

1:xrootd-server-devel-3.3.1-1.slc5.xu.x86_64 from installed has depsolving problems   --> Missing Dependency: xrootd-libs-devel = 1:3.3.1-1.slc5.xu is
needed by package 1:xrootd-server-devel-3.3.1-1.slc5.xu.x86_64 (installed)

1:xrootd-client-devel-3.3.1-1.slc5.xu.x86_64 from installed has depsolving problems   --> Missing Dependency: xrootd-libs-devel = 1:3.3.1-1.slc5.xu is
needed by package 1:xrootd-client-devel-3.3.1-1.slc5.xu.x86_64 (installed)

1:xrootd-devel-3.3.1-1.2.osg.el5.x86_64 from osg has depsolving problems   --> Missing Dependency: xrootd-libs = 1:3.3.1-1.2.osg.el5 is needed by
package 1:xrootd-devel-3.3.1-1.2.osg.el5.x86_64 (osg)

[...]

Issue #11, opened 2013-05-01
xrootdfs crash


dougbenjamin:
Hi,

  We are seeing this error in the system logs

May  1 14:54:48 atl008 kernel: xrootdfs[3249] trap divide error rip:336920819a rsp:2af4fdc67710 error:0

for a static mount xrootdfs mount point -
xrootdfs  /atlfs03/atlas fuse  rdr=root://atlfs03.phy.duke.edu:1094//atlas,uid=54657 0 0

[root@atl008 ~]# yum list installed *xrootd-fuse*
Loaded plugins: downloadonly, kernel-module
Excluding Packages from Extra Packages for Enterprise Linux 5 - x86_64
Finished
Excluding Packages from VDT RPM repository - development versions for Redhat Enterprise Linux 5 and compatible
Finished
Installed Packages
xrootd-fuse.x86_64                                                   1:3.3.1-1.slc5.xu                                                   installed

Here is the OS details -

[benjamin@atl008 ~]$ uname -a
Linux atl008.phy.duke.edu 2.6.18-348.4.1.el5 #1 SMP Tue Apr 16 15:42:58 EDT 2013 x86_64 x86_64 x86_64 GNU/Linux
[benjamin@atl008 ~]$ cat /etc/redhat-release
Scientific Linux release 5.9 (Boron)

Any ideas how I can get around this issue.

Thanks,

Doug

PR #2245, opened 2024-04-22
[XrdCrypto] Avoid some repated calls of EVP_PKEY_check on the same key

smithdh:
With openssl3 the EVP_PKEY_check() may be more computationally expensive than with previous openssl versions. Currently XrdCl can call this check multiple times for the same key, e.g. when copying constructing an XrdCryptosslRSA object. This patch is to avoid making repeated, but similar, calls to EVP_PKEY_check().

### DIFF
--- a/src/XrdCrypto/XrdCryptosslAux.cc
+++ b/src/XrdCrypto/XrdCryptosslAux.cc
@@ -49,32 +49,6 @@
 static int gErrVerifyChain = 0;
 XrdOucTrace *sslTrace = 0;
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
-static RSA *EVP_PKEY_get0_RSA(EVP_PKEY *pkey)
-{
-    if (pkey->type != EVP_PKEY_RSA) {
-        return NULL;
-    }
-    return pkey->pkey.rsa;
-}
-#endif
-
-static int XrdCheckRSA (EVP_PKEY *pkey) {
-   int rc;
-#if OPENSSL_VERSION_NUMBER < 0x10101000L
-   RSA *rsa = EVP_PKEY_get0_RSA(pkey);
-   if (rsa)
-      rc = RSA_check_key(rsa);
-   else
-      rc = -2;
-#else
-   EVP_PKEY_CTX *ckctx = EVP_PKEY_CTX_new(pkey, 0);
-   rc = EVP_PKEY_check(ckctx);
-   EVP_PKEY_CTX_free(ckctx);
-#endif
-   return rc;
-}
-
 //____________________________________________________________________________
 int XrdCryptosslX509VerifyCB(int ok, X509_STORE_CTX *ctx)
 {
@@ -552,7 +526,7 @@ int XrdCryptosslX509ParseFile(FILE *fcer,
          DEBUG("found a RSA private key in file " << fname);
          // We need to complete the key
          // check all the public keys of the loaded certificates
-         if (XrdCheckRSA(rsa) == 1) {
+         if (rsa) {
             // Loop over the chain certificates
             XrdCryptoX509 *cert = chain->Begin();
             while (cert && cert->Opaque()) {
@@ -568,12 +542,12 @@ int XrdCryptosslX509ParseFile(FILE *fcer,
 #endif
                         EVP_PKEY_free(evpp);
                         if (rc == 1) {
-                           DEBUG("RSA key completed");
-                           // Update PKI in certificate
+                           // Update PKI in certificate; also tests if the key is complete
                            cert->SetPKI((XrdCryptoX509data)rsa);
-                           // Update status
-                           cert->PKI()->status = XrdCryptoRSA::kComplete;
-                           break;
+                           if (cert->PKI()->status == XrdCryptoRSA::kComplete) {
+                              DEBUG("RSA key completed");
+                              break;
+                           }
                         }
                   }
                }
@@ -662,7 +636,7 @@ int XrdCryptosslX509ParseBucket(XrdSutBucket *b, XrdCryptoX509Chain *chain)
          DEBUG("found a RSA private key in bucket");
          // We need to complete the key
          // check all the public keys of the loaded certificates
-         if (XrdCheckRSA(rsa) == 1) {
+         if (rsa) {
             // Loop over the chain certificates
             XrdCryptoX509 *cert = chain->Begin();
             while (cert && cert->Opaque()) {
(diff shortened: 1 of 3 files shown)


abh3 (commented):
Well, I don't completely understand how this will avoid repeated calls but it does look reasonable. Presumably, deletion of OpenSSL version 1.01 dependent code is also correct. However, it would be nice to explain why this is the case. Any reason why this is still in draft mode?

smithdh:
I'll try to motivate this change done with some measurements:

A single client was used to start many (8000) threads, each uses XrdCl to open an xroot connection to a server and does a brief operation (stat). GSI authentication is used; I time how much elapsed time the client takes, and amount of CPU time is accumulated during that time by the xroot server process. (Server has a 4096 bit host cert, client is using a 2048 bit voms-proxy). For openssl 3.0/3.3 an ALMA9 EOS server was used, for openssl 1.0 test a pure xrootd (5.6.9) server on Centos7 was used. The two servers do have different number of cores/cpu type.

Openssl 3.0: standard xrootd
1060s elapsed:	15704s cpu
997s   elapsed:	14816s cpu

Openssl 3.0: xrootd + #2245
77s elapsed:		858s cpu
77s elapsed:		886s cpu

Openssl 3.3: standard xrootd
998s elapsed:		15199s cpu
949s elapsed:		14445s cpu

Openssl 3.3: xrootd + #2245
72s elapsed:		613s cpu
69s elapsed:		604s cpu

--

openssl 1.0;  standard xrootd
78s elapsed		1278s cpu
85s elapsed		1278s cpu

openssl 1.0; xrootd + #2245
77s elapsed		1005s cpu
75s elapsed		1016s cpu

The indication was that, in this test, this patch appears to have give a large reduce in CPU time on the server to process the requests when using openssl 3.0 or later.

smithdh:
Typical busy thread, server side, AL9 openssl-3.0.7-25:

#0  0x00007f4b122afe49 bn_sqrx8x_internal
#1  0x00007f4b122afa87 bn_powerx5
#2  0x00007f4b120ec691 BN_mod_exp_mont_consttime
#3  0x00007f4b120eefde ossl_bn_miller_rabin_is_prime.part.0
#4  0x00007f4b120ef3a7 bn_is_prime_int
#5  0x00007f4b121eec88 rsa_validate_keypair_multiprime.lto_priv.0
#6  0x00007f4b122759ba rsa_validate.lto_priv.0
#7  0x00007f4b121a1ec1 try_provided_check
#8  0x00007f4b121a21f5 EVP_PKEY_pairwise_check
#9  0x00007f4b12e40062 XrdCryptosslRSA::XrdCryptosslRSA(XrdCryptosslRSA const&)
#10 0x00007f4b125c30b2 XrdCryptosslFactory::RSA(XrdCryptoRSA const&)
#11 0x00007f4b125edd98 XrdSecProtocolgsi::ServerDoCertreq(XrdSutBuffer*, XrdSutBuffer**, XrdOucString&)
#12 0x00007f4b125ee08d XrdSecProtocolgsi::ParseServerInput(XrdSutBuffer*, XrdSutBuffer**, XrdOucString&)
#13 0x00007f4b125ee52f XrdSecProtocolgsi::Authenticate(XrdSecBuffer*, XrdSecBuffer**, XrdOucErrInfo*)
#14 0x00007f4b12f65837 XrdXrootdProtocol::do_Auth()
#15 0x00007f4b12f58296 XrdXrootdProtocol::Process2()

abh3 (approved):
The statistics are motivating enough since it's even faster than xrootd with OpenSsl 1.0.

src/XrdCrypto/XrdCryptosslX509.cc
```diff
@@ -432,14 +408,26 @@ void XrdCryptosslX509::CertType()
...
+   // if newpki is null does nothing
+   // if newpki contains a consistent private & public key we take ownership
+   //   so that this->PKI()->status will be kComplete.
+   // otherwise, newpki is not consistent:
+   // if the previous PKI() was null or was already kComplete it is and reset
+   //   so that this->PKI()->status will be kInvalid.
+
+   if (!newpki) return;
+
+   std::unique_ptr<XrdCryptoRSA> tmprsa(new XrdCryptosslRSA((EVP_PKEY *)newpki, 1));
```
amadio, line 421: After looking at the changes more thoroughly, I think the changes are all fine. The only suggestion I have is to use `std::make_unique` rather than calling new directly (maybe without using `auto` if you don't like that):
```suggestion
   auto tmprsa = std::make_unique<XrdCryptosslRSA>((EVP_PKEY*)newpki, 1));
```
Note that there is a mismatch of types between `std::unique_ptr<XrdCryptoRSA>` and `new XrdCryptosslRSA` in the code above, which should be fixed in any case.

src/XrdCrypto/XrdCryptosslX509.cc
```diff
@@ -175,9 +150,10 @@ XrdCryptosslX509::XrdCryptosslX509(const char *cf, const char *kf)
       if ((evpp = PEM_read_PrivateKey(fk,0,0,0))) {
          DEBUG("RSA key completed ");
          // Test consistency
-         if (XrdCheckRSA(evpp) == 1) {
+         std::unique_ptr<XrdCryptoRSA> tmprsa(new XrdCryptosslRSA(evpp, 1));
```
amadio, line 153: Same here, maybe use `std::make_unique` and fix mismatch of pointer types.

src/XrdCrypto/XrdCryptosslAux.cc
```diff
@@ -568,12 +542,12 @@ int XrdCryptosslX509ParseFile(FILE *fcer,
...
-                           // Update PKI in certificate
+                           // Update PKI in certificate; also tests if the key is complete
                            cert->SetPKI((XrdCryptoX509data)rsa);
-                           // Update status
-                           cert->PKI()->status = XrdCryptoRSA::kComplete;
-                           break;
+                           if (cert->PKI()->status == XrdCryptoRSA::kComplete) {
+                              DEBUG("RSA key completed");
+                              break;
+                           }
```
amadio, line 550: > Well, I don't completely understand how this will avoid repeated calls but it does look reasonable.

Just for the sake of documenting the change for later, the check is removed from `if (XrdCheckRSA(rsa) == 1)` which becomes just `if (rsa)` because later on in `SetPKI(rsa)` the check is performed when calling `new XrdCryptosslRSA(rsa, 1)`. The constructor ensures that the key being created is checked.

src/XrdCrypto/XrdCryptosslRSA.cc
```diff
@@ -236,10 +236,10 @@ XrdCryptosslRSA::XrdCryptosslRSA(const XrdCryptosslRSA &r) : XrdCryptoRSA()
...
           } else {
             if ((fEVP = PEM_read_bio_PrivateKey(bcpy,0,0,0))) {
-               // Check consistency
-               if (XrdCheckRSA(fEVP) == 1) {
-                  // Update status
-                  status = kComplete;
+               // Check consistency only if original was not marked complete
+               if (r.status == kComplete || XrdCheckRSA(fEVP) == 1) {
+                 // Update status
+                 status = kComplete;
```
amadio, line 242: This is the other place where a redundant check is removed. If the key being copied has already been checked, there's no need to re-check it when copying it.

smithdh:
Thanks. I added the use of make_unique.

amadio:
Thanks, @smithdh, I'm merging! The CI failure is not related. I will fix that afterwards.

PR #2246, opened 2024-04-22
[XrdCms] add a tunable weighed random load balancing algorithm

Jo-stfc:
Add a tunable weighed random load balancing algorithm for SelbyLoad based selection.

This can be enabled by adding the configuration option: cms.sched randlb 1.

This prevents load hotspotting and aims to distribute load somewhat evenly across all available servers, with higher chance of picking less loaded gateways. This is currently the load balancing algorithm used at RAL with good performance and stability under heavy loads


src/XrdCms/XrdCmsCluster.cc
```diff
@@ -1786,10 +1788,15 @@ XrdCmsNode *XrdCmsCluster::SelbyLoad(SMask_t mask, XrdCmsSelector &selR)
...
-//
-   selR.Reset(); SelTcnt++;
-   for (int i = 0; i <= STHi; i++)
+    // Scan for a node (preset possible, suspended, overloaded, full, and dead)
+    //
+    selR.Reset(); SelTcnt++;
+    int selCap = 1;
+    int randomSel=1;
+    //default 0 to skip the node in random selection if the below checks fail
+    int *weighed = new int[STHi] {0};
```
abh3, line 1797: I'd change this to allow for better optimization to:
int weighed = (Config.P_randlb==1 ? new int[STHi] {0} : 0);

Then subsequently substitute if (Config.P_randlb==1) with if (weighted)

src/XrdCms/XrdCmsCluster.cc
```diff
@@ -1816,12 +1833,29 @@ XrdCmsNode *XrdCmsCluster::SelbyLoad(SMask_t mask, XrdCmsSelector &selR)
...
+    std::uniform_int_distribution<int> distr(randomSel,selCap);
+    randomSel = distr(generator);
+    for(int i=0;i<=STHi;i++){
+      if(randomSel<=weighed[i]){
+        sp=NodeTab[i];
+        break;
+      }
+    }
+    }
+   delete [] weighed;
```
abh3, line 1852: Given that allocation only occurs when needed this delete can now move into the scope of the if statement,

abh3 (changes requested):
Just a small improvement in this time sensitive code is needed, Also, could you post what the documentation should say? See https://xrootd.slac.stanford.edu/doc/dev54/cms_config.htm#_Toc53611076

Specifically, explanation of the algorithm being used and how it alters the selection process for load balanced selection.

src/XrdCms/XrdCmsConfig.cc
```diff
@@ -735,6 +735,7 @@ void XrdCmsConfig::ConfigDefaults(void)
    P_load   = 0;
    P_mem    = 0;
    P_pag    = 0;
+   P_randlb = 0;          // SelbyLoad algoruthm choice
```
amadio, line 738: `s/algoruthm/algorithm/`

src/XrdCms/XrdCmsCluster.cc
```diff
@@ -1800,7 +1807,17 @@ XrdCmsNode *XrdCmsCluster::SelbyLoad(SMask_t mask, XrdCmsSelector &selR)
...
-              else{if (selR.needSpace)
+              else{
+                 if (Config.P_randlb==1){
+                   //add 1 to the inverse load, this is to allow some selection in case reported loads hit 100
+                   weighed[i] = selCap + static_cast<int>(101 - np->myLoad +
+                     std::pow(101 - np->myLoad,
+                     std::log(100)-std::log(std::min(std::max(Config.P_fuzz,1),100)))/2);
+                   selCap += static_cast<int>(101 - np->myLoad +
+                     std::pow(101 - np->myLoad,
+                     std::log(100)-std::log(std::min(std::max(Config.P_fuzz,1),100)))/2);
```
amadio, line 1818: You should assign `std::log(100)-std::log(std::min(std::max(Config.P_fuzz,1),100)))/2` to a variable and reuse it instead of computing this twice. You can also use [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp) here instead of the combination of `std::min` + `std::max`, since we have moved to C++17 as our baseline C++ standard on all platforms. Since this expression also doesn't depend on the loop variable, it should be hoisted outside of the `for` loop to be computed only once. But also, is there no way to have this done without `log` and `pow`? These are expensive things to be called frequently, and will burn lots of CPU when there is heavy load in the system. If the range is limited, you should consider using a polynomial approximation of the `log`, I think the exact shape here is not very important, just the general behavior in the range 1 to 100.

src/XrdCms/XrdCmsCluster.cc
```diff
@@ -1816,12 +1833,29 @@ XrdCmsNode *XrdCmsCluster::SelbyLoad(SMask_t mask, XrdCmsSelector &selR)
...
                    Multi = true;
                   }
+               }
           }
-
+    if (Config.P_randlb==1){ 
+    // pick a random weighed node
+    //
+    std::random_device rand_dev;
+    std::mt19937 generator(rand_dev());
```
amadio, line 1842: This function must be called under a lock, so these can probably be created at initialization only and be reused afterwards. Also, I'd avoid hard-coding [`mt19937`](https://en.cppreference.com/w/cpp/header/random) here, which is 32-bit only, and just use `default_random_engine` instead. Frequent initialization with hardware-based randomness can create unnecessary slowness.

src/XrdCms/XrdCmsCluster.cc
amadio: Or even better, if this is called frequently, it should not contain any allocations. Allocations should happen only at initialization.

amadio (changes requested):

src/XrdCms/XrdCmsCluster.cc
Jo-stfc: thanks, included in the second commit

src/XrdCms/XrdCmsCluster.cc
Jo-stfc: thanks, included in the second commit

src/XrdCms/XrdCmsCluster.cc
```diff
@@ -35,6 +35,8 @@
 #include <unistd.h>
 #include <netinet/in.h>
 #include <sys/types.h>
+#include <cmath>
+#include <random>
```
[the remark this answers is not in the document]
Jo-stfc: thanks, included in the second commit

src/XrdCms/XrdCmsCluster.cc
Jo-stfc: I've moved the calculation outside the loop and used std::clamp. Exponentials are needed to show significant differences in the load distribution according to the specified fuzz. I've replaced the log function with a scaled percentile, it was used to prevent the exponentials from triggering an integer overflow.

src/XrdCms/XrdCmsConfig.cc
Jo-stfc: fixed

src/XrdCms/XrdCmsCluster.cc
Jo-stfc: switched the random number generator to static

src/XrdCms/XrdCmsCluster.cc
```diff
@@ -1800,7 +1808,17 @@ XrdCmsNode *XrdCmsCluster::SelbyLoad(SMask_t mask, XrdCmsSelector &selR)
...
-              else{if (selR.needSpace)
+              else{
+                 if (weighed){
+                   //add 1 to the inverse load, this is to allow some selection in case reported loads hit 100
+                   weighed[i] = selCap + static_cast<int>(101 - np->myLoad +
+                     std::pow(101 - np->myLoad,
+                     scalingFactor)/2);
+                   selCap += static_cast<int>(101 - np->myLoad +
+                     std::pow(101 - np->myLoad,
+                     scalingFactor)/2);
```
amadio, line 1819: I find these split lines quite confusing to read, maybe it's better to just join the assignments for `weighed[i] =` and `selCap +=` into a single line. You are still computing `static_cast<int>(101 - np->myLoad + std::pow(101 - np->myLoad, scalingFactor)/2)` twice. You could just do this:
```suggestion
                   int nload = 101 - np->myLoad;
                   selCap += static_cast<int>(nload + std::pow(nload, scalingFactor)/2);
                   weighed[i] = selCap;
```

Jo-stfc: thanks. I've applied that suggestion

amadio:
Hi @Jo-stfc, Thanks for pushing the new version. Could you please also edit `tests/cluster/common.cfg` and enable your new algorithm so it's tested in the CI? I see that my version which is based on yours is not quite working (likely my own fault, for sure). The problem I see is that not all nodes get weights associated with them in the first loop, and the selection in the second loop doesn't take that into account. I will continue to work on this tomorrow. Cheers,

Jo-stfc:
Thanks Guilherme. I've enabled it on the tests. Are uninitialized weights set as 0? if it's undefined that's my bad, I didn't put in a check that the current node weight is valid.

amadio:
@Jo-stfc I was playing with this a bit more, and created an alternative version of the algorithm, which you can find here: https://github.com/amadio/xrootd/commit/478333f5d23d8220c7a578c0bde1a342990192d7

Could you please give it some testing and see if it fits your needs? I was also wondering if the existing load-based algorithms with a fuzz factor would not be able to do what you want without the need to add a new algorithm. Note that in order to really enable the load-balancing algorithm, I had to specify some form of load formula, so in this pull request, if you want the algorithm as you have it, you should add something like `cpu 100` or other kind of mix in the `cms.sched` configuration line. Otherwise I didn't see the function `SelbyLoad` get called in my tests. Cheers,

amadio:
@abh3 Could you please take a look at my commit above and let me know what you think? Thank you!

amadio:
@Jo-stfc Please note that there was a [compilation error](https://my.cdash.org/viewBuildError.php?buildid=2589076) on Ubuntu. You may want to fix it:

Jo-stfc:
> @Jo-stfc I was playing with this a bit more, and created an alternative version of the algorithm, which you can find here: [amadio@478333f](https://github.com/amadio/xrootd/commit/478333f5d23d8220c7a578c0bde1a342990192d7)
> 
> Could you please give it some testing and see if it fits your needs? I was also wondering if the existing load-based algorithms with a fuzz factor would not be able to do what you want without the need to add a new algorithm. Note that in order to really enable the load-balancing algorithm, I had to specify some form of load formula, so in this pull request, if you want the algorithm as you have it, you should add something like `cpu 100` or other kind of mix in the `cms.sched` configuration line. Otherwise I didn't see the function `SelbyLoad` get called in my tests. Cheers,

Thanks, I've added the load formula to the test. Your commi

> @Jo-stfc I was playing with this a bit more, and created an alternative version of the algorithm, which you can find here: [amadio@478333f](https://github.com/amadio/xrootd/commit/478333f5d23d8220c7a578c0bde1a342990192d7)
> 
> Could you please give it some testing and see if it fits your needs? I was also wondering if the existing load-based algorithms with a fuzz factor would not be able to do what you want without the need to add a new algorithm. Note that in order to really enable the load-balancing algorithm, I had to specify some form of load formula, so in this pull request, if you want the algorithm as you have it, you should add something like `cpu 100` or other kind of mix in the `cms.sched` configuration line. Otherwise I didn't see the function `SelbyLoad` get called in my tests. Cheers,

Thanks Amadio, your implementation looks good. The existing load balancing algorithm did not work for us as it's not making optimal use of the available capacity (it always sends transfers to the N best nodes, resulting in load bouncing between different hosts) as well as having some ladder behaviour (since the nodetab is static, if the middle node has much higher load, the RR would only go trough unloaded nodes on one side of the array, sometimes locking transfers to a subset of the nodes)

amadio:
@Jo-stfc I'd appreciate it if you could let me know by Wednesday this week if the algorithm I implemented above is good enough for your use case. If yes, I will include it into the next release (XRootD 5.7.0). Otherwise, we will keep the patch and include it in the next minor or major release, which should happen by the end of the year. Thank you!

Jo-stfc:
Thanks Guillherme, I'll test it out tomorrow morning and let you know.

Jo-stfc:
Hi Guilherme, yes the algorithm above works on our use case

amadio:
Thank you @Jo-stfc for the idea for the new load balancing algorithm. Since we released the alternative implementation above in XRootD 5.7.0, I'm closing this pull request. Best regards,

PR #2253, opened 2024-04-29
[XrdHttp] External handlers can be loaded without TLS

### LINKED ISSUE #2099: [XrdHttp] Add an option to allow the loading of external handlers if TLS is not configured
@amadio found an interesting issue when he tried to setup XrdHttpTPC on his devbox.
Without TLS configured, the XrdHttpTPC library could not be loaded.

It comes from the following line: https://github.com/xrootd/xrootd/blob/master/src/XrdHttp/XrdHttpProtocol.cc#L1085

Some external HTTP plugins don't necessary need TLS to be configured. E.g: XrdHttpTPC should work without the need for TLS.

We will add a new option to the http.exthandler configuration, named `notls` to allow a user to specify that the plugin to load does not require TLS.

Implementation idea: 
- create a second vector for plugins to be loaded without TLS support.
- Don't remember to update the doc: https://xrootd.slac.stanford.edu/doc/dev56/xrd_config.htm#_Toc135684370

ccaffy:
The XrootD server administrator can now configure HTTP handlers without having configured TLS.
They can do so by adding "notls" after the path of the library. Example: `http.exthandler xrdtpc libXrdHttpTPC.so notls`
By default, HTTP external handlers can only be loaded if TLS is enabled.

Fixes #2099

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -2590,10 +2590,17 @@ int XrdHttpProtocol::xexthandler(XrdOucStream &Config,
   }
   strncpy(namebuf, val, sizeof(namebuf));
   namebuf[ sizeof(namebuf)-1 ] = '\0';
-  
+
+  // Get the +notls option if it was provided
+  val = Config.GetWord();
+
+  if(val && !strcmp("+notls",val)) {
+    noTlsOK = true;
+    val = Config.GetWord();
+  }
+
   // Get the path
   //
-  val = Config.GetWord();
   if (!val || !val[0]) {
     eDest.Emsg("Config", "No http external handler plugin specified.");
     return 1;
@@ -2604,12 +2611,6 @@ int XrdHttpProtocol::xexthandler(XrdOucStream &Config,
   }
 
   strcpy(path, val);
-
-  val = Config.GetWord();
-
-  if(val && !strcmp("notls",val)) {
-    noTlsOK = true;
-  }
   
   // Everything else is a free string
   //
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1105,6 +1105,7 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
        eDest.Say("Config warning: HTTPS functionality ", why);
        httpsmode = hsmOff;
 
+       LoadExtHandlerNoTls(extHIVec, ConfigFN, *myEnv);
        if (what)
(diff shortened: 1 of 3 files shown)


amadio:
Ignore macOS builds for now, they are affected by https://github.com/actions/runner-images/issues/8649 If I cannot fix tomorrow, I will hard-code macos-13 for now instead of latest (which switched to 14).

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2601,6 +2604,12 @@ int XrdHttpProtocol::xexthandler(XrdOucStream &Config,
   }
 
   strcpy(path, val);
+
+  val = Config.GetWord();
+
+  if(val && !strcmp("notls",val)) {
```
abh3, line 2610: So, this introduces a problem. If the token is not 'notls' then it is ignored but according to the specification: https://xrootd.slac.stanford.edu/doc/dev56/xrd_config.htm#_Toc152596904 it could have been a parameter token to the plugin-in. Of course, the obvious solution is if it is 'notls' set the variable true and get another token. Otherwise, just pass it through (i.e. delete the subsequent 'get'). However, this creates another problem. If the token is notls then the exthandler will not see it and maybe notls is a valid exthandler parameter that now is technically missing. Unfortunately, the solution is to adopt the xroot style of of system parameters vs plug-in parameters which is somewhat more complicated. See my summary comments on how to do this.

abh3 (changes requested):
So, in general plug-in directives have a common format:
<directive> <system parameters> <library path> <plug-in parameters>

For the exthandler directive is a little squishy following that format, However, there is a way out as shown on what the osslib directive which had a similar problem:
https://xrootd.slac.stanford.edu/doc/dev56/ofs_config.htm#_Toc136617299

Notice that the system parameters are differentiated from the path by prefixing them with something that a name (exthandler specific)  or path would never start with (here is it a '+' but a '-' could also have been used). This avoids messing with possible plug-in parameters and still keeps backward compatibility. So, I would suggest using something along these lines. It is a bit more complicated to implement but it will not conflict with anyone's config file.

src/XrdHttp/XrdHttpProtocol.cc
ccaffy: Thanks for your review. I've renamed the parameter as "+notls" and added it after the name of the external handler to load. That way, the path of the external handler can still be provided with extra parameters if necessary. https://github.com/xrootd/xrootd/pull/2253/commits/1b4aebd411bde17841c850c5598d8761ee1003a3

To be squashed once everything is OK before merging

abh3 (approved):
Thank you!

PR #2255, opened 2024-05-02
[XrdHttp] Improved first-line and header parsing

### LINKED ISSUE #1964: XrdHttp parsing of the headers is "fragile"
This issue was discovered while debugging the CMS SAM tests which run against the eoscms.cern.ch instance using scitokens over HTTP. Basically the symptom was that the first PROPFIND request would succeed while the second one would fail. For subsequent requests this pattern would repeat itself.

The requests were submitted using gfal python bindings in the same context (TCP connection). Basically the problem came from the way the bearer token was formatted in the sense that it contained an extra new line character at the end `\n` (`0xA` in hex). 

This completely messes up the parsing of the HTTP headers in the XrdHttp layer and causes predictable failures. Thanks to Joao from the FTS team, I attach a simple python reproducer using the gfal client to trigger this issue and the corresponding client logs. The important thing in this script is the reading of the token from the file which appends a `0xA` char at the end of it.

The original issue is now being fixed in the CMS SAM tests but still, I think the XrdHttp parsing of the headers should be more robust and handle such cases better by for example failing the initial request which has "malformed" headers rather than cascading the error to subsequent requests. 

[gfal_log.txt](https://github.com/xrootd/xrootd/files/10981328/gfal_log.txt)
[gfal_test.py.txt](https://github.com/xrootd/xrootd/files/10981336/gfal_test.py.txt)

ccaffy:
When we detect the end of header in the client request ("\r\n"), we will actually continue to read data from the buffer if there are still some to read.

The case when a client submits a HTTP request like `\r\nGET ...` is also covered.

Fixes #1964

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -616,6 +616,19 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
       TRACE(DEBUG, " rc:" << rc << " got hdr line: " << tmpline.c_str());
 
       if ((rc == 2) && (tmpline.length() > 1) && (tmpline[rc - 1] == '\n')) {
+        // A \r\n was detected
+        if(BuffUsed()) {
+          if(CurrentReq.request == CurrentReq.rtUnset){
+            // The first line has a "\n" at the beginning
+            // Bad request
+            return SendSimpleResp(400, nullptr, nullptr, "Got first line break before the HTTP verb", 0, false);
+          } else {
+            // We are getting the headers from the buffer and we detected the header end.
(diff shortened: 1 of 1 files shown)


abh3 (approved):
I am going to leave this as OK assuming this was tested to the hilt. Unfortunately, the code assumes you know exactly what the buffer reader returns and why a length of two is important. In essence, BuffgetLine() is incomprehensible and likely should be on the list of things to get rewritten; though is may be too fragile to tamper with.

amadio:
I'm merging this, but let's include `int XrdHttpProtocol::Process(XrdLink *lp)` in the refactoring campaign of #2256, it doesn't look so good either at the moment.

bbockelm:
Note: this caused a regression as documented in #2273.

Issue #2067, opened 2023-07-31
XrdHttpTpc better reporting of underlying curl errors


esindril:
It would be good, especially for end-users, to get more information when the underlying curl copy that drives HTTP TPC transfers fails. For example the following two transfers report the same error message, but the actual reason is very different:

1. `https://fts3-atlas.cern.ch:8449/fts3/ftsmon/#/job/067ed092-2c51-11ee-a2d9-fa163e087e6b`
**Reported error**
reason_text:TRANSFER ERROR: Copy failed (3rd pull). Last attempt: Transfer failure: Error when interacting with local filesystem: Failure when closing file handle: Unable to store file - file has been cleaned because the stored file does not match the provided targetsize /eos/atlas/atlasscratchdisk/rucio/user/skadir/b4/19/user.skadir.34199691.EXT0._001551.IDPVM.root; input/output error (code=5)
**Actual error message**
TPC_PullRequest: event=SIZE_FAIL, local=/eos/atlas/atlasscratchdisk/rucio/user/skadir/b4/19/user.skadir.34199691.EXT0._001551.IDPVM.root, remote=https://ccdavatlas.in2p3.fr:2880/atlasscratchdisk/rucio/user/skadir/b4/19/user.skadir.34199691.EXT0._001551.IDPVM.ro ot?copy_mode=pull, user=(anonymous), status=500; HTTP library failed: Failed to load CRL file (path? access rights?, format?)

2. `https://fts3-atlas.cern.ch:8449/fts3/ftsmon/#/job/fb91fad2-2d8c-11ee-9e19-fa163eb04ff0`
**Reported error**
reason_text:TRANSFER ERROR: Copy failed (3rd pull). Last attempt: Transfer failure: Error when interacting with local filesystem: Failure when closing file handle: Unable to store file - file has been cleaned because the stored file does not match the provided targetsize /eos/atlas/atlasdatadisk/rucio/mc16_13TeV/65/99/HITS.34202576._005271.pool.root.1; input/output error (code=5)
**Actual error**
This is actually not visible in the logs, as it comes from a 120 seconds timeout in curl when no data moves. Or after a 60 seconds timeout after the transfer starts but then it blocks for whatever reason.

It would be extremely helpful for operators to distinguish between such types of errors as they have really different underlying causes.

ccaffy:
Done and merged via https://github.com/xrootd/xrootd/pull/2260

PR #2260, opened 2024-05-14
[XrdTpcTPC] Improved curl error reporting to the client

ccaffy:
All the errors returned to the client now start with "failure:". If the failure is due to a curl error, the error message will contain "HTTP library failure=X" where X is the error reported by libcurl. In addition, all failures indicate the local and the remote resources involved in the TPC transfer that failed.

### DIFF
--- a/src/XrdTpc/XrdTpcMultistream.cc
+++ b/src/XrdTpc/XrdTpcMultistream.cc
@@ -374,8 +374,10 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
             }
         } while (msg);
         if (res != static_cast<CURLcode>(-1) && res != CURLE_OK) {
+            std::stringstream ss;
+            ss << "Breaking loop due to failed curl transfer: " << curl_easy_strerror(res);
             logTransferEvent(LogMask::Debug, rec, "MULTISTREAM_CURL_FAILURE",
-                "Breaking loop due to failed curl transfer");
+                ss.str());
             break;
         }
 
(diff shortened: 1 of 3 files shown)


ccaffy:
This PR is meant to solve https://github.com/xrootd/xrootd/issues/2067

abh3:
I will certainly review this. In addition could you check the WLCG message requirements that may affect this? The document can be found at https://twiki.cern.ch/twiki/bin/view/LCG/WebdavErrorImprovement and it may not apply to the current patch set.

ccaffy:
I indeed had a look at that document, but I also talked to FTS about this as I wanted to know how they report the error to the user. They parse the body of the server reply and look for `failure` or `failed`. https://github.com/cern-fts/davix/blob/R_0_8_6/src/modules/copy/copy.cpp#L580-L587 

So at the end I just harmonized this to every message in addition to the performance marker. We may need a separate issue to address what's in the Webdav error message improvement project (https://twiki.cern.ch/twiki/bin/view/LCG/WebdavErrorImprovement) E.g: if one tries a TPC PULL where the source is offline, in XRootD we return 500, instead of 404 as indicated in this document.
So I'll create a separate issue to address this.

abh3 (approved):
It would be interesting if we could somehow merge the functionality of logTransferEvent() with generateClientErr() to simplify the message generation code. It certainly would decrease the visual code noise in the error handling sections. However, that's easy to say as it looks like a lot of rewriting would be needed. to make that happen. So, let's leave it the way it is for now.

PR #2262, opened 2024-05-15
Fix timing on the throttle plugin

bbockelm:
When examining some plots of the new g-stream data, we were surprised to see the aggregate I/O time was non-monotonic (by definition, the sum of all time since start should always be monotonic!).

Two bugs were discovered:
- Incorrect handling of the timespec normalization.
- Incorrect clock used for the timer.  Switched to `CLOCK_MONOTONIC`.

Once the second item was handled, the code only uses POSIX functions so enabling Mac OS X (my main development platform) was trivial and I adjusted the defines accordingly.

I would suggest considering the patch for backport to a current release series if a subsequent release is planned.


abh3 (approved):
Looks OK to me.

bbockelm:
> Technically, the fix and the MacOS enable should be two commits as they are very different things

Yes, there's two commits here.  If you'd like, I can do two PRs of one commit as well -- that just felt a bit excessive.

PR #2265, opened 2024-05-16
[XrdHttp/TPC/Macaroons] Client header parsing is now case-insensitive

### LINKED ISSUE #2259: [XrdHttp] Header parsing should be made case-insensitive
```

> GET /eos HTTP/2
> Host: xrootd.server.cern.ch
> user-agent: curl/7.76.1
> accept: */*
> authorization: Bearer token
>

```

```cpp
      trim(m_req_digest);
    }
```

ccaffy:
Fixes #2259

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -154,7 +154,7 @@ int XrdHttpReq::parseLine(char *line, int len) {
     // The value is val
     
     // Screen out the needed header lines
-    if (!strcmp(key, "Connection")) {
+    if (!strcasecmp(key, "Connection")) {
 
       if (!strcasecmp(val, "Keep-Alive\r\n")) {
         keepalive = true;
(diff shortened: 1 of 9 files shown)


abh3 (commented):
Given that comparisons that ignore case are rather inefficient, we can avoid it (at least for external handlers) by always converting the "key" to lowercase and also using that as the insert key in the allheaders map (XrdHttpReq.cc:149). Then we make sure that whenever we try to see if a header is present in the map or when seeing if the header key is one we are interested in, we always use the lowercase version of the value. That would remove the need for a case insensitive template for find and avoid converting the case for each find() invocation (i.e. it's only done once)., and we continue to use the more efficient strcmp()  What do you think?

I'll re-review this once you consider the alternative approach. Also see my question on the perhaps misplaced include.

ccaffy:
I also considered it but refrained to do that because if there are external handlers that are looking at header key without lower-casing them, we will break those handlers. I'm fine with that, of course as it simplifies everything for us, but quid for other external handlers? EOS already does a lower-case conversion of header keys so we are safe with that respect: https://gitlab.cern.ch/dss/eos/-/blob/master/mgm/http/xrdhttp/EosMgmHttpHandler.cc?ref_type=heads#L232

abh3:
Unfortunately, we will break something, I am sure. Fortunately, I don't think there are any exthandlers that are supplied by third parties as far as I know. The only solution for that problem is to include both versions of the key in the map. If we go down that route then we should have an option that can enable it as an emergency workaround if it breaks an exthandler (i.e. the default is to not include both versions). We don't even need to document it as it would be used t provide time to fix the exthandler.

ccaffy:
But with my current implementation, we don't break anything right? I do not touch what is in `allheaders` nor the headers passed to the extHandlers. I just compare using strcasecmp and wherever `find` is needed, I use a case-insensitive `find`. This allows exthandlers not implemented in XRootD to still work. Are we really worried about the efficiency of header parsing?
If yes, then we could go for what you proposed initially which is, lower-case every header that we receive from the client (with the risk of breaking unknown external handlers)

abh3:
I talked to @bbockelm and he doesn't know of any 3rd party exthandlers and said doing the lowercase implementation would be the approach he would take and would not bother with the fallback option (that's a bit more severe than I like so I'd put in the fallback). So, he's convinced.

ccaffy:
OK thanks, will then do this fallback mechanism in addition so we are extra sure!

abh3:
Thanks Cedric! Please give me a poke when you have V2 ready. I might not be able to get to it until Sunday depending on when it's ready.

ccaffy:
@abh3, here's a new PR with the new functionality (https://github.com/xrootd/xrootd/pull/2266) , we can close this one if you want.

abh3:
Superseded by 2266 so we can close this one.

PR #2266, opened 2024-05-17
[XrdHttp/TPC/Macaroons] Client header parsing is now case-insensitive

### LINKED ISSUE #2259: [XrdHttp] Header parsing should be made case-insensitive

ccaffy:
Fixes #2259

Did a separate PR, please feel free to reject the previous one :)

There's a new option that one can set in the server config file `http.keepheadercase yes`. If it is set, the headers coming from the client that are **not** lower-case will be inserted as-is in the `allheaders` map in addition to their lower-case version.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -113,6 +113,7 @@ XrdNetPMark * XrdHttpProtocol::pmarkHandle = nullptr;
 XrdHttpChecksumHandler XrdHttpProtocol::cksumHandler = XrdHttpChecksumHandler();
 XrdHttpReadRangeHandler::Configuration XrdHttpProtocol::ReadRangeConfig;
 bool XrdHttpProtocol::tpcForwardCreds = false;
+bool XrdHttpProtocol::keepHeaderCase = false;
 
 XrdSysTrace XrdHttpTrace("http");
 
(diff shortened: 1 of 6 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2850,6 +2852,17 @@ int XrdHttpProtocol::xauth(XrdOucStream &Config) {
...
 }
 
+int XrdHttpProtocol::xkeepheadercase(XrdOucStream &Config) {
+  char *val = Config.GetWord();
+  if (!val || !val[0])
+  {eDest.Emsg("Config", "keepheadercase argument not specified"); return 1;}
+
+  if(!strcmp("yes",val)) {
+    keepHeaderCase = true;
+  }
```
abh3, line 2862: There should be a else keepHeaderCase = false; to take care of the case
http.keepheadercase yes
http.keepheadercase no
Yes, people do that.

abh3 (changes requested):
Looks good. I am asking for one minor change to take care of some eccentric users and their use cases.

src/XrdHttp/XrdHttpProtocol.cc
ccaffy: That's totally correct, just force-pushed with that change included!

abh3: Something wennt wrong. The change does not appear in your new ush. In fact, the commit count is still 1. So, it shows the change has not been done. Can you try again?

ccaffy: Sorry yes, pushed the wrong branch ;) Now it should be good...

abh3 (approved):
Looks good, thanks.

PR #2269, opened 2024-05-17
Defer or disable TLS client authentication

bbockelm:
When users visit XRootD-based HTTP servers in the browser, they get a pop-up request for selecting a client certificate.  This is fairly disruptive for users not working in X.509-based environments -- and completely unnecessary if X.509 authentication is not in use!

This PR adds two new options, `tlsclientauth` and `tlsrequiredprefix`, that control when the HTTPS server may request a client certificate.  Note that this only makes sense for the HTTPS protocol and hence isn't part of the generic XrdTls configuration.

<hr>

`http.tlsclientauth [on|off|defer]`

Controls the use of client certificate authentication *only* for the HTTPS protocol.  When *on* (default), the HTTPS protocol server will always request a client certificate from the client.  When *off*, it will never request a certificate.  When set to *defer*, it will only request a client certificate is it is in a path specified by `https.tlsrequiredprefix`.

It is not an error if the client does not provide a certificate.

`http.tlsrequiredprefix /prefix`

This option, which can be specified multiple times, specifies a prefix that requires the use of TLS client certificate authentication.  If a HTTP request is made for a path starting with this prefix and `http.tlsclientauth` is in defer mode, then the HTTP server will request a client certificate from the client before a response is made.

Note this only works with clients that support the post-handshake authentication option in TLS 1.3 (these include curl 7.62 or later and the curl version shipped with RHEL8 or later).

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -139,6 +139,7 @@ static const int hsmOn   =  1; // Dual purpose but use a meaningful varname
 
 int  httpsmode = hsmAuto;
 int  tlsCache  = XrdTlsContext::scOff;
+bool tlsClientAuth = true;
 bool httpsspec = false;
 bool xrdctxVer = false;
 }
@@ -557,7 +558,7 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
       strcpy(SecEntity.prot, "https");
 
       // Get the voms string and auth information
-      if (HandleAuthentication(Link)) {
+      if (tlsClientAuth && HandleAuthentication(Link)) {
           SSL_free(ssl);
           ssl = 0;
           return -1;
@@ -1082,6 +1083,7 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
       else if TS_Xeq("httpsmode", xhttpsmode);
       else if TS_Xeq("tlsreuse", xtlsreuse);
       else if TS_Xeq("auth", xauth);
+      else if TS_Xeq("tlsclientauth", xtlsclientauth);
       else {
         eDest.Say("Config warning: ignoring unknown directive '", var, "'.");
         Config.Echo();
@@ -1910,6 +1912,9 @@ bool XrdHttpProtocol::InitTLS() {
       {eDest.Say("Config failure: ", "Unable to set allowable https ciphers!");
        return false;
       }
+
+// Enable or disable the config in the context
+   xrdctx->SetTlsClientAuth(tlsClientAuth);
 
 // All done
 //
@@ -2963,6 +2968,24 @@ int XrdHttpProtocol::xtlsreuse(XrdOucStream & Config) {
 //
    eDest.Emsg("config", "invalid tlsreuse parameter -", val);
    return 1;
+}
+
+int XrdHttpProtocol::xtlsclientauth(XrdOucStream &Config) {
+  auto val = Config.GetWord();
+  if (!val || !val[0])
+     {eDest.Emsg("Config", "tlsclientauth argument not specified"); return 1;}
+
+  if (!strcmp(val, "off"))
+     {tlsClientAuth = false;
+      return 0;
+     }
+  if (!strcmp(val, "on"))
+     {tlsClientAuth = true;
+      return 0;
+     }
+
+  eDest.Emsg("config", "invalid tlsclientauth parameter -", val);
+  return 1;
 }
 
 int XrdHttpProtocol::xauth(XrdOucStream &Config) {
--- a/src/XrdTls/XrdTlsContext.cc
+++ b/src/XrdTls/XrdTlsContext.cc
@@ -706,7 +706,7 @@ XrdTlsContext::XrdTlsContext(const char *cert,  const char *key,
 // depth and turn on peer cert validation. For now, we don't set a callback.
 // In the future we may to grab debugging information.
 //
-   if (caDir || caFile)
+   if ((caDir || caFile) && !(opts & clcOF))
      {if (!SSL_CTX_load_verify_locations(pImpl->ctx, caFile, caDir))
          FATAL_SSL("Unable to load the CA cert file or directory.");
 
@@ -1139,4 +1139,15 @@ bool XrdTlsContext::newHostCertificateDetected() {
         }
     }
     return false;
+}
+
+void XrdTlsContext::SetTlsClientAuth(bool setting) {
+    bool LogVF = (pImpl->Parm.opts & logVF) != 0;
+    if (setting)
+       {pImpl->Parm.opts &= ~clcOF;
+        SSL_CTX_set_verify(pImpl->ctx, SSL_VERIFY_PEER, (LogVF ? VerCB : 0));
+       } else
+       {pImpl->Parm.opts |= clcOF;
+        SSL_CTX_set_verify(pImpl->ctx, SSL_VERIFY_NONE, 0);
+       }
 }
--- a/src/XrdTls/XrdTlsContext.hh
+++ b/src/XrdTls/XrdTlsContext.hh
@@ -174,6 +174,21 @@ void            SetDefaultCiphers(const char *ciphers);
       bool      SetCrlRefresh(int refsec=-1);
 
 //------------------------------------------------------------------------
+//! Indicate how the server should handle TLS client authentication.
+//!
+//! @param  setting true:  All clients will be asked to send a TLS client
+//!                        certificate.
+//!                 false: No clients will be asked to send a TLS client
+//!                        certificate.
+//!
+//! Note the TLS connection will not fail if the client is asked for a cert
+//! but none are provided.
+//!
+//------------------------------------------------------------------------
+
+       void     SetTlsClientAuth(bool setting);
+
+//------------------------------------------------------------------------
 //! Check if certificates are being verified.
 //!
 //! @return True if certificates are being verified, false otherwise.
@@ -239,6 +254,7 @@ static const uint64_t crlFC = 0x000000C000000000; //!< Full crl chain checking
 static const uint64_t crlRF = 0x00000000ffff0000; //!< Mask to isolate crl refresh in min
 static const int      crlRS = 16;                 //!< Bits to shift   vdept
 static const uint64_t artON = 0x0000002000000000; //!< Auto retry Handshake
+static const uint64_t clcOF = 0x0000010000000000; //!< Disable client certificate request
 
        XrdTlsContext(const char *cert=0,  const char *key=0,
                      const char *cadir=0, const char *cafile=0,
--- a/src/XrdHttp/XrdHttpProtocol.hh
+++ b/src/XrdHttp/XrdHttpProtocol.hh
@@ -225,6 +225,7 @@ private:
   static int xhttpsmode(XrdOucStream &Config);
   static int xtlsreuse(XrdOucStream &Config);
   static int xauth(XrdOucStream &Config);
+  static int xtlsclientauth(XrdOucStream &Config);
   
   static bool isRequiredXtractor; // If true treat secxtractor errors as fatal
   static XrdHttpSecXtractor *secxtractor;
--- a/tests/XRootD/CMakeLists.txt
+++ b/tests/XRootD/CMakeLists.txt
@@ -12,8 +12,9 @@ endif()
 # Note: HAVE_SCITOKEN_CONFIG_SET_STR is required to override the CA file
 # for the scitokens library; otherwise, the test will fail with TLS errors.
 if(BUILD_SCITOKENS AND HAVE_SCITOKEN_CONFIG_SET_STR)
-  list(APPEND XROOTD_CONFIGS scitokens)
+  list(APPEND XROOTD_CONFIGS scitokens httpnoclient)
   list(APPEND scitokens_FIXTURES SciTokens)
+  list(APPEND httpnoclient_FIXTURES SciTokens)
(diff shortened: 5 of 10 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -672,6 +709,49 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
 
   }
 
+
+#if OPENSSL_VERSION_NUMBER >= 0x10100010L
+  if (postheaderwait) {
+    postheaderwait = false;
+    if (SSL_verify_client_post_handshake(ssl) != 1) {
```
abh3, line 716: We tried very very hard to keep all OpenSSL calls from not spreading outside of the XrdTls class. This destroys that principal. Could you please come up with a containment mechanism that moves this code as a function of TlsContext or TlsSocket, as appropriate? This really eases the maintenance burden,.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -446,6 +448,23 @@ static long BIO_XrdLink_ctrl(BIO *bio, int cmd, long num, void * ptr)
   case BIO_CTRL_FLUSH:
     ret = 1;
     break;
+  case BIO_C_SET_NBIO:
```
abh3, line 451: I see this addition but I don't see any use of it. Could you point me to the code that uses this case?

abh3 (changes requested):
In addition to the requested changes. Since "Note this only works with clients that support the post-handshake authentication option in TLS 1.3 (these include curl 7.62 or later and the curl version shipped with RHEL8 or later)." the obvious question is what happens if you set this option but the client doesn't support it? Failure,? Odd error? Since we have to support a wide mix of clients , getting into a situation where sometimes this works and sometimes not is not particularly appealing. What is the mitigation here?

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Look around line 730 in this file -- `BIO_set_nbio(sbio, 1);`.

`SSL_peek_ex` relies on the BIO implementing non-blocking IO to function (well, to function without blocking).

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: This is all manipulation of the SSL socket so the natural place for this would be to live in `XrdTlsSocket`.

However, there's zero use of `XrdTlsSocket` in `XrdHttp` - it utilizes the SSL context directly.  Introducing and using it for only this small part of the handshake doesn't seem practical.  If we bring `XrdTlsSocket` in, it wants to own the SSL context object and manipulates its internals.

So, short of a reimplementation of the guts of the XrdHttpProtocol object to use XrdTlsSocket, is there any way to provide a containment mechanism?

xrootd-dev:
The short answer is not yet. Actually, Http uses the TlsContext for certain things. What you are seeing is that it is passing the SSL context to the xrtractor plugins which was code written before we had a TLSContext class. It was unfortunate but something we could not eliminate until a major release. So, that was what we planned to do to finally centralize  OpenSsl code dependencies.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2851,6 +2908,39 @@ int XrdHttpProtocol::xtlsreuse(XrdOucStream & Config) {
...
+  if (!strcmp(val, "off"))
+     {tlsClientAuth = XrdTlsContext::ClientAuthSetting::kOff;
+      return 0;
+     }
+  if (!strcmp(val, "on"))
+     {tlsClientAuth = XrdTlsContext::ClientAuthSetting::kOn;
+      return 0;
+     }
+  if (!strcmp(val, "defer"))
+     {tlsClientAuth = XrdTlsContext::ClientAuthSetting::kDefer;
```
turetske: There needs to be a `return 0;` below this line.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2833,6 +2911,45 @@ int XrdHttpProtocol::xtlsreuse(XrdOucStream & Config) {
...
+      return 0;
+     }
+  if (!strcmp(val, "on"))
+     {tlsClientAuth = XrdTlsContext::ClientAuthSetting::kOn;
+      return 0;
+     }
+  if (!strcmp(val, "defer"))
+     {
+#if OPENSSL_VERSION_NUMBER >= 0x10100010L
+     tlsClientAuth = XrdTlsContext::ClientAuthSetting::kDefer;
```
turetske, line 2930: There's no return if "defer" is supported, there needs to be a `return 0` line.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -685,6 +697,47 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
...
   }
 
+
+  if (postheaderwait) {
+    postheaderwait = false;
+    if (SSL_verify_client_post_handshake(ssl) != 1) {
+      // This is hit if the remote client doesn't support the post-handshake authentication
+      // (curl, Mac OSX) or TLS v1.3 (RHEL7).
+      TRACEI(ALL, "Unable to request client X.509 authentication");
+      ERR_print_errors(sslbio_err);
```
turetske, line 709: Should this have a `return -1;`? Without it, the code will treat it as authentication succeeded for the requested prefixes even when the remote client doesn't support the post-handshake authentication.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Fixed in the latest patch revision.

bbockelm:
(Note: converting the PR back to "draft" status until Emma has a chance to re-test)

bbockelm:
Summary of in-person discussion:
- Brian will move new OpenSSL API invocation into `XrdTls`, even if it's a static function in the directory.
- Similarly, any `#ifdef` around OpenSSL version in this PR will be migrated to the `XrdTls` directory.
- The implementation of the BIO nonblocking, since it's in the middle of the existing BIO configuration function in HTTP, will be kept where it is.

bbockelm:
@abh3 - requested changes were completed.

abh3:
So our reviewers have raised concerns tat this patch is not a practical solution and would lead to even more confusion. Specifically, the patch relies on using certain browsers and is not a uniform solution. As David Smith says....

This is what Louis Regnier of the FTS team found when confronted with "Cannot perform Post-Handshake Authentication" errors when using newer browsers against the FTS web-monitoring interface using TLS version 1.3:

https://its.cern.ch/jira/browse/FTS-2037

bbockelm:
@abh3 - as the old saying goes, "that's a feature, not a bug".

FTS is getting into trouble because they are deferring the post-client auth *but* making it mandatory for some requests(meaning any client that doesn't support the feature will get an error).  As they discovered in that ticket, it's a bad idea because browsers can do global TLS at handshake time but not post-handshake (unless you enable a hidden feature no one enables), meaning what was previously working fine is now broken.

This approach is different and it's part of slowly phasing out X.509 auth in our infrastructure.  We make the post-handshake authentication _optional_, meaning there's no error served up to the client.  Further, we only want X.509 to be used from command-line clients (e.g., libcurl) and _never_ in the browser.

src/XrdTls/XrdTlsContext.cc
```diff
@@ -1151,7 +1151,13 @@ void XrdTlsContext::SetTlsClientAuth(ClientAuthSetting setting) {
          SSL_CTX_set_verify(pImpl->ctx, SSL_VERIFY_NONE, 0);
          break;
       case kDefer:
+#if OPENSSL_VERSION_NUMBER < 0x10100010L
```
amadio, line 1154: I'd use here `#ifdef SSL_VERIFY_POST_HANDSHAKE`...

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -652,6 +657,20 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
           TRACE(DEBUG, " Parsing of first line failed with " << result);
           return -1;
         }
+
+
+        // We permit TLS client auth to be deferred until after the request path is sent.
+        // If this is a path requiring client auth, then do that now.
+        if (tlsClientAuth == XrdTlsContext::ClientAuthSetting::kDefer)
+           {for (const auto &prefix : tlsAuthRequestPrefixes) {
+               {if (!strncmp(prefix.c_str(), CurrentReq.resource.c_str(), prefix.length()))
```
amadio, line 670: This looks like it could be pretty slow in practice if there are several prefixes to be checked and every new request forces the server to go through this loop. Also, this seems to have the problem that `/foo` prefix configured would force authentication on `/foobar`, which may not be intended (similar to the discussion we've had with tokens). I'm not sure what better alternative we could have to handle this, but I'd like to have tests in the test suite before merging, to ensure that this bit really works as it's a security issue if we get it wrong.

src/XrdHttp/XrdHttpProtocol.cc
amadio: I tend to agree with @bbockelm here, we need to live with the SSL calls as they are.

amadio (changes requested):
Hi @bbockelm, I left some comments for this. I think that we need at least a set of tests for the part that checks if we need auth or not to ensure we don't have a security issue, and I think that it might be good to discuss a bit further how to offer this configuration (i.e. the prefixes that need auth) before adding this to a release.

Since `XrdTlsContext::ClientAuthSetting` is only used by `XrdHttpProtocol` it might not make sense to add this into `XrdTlsContext`, we could keep it local to `XrdHttpProtocol` itself.

A final thing is that it would be better to confine this feature as much as possible to the `HandleAuthentication` method, maybe adding new arguments or another method with extra arguments for handling possibly deferred authentications. We could pass in the path and a state which encodes what stage we are in (auth not done, auth deferred, auth done).

amadio:
One more comment, please restructure the commits such that each still keeps everything working everywhere. I see that some commits add changes that break on some versions of OpenSSL, and ifdefs are added only later. We'd like to avoid that if possible (you can just combine that commit with the previous one). Thanks!

bbockelm:
Some good news on this PR --

We've carried this out-of-band for so long that the driving use case was able to retire their use of X.509 client authentication.  That means we don't need to use the TLS v1.3 extension anymore -- we just need an "on/off switch" for X.509 client authentication, which will drastically simplify things (I think all the pending comments above were related to the deferred TLS auth).

Haven't had the time to revise the PR but wanted to at least share the good news.

bbockelm:
As mentioned a few weeks back, I'm quite happy to share that it seems like we made the needed transition with this patch out-of-tree, *greatly* simplifying this to be an on/off and dropping the use of the TLS 1.3 extension.

Accordingly, as this is pretty much a rewrite (particularly, removing the most controversial pieces), I'm dismissing the old reviews and re-requesting a fresh review.

bbockelm:
@abh3 / @amadio - kind reminder of this PR.  In case it wasn't clear from the previous comments, it's been completely reworked since last discussed.  Hence, I re-requested a review in GitHub.

amadio:
Hi @bbockelm, we will review it and consider it for 5.8. Thanks!

bbockelm:
@amadio - thanks for the review!  I just wanted to ACK I saw it and will try to get to this over the weekend.

src/XrdTls/XrdTlsContext.cc
```diff
@@ -591,6 +592,8 @@ XrdTlsContext::XrdTlsContext(const char *cert,  const char *key,
          SSL_CTX **ctxLoc;
         } ctx_tracker(&pImpl->ctx);
 
+   pImpl->opts = opts;
```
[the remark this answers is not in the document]
bbockelm: Fixed in the latest revision.

src/XrdTls/XrdTlsContext.hh
```diff
@@ -173,6 +173,27 @@ void            SetDefaultCiphers(const char *ciphers);
 
       bool      SetCrlRefresh(int refsec=-1);
 
+enum ClientAuthSetting {
+      kOn,
+      kOff,
+};
```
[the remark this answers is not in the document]
bbockelm: Changed in the latest revision (previously, this was ternary logic); now a boolean.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -510,6 +511,10 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
       if (!ssl) {
           sbio = CreateBIO(Link);
           BIO_set_nbio(sbio, 1);
+          if (!xrdctx->SetTlsClientAuth(tlsClientAuth)) {
```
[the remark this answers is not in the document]
bbockelm: Unfortunately, the periodic update of SSL contexts were changing the `tlsClientAuth`.

In the latest revision, I took a new tack: I put the value in the context's configuration instead so newly-created contexts should have the same parameters.

bbockelm:
@amadio - in the latest revision, I overhauled the patch so setting the context's configuration each time was no longer needed and to include integration tests that the patch was working.  Life is better, of course, with test coverage.

bbockelm:
@amadio - Performed the investigation requested above.

There's no way for the client to "send unprompted" the certificate (it's a protocol violation).  Thus, there's no way to disable the server from prompting for a client certificate but verifying one if the client sends it regardless.

Per our conversation last month, I think this is the best we can do.  I think it's ready to merge for the next feature release.

Issue #2273, opened 2024-05-19
[XrdHttp] Regression in header parsing


bbockelm:
In 7514bcb4dbf9c83110c1c28b8f531f32ebf8e6a7 / #2255, @ccaffy introduced the following code:

```c
        // A \r\n was detected
        if(BuffUsed()) {
          if(CurrentReq.request == CurrentReq.rtUnset){
            // The first line has a "\n" at the beginning
            // Bad request
            return SendSimpleResp(400, nullptr, nullptr, "Got first line break before the HTTP verb", 0, false);
          } else {
            // We are getting the headers from the buffer and we detected the header end.
            // Though, there is still data to read from the buffer, continue to read!
            continue;
          }
        }
        // We have detected the header end and no data remains in the buffer, we stop this loop
```

The `else` case here is not correct.  After the `\r\n`, the client has indicated the end-of-headers and the remainder of the buffer is the body of the request.

I suspect it might have worked in hand-testing -- probably many clients have a slight pause between the headers and the body that cause the request to be split into two packets.  However, testing with my `golang` client causes an immediate failure.

Symptoms of issue: after issuing the request, the client hangs indefinitely because the server believes there are more headers coming (it's not handling the body) while the client is waiting for the response.

Example log output (removing some Pelican wrapper text and empty lines):
```
240519 09:57:32 5379 http_Protocol: read 525 of 1048576 bytes
240519 09:57:32 5379 http_Protocol:  rc:60 got hdr line: PROPFIND /test/SomeDirectoryThatDoesNotExist:%29/ HTTP/1.1
240519 09:57:32 5379 http_Protocol:  Parsing first line: PROPFIND /test/SomeDirectoryThatDoesNotExist:%29/ HTTP/1.1
240519 09:57:32 5379 http_Protocol:  rc:30 got hdr line: Host: F4HP7QL65F.local:65381
240519 09:57:32 5379 http_Protocol:  rc:32 got hdr line: User-Agent: pelican-client/dev
240519 09:57:32 5379 http_Protocol:  rc:21 got hdr line: Content-Length: 203
240519 09:57:32 5379 http_Protocol:  rc:34 got hdr line: Accept: application/xml,text/xml
240519 09:57:32 5379 http_Protocol:  rc:23 got hdr line: Accept-Charset: utf-8
240519 09:57:32 5379 http_Protocol:  rc:19 got hdr line: Accept-Encoding: 
[...]

PR #2275, opened 2024-05-22
2273 regression in header parsing

ccaffy:
I've reverted the previous commit creating a regression and just added a check that ensures each header line provided by the client terminates with `\r\n`. If that is the case, a 400 Bad request error is replied with an error message saying that the headers are malformed.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -125,7 +125,7 @@ int XrdHttpReq::parseLine(char *line, int len) {
   if (!p) {
 
     request = rtMalformed;
-    return 0;
+    return -1;
   }
 
   pos = (p - line);
@@ -145,6 +145,10 @@ int XrdHttpReq::parseLine(char *line, int len) {
     // We memorize the headers also as a string                                                                                                                                              
     // because external plugins may need to process it differently                                                                                                                          
     std::string ss = val;
+    if(ss.length() >= 2 && ss.substr(ss.length() - 2, 2) != "\r\n") {
+      request = rtMalformed;
+      return -3;
+    }
     trim(ss);
     allheaders[key] = ss;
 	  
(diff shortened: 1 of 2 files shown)


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -145,6 +145,10 @@ int XrdHttpReq::parseLine(char *line, int len) {
     // We memorize the headers also as a string                                                                                                                                              
     // because external plugins may need to process it differently                                                                                                                          
     std::string ss = val;
+    if(ss.length() >= 2 && ss.substr(ss.length() - 2, 2) != "\r\n") {
```
abh3, line 148: So, I am confused. If the recommendation is to allow both \r\n and simple \n, why is the added code ignoring the recommendation in RFC-2616 19.3? Is it to be consistent with how weirdly the Connection header value is parse? Anyway, looking closely at the pre-existing code (i.e. not part of this pull request), I can see lots of issues that may bite us in the future.

abh3 (commented):
Before I approve, could you comment on the issue I raised so we both are clear on what is happening here?

src/XrdHttp/XrdHttpReq.cc
ccaffy: The reason for that is that Brian said that we can also ignore the RFC-2616 19.3 and send a bad request in case we see header lines not finishing with `\r\n`: https://github.com/xrootd/xrootd/issues/2273#issuecomment-2122679734

abh3 (approved):
Thanks, we'll see how hat works out but should be no worse than what we had before anyway.

ccaffy:
@bbockelm , this got merged, please try the devel branch with your go client :)

src/XrdHttp/XrdHttpReq.cc
bbockelm: I think my comment was a bit more subtle -- but I'm OK with this implementation.  My comment suggested is to  permit `\n` only if the client consistently uses it (that is, no mixing of `\n` and `\r\n` in the same request).

In fact, I looked up RFC 7230 which is a 2014 update that obsoletes RFC 2616.  In [section 3.5](https://datatracker.ietf.org/doc/html/rfc7230#section-3.5), they say that clients *MAY* permit use of `\n`-only.  This is quite different than the prior, circa 1999, advice of *SHOULD* accept such requests.

So yes: I'm OK with saying we've evolved beyond the advice from the late 1990's and can assume clients correctly frame requests by default.  Especially if assuming so allows us to better detect malformed requests or client bugs (such as this one!).

Issue #1730, opened 2022-06-29
Stackable OSS plugins silently fail in cmsd


bbockelm:
I noticed the following in the cmsd logs:

> Config warning: stacked plugins are not supported in this context; directive ignored!
> =====> ofs.osslib ++

It appears that the cmsd silently ignores the stacking, causing the cmsd and xrootd daemons to have differing views of the filesystem.  Could we:

- Enable stacking for cmsd?  It's truly needed here and a few sites have only been functioning by chance.  Right now, I'm just going to give them instructions on how to configure the plugins to do the stacking internally (like what occurred in XRootD 4) which is quite unpleasant.
- Upgrade the stacked plugin failure from a warning to a fatal error?

abh3:
Assuming you are not using forwarding (which we don't really recommend) The fastest way to solve this is to use the "stat" oss plugin and provide whatever view you want from there. The cmsd only relies on Stat() and StatVS() (for writable storage).

bbockelm:
Hi @abh3!

Does the stat oss plugin support stacking?  Using a regular oss plugin works fine here -- I just have to dlopen the underlying plugin and stack myself (which is precisely what we were trying to avoid in V5, no?).

What are your thought on making a stacked plugin failure a fatal error for startup so users aren't surprised?

Brian

abh3:
I agree, we don't want people to manually load stuff. My suggestion here was with the thought that stacking is overkill since all you need is to provide a custom stat() which is what the stat plugin allows you to do. Doesn't seem like that would be complicated to the point of needing to stack anything.

As far as making it a fatal error, well it depends. That is a rather big change in config behavior and we generally like to avoid that kind of thing. I don't have a feel of how many people have tried stacking but I suspect it's only you as no one else has brought this up.

bbockelm:
> I suspect it's only you as no one else has brought this up.

I would be anyone using multiuser + data federation is hitting this.  Once I realized that this was a potential problem, I reached out to Nebraska and noticed they were hitting it as well but had just marked all files world-readable (which is a way around things, I suppose!).

> My suggestion here was with the thought that stacking is overkill since all you need is to provide a custom stat() which is what the stat plugin allows you to do.

In theory, the multiuser plugin can go on top of anything ... in practice, I suppose that it's only stacked on top of the default (POSIX) plugin.  This would require copy/pasting a bunch of code from the default plugin ... but suppose it beats copy/pasting the entire OSS plugin.

[...]

PR #2279, opened 2024-05-31
Add OSS statistics/monitoring plugin

bbockelm:
This PR adds a new OSS implementation to XRootD for gathering statistics about the underlying OSS performance.

It keeps counters about the number of operations performed and the overall time spent performing these operations then sends them out to the g-stream.  It keeps a separate set of counters for "slow" operations (configurable), providing insight to administrators about the user experience beyond the averages.  The intent is this information can easily be passed to time-series monitoring system such as Prometheus.

This adds two configuration options:

> `fsstats.trace [all|debug|info|warning|error]`
>
> `fsstats.trace` manages the verbosity of the logging done by the OSS statistics library.  When debug
> information is enabled, it will print out the performance counters once per second.
>
> If `fsstats.trace` is not specified, then the default is `fsstats.trace warning`.

and

> `fsstats.slowops [duration]`
> 
> The fsstats.slowops directive determines the threshold where a given storage operation is
> considered "slow".  Slow operations are kept as a separate set of monitoring counters, allowing
> the administrator to see how often a storage system has poor performance.
>
> The duration must be specified as a number with a unit such as `1.5s` or `1s50ms`; valid units
> include `h` (hours), `m` (minutes), `s` (seconds), `ms` (milliseconds), `us` (microseconds), or
> `ns` (nanoseconds).
>
> If `fsstats.slowops` is not specified, then the default is `fsstats.slowops 1s`.

While the intent is the implementation is lightweight statistics gathering, everything has a cost; hence, this is kept as a loadable plugin instead of in the XRootD core.

The initial use case is to understand the performance of the storage subsystems in XCache.  Currently our only performance monitoring is from looking at the throttle plugin.  The throttle plugin includes both the storage performance for cache hits and the proxy performance for cache misses; I want to exclude the latter.


amadio:
I've requested a review from @abh3, but first the CI needs to be fully green.

abh3:
Sure, I'll review it but it will take me a bit longer than normal as I will be busy with an ATLAS meeting in Oslo.

amadio:
> Sure, I'll review it but it will take me a bit longer than normal as I will be busy with an ATLAS meeting in Oslo.

This needs a major release, I believe, so no need to rush. Cheers,

abh3 (changes requested):
So I am confused as to what this is. A real storage system or a wrapper to a storage system. My inclination is that this is a wrapper and, if so, i9t does not follow the conventions for wrapping an oss plugin. First, it should use the class defined in XrdOssWrapper.hh because doing so alleviates us from having to chase down  yet another plugin API when  we make future changes in the Oss API. Second, as a wrapper it should not have a GetFileSystem entry point as it can only be added using the "++" convention. Thirdly, a wrapper should not load or otherwise initialize the file system it is wrapping. That is done by the plugin system. So, that code is not appropriate here.  So, could this be rewritten to follow the standard way Oss plugins wrappers should be written?

bbockelm:
> First, it should use the class defined in XrdOssWrapper.hh because doing so alleviates us from having to chase down yet another plugin API when we make future changes in the Oss API

Are there any examples of using these classes?  I looked at the wrapper and it appears simple enough but it'd be good to know if I'm the only user of the class.

> Second, as a wrapper it should not have a GetFileSystem entry point as it can only be added using the "++" convention. Thirdly, a wrapper should not load or otherwise initialize the file system it is wrapping.

Sadly, it's not possible to rely on the `++` syntax as you suggest due to https://github.com/xrootd/xrootd/issues/1730.  Unfortunately, that issue means a wrapper writer needs to implement exactly what you say the wrapper should not do.  If you are able tackle issue #1730, then I'd happily drop the other entry points (on this and other plugins).

abh3:
Hi Brian,

No, you are not the only user.  We have an upcoming tape RSE plugin, and we will be converting existing wrapper plugins into using the official wrapper.

Andy

abh3:
As for 1730 that is a cmsd problem and as outlined in the issue, the only thing the cmsd wants to use is stat() which itself is replaceable via the stat plugin. So practically there is no reason to stack the cmsd plugin. I suspect this plugin would not be used for the cmsd as well. So, I think you can ignore issue 1730. I agree that the cmsd should not be silent in this case.

bbockelm:
@abh3 - updated to rebase on the `XrdOssWrapper` and removed the support for running as part of the `cmsd`.

Is the `XrdOssWrapper` part of the public API/ABI yet?  The design is awkward to utilize because it does not own the `wrapDF` object.  Hence, I have to keep two copies of the pointer, once in `StatsFile` (to own the lifetime) and once in the `XrdOssWrapDF` (a reference).

If the API is not frozen, it'd be better to change the wrapper to own the object to manage the lifetime.

xrootd-dev:
Yes, it is part of the public API. Yes,  you cannot control the lifetime of Oss system object because you should be using the passed pointer whose lifetime you do not control. However, you are correct that the lifetime of the DF object is something you do control. We need to wait for R6 to change that. Unfortunately, keeping a separate pointer is the only way to handle the lifetime situation for now.

bbockelm:
@abh3 - just a bump on this PR (I see I didn't press the "re-review" button; I'll do that now).  If it wasn't obvious from the history, I did the refactoring you requested back in July.  My follow-up questions were simply around the API design.

abh3:
Will do, it's a bit hectic at the moment with the workshop less than 2 weeks away.

src/XrdOssStats/XrdStatsFileSystem.hh
```diff
@@ -0,0 +1,128 @@
...
+            std::atomic<uint64_t> &m_op_count;
+            std::atomic<uint64_t> &m_slow_op_count;
+            std::atomic<uint64_t> &m_timing;
+            std::atomic<uint64_t> &m_slow_timing;
+            std::chrono::steady_clock::time_point m_start;
+            std::chrono::steady_clock::duration m_slow_duration;
+    };
+
+    struct OpRecord {
+        std::atomic<uint64_t> m_read_ops{0};
```
abh3, line 92: OK, I understand why you may want to intercept to get timing data which we don't do. But much of this information (i.e. counts) is already reported in the f-stream. So, now we have a duplicate way of doing it and potentiually collecting the majority of the same statistics (and reporting them) twice. So, what's the actual motivation here?

Frankly, if this is a preferred way of doing things now then I would make oss gStream incompatible with f-stream you can do one or the other but not both.

src/XrdPlugins.cmake
```diff
@@ -133,6 +134,22 @@ target_link_libraries(
   XrdServer
   XrdUtils )
 
+#-------------------------------------------------------------------------------
+# An OSS plugin for calculating storage performance statistics
+#-------------------------------------------------------------------------------
+add_library(
+  ${LIB_XRD_OSSSTATS}
+  MODULE
+  XrdOssStats/XrdStatsConfig.cc     XrdOssStats/XrdStatsConfig.hh
```
abh3, line 143: The names here do not follow the standard naming conventopn. The directory name is the prefix of whatever files are in the directory (well, hh and cc files). So, if youi want XrdStatsConfig,cc in XrdOssStats then it really should be named XrdOssStatsConfig.cc. So please introducs the"Oss" into the file names to be consistent with that policy.

And, yes, there are some inconsistencies in the repo that are historical in nature and not worth changing but we don't want to add to it.

src/XrdOssStats/XrdStatsConfig.cc
```diff
@@ -0,0 +1,128 @@
...
+XrdOss *XrdOssAddStorageSystem2(XrdOss       *curr_oss,
+                                XrdSysLogger *logger,
+                                const char   *config_fn,
+                                const char   *parms,
+                                XrdOucEnv    *envP)
+{
+                   
+    XrdSysError log(logger, "fsstats_");
+    try {
+        return new StatsFileSystem(curr_oss, logger, config_fn, envP);
```
abh3, line 119: This where you would hand off curr_oss to the wrapper.

src/XrdOssStats/XrdStatsFileSystem.cc
```diff
@@ -0,0 +1,337 @@
...
+        throw std::runtime_error("Failed to configure the storage statistics plugin.");
+    }
+    pthread_t tid;
+    int rc;
+    if ((rc = XrdSysThread::Run(&tid, StatsFileSystem::AggregateBootstrap, static_cast<void *>(this), 0, "FS Stats Compute Thread"))) {
+      m_log.Emsg("StatsFileSystem", rc, "create stats compute thread");
+      throw std::runtime_error("Failed to create the statistics computing thread.");
+    }
+    if (envP) {
+       m_gstream = reinterpret_cast<XrdXrootdGStream*>(envP->GetPtr("oss.gStream*"));
```
abh3, line 34: I would really move the extraction of the gStream into the AddStorageSystem() method and return failure if you cannot get a gStream. Nothing worse than collecting statistics and not being able to report them. If you don't want to fail, then you could simply issue a message and return the current_ossp pointer and effectively you have disabled yourself with no harm done. You can't do any of that here.

abh3 (changes requested):
OK, let's resolve the comments.

src/XrdOssStats/XrdStatsFileSystem.hh
```diff
@@ -0,0 +1,128 @@
...
+#include <chrono>
+
+class XrdXrootdGStream;
+
+// The "stats" filesystem is a wrapper that collects information
+// about the performance of the underlying storage.
+//
+// It allows one to accumulate time spent in I/O, the number of operations,
+// and information about "slow" operations
+class StatsFileSystem : public XrdOss {
```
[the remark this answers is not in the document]
bbockelm: Strange.  Something want awry with my rebase as the switch to the wrapper (and cleaning up the `#pragma once`) was the core of my second round of changes -- yet they certainly have disappeared!

I'll try and figure out what went wrong with commit.

src/XrdOssStats/XrdStatsFileSystem.hh
bbockelm: Is this a complete overlap though?

My understanding is the f-stream works at the xrootd protocol handler level; this works at the OSS level.  The OSS level, for example, is used directly by XrdPfc and XrdTpc, meaning that this is seeing many reads that aren't reflected in the f-stream.

src/XrdPlugins.cmake
bbockelm: Fixed in latest commit!

src/XrdOssStats/XrdStatsFileSystem.hh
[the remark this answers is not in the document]
bbockelm: Fixed in the latest commit!

src/XrdOssStats/XrdStatsFileSystem.cc
bbockelm: Done in the latest push.

src/XrdOssStats/XrdStatsConfig.cc
bbockelm: I believe this is taken care of in the latest push (though I think the comment is wrong, no?  It's handed off to the base class in the constructor of `StatsFileSystem`).

bbockelm:
Ok, @abh3 - I think I have all the items you highlight in your review addressed in the latest push.

One item you may want to revisit is my comment about f-stream versus g-stream.

amadio:
@bbockelm Could you please look into the CI failures after the latest changes? Thanks!

bbockelm:
@amadio - looks like there was a cdash failure (and an EPEL outage for CentOS 7?  Is there some branch I need to sync again)?

My level of triage doesn't permit me to try and re-run manually, unfortunately.  Can you hit the button?

abh3 (changes requested):
Just a request to cleanup unneeded code.

src/XrdOssStats/XrdOssStatsFileSystem.cc
```diff
@@ -0,0 +1,344 @@
...
+    return wrapPI.Truncate(path, fsize, env);
+}
+
+int       StatsFileSystem::Unlink(const char *path, int Opts, XrdOucEnv *env)
+{
+    OpTimer op(m_ops.m_unlink_ops, m_slow_ops.m_unlink_ops, m_times.m_unlink, m_slow_times.m_unlink, m_slow_duration);
+    return wrapPI.Unlink(path, Opts, env);
+}
+
+int       StatsFileSystem::Lfn2Pfn(const char *Path, char *buff, int blen)
```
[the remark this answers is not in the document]
bbockelm: Good catch!  In the latest patch, I cleaned up these plus a handful of other cases in the other classes where the implementation was redundant.

bbockelm:
@abh3 - I never heard back on this after the latest round of fixes.  Anything remaining that I had missed?

abh3:
Correct, the long closure got in the way. It's on my list of things to do. I also thought of another addition to avoid double statistics which will inevitably happen. We can add the "export-only" when stacking plugins. This would only stack the plugin for exporting and not when using it in the mainline code path. In practice, the need for this plugin was triggered by the fact that it was exported for external use and that bypassed the collection of usage information. It, however, would be better if we could do this programmatically. That is, say the plugin defined the symbol XRD_EXPORT-ONLY then the plugin loaded would know to include that plugin in the export chain not in the mainline chain and we could dispense with yet another config option. What do you think?

src/XrdOssStats/XrdOssStatsFileSystem.cc
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    // While the plugin _does_ print its activity to the debugging facility (if enabled), its relatively useless
+    // unless the g-stream is available.  Hence, if it's _not_ available, we should fail to startup.
+    if (envP) {
+       m_gstream = reinterpret_cast<XrdXrootdGStream*>(envP->GetPtr("oss.gStream*"));
+       if (m_gstream) {
+         m_log.Say("Config", "Stats monitoring has been configured via xrootd.mongstream directive");
+       } else {
+         throw std::runtime_error("XrdOssStats plugin is loaded but it requires the oss monitoring g-stream to also be enabled to be set; try adding `xrootd.mongstream oss ...` to your configuration");
```
abh3, line 39: Hmmm, this prevents me from having a base configuration and able to turn this monitoring on/off via mongstream directive. Instead I have to comment out the plugin stacking to turn stuff off which isn't particularly intuitive. Why not issue a message and return the underneath oss pointer which effectively doesn't load the plugin (well use it anyway). Plus, we do not use throw in the server unless there is no alternative. In your case, if you really want to kill server initialization return a nil pointer. Anyway, this test should be moved to XrdOssAddStorageSystem2() where you can properly handle this case, as described above, without using a throw.

src/XrdOssStats/XrdOssStatsConfig.cc
```diff
@@ -0,0 +1,128 @@
...
+#include "XrdOssStatsConfig.hh"
+#include "XrdOssStatsFileSystem.hh"
+#include "XrdSys/XrdSysError.hh"
+
+#include <sstream>
+
+XrdVERSIONINFO(XrdOssGetFileSystem, Stats);
+
+
+std::string LogMaskToString(int mask) {
```
abh3, line 13: This method and all subsequent file-level methods should reside in an anonymous namespace.

src/XrdOssStats/XrdOssStatsDirectory.hh
```diff
@@ -0,0 +1,45 @@
...
+#define __XRDSTATS_DIRECTORY_H
+
+#include "XrdOuc/XrdOucEnv.hh"
+#include "XrdOss/XrdOssWrapper.hh"
+#include "XrdOssStatsFileSystem.hh"
+#include "XrdSys/XrdSysError.hh"
+
+#include <memory>
+
+class StatsDirectory : public XrdOssWrapDF {
```
abh3, line 12: OK, I see you did move the plugins into XrdOssStats directory. This always forms the basis of subsequent names. For instance, namespace XrdOssStats { class Directory...} which is the closest you can come here. Normally, classes would be named XrdOssStatsDirectory for consistency. However, namespaces to keep the naming consistent are also acceptable. Why do we do this? To ease debugging across multiple compilation units by using a consistent naming scheme. So, I would suggest jiggering the names here.

abh3 (changes requested):
Not quite there yet.

src/XrdOssStats/XrdOssStatsConfig.cc
bbockelm: This is not possible as these functions are (exclusively) used outside the translation unit; an anonymous namespace causes the symbols to only be available within the translation unit.

However, I can use the same trick as the `XrdMacaroons` and `XrdTpc` currently does to manage symbol visibility, preventing these from being seen outside the module.

src/XrdOssStats/XrdOssStatsDirectory.hh
bbockelm: Rename is done in the latest patch.

src/XrdOssStats/XrdOssStatsConfig.cc
bbockelm: The latest patch hides all the appropriate symbols which, I suspect, is the best that can be done for now.

bbockelm:
> We can add the "export-only" when stacking plugins. This would only stack the plugin for exporting and not when using it in the mainline code path. In practice, the need for this plugin was triggered by the fact that it was exported for external use and that bypassed the collection of usage information

I'm not entirely following this comment?  I don't know what it means when a plugin is stacked "for exporting" versus "mainline code path".

The module has to be specifically configured and enabled.  At least for now, I don't think we'll be pushing to have it turned on by default.

abh3:
What I mean marking the plugin for export-only is that plugins are either used internally by direct linkage or via a pointer in the XrdOucEnv.  We call the mechanism of placing the pointer to the plugin in XrdOucEnv as exporting the plugin because now the plugin can be used by modules that did not request the plugin to be actually loaded. Hence the loaded plugin was exported to the external environment. So, an export-only wrapper would only be included in the plugin exported to the XrdOucEnv and not added to the one used internally. In the oss case, we already effectively collect statistics on the Oss plugin use because we collect detailed statistics on the filesystem chain from xrootd's perspective. Using the plugin from the XrdOucEnv bypasses that whole collection chain. So, that was the idea why the statistics plugin would be needed to capture usage of the plugin by external userts. That is what I mean.

src/XrdOssStats/XrdOssStatsConfig.cc
abh3: Could you explain what the proposed solution would be? I would have expected that if anonymous namespaces would not work the alternative is to use a related namespace (e.g. XrdOssStatsUtils) which would at least isolate the various symbols. Anyway, I don't see any material change here so I am a bit puzzled.

src/XrdOssStats/XrdOssStatsConfig.cc
```diff
@@ -0,0 +1,134 @@
...
+                   
+    XrdSysError log(logger, "fsstats_");
+    auto new_oss = new XrdOssStats::FileSystem(curr_oss, logger, config_fn, envP);
+    if (!new_oss) {
+        return nullptr;
+    }
+    std::string errMsg;
+    if (!new_oss->InitSuccessful(errMsg)) {
+        if (errMsg.empty()) { // Initialization failure was non-fatal; just bypass this module.
+            return curr_oss;
```
abh3, line 123: Would be nice if new_oss were deleted as it will never be used.

src/XrdOssStats/XrdOssStatsFileSystem.hh
```diff
@@ -0,0 +1,138 @@
...
+    XrdOucEnv *m_env;
+    XrdSysError m_log;
+
+    class OpTimer {
+        public:
+            OpTimer(std::atomic<uint64_t> &op_count, std::atomic<uint64_t> &slow_op_count, std::atomic<uint64_t> &timing, std::atomic<uint64_t> &slow_timing, std::chrono::steady_clock::duration duration);
+            ~OpTimer();
+
+        private:
+            std::atomic<uint64_t> &m_op_count;
```
abh3, line 89: We still have the std::atomic issue. Changing this is RAtomics is trivial and no other code changes are needed. What's the issue here?

abh3 (changes requested):
No everything was resolved yet. Also a new question on this round of fixes.

src/XrdOssStats/XrdOssStatsFileSystem.hh
bbockelm: Well, this never was brought up previously which is why it was never changed.

Hadn't previously `RAtomic` before, seems like it'd fit this use case well.  Will refactor.

src/XrdOssStats/XrdOssStatsConfig.cc
bbockelm: Ah - the material change is [here](https://github.com/xrootd/xrootd/pull/2279/files#diff-d870c68188db9cdc0db2add613b44f7f158c6d87a5a13b77e257bde565aeb9dcR150).  That manages symbol visibility according to the policy set [here](https://github.com/xrootd/xrootd/pull/2279/files#diff-36ff6e5d1bc1f58e95b09ff08a5f5ee3489330376f8416a65bccbff1e5a148f8).

If you're not familiar with the syntax, it basically says "make any object prefixed with `XrdOssAddStorageSystem2` visible at the global namespace (of the shared module) and otherwise local".  With that, there's no way to use the symbol from outside the shared module (you can run `nm -D` to list all symbols and note it's quite empty).

In terms of interacting with humans, I like the idea of using a separate namespace to clearly indicate "don't touch this".  Projects I've worked with use a sub-namespace called `detail` (as in `XrdOssStats::detail`) and I don't see any clear common practice grepping through the code.  Does `XrdOssStats::detail` sound good to you?

src/XrdOssStats/XrdOssStatsFileSystem.hh
bbockelm: Refactoring done in the latest commit.

src/XrdOssStats/XrdOssStatsConfig.cc
bbockelm: Ah, good spot!  Switched to a `unique_ptr` to make the ownership simpler.

bbockelm:
@abh3 - any news on this front?

I left the one [remaining discussion thread](https://github.com/xrootd/xrootd/pull/2279#discussion_r1919019734) open because I wanted to make sure you understood how the visibility changes work.

src/XrdOssStats/export-lib-symbols
```diff
@@ -0,0 +1,7 @@
+{
```
abh3, line 1: Could you explain why this is needed? By default, the way XrdOssAddStorageSystem2 has been declared automatically makes it visible. Any other symbols are visible when made extern, making everything visible is generally not good practice.

abh3 (commented):
@bbockelm Almost done here. Only one question remains (no code changes needed). We don'y like putting things in the repo that a) are not explainable, b) look like they are not needed. So, could you address the question posed here.

src/XrdOssStats/export-lib-symbols
bbockelm: You might be reading this backward.  This makes all symbols local to the library (not appearing in the symbol table) _except_ for `XrdOssAddStorageSystem2*`.  If I didn't have the exception listed in the `global` section, then it would hide `XrdOssAddStorageSystem2*` as well -- not what we want!

> Any other symbols are visible when made extern, making everything visible is generally not good practice.

That's not how C++ works.  The `extern` keyword specifies the storage class for the variable or the language linkage property for functions.  It *does not* affect symbol visibility in the resulting module.

For example, consider the SciTokens library (which doesn't manage symbol visibility:

```
$ nm -D lib64/libXrdAccSciTokens-5.so | c++filt | grep -v ' U ' | wc -l
176
```

Quite a bit of "stuff" -- all not needed to be visible from the outside because, really, all that's needed is `XrdAccAuthorizeObjAdd` and friends.  Loading it pollutes up the global symbol table.

This is not because `extern` is or is not used -- but because C++ exports all the namespace-level symbols globally by default when dynamic linking.

Compare this to libXrdMacaroons where we do manage the symbol visibility:

```
$ nm -D lib64/libXrdMacaroons-5.so | c++filt | grep -v ' U ' | wc -l
14
```

The list is short enough I can paste it into the comment -- and it looks reasonable:

```
$ nm -D build/release_dir/lib64/libXrdMacaroons-5.so | c++filt | grep -v ' U ' 
                 w __cxa_finalize
000000000000e678 t _fini
                 w __gmon_start__
0000000000004000 t _init
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 w __pthread_key_create
0000000000014618 B SciTokensHelper
0000000000005be0 T XrdAccAuthorizeObjAdd
0000000000014540 D XrdAccAuthorizeObjAdd_
0000000000005c50 T XrdAccAuthorizeObject
00000000000145a0 D XrdAccAuthorizeObject_
0000000000005f20 T XrdHttpGetExtHandler
00000000000144e0 D XrdHttpGetExtHandler_
```

You only have things that are desired to be exported -- and basically required for a shared library (`_init`, `_fini`).

Most of the Xrootd modules export way more than intended: based on your review comment, I wanted to tidy up this one.

abh3 (approved):
You are correct, I read it backwards. Thanks for the explanation.

amadio:
This looks like it's ready for merging. However, if you don't want me to squash commits, please merge fixes into the appropriate commits they are fixing (i.e. the commit with "Address review comments" should just be a fixup of the previous commit, most likely).

amadio:
Updated this to the current `devel` branch. Will merge it once the CI is green.

amadio:
It was easy to fix, so fixed in dcaa75fbed59f367aef4f1f0bcb397837f0705a9, 86ef56b1fb12e52bcae460bd12c2e62ed320a0b4 and 560e360653c5b472e9eea759ee1e96b7254bc7ff.

Issue #2222, opened 2024-03-23
Tokens leaking to the logs


bbockelm:
@abh3 previously requested that I drop an issue if I see a token leaking to the logfiles.  Doing some token work this morning and decided to poke around.

The token in this message is from demo.scitokens.org and hence not a "real" token.

Below are randomly-selected log snippets from two different test servers:

```
240323 07:02:48 5891 http_Protocol:  rc:707 got hdr line: Authorization: eyJhbGciOiJSUzI1NiIsImtpZCI6ImtleS1yczI1NiIsInR5cCI6IkpXVCJ9.eyJzY29wZSI6InJlYWQ6L3Byb3RlY3RlZCIsImF1ZCI6Imh0dHBzOi8vZGVtby5zY2l0b2tlbnMub3JnIiwidmVyIjoic2NpdG9rZW46Mi4wIiwiaXNzIjoiaHR0cHM6Ly9kZW1vLnNjaXRva2Vucy5vcmciLCJleHAiOjE3MjExNzQzMjEsImlhdCI6MTcxMTE3MzczMSwibmJmIjoxNzExMTczNzMxLCJqdGkiOiJkZjk2YzQxYi01ZWNjLTQ4Y2QtYTliZi1lMGU0ZWJkOWU0M2MifQ.Uc-jWy8Xb8cO_qd9pXvlUIf-FlaISKpED9X62i9rURRgq43JQ_8AtnPgxo5HLprzM1wM1jzNjYvRPElZHOr8Z8_fpiZhiCywfi5M9ZBR9MI6qEDlXTDCASt-D6-JUUXEpTk0rP0lh8xvoCfduw6FgdPJfRUg1UBDbCjSQPbTQo9_9JLp3mFJb8rr7IrKIi0R2KuHEcFmVJna7YmRbYTgCk0m2WFEqKyG9woL2mghFLJC8c5q1-ONmvnPizgYVtlYXFSivYveDvv5lfrbKr9VaRc_qWZZxUDPRVhGCIjAlWPCKZ9vbPxreVv0X22bP_6vUtHy1iShIghyVUppbm6-fQ
[...]

PR #2285, opened 2024-06-17
[XrdTpcTPC] Added tpc.header2cgi configuration

ccaffy:
The administrator can now use this option in order to allow some client HTTP headers to be passed to the OFS layer as opaque information. It works exactly the same way as the http.header2cgi configuration. E.g: the active server of a HTTP/TPC PULL transfer receives the header 'ArchiveMetadata' and 'tpc.header2cgi ArchiveMetadata archivemetadata' is configured. If a client submits a 'ArchiveMetadata: test' header with their request, the opaque archivemetadata=test will be provided to the XrdSfsFile open() function.

The implementation was done as follow:
- Put the HTTP http.header2cgi parsing logic to a public static function of the XrdHttpProtocol class (`parseHeader2CGI()`)
- Add a hdr2cgimap attribute to the TPCHandler
- Call this function in both `http.header2cgi` and `tpc.header2cgi` config parsing function
- Refactor and add the logic to append opaque infos to the URL in the `TPCHandler::prepareURL(...)` function

Thanks ahead for your review!

### DIFF
--- a/src/XrdTpc/XrdTpcTPC.cc
+++ b/src/XrdTpc/XrdTpcTPC.cc
@@ -161,6 +161,7 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
 // We need to utilize the full URL (including the query string), not just the
 // resource name.  The query portion is hidden in the `xrd-http-query` header;
 // we take this out and combine it with the resource name.
+// We also append the value of the headers configured in tpc.header2cgi to the resource full URL
 //
 // One special key is `authz`; this is always stripped out and copied to the Authorization
 // header (which will later be used for XrdSecEntity).  The latter copy is only done if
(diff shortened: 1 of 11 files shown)


src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -160,43 +160,56 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
...
+        }
+      }
+    }
+
+    // Append CGI coming from the tpc.header2cgi parameter
+    for(auto & hdr2cgi : hdr2cgimap) {
+        auto it = std::find_if(req.headers.begin(),req.headers.end(),[&hdr2cgi](const auto & elt){
+            // We exclude the Authorization header from the hdr to cgi conversion, as it will be appended later on to the opaque before the open
+            // happens
+            return elt.first != "authorization" && !strcasecmp(elt.first.c_str(),hdr2cgi.first.c_str());
```
amadio, line 201: Maybe you need `strcasecmp(elt.first.c_str(), "authorization")` here, no?

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -160,43 +160,56 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
...
         } else if (!strncmp(token.c_str(), "authz=", 6)) {
-            if (!has_authz_header) {
-                req.headers["authorization"] = token.substr(6);
-                has_authz_header = true;
-            }
-        } else if (!found_first_header) {
-            result << "?" << token;
-            found_first_header = true;
+          if (!has_authz_header) {
+            req.headers["authorization"] = token.substr(6);
```
amadio, line 186: This may need to be `"Authorization"` to work with older servers.

ccaffy:
Correct, I need to correct this one after the other PR is merged :) Will do so!

src/XrdTpc/XrdTpcTPC.cc
ccaffy: No no I cannot do this, Brian seems to use it ;) see https://github.com/xrootd/xrootd/issues/2222#issue-2203719451 the second line

bbockelm: I'm not quite following the comment?

Indeed, we translate `Authorization` -> `authz` in many places.  However, what's the harm in allowing someone to set it again here?  Isn't the worst case that the token is appended twice?

ccaffy: Yes, that's what would happen and I do not know if this is something we care about or not...

ccaffy: Here's what we will get in the open opaque when I pass the `Authorization: Bearer test` header with my request and when the `tpc.header2cgi Authorization authz` is set: `"authz=Bearer test&nbstreams=10&oss.asize=5242880&authz=Bearer%20test"`

ccaffy: Now with the 'authorization' exclusion:
```
nbstreams=10&oss.asize=5242880&authz=Bearer%20test"
```

amadio: What I meant is that the map `tpc.header2cgi` should never contain `Authorization` as key, isn't that the intention in any case?

ccaffy: Yes indeed, we can remove it from the map actually because if it is there, it will be automatically be appended to the opaque of the open no matter what...

ccaffy: OK done. Anyway, this should not be merged until rebased on top of the case-insensitive header parsing commit

amadio:
All good, merging! Thanks!

Issue #2283, opened 2024-06-10
[XrdHttpTPC] PULL - allow to pass header values to the file open (as CGI)


ccaffy:
This use-case comes from the CTA project where they would like tape archive-metadata to be passed to them when a file is open.

The only possible way we can pass archive metadata to CTA is to append the archivemetadata CGI to the resource URL of the file uploaded to them: `xrdcp myLocalFile root://eosctacern.cern.ch//path/to/file?archivemetadata=base64encodedarchivemetadata`.

There's a commit in EOS that allows to pass archive metadata to CTA via a normal HTTP PUT request with the header `ArchiveMetadata: base64encodedarchivemetadata`. This header is transformed into a CGI before the file is opened.

Unfortunately, transfers towards CTA almost always use HTTP/TPC PULL. EOS only uses the vanilla XRootD HTTP TPC plugin and there is currently no mechanism in XrdHTTP-TPC that allows to transform the `ArchiveMetadata` header into a cgi `archivemetadata=` in the file open(). Only `oss.asize` is passed as an opaque (after a SIZE request is performed). I tried using the `http.header2cgi` configuration, without success (the URL with opaque information is only passed if a redirect happens, but not to the `open()` call!).

We need to have a mechanism to allow such information coming from the header to be transformed into a XrootD CGI for the open call. @abh3, probably something like the `http.header2cgi` but with another name (`tpc.header2cgi` for example) ?

ccaffy:
OK thanks, will add this `tpc.header2cgi` then!

PR #2286, opened 2024-06-19
[XrdHttp/TPC/Macaroons] Improved the case-insensitive client header parsing

### LINKED ISSUE #2259: [XrdHttp] Header parsing should be made case-insensitive

ccaffy:
We do not modify the header key coming from the client anymore. The reason for that is for HTTP/TPC where we may need to pass headers to the non-active server. If they are lower-cased, then we will break the compatibility between old xrootd server and new ones as the old ones have a case-sensitive header parsing logic!

Now we use strcasecmp and some custom case-insensitive std::map find function

Fixes #2259


ccaffy:
@abh3, @amadio , Please do not merge yet, I'd like to test all of this of course and add unit tests. But this is just to agree on the principle!

ccaffy:
Alright done, feel free to review and merge to devel so I can then finish #2283  :)

amadio (approved):

Issue #2292, opened 2024-07-05
Problems with identity/certificate during FTS tape polling calls when using XRootd with "xrd.gsiusrpxy" opaque


mpatrascoiu:
Hello,

This problem started manifesting itself on the FTS3-Public the past days, as here we have different clients interacting with the same tape storage (EOSCTAPUBLIC).

We observe that during our tape polling calls (stage poll or archive poll), if the requests are within a short time window, the identity used for the polls may be mixed. FTS sets the proxy certificate to be used via the `xrd.gsiusrpxy=<cert-file-path>`. Example:
1. Perform tape poll for (batch#1, vo#1) with `xrd.gsiusrpxy=<cert#1>`
  a. EOSCTA sees the correct certificate
2. Within a short time frame (but in the span of tens of seconds), perform second tape poll for (batch#2, vo#2) with `xrd.gsiusrpxy=<cert#2>`
  b. EOSCTA reports seeing the previous `<cert#1>` certificate and maps us to the wrong identity

We've investigated this problem quite a bit on our side. One suggestion we had was to add the username when creating the new batch. Now, for the `query prepare` command, we craft the endpoint URL as follows:
`root://fts<counter-id>@eosctapublic.cern.ch:1094:/?xrd.gsiusrpxy=<cert>`

This has been suggested to us in order to force XRootd to create a new channel with the storage endpoint. This seems to have reduced the amount of errors, but we still see problems for polling requests that happen in parallel. At this point, we believe there might still be a caching mechanism in the XRootd library which doesn't work correctly.

Can you have a look at this?
If needed, I can provide XRootd debug logs for the investigation.

Cheers,
Mihai

smithdh:
I'll assign this to myself for now, as I'd already heard about this issue and had started to look into it.

amadio:
Fixed by #2293 (i.e. 921a65e2ffaf1a7b369d7c34a056555f58ae6076 and 459967d223aa3e6a79bfd42fdb8228888a03384d).

PR #2293, opened 2024-07-05
SecProtocolgsi: Local copy of url supplied proxy location

smithdh:
This is a draft patch for #2292

### DIFF
--- a/src/XrdCl/XrdClThirdPartyCopyJob.cc
+++ b/src/XrdCl/XrdClThirdPartyCopyJob.cc
@@ -570,7 +570,7 @@ namespace XrdCl
     //--------------------------------------------------------------------------
     // Verify if the destination supports TPC / TPC-lite
     //--------------------------------------------------------------------------
-    st = Utils::CheckTPCLite( realTarget.GetHostId() );
+    st = Utils::CheckTPCLite( realTarget.GetURL() );
     if( !st.IsOK() )
     {
       // we still want to send a close, but we time it out fast
@@ -607,11 +607,11 @@ namespace XrdCl
     //--------------------------------------------------------------------------
     if( !tpcLite )
     {
-      st = Utils::CheckTPC( URL( tpcSource ).GetHostId(), timeLeft );
+      st = Utils::CheckTPC( URL( tpcSource ).GetURL(), timeLeft );
       if( !st.IsOK() )
       {
         log->Error( UtilityMsg, "Source (%s) does not support TPC",
-                    tpcSource.GetHostId().c_str() );
+                    tpcSource.GetURL().c_str() );
         return XRootDStatus( stError, errNotSupported, 0, "Source does not "
                              "support third-party-copy" );
       }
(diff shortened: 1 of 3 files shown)


smithdh:
The issue is thought to be around the extraction and storage of the value of a "xrd.gsiusrpxy" url-option (or one of the other similar url parameters). The extraction is done by an XrdSecProtocolgsi object. Values are extracted from the url during the method XrdSecProtocolgsi::getCredentials() and used during XrdSecProtocolgsi::ClientDoInit(). The url value is stored in the variable UsrProxy (overwriting the existing value). UsrProxy is static (static class member), and may be initialised during XrdSecProtocolgsi::Init (a static class method): so updating it during a non-static method leads to potential concurrent update of the member (a String) by different class instances, and also means one request can affect other requests in an unexpected way.

The PR changes the approach to store the url value in a non-static member, and then evaluate the actual value to use during ClientDoInit() based on the static UsrProxy value and the actual url-option value.

smithdh:
Hi. I'll update with more information about the testing of this PR: our FTS team did try to test this patch in their service, but encountered some problems. For the time being there's been no chance to understand what exactly happened (and there won't be chance to discuss with the person who ran the test for a little while). So for now I don't think I'll move this PR out of draft.

The patch itself seems alright to me, and I've done some tests myself that show a problem and the expected fix. However as it does change some observed behaviour it's possible a client could have depended on the previous behaviour, so perhaps caution is needed.

amadio:
The problem in FTS was due to the changes in `devel` which break APIs as far as I know. They should apply this patch on top of 5.7.0 and try again.

abh3 (approved):
I am not completely familiar with the original issue here. From my petspective. It would appear that the attempt here is to use the proxy value in the url's cgi to override the one that came from elsewhere. If so, that is what has been implemented here. What I don't understand, based on various comments, why/how this problem was introduced in the first place. So, I cannot comment on whether this is a good fix, though according to @smithdh itr appears to work.

smithdh:
Hi. I'll update this ticket with the current status & summary. These were already discussed within the team but I add an update here for clarity. There were two connected aspects to the problem that this PR should fix:

(i) a problem that was causing #2292, which is an issue that may show when there are concurrent connections using gsi, and where each connection has a different proxy location specified by &xrd.gsiusrpxy.

(ii) The default location to look for gsi proxy is changed after a processing a connection which specified a particular location with a &xrd.gsiusrpxy. (Subsequent connections that don't specify &xrd.gsiusrpxy look in the last specified location). No concurrency is required to show this behaviour.

This patch fixes/changes both (i) & (ii). But the change for issue (ii) had caused a first test of this patch by FTS to fail, due to the changed in behaviour, within XrdCl/XrdClThirdPartyCopyJob.cc. A second commit was added to this PR to fix that.

I'll move this PR out of draft status now. But I believe the preference is to keep this unmerged for now.

amadio:
@smithdh Could you please rebase this locally and push, since it has conflicts previnting automatic rebase? Thank you!

PR #2294, opened 2024-07-06
[Server] IOV_MAX can be undefined

ellert:
Use the same code as in other places where IOV_MAX is needed (e.g. src/XrdXrootd/XrdXrootdXeqFAttr.cc)

### DIFF
--- a/src/XrdXrootd/XrdXrootdXeqPgrw.cc
+++ b/src/XrdXrootd/XrdXrootdXeqPgrw.cc
@@ -205,8 +205,20 @@ int XrdXrootdProtocol::do_PgRIO()
 // We restrict the maximum transfer size to generate no more than 1023 iovec
 // elements where the first is used for the header.
 //
-   static const int maxIOVZ = IOV_MAX;
-   static const int maxCSSZ = IOV_MAX/2 - 1;
+   static int iovmax = -1;
+   if (iovmax == -1) {
+#ifdef _SC_IOV_MAX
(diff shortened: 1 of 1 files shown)


abh3 (changes requested):
This is actually bery bad. Every time a read occurs we ned to do a sys_conf call. This logic should be moved to configuration time (I am not a fan of onetime as that is still an ctive test). The correct value can then be placed in a an accessible place and used therafter.

ellert:
> Every time a read occurs we ned to do a sys_conf call.

No, it is done once, After the first call the static variableisn't -1 ny more.

amadio:
How about defining `IOV_MAX` to 1024 in `XrdSysPlatform.hh` if it's not already defined? I wonder in which case it was not defined as well (and if we should include a build in CI for it).

abh3:
That's effectively what is done but it has to be an active test because IOV_MAX might not be defined ibecause you have to use SC_Conf() to get the value. If you suggest defining the static value in  the anonymous namespace, it seems like overkill since you would have to redo the calulation everywhere SysPlatform is included and we need IOV_MAX in very few places. Making it a global static exposes you to static initialization ordering (though it's unlikely here). So, what Matias had is correct but that code is duplicated in various places. It really should be considildated into a single place. Hence my proposal which would work everywhere IOV_MAX is needed and the change where it is used is pretty trivial.

ellert:
See #2298.

PR #2298, opened 2024-07-20
Remove code duplication. Determine IOV_MAX in one place.

ellert:


amadio:
@abh3 Are you ok with this implementation? Should we use the former name of the function (i.e. `getIovMax()` from `XrdLinkXeq.cc`? What do you think of the version below? In other words, do we really need the struct wrapping the initialization?

```cpp
int getIovMax()
{
#ifndef IOV_MAX
#define IOV_MAX 1024;
#endif

#ifdef _SC_IOV_MAX
    static int iov_max = sysconf(_SC_IOV_MAX) > 0 ? sysconf(_SC_IOV_MAX) : IOV_MAX;
#else
    static int iov_max = IOV_MAX;
#endif

    return iov_max;
}
```

amadio:
I created a local commit with the change suggested above (just adding myself as co-author, keeping @ellert as main author). I will likely merge that into `master` if there's no objection.

PR #2300, opened 2024-07-23
Avoid calling 'stat' on HTTP GET

bbockelm:
This is a first attempt at "fixing" #2299.  Instead of invoking a filesystem `stat`, then opening the path, it changes the GET code to _always_ try the 'open' first.

If the open fails with an `kXR_isDirectory` (i.e., `EISDIR`), then the code falls back to doing a directory listing.

At the minimum, this will halve the number of `stat` requests the underlying storage will see as `stat` is done twice per request if the requests are sent sequentially.  At best, it'll eliminate nearly all `stat`s under high concurrency given the OFS will collapse the open files into a single handle and some OSS's will cache the results of the first `fstat`.

This state machine for GET is notoriously tricky so I'd appreciate a review by @ccaffy or @smithdh.  Tests done by hand:
- Download full file.
- Download file in single `Range` request.
- Download file in multiple `Range` requests.
- `GET` on a directory.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2280,31 +2280,8 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
           // On error, we can only send out a message if trailers are enabled and the
           // status response in trailer behavior is requested.
           if (xrdresp == kXR_error) {
-            if (m_transfer_encoding_chunked && m_trailer_headers && m_status_trailer) {
-              // A trailer header is appropriate in this case; this is signified by
-              // a chunk with size zero, then the trailer, then a crlf.
-              //
-              // We only send the status trailer when explicitly requested; otherwise a
-              // "normal" HTTP client might simply see a short response and think it's a
-              // success
-              if (prot->ChunkRespHeader(0))
-                return -1;
-
-              const std::string crlf = "\r\n";
-              std::stringstream ss;
-              ss << "X-Transfer-Status: " << httpStatusCode << ": " << httpStatusText << crlf;
-
-              const auto header = ss.str();
-              if (prot->SendData(header.c_str(), header.size()))
-                return -1;
-
-              if (prot->ChunkRespFooter())
-                return -1;
-
-              return -1;
-            } else {
-              return -1;
-            }
+            sendFooterError("");
+            return -1;
           }
 
 
@@ -2736,6 +2713,13 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
 void
 XrdHttpReq::sendFooterError(const std::string &extra_text) {
   if (m_transfer_encoding_chunked && m_trailer_headers && m_status_trailer) {
+    // A trailer header is appropriate in this case; this is signified by
+    // a chunk with size zero, then the trailer, then a crlf.
+    //
+    // We only send the status trailer when explicitly requested; otherwise a
+    // "normal" HTTP client might simply see a short response and think it's a
+    // success
+
     if (prot->ChunkRespHeader(0))
       return;
 
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -526,8 +526,10 @@ bool XrdHttpReq::Error(XrdXrootd::Bridge::Context &info, //!< the result context
 
   if (PostProcessHTTPReq()) reset();
 
-  // Second part of the ugly hack on stat()
-  if ((request == rtGET) && (xrdreq.header.requestid == ntohs(kXR_stat)))
+  // If we are servicing a GET on a directory, it'll generate an error for the default
+  // OSS (we don't assume this is always true).  Catch and suppress the error so we can instead
+  // generate a directory listing (if configured).
+  if ((request == rtGET) && (xrdreq.header.requestid == ntohs(kXR_open)) && (xrderrcode == kXR_isDirectory))
     return true;
   
   return false;
@@ -1098,22 +1100,36 @@ int XrdHttpReq::ProcessHTTPReq() {
           }
       
       // The reqstate parameter basically moves us through a simple state machine.
-      // - 0: Perform a stat on the resource
+      // To optimize things, we start off by opening the file; if it turns out to be a directory, then
+      // we close the file handle and switch to doing a HTML-based rendering of the directory.  This
+      // avoids needing to always to do "stat" first to determine the next step (since the file-open also
+      // does a "stat").
+      // - 0: Perform an open on the resource
       // - 1: Perform a checksum request on the resource (only if requested in header; otherwise skipped)
-      // - 2: Perform an open request (dirlist as appropriate).
-      // - 3+: Reads from file; if at end, perform a close.
+      // - 2: Perform a close (for dirlist only)
+      // - 3: Perform a dirlist.
+      // - 4+: Reads from file; if at end, perform a close.
       switch (reqstate) {
-        case 0: // Stat()
-          
-          // Do a Stat
-          if (prot->doStat((char *) resourceplusopaque.c_str())) {
-            XrdOucString errmsg = "Error stating";
-            errmsg += resource.c_str();
-            prot->SendSimpleResp(404, NULL, NULL, (char *) errmsg.c_str(), 0, false);
+        case 0: // Open the path for reading.
+        {
+          memset(&xrdreq, 0, sizeof (ClientRequest));
+          xrdreq.open.requestid = htons(kXR_open);
+          l = resourceplusopaque.length() + 1;
+          xrdreq.open.dlen = htonl(l);
+          xrdreq.open.mode = 0;
+          xrdreq.open.options = htons(kXR_retstat | kXR_open_read);
+
+          if (!prot->Bridge->Run((char *) &xrdreq, (char *) resourceplusopaque.c_str(), l)) {
+            prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run request.", 0, false);
             return -1;
           }
 
+          // Prepare to chunk up the request
+          writtenbytes = 0;
+
+          // We want to be invoked again after this request is finished
           return 0;
+        }
         case 1: // Checksum request
           if (!(fileflags & kXR_isDir) && !m_req_digest.empty()) {
             // In this case, the Want-Digest header was set.
@@ -1143,16 +1159,23 @@ int XrdHttpReq::ProcessHTTPReq() {
             reqstate += 1;
           }
         // fallthrough
-        case 2: // Open() or dirlist
-        {
+        case 2: // Close file handle for directory
+          if ((fileflags & kXR_isDir) && fopened) {
+            memset(&xrdreq, 0, sizeof (ClientRequest));
+            xrdreq.close.requestid = htons(kXR_close);
+            memcpy(xrdreq.close.fhandle, fhandle, 4);
 
-          if (!prot->Bridge) {
-              prot->SendSimpleResp(500, NULL, NULL, (char *) "prot->Bridge is NULL.", 0, false);
+            if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
+              prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run close request.", 0, false);
               return -1;
             }
-
+            return 0;
+          } else {
+            reqstate += 1;
+          }
+        // fallthrough
+        case 3: // List directory
           if (fileflags & kXR_isDir) {
-
             if (prot->listdeny) {
               prot->SendSimpleResp(503, NULL, NULL, (char *) "Listings are disabled.", 0, false);
               return -1;
@@ -1172,10 +1195,8 @@ int XrdHttpReq::ProcessHTTPReq() {
               return -1;
             }
 
-
             std::string res;
             res = resourceplusopaque.c_str();
-            //res += "?xrd.dirstat=1";
 
             // --------- DIRLIST
             memset(&xrdreq, 0, sizeof (ClientRequest));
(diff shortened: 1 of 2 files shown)


bbockelm:
@amadio - I see the following unit test fails:
```
  79: Test command: /__w/xrootd/xrootd/build/tests/XrdEc/xrdec-unit-tests "--gtest_filter=XrdEcTests.IllegalVectorReadTest" "--gtest_also_run_disabled_tests"
  79: Working Directory: /__w/xrootd/xrootd/build/tests/XrdEc
  79: Test timeout computed to be: 600
  79: Running main() from gtest_main.cc
  79: Note: Google Test filter = XrdEcTests.IllegalVectorReadTest
  79: [==========] Running 1 test from 1 test case.
  79: [----------] Global test environment set-up.
  79: [----------] 1 test from XrdEcTests
  79: [ RUN      ] XrdEcTests.IllegalVectorReadTest
   40/108 Test  #79: XrdEc::XrdEcTests.IllegalVectorReadTest ...................................***Exception: SegFault  0.20 sec
```

Is that expected to succeed?  I'm having a hard time figuring out how changes exclusively in `XrdHttp` could cause this :/

smithdh:
Hi. I can reproduce an occasional segfault during XrdEcTests.IllegalVectorReadTest, e.g. using devel branch without this PR. So indeed it seems the crash is probably not related to this http change. (I'd have to confirm with Guilherme if he knows about any existing outstanding issue with the test).

amadio:
@bbockelm @smithdh I confirm that this test fails sporadically in the `master` branch. I've looked at it before, but didn't manage to identify the root cause yet. It's a lifetime management issue of Zip archive objects. Now I'm back from vacations and will take a look at this and other issues. Feel free to ignore for now.

ffurano:
I managed to make the directory listing hang with a sufficiently large directory
> GET / HTTP/1.1
> User-Agent: libdavix/0.8.5 neon/0.0.29
> Keep-Alive: 
> Connection: Keep-Alive
> TE: trailers
> Host: littlexrdhttp.cern.ch:1094
> 

Server just says:
240912 10:03:44 79607 XrootdBridge: unknown.149234:27@pcitsdcfab.dyndns login as nobody
240912 10:03:44 79607 ofs_open: unknown.149234:27@pcitsdcfab.dyndns Unable to open /; is a directory
240912 10:04:38 79595 XrootdXeq: unknown.149234:27@pcitsdcfab.dyndns disc 0:00:54

smithdh:
H. (Adding a comment as I'd said something in person during the workshop, of a possible reason for Fabrizio's observation of hang) - I had in mind an existing bad behaviour during header parsing (my last comment on https://github.com/xrootd/xrootd/pull/2072 ). However reading Fabrizio's description of the problem, that issue may not be relevant.

bbockelm:
@ffurano -- for recreating the failure you observe, what's the approximate size of directory for your directory listing?

ffurano:
mine is 50k entries, and interestingly enough it enters a dodgy CPU-eating mode. However I think I have found the fix, it will likely come today after I finish testing.

bbockelm:
FWIW -- I created 100k files in a directory doing the following:

```
for num in {1..100000}; do echo "hello world $num" > /tmp/xrootd/public/listdir/hello.txt.listdir.$num; done
```

and then invoked 

```
for num in {1..10}; do curl -v -H 'depth: 1' -X PROPFIND http://localhost:59906/public/listdir/ > /dev/null; done
```

I didn't see any obvious failures.  I _did_ notice that the PROPFIND response buffers the entire 50MB response string in memory until all 100k entries were processed, then sent it at once.  So, if you're doing something like a directory with 4M files in it, you need at least 20GB of memory for the response.  (That's something that seems worth fixing but outside the purview of this particular PR.)

@smithdh mentioned he has previously seen something like this so I'm summoning him on this ticket.

ffurano:
The issue appears with GET, not PROPFIND

ffurano:
The fix for this hang is in
https://github.com/bbockelm/xrootd/pull/8

In general I like the idea behind this PR very much, and I like also the refactoring bits that it comes with. I'll test a bit further

bbockelm:
The latest force-push includes @ffurano's fix for directory listing.

ffurano (approved):
To me the thing works and could provide a measurable benefit in some use cases.

Since the mod is based on a correct kXR_isDir behaviour, it may be subjected to the correct implementation of other Xrootd plugins. I would suggest someone else to test this with an xrootd in caching mode, whose behaviour and use cases I don't entirely master.

bbockelm:
@osschar - if you want an endpoint to test, this patch should be on all the OSDF cache servers (I believe Diego has CMS files present if you want CMS files to test).  Alternatively, I can give you an RPM if you'd like.

bbockelm:
FWIW -- I've been exclusively testing and running this branch on all the OSDF caches (so, around 1PB a day of data delivery).  I think it's in pretty good shape for that.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1292,22 +1296,25 @@ int XrdHttpReq::ProcessHTTPReq() {
             if ((offs >= filesize) || (offs+l > filesize)) {
               TRACE(ALL, " Requested range " << l << "@" << offs <<
               " is past the end of file (" << filesize << ")");
-              //prot->SendSimpleResp(522, NULL, NULL, (char *) "Invalid range request", 0);
-              return -1;
+              httpStatusCode = 552;
```
ccaffy, line 1299: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1292,22 +1296,25 @@ int XrdHttpReq::ProcessHTTPReq() {
...
+              httpStatusText = "Invalid range request";
+              return sendFooterError("Requested range is past end of file");
             }
             
             if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
-              prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run read request.", 0, false);
-              return -1;
+              TRACEI(REQ, " Failed to run read request on the bridge.");
+              XrdHttpReq::mapXrdErrorToHttpStatus();
```
ccaffy, line 1306: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1292,22 +1296,25 @@ int XrdHttpReq::ProcessHTTPReq() {
...
             }
           } else {
             // --------- READV
 
             length = ReqReadV(readChunkList);
 
             if (!prot->Bridge->Run((char *) &xrdreq, (char *) &ralist[0], length)) {
-              prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run read request.", 0, false);
-              return -1;
+              TRACEI(REQ, " Failed to run ReadV request on the bridge.");
```
ccaffy, line 1315: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1443,8 +1450,9 @@ int XrdHttpReq::ProcessHTTPReq() {
 
             TRACEI(REQ, "XrdHTTP PUT: Writing chunk of size " << bytes_to_write << " starting with '" << *(prot->myBuffStart) << "'" << " with " << chunk_bytes_remaining << " bytes remaining in the chunk");
             if (!prot->Bridge->Run((char *) &xrdreq, prot->myBuffStart, bytes_to_write)) {
-              prot->SendSimpleResp(500, NULL, NULL, (char *) "Could not run write request.", 0, false);
-              return -1;
+              TRACEI(REQ, " Failed to run write request on the bridge.");
```
ccaffy, line 1453: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1466,8 +1474,9 @@ int XrdHttpReq::ProcessHTTPReq() {
 
           TRACEI(REQ, "Writing " << bytes_to_read);
           if (!prot->Bridge->Run((char *) &xrdreq, prot->myBuffStart, bytes_to_read)) {
-            prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run write request.", 0, false);
-            return -1;
+            TRACEI(REQ, " Failed to run write request on the bridge.");
```
ccaffy, line 1477: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1489,8 +1498,9 @@ int XrdHttpReq::ProcessHTTPReq() {
 
 
           if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
-            prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run close request.", 0, false);
-            return -1;
+            TRACEI(REQ, " Failed to run close request on the bridge.");
```
ccaffy, line 1501: Maybe you can put the error message in a string allowing you to not repeat the same string in the two calls?

ccaffy (changes requested):

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1180,43 +1201,28 @@ int XrdHttpReq::ProcessHTTPReq() {
             xrdreq.dirlist.dlen = htonl(l);
 
             if (!prot->Bridge->Run((char *) &xrdreq, (char *) res.c_str(), l)) {
-              prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run request.", 0, false);
-              return -1;
+              XrdHttpReq::mapXrdErrorToHttpStatus();
```
amadio, line 1204: You can use just `mapXrdErrorToHttpStatus();` here, as we are inside `XrdHttpReq` anyway.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1867,488 +2163,167 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
...
+          return PostProcessListing(final_);
+        }
+        default: //read or readv, followed by a close.
+        {
+          // If we are postprocessing a close, potentially send out informational trailers
+          if ((ntohs(xrdreq.header.requestid) == kXR_close) || readClosing)
           {
-            // Ugly hack. Be careful with EOS! Test with vanilla XrdHTTP and EOS, separately
-            // A 404 on the preliminary stat() is fatal only
-            // in a manager. A non-manager will ignore the result and try anyway to open the file
```
amadio, line 2091: @ccaffy I wonder if the changes introduced here will affect EOS, do you know? If that's possible, we should test this with EOS to make sure nothing there is relying on the previous behavior of stat + (open | dirlist).

amadio (changes requested):
@bbockelm Thanks for these changes, they look good. I'd like to request a few changes, however, in how the commits are structured. I think it makes sense to put first a commit adding `sendFooterError` and refactoring current code to use it where it makes sense, then add another commit refactoring the listing into the newly added `PostProcessListing` (including @ffurano's fix in the initial version, and with a `Reviewed-by` or `Co-authored-by` tag to acknowledge it). Then the commit refactoring the handling of `GET` requests to avoid calling `stat`. This will make it much easier to understand later. Thank you!

abh3:
Can we get this pull request resolved? It is now blocking work on pull request 2348 as any changes in that one rely on the changes in this pull request (i.e. returning the correct "AGE" response). It's not clear what all the holdups are, one certainly is that @bbockelm needs to respond to some change requests the other is a concern from @amadio, that said can we get some traction on this?

src/XrdHttp/XrdHttpReq.cc
bbockelm: Addressed in the latest commit (plus other, similar locations).

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done in the latest commit.  Note the `httpStatusCode` in the commit (552) is different from the original one (522).

Both are wrong.  Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses, it should be `416 Range Not Satisfiable`.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done.  Moved the trace to the common `sendFooterError` at this and other call sites.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done in the latest commit.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done in latest commit.

src/XrdHttp/XrdHttpReq.cc
bbockelm: Done in latest commit.

src/XrdHttp/XrdHttpReq.cc
bbockelm: I would indeed appreciate a test from the EOS side as well!

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2750,6 +2725,29 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
   return 0;
 }
 
+int
+XrdHttpReq::sendFooterError(const std::string &extra_text) {
```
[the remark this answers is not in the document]
bbockelm: I consider the `return sendFooterError(...)` form as more understandable but will defer to you as the reviewer!  Changing in the latest commit.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2715,6 +2725,29 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
...
+  if (m_transfer_encoding_chunked && m_trailer_headers) {
```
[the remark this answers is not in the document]
bbockelm: Good catch!  I prefer being a bit more conservative in what I send here and added the extra `&& m_status_trailer` in the latest commit.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2750,6 +2725,29 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
...
+    const auto header = ss.str();
+    if (prot->SendData(header.c_str(), header.size()))
+      return -1;
+
+    if (prot->ChunkRespFooter())
+      return -1;
+  }
+  return -1;
+}
+
```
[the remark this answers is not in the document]
bbockelm: I didn't use this directly but took at least the `crlf` suggestion -- from other suggestions in the PR, I added more functionality meaning I couldn't use things directly.

amadio:
I'm ok with merging the current version of the code. However, in its current state I'd squash everything into a single commit. If you prefer to keep separate commits for this, please refactor as I suggested [above](https://github.com/xrootd/xrootd/pull/2300#pullrequestreview-2404575191), or let me know and I will squash and merge into `devel`. Thanks.

bbockelm:
I'm OK with squashing!

(and it'll probably help with the timeline -- I'm hosting a workshop this week so my next technical context switch is potentially a week or two out)

PR #2301, opened 2024-07-26
Add support for "Bad Gateway" HTTP status code

bbockelm:
Map various connection-related errors in XrdCl in XCache all the way through to the HTTP layer.

This allows a failure to connect to the origin to be represented with the correct status code ("Bad Gateway"); currently, a down origin results in a 500 ("Internal Server Error") which is misleading for users.

### DIFF
--- a/src/XProtocol/XProtocol.hh
+++ b/src/XProtocol/XProtocol.hh
@@ -1368,6 +1368,8 @@ static int mapError(int rc)
            case ENOSPC:        return kXR_NoSpace;
            case ENAMETOOLONG:  return kXR_ArgTooLong;
            case ENETUNREACH:   return kXR_noserver;
+           case EHOSTUNREACH:  return kXR_noserver;
+           case ECONNREFUSED:  return kXR_noserver;
            case ENOTBLK:       return kXR_NotFile;
            case ENOTSUP:       return kXR_Unsupported;
            case EISDIR:        return kXR_isDirectory;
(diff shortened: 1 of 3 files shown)


src/XProtocol/XProtocol.hh
```diff
@@ -1368,6 +1368,8 @@ static int mapError(int rc)
            case ENOSPC:        return kXR_NoSpace;
            case ENAMETOOLONG:  return kXR_ArgTooLong;
            case ENETUNREACH:   return kXR_noserver;
+           case EHOSTUNREACH:  return kXR_noserver;
+           case ECONNREFUSED:  return kXR_noserver;
```
amadio, line 1372: I think connection refused might be better mapped to something else, no? Maybe `kXR_NotAuthorized` or `kXR_ServerError`?

bbockelm: I picked these after looking at https://xrootd.slac.stanford.edu/doc/dev56/XRdv520.htm; there's no "perfect match", unfortunately.  Here's my thoughts:

- Definitely not `kXR_NotAuthorized`.  "Connection refused" is what you get when the TCP port is not open.
- `kXR_ServerError` is described as "An internal server error has occurred".  That's quite generic; if I look at other uses in the code base, it tends to really be problems with the server itself ("Logic error computing checksum").

`kXR_noserver` is described as "There are no servers available to process the request"; seems that a closed TCP port falls under "no server available".

bbockelm: I should add that the other piece of this is in `XrdPosixMap.cc` which maps the XrdCl errors to POSIX errno's:
- `XrdCl::errInvalidAddr` -> `EHOSTUNREACH`
- `XrdCl::errConnectionError` -> `ECONNREFUSED`

It's a fairly rickety setup:
- libcurl errors in the plugin are translated to XrdCl errors
- XrdCl errors are translated to POSIX errno's
- POSIX errno's are translated to `xrootd` protocol error codes
- `xrootd` protocol error codes are translated to HTTP status codes.

amadio: Yes, what is there is not ideal, this is definitely an area that should be looked at for XRootD 6.0. From the comments above, it seems that indeed `kXR_noserver` is a good mapping. There should be a single mapping between XRootD and POSIX error codes and vice-versa.

PR #2311, opened 2024-08-15
[XrdPfc] fix ReadV max_ior excess on pfc request forwarding, fixes #2308

### LINKED ISSUE #2308: Read requests fail for proxy+origin setup under heavy load
Dear xrootd developers,

At RAL we noticed that sometimes read requests fail with error `[ERROR] Server responded with an error: [3008] cannot allocate memory (source)`. After an investigation, it turned out that the error is not a simple lack of memory.

The error can be reproduced in the following way:
1. Start xrootd server and xrootd proxy on the same host, with the following configs:

proxy:
```
all.export *?
all.export /

# The adminpath and pidpath variables indicate where the pid and various
# IPC files should be placed
#

xrootd.async off
xrd.buffers maxbsz 8m
# Configure proxy disk caching
#
ofs.osslib   libXrdPss.so
pss.cachelib libXrdPfc.so
pss.origin 127.0.0.1:1094

oss.localroot /xcache

xrd.port 1095

# Proxy settings
#
pfc.ram 1g
pfc.diskusage 0.89 0.90 purgecoldfiles 2d 5
pfc.blocksize 8m
pfc.prefetch 0
pfc.cschk off
```
server:
```
all.pidpath /run/xrootd
```

2. Copy enough files of reasonable size to the server. For me 128 files of ~128MB in size work:
```
[root@5591c95906fc lhcb_job]# for i in {1..128}; do xrdcp testfile root://127.0.0.1:1094//tmp/file${i} ; done
[129.1MB/129.1MB][100%][==================================================][129.1MB/s]
[129.1MB/129.1MB][100%][==================================================][129.1MB/s]
[..]
```

3. Make sure that proxy's cache dir (/xcache in the above config) is empty and  for every copied file request random chunks of at least 8MiB in size via proxy. That should reproduce the error.
E.g. something like this:
```
#!/usr/bin/env python3
import sys
import random
from pyxrootd import client

if __name__ == '__main__':
    url = sys.argv[1]
    block_size = 8*1024*1024
    with client.File() as fd:
        res, _ = fd.open(url)
        if res['ok']:
            res, stat = fd.stat()
            size = stat['size']
            while True:
                offset = random.randint(0, size - 1 )
                status, res = fd.read(offset=offset, size=block_size)
                if not status['ok']:
                    print(status)
[root@5591c95906fc lhcb_job]# for i in {1..128}; do python3 test_read.py root://127.0.0.1:1095//tmp/file${i} & done
[1] 37579
[2] 37580
....
{'status': 1, 'code': 400, 'errno': 3008, 'message': '[ERROR] Server responded with an error: [3008] Unable to read /tmp/file108; cannot allocate memory\n', 'shellcode': 54, 'error': True, 'fatal': False, 'ok': False}
{'status': 1, 'code': 400, 'errno': 3008, 'message': '[ERROR] Server responded with an error: [3008] Unable to read /tmp/file86; cannot allocate memory\n', 'shellcode': 54, 'error': True, 'fatal': False, 'ok': False}
{'status': 1, 'code': 400, 'errno': 3008, 'message': '[ERROR] Server responded with an error: [3008] Unable to read /tmp/file39; cannot allocate memory\n', 'shellcode': 54, 'error': True, 'fatal': False, 'ok': False}
{'status': 1, 'code': 400, 'errno': 3008, 'message': '[ERROR] Server responded with an error: [3008] Unable to read /tmp/file85; cannot allocate memory\n', 'shellcode': 54, 'error': True, 'fatal': False, 'ok': False}
```

Obviously, in the example above proxy runs out of allocated memory, however it should be able to forward the incoming requests to the origin server, as in understand. Furthermore, it looks like the original error message coming from the server is different from the one client receives (this complicates debugging a lot...):
```
Thread 56 "xrootd" hit Breakpoint 1, XrdPosixFileRH::HandleResponse (this=0x7f9ec8001020, status=0x7faa9c001cf0, response=0x0)
    at /usr/src/debug/xrootd-5.7.0-2.el9.x86_64/src/XrdPosix/XrdPosixFileRH.cc:115
115                result = XrdPosixMap::Result(*status, theFile->ecMsg, false);
(gdb) p *status
$2 = {<XrdCl::Status> = {status = 1, code = 400, errNo = 3008}, pMessage = "Single readv transfer is too large"}
(gdb)
```

I believe, the following happens here:
1.Proxy eventually processes  Read and ReadV requests in [ReadOpusCoalescere](https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfcFile.cc#L701) function.
2. When there is lack of memory, requests are forwarded to the origin server as ReadV, even if the original request was Read -- see [here](https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfcFile.cc#L820), [here ](https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfcFile.cc#L851) and [here](https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfcFile.cc#L613).
3. ReadV requests have limit on maximum chunk size, which is defined [here ](https://github.com/xrootd/xrootd/blob/master/src/XrdXrootd/XrdXrootdProtocol.cc#L116), as far as i can see. The limit is lower than 8MiB. Read requests do not have this limit. Therefore, for example, client can submit a Read request for 8MiB block to the proxy, and if there is lack of memory, it will be forwarded to the origin server as ReadV request with one 8MiB chunk.
4. If ReadV chunk size is bigger than the limit, server will return `Single readv transfer is too large` error to the proxy, which becomes (annoyingly enough) `Cannot allocate memory` when forwarded to the client.

The above example was produced with xrootd 5.7.0, but i believe all versions from at least 5.5.4 (the one we are using in prod) are affected.

Would it be possible to amend proxy code, so that it takes into account ReadV limits of the origin server?

Best Regards,
Alex

alex-rg:
This is an attempt to fix #2308. If the direct read vector has only one element, let's forward it as `Read` request, not `ReadV`. The idea being that the original request may be`Read`, and therefore can have "length" bigger then `ReadV's` `max_ior`.

If max buffer size is ~2 times bigger than the proxy's block size, it will still be possible to have a `ReadV` request which exceeds `max_ior` limit. For example, it can happen in the following way:
 
1. Client requests bytes from `a` to `b`, where bytes `a`and  `b` correspond exactly to the beginning and end of a single block; this block is fecthed and then stored to disk.

2. Proxy runs out of memory.

3. Client requests bytes from `c to d`, where `c < a`, `d >= b + Proxy_block_size` and `Buffer_size >= d - c + 1`.
 

In the above scenario it looks like the read vector (that is going to be submitted to the origin) will have the following coordinates (offsets first, then lengths): `[(c, a-c); (b+1, d-b)]`. The second block will have length bigger than `max_ior`.
 

That said, there are a lot of conditions to be met for the error to be reproduced (particular buffer size, particular alignment of requested bytes). Without the fix it is much simpler -- just a lack of memory and a big enough read of not fetched block(s) is enough.

### DIFF
--- a/src/XrdPfc/XrdPfcFile.cc
+++ b/src/XrdPfc/XrdPfcFile.cc
@@ -608,9 +608,20 @@ void File::ProcessBlockRequests(BlockList_t& blks)
 
 void File::RequestBlocksDirect(IO *io, DirectResponseHandler *handler, std::vector<XrdOucIOVec>& ioVec, int expected_size)
 {
-   TRACEF(DumpXL, "RequestBlocksDirect() issuing ReadV for n_chunks = " << (int) ioVec.size() << ", total_size = " << expected_size);
+   int iovec_sz = (int) ioVec.size();
+   TRACEF(DumpXL, "RequestBlocksDirect() issuing ReadV for n_chunks = " << iovec_sz << ", total_size = " << expected_size);
 
-   io->GetInput()->ReadV( *handler, ioVec.data(), (int) ioVec.size());
+   /* If there is only one element in the read vector, it may come from a Read request.
+    * Read requests do not respect ReadV max_ior limit, so to avoid erros it should not
+    * be forwarded as ReadV. */
+   if (iovec_sz == 1)
+   {
+      io->GetInput()->Read(*handler, ioVec[0].data, ioVec[0].offset, ioVec[0].size);
+   }
+   else
+   {
+      io->GetInput()->ReadV(*handler, ioVec.data(), iovec_sz);
+   }
 }
 
 //------------------------------------------------------------------------------


osschar:
Replaced by #2315 that provides full ReadV conformance to the protocol ... closing.

PR #2314, opened 2024-08-22
[XrdOucUtils] Speed-up authz obfuscation

ccaffy:
@amadio discovered that the XrdEc tests were 10 times slower than usual. This slowness came from the fact that for each call to XrdOucUtils::obfuscateAuth(...) 2 std::regex were instanciated. The instanciation of those two std::regex objects is expensive as there is a compilation logic to ensure the regex is correct. Those 2 regexes are now constructed only once in the lifetime of the XRootD process by creating a static std::vector<std::regex>.

### DIFF
--- a/src/XrdOuc/XrdOucUtils.cc
+++ b/src/XrdOuc/XrdOucUtils.cc
@@ -114,7 +114,15 @@ int LookUp(idMap_t &idMap, unsigned int id, char *buff, int blen)
    return luRet;
 }
 }
-  
+
+std::vector<std::regex> XrdOucUtils::authObfuscationRegexes {
+  //authz=xxx&... We deal with cases like "(message: kXR_stat (path: /tmp/xrootd/public/foo?pelican.timeout=3s&authz=foo1234, flags: none)" where we do not want to obfuscate
+  // ", flags: none)" + we deal with cases where the 'authz=Bearer token' when an admin could set 'http.header2cgi Authorization authz' in the server config
+  std::regex(R"((authz=)(Bearer\s)?([^ &\",<>#%{}|\^~\[\]`]*))",std::regex_constants::optimize),
+  // HTTP Authorization, TransferHeaderAuthorization headers that with the key that can be prefixed with spaces and value prefixed by spaces
+  std::regex(R"((\s*\w*Authorization\s*:\s*)[^$]*)", std::regex_constants::icase | std::regex_constants::optimize)
+};
+
 /******************************************************************************/
 /*                               a r g L i s t                                */
 /******************************************************************************/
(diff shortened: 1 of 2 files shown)


amadio (approved):
Looks good, performance is back to what it was before. :clap:

abh3:
The bigger question is why is XrdEc printing out so much stuff that involves onfuscation? That just doesn't sound right or at least well thought out.

ccaffy:
Well, it's not the fault of XrdEc' logging logic ;) It's because the obfuscation logic happens in the construction of the XrdClUrl object (avoiding to call the obfuscation each time we log the URL in a transfer log line). As XrdEc uses zip archives with plenty of Url construction (one URL per stripe), that's why it was slower...

abh3:
Oh, well why isn't being done in the logging object but only when a URL is being printed? OK, that is complicated but there has to be a way of avoiding obfusatiion logic unless a url is being logged; which should happen infrequently. I guess I am confused where the obfuscation actually occurs.

ccaffy:
For a simple xrdcp transfer with debug logging enabled:
```
$ xrdcp 'root://localhost:1094//tmp/passwd?authz=testabcd&hello=world' - 2>&1 | grep -c REDACTED
30
$ 
```

There are 30 places for one transfer where the URL should be obfuscated so that it does not leak to the logs. I believe that it is better to obfuscate the URL during construction (here only once) and assign the obfuscated URL to a string attribute in the URL class rather than obfuscating each time one will print the log :)

abh3:
Could we create one upon fetching the value if ity has not been created already? I know that is a tricky proposition. I suppose the additonal memory isn't a big deal but it seems always creating a obfuscated version when 99% of the time it won't be used seems just wrong. Anyway, something to think about.

amadio:
Hi Andy, indeed, it would be better to follow your suggestion. However, after the performance fix, I couldn't anymore notice any performance impact on the tests. So, I think that even in the current state, things are ok.

amadio:
Actually, reviewing the ABI compatibility report, adding the `pObfuscatedURL` new member to `XrdCl::URL` changes its size (and therefore libraries linked against XrdCl allocate the wrong size when calling new, or have too little space if using it on the stack). We will need to remove the new member and apply the obfuscation when needed if we want to have this in the next patch release (i.e. we need to do it as suggested by Andy above). In any case it will not be possible to be 100% backward compatible because the signature of `XrdOucUtils::obfuscate()` changed from what it was in 5.7.0 as well. I think we should simply not expose this in our public interface.

bbockelm:
I think @abh3 was heading in the right direction [in his comment](https://github.com/xrootd/xrootd/pull/2314#issuecomment-2306513807) (it's also what we ended up doing in Pelican):

Do the obfuscation in the logging object, prior to emitting the log line.

That way, the patch goes in once per logging subsystem (XrdCl, core xrootd) and regressions are far less likely.

Logically, if the administrator has already cranked up the logging enough that the cost of the regex is relevant, timing-wise, they've basically already accepted that they aren't performance-sensitive.  One can't argue that they care about microseconds of extra latency _and_ ask for gigabytes of logs at the same time.

In Pelican, the setup is perhaps slightly better because the logging goes into a kernel buffer (a pipe) and asynchronously redacted by a separate thread.  However, the logging object could do the same thing here as well.

abh3:
Well, yes and no. What I don't like about implementing this at the logging level is that 90% of the effort will be devoted to snanning lines that simply don't apply. That's a waste of effort (well Google would disagree but that's another story). What I meant to say that if debugging is not enabled we simply don't generrate the extra member. But as Cedric points out it's a miniscule hit that you would b hard pressed to measure. To summarize, I don't think doing all the hard work in the logging plugin is a good idea unless you have a magic formula that can determine whether obfuscation is required in a simple test on the log line (I can't think of one). So, in the end all this means is the patch is postponed to 6.0, which is not ideal but tolerable.

PR #2315, opened 2024-08-22
[XrdPfc] Make sure direct vread requests conform to protocol limits.

### LINKED ISSUE #2308: Read requests fail for proxy+origin setup under heavy load

osschar:
Fixes #2308.

### DIFF
--- a/src/XrdPfc/XrdPfcFile.cc
+++ b/src/XrdPfc/XrdPfcFile.cc
@@ -606,11 +606,23 @@ void File::ProcessBlockRequests(BlockList_t& blks)
 
 //------------------------------------------------------------------------------
 
-void File::RequestBlocksDirect(IO *io, DirectResponseHandler *handler, std::vector<XrdOucIOVec>& ioVec, int expected_size)
+void File::RequestBlocksDirect(IO *io, ReadRequest *read_req, std::vector<XrdOucIOVec>& ioVec, int expected_size)
 {
-   TRACEF(DumpXL, "RequestBlocksDirect() issuing ReadV for n_chunks = " << (int) ioVec.size() << ", total_size = " << expected_size);
+   int n_chunks    = ioVec.size();
+   int n_vec_reads = (n_chunks - 1) / XrdProto::maxRvecsz + 1;
 
-   io->GetInput()->ReadV( *handler, ioVec.data(), (int) ioVec.size());
+   TRACEF(DumpXL, "RequestBlocksDirect() issuing ReadV for n_chunks = " << n_chunks <<
+          ", total_size = " << expected_size << ", n_vec_reads = " << n_vec_reads);
+
+   DirectResponseHandler *handler = new DirectResponseHandler(this, read_req, n_vec_reads);
+
+   int pos = 0;
+   while (n_chunks > XrdProto::maxRvecsz) {
+      io->GetInput()->ReadV( *handler, ioVec.data() + pos, XrdProto::maxRvecsz);
+      pos      += XrdProto::maxRvecsz;
+      n_chunks -= XrdProto::maxRvecsz;
+   }
+   io->GetInput()->ReadV( *handler, ioVec.data() + pos, n_chunks);
 }
 
 //------------------------------------------------------------------------------
@@ -820,12 +832,25 @@ int File::ReadOpusCoalescere(IO *io, const XrdOucIOVec *readV, int readVnum,
             {
                TRACEF(DumpXL, tpfx << "direct block " << block_idx << ", blk_off " << blk_off << ", size " << size);
 
-               if (lbe == LB_direct)
-                  iovec_direct.back().size += size;
-               else
-                  iovec_direct.push_back( { block_idx * m_block_size + blk_off, size, 0, iUserBuff + off } );
                iovec_direct_total += size;
                read_req->m_direct_done = false;
+
+               // Make sure we do not issue a ReadV with chunk size above XrdProto::maxRVdsz.
+               // Number of actual ReadVs issued so as to not exceed the XrdProto::maxRvecsz limit
+               // is determined in the RequestBlocksDirect().
+               if (lbe == LB_direct && iovec_direct.back().size + size <= XrdProto::maxRVdsz) {
+                  iovec_direct.back().size += size;
+               } else {
+                  long long  in_offset = block_idx * m_block_size + blk_off;
+                  char      *out_pos   = iUserBuff + off;
+                  while (size > XrdProto::maxRVdsz) {
+                     iovec_direct.push_back( { in_offset, XrdProto::maxRVdsz, 0, out_pos } );
+                     in_offset += XrdProto::maxRVdsz;
+                     out_pos   += XrdProto::maxRVdsz;
+                     size      -= XrdProto::maxRVdsz;
+                  }
+                  iovec_direct.push_back( { in_offset, size, 0, out_pos } );
+               }
 
                lbe = LB_direct;
             }
@@ -847,8 +872,7 @@ int File::ReadOpusCoalescere(IO *io, const XrdOucIOVec *readV, int readVnum,
    // Second, send out remote direct read requests.
    if ( ! iovec_direct.empty())
    {
-      DirectResponseHandler *direct_handler = new DirectResponseHandler(this, read_req, 1);
-      RequestBlocksDirect(io, direct_handler, iovec_direct, iovec_direct_total);
+      RequestBlocksDirect(io, read_req, iovec_direct, iovec_direct_total);
 
       TRACEF(Dump, tpfx << "direct read requests sent out, n_chunks = " << (int) iovec_direct.size() << ", total_size = " << iovec_direct_total);
    }
--- a/src/XrdPfc/XrdPfcFile.hh
+++ b/src/XrdPfc/XrdPfcFile.hh
@@ -390,7 +390,7 @@ private:
    void   ProcessBlockRequest (Block       *b);
    void   ProcessBlockRequests(BlockList_t& blks);
 
-   void   RequestBlocksDirect(IO *io, DirectResponseHandler *handler, std::vector<XrdOucIOVec>& ioVec, int expected_size);
+   void   RequestBlocksDirect(IO *io, ReadRequest *read_req, std::vector<XrdOucIOVec>& ioVec, int expected_size);
 
    int    ReadBlocksFromDisk(std::vector<XrdOucIOVec>& ioVec, int expected_size);
 


abh3:
Hmmm, where did codecov post come from? Sort of wierd I'd say as I never asked for it.

amadio:
Codecov is the coverage tool I integrated recently. The comment is weird, I will see if I can turn it off. Just ignore it for now.

src/XrdPfc/XrdPfcFile.cc
```diff
@@ -820,13 +832,26 @@ int File::ReadOpusCoalescere(IO *io, const XrdOucIOVec *readV, int readVnum,
...
-                  iovec_direct.back().size += size;
-               else
-                  iovec_direct.push_back( { block_idx * m_block_size + blk_off, size, 0, iUserBuff + off } );
                iovec_direct_total += size;
                read_req->m_direct_done = false;
 
+               // Make sure we do not issue a ReadV with chunk size above XrdProto::maxRVdsz.
+               // Number of actual ReadVs issued so as to not exceed the XrdProto::maxRvecsz limit
+               // is determined in the RequestBlocksDirect().
+               if (lbe == LB_direct && iovec_direct.back().size + size <= XrdProto::maxRVdsz) {
```
alex-rg, line 841: Looks like the variable `XrdProto::maxRVdsz` is undefined? At least i was not able to find its definition.

Probably something like `XrdXrootdProtocol::maxReadv_ior` should be here (though it is protected...).

amadio: This was defined in 5dff60cdd4a8b5ec62c52a9dc0de1590417ed6c6.

src/XrdPfc/XrdPfcFile.cc
```diff
@@ -820,13 +832,26 @@ int File::ReadOpusCoalescere(IO *io, const XrdOucIOVec *readV, int readVnum,
...
+                  iovec_direct.back().size += size;
+               } else {
+                  long long  in_offset = block_idx * m_block_size + blk_off;
+                  char      *out_pos   = iUserBuff + off;
+                  while (size > XrdProto::maxRVdsz) {
+                     iovec_direct.push_back( { in_offset, XrdProto::maxRVdsz, 0, out_pos } );
```
amadio, line 847: This is a good opportunity to use [`emplace_back`](https://en.cppreference.com/w/cpp/container/vector/emplace_back) to avoid creating a temporary object:
```suggestion
                     iovec_direct.emplace_back(in_offset, XrdProto::maxRVdsz, 0, out_pos);
```

src/XrdPfc/XrdPfcFile.cc
```diff
@@ -820,13 +832,26 @@ int File::ReadOpusCoalescere(IO *io, const XrdOucIOVec *readV, int readVnum,
...
+                     in_offset += XrdProto::maxRVdsz;
+                     out_pos   += XrdProto::maxRVdsz;
+                     size      -= XrdProto::maxRVdsz;
+                  }
+                  iovec_direct.push_back( { in_offset, size, 0, out_pos } );
```
amadio, line 852: Same here
```suggestion
                  iovec_direct.emplace_back(in_offset, size, 0, out_pos);
```

amadio (commented):
Minor comments from me, if you could fix it would be nice. I disabled codecov comments now, it should not come back. The complaint from it is due to the fact that the lines being touched here are not covered by any tests, something we could look into in the near future (adding XCache tests).

src/XrdPfc/XrdPfcFile.cc
osschar: For this to work I have to add ctors to XrdOucIOVec:
```
       XrdOucIOVec() = default;
       XrdOucIOVec(long long offs, int sz, int inf, char *buff) :
         offset(offs), size(sz), info(inf), data(buff) {}
```
here: https://github.com/xrootd/xrootd/blob/master/src/XrdOuc/XrdOucIOVec.hh#L46

Do you want me to go forward with that?

abh3:
Presumably those ctors will do nothing, right?

src/XrdPfc/XrdPfcFile.cc
abh3: Presumably the ctors will do nothing, right?

osschar: Yes, they do nothing ... and they keep the automatic move ctor and operator=, the copy ctor and operator=.

amadio: If the replacement I suggested doesn't work out of the box, then consider my suggestions optional. I wonder why the initializer list works then, need to take a look at this later.

amadio:
I will merge this as is when I'm back at work. I can test if the addition of the constructors to use emplace_back are ABI compatible and do that change later if the answer is yes. Now we have tools in place to check it: https://xrootd.web.cern.ch/abi-tracker/timeline/xrootd. Have a nice weekend!

osschar:
Thank you x 2, likewise! :)

src/XrdPfc/XrdPfcFile.cc
osschar: I've been wondering about this myself ... and concluded that it needs the argument-ed constructor, as done above. Initializer list was automatically promoted to move-constructor. Initializer lists are very eager to become types, too eager, I hear people say ;)
```
void push_back( T&& value );
// vs.
template< class... Args >
void /* reference from c++17 on */ emplace_back( Args&&... args );
```

PR #2319, opened 2024-08-26
Defer TLS client auth (backport)

bbockelm:
This is the V5 backport branch for deferring TLS client authentication until needed.


bbockelm:
Shoot, sent the PR to the wrong repo -- meant to keep this one internal (I don't think we want to upstream this for V5).

PR #2325, opened 2024-08-29
2296 tokens leaking into logfiles

ccaffy:


src/XrdOuc/XrdUtils.hh
```diff
@@ -0,0 +1,59 @@
...
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+
+
+/**
+ * PRIVATE HEADER for utility functions, implementation in XrdOucUtils.cc
+ */
+#ifndef XROOTD_XRDUTILS_HH
+#define XROOTD_XRDUTILS_HH
```
amadio, line 24: @abh3 Please let us know your opinion on the naming here. I'd suggest `XrdOucPrivUtils.hh`, and we only install into `include/xrootd/private/XrdOuc` if asked by someone to keep it truly private for as long as possible.

ccaffy: I've renamed it to `XrdOucPrivateUtils.hh`, then we know what is this header about

src/XrdOuc/XrdOucUtils.hh
```diff
@@ -37,6 +37,8 @@
 #include <vector>
 #include <regex>
 
+#include "XrdOuc/XrdOucPrivateUtils.hh"
```
amadio, line 40: This is the problem. A public header cannot include a private header.

src/XrdOuc/XrdOucPrivateUtils.hh
```diff
@@ -0,0 +1,58 @@
...
+
+#include <string>
+#include <vector>
+#include <regex>
+
+static const std::string OBFUSCATION_STR = "REDACTED";
+
+// As the compilation of the regexes when the std::regex object is constructed is expensive,
+// we initialize the auth obfuscation regexes only once in the XRootD process lifetime
+static const std::vector<std::regex> authObfuscationRegexes = {
```
amadio, line 34: Shouldn't this go in the .cc file instead of in the header? Same for `"REDACTED"` above.

ccaffy: That's absolutely right. Thought that it would be enough to declare and initialize them on the same file, but I'll modify. Thanks @amadio !

ccaffy: Done

amadio:
Alpine fails due to some git problem. Merging anyway, as the next run of the CI should have the base commit and work fine.

PR #2335, opened 2024-09-13
time_t can be 64 bit

ellert:
In a previous commit the timeout variables were changed from uint16_t to time_t. Since time_t can be 64 bits the Python bindings need to use unsigned long long.

Related: #2236, #2237


amadio:
Indeed, nice catch. I ran the tests and everything looks alright after I fixed a couple of problems in the tests themselves (overflow test with value that doesn't overflow in 64 bits). Thanks for this!

PR #2337, opened 2024-09-16
Reproducible doxygen output

ellert:
The files generated by doxygen whose file names contain the string unnamedN where N is a number are not generated using predictable names.

This happens when running doxygen using multiple threads.

Set NUM_PROC_THREADS=1 to work around this issue.

See: https://github.com/doxygen/doxygen/issues/11138

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -492,7 +492,7 @@ LOOKUP_CACHE_SIZE      = 2
 # DOT_NUM_THREADS setting.
 # Minimum value: 0, maximum value: 32, default value: 1.
 
-NUM_PROC_THREADS       = 0
+NUM_PROC_THREADS       = 1
 
 # If the TIMESTAMP tag is set different from NO then each generated page will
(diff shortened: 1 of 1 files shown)


amadio:
How important is this for you? If consistency is important in the distro DEB/RPM packages, I'm ok to take this, although then users building the docs with `doxygen Doxydoc` will have slower behavior by default, which is not ideal.

ellert:
The RPM package build fails on Fedora/EPEL if the documentation package is declared to be architecture independent (noarch) in the spec file without this change. The build server (koji) checks that the list of files in an architecture independent package is the same for the builds on all architectures.

Debian checks package build reproducibility by building a package twice with slightly different conditions (e.g. different date and time, different path to the source directory) and checks that produced packages are the same. If these tests fail the package is flagged as having non-reproducible builds. This is not a critical error that gets the package kicked out od Debian, but its nice to avoid.

I can of course keep this as a patch in the package builds if the change is not suitable for upstream.

PR #2340, opened 2024-09-17
Increase client timeouts for XRootD server tests

ellert:
The XRootD server tests currently uses very short settings for the client timeouts:

https://github.com/xrootd/xrootd/blob/ae3391f39bf3993dc9d767c00bf55611bd20f6a5/tests/XRootD/test.sh#L50-L52

This is a major reduction from the default values:

https://github.com/xrootd/xrootd/blob/ae3391f39bf3993dc9d767c00bf55611bd20f6a5/src/XrdCl/XrdClConstants.hh#L54-L56

According to the comment in the test script, this is done in order to "catch errors quickly and prevent the test suite from getting stuck waiting for timeouts while running".

However, the very short timeouts make some tests fail with timeout errors when run on small or slow systems.

This commit increases the timeout values to avoid errors:

 - XRD_REQUESTTIMEOUT: 180
 - XRD_STREAMTIMEOUT: 30
 - XRD_TIMEOUTRESOLUTION: 10

### DIFF
--- a/tests/XRootD/test.sh
+++ b/tests/XRootD/test.sh
@@ -47,9 +47,9 @@ export XRD_LOGLEVEL XRD_LOGFILE
 # Reduce default timeouts to catch errors quickly and prevent the test
 # suite from getting stuck waiting for timeouts while running.
 
-: "${XRD_REQUESTTIMEOUT:=2}"
-: "${XRD_STREAMTIMEOUT:=2}"
-: "${XRD_TIMEOUTRESOLUTION:=1}"
+: "${XRD_REQUESTTIMEOUT:=180}"
+: "${XRD_STREAMTIMEOUT:=30}"
+: "${XRD_TIMEOUTRESOLUTION:=10}"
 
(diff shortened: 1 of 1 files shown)


amadio:
@abh3 My plan is to increase the timeouts, but by somewhat smaller amounts.

@ellert, if you could try with timeout resolution at 1s (this is just the frequency at which timeouts are checked), request timeout of 15s, and stream timeout of 5s, I'd merge it (given that these numbers would work for your slow hardware too). Otherwise, let us know what are the minimum numbers that would work for your case. Thanks!

amadio:
@ellert Could you please let us know if the timeouts suggested above would work for your slow system? The other option is that you just set them in your environment for your slow system. I've done the timeout settings like that precisely to allow overrides from outside.

ellert:
The 5.7.2 version built on all architectures with the values in the source tarball, so it looks OK.

amadio:
@ellert Great, thank you for letting me know!

PR #2341, opened 2024-09-17
Merge back XrdCeph branch from RAL into main XRootD repository

Jo-stfc:
features:
vector read improvements
variable object size read
I/O buffers for avoiding small reads
logging improvements
pool space info reporting

### DIFF
--- a/tests/XrdCeph/XrdCeph.cc
+++ b/tests/XrdCeph/XrdCeph.cc
@@ -7,15 +7,6 @@
 
 #define MB 1024 * 1024
 
-struct CephFile {
-  std::string name;
-  std::string pool;
-  std::string userId;
-  unsigned int nbStripes;
-  unsigned long long stripeUnit;
-  unsigned long long objectSize;
-};
-
 void fillCephFile(const char *path, XrdOucEnv *env, CephFile &file);
 void fillCephFileParams(const std::string &params, XrdOucEnv *env,
                         CephFile &file);


amadio:
In its current form, this is quite difficult to review. Could you please maybe try to cherry-pick into a branch the commits as you had them before such that each commit is a single logical change? We'd also like to avoid merge commits, so the branch should be rebased on the current master branch if possible. I can, of course, give a hand with that. Fixups should also be combined with previous commits such that only the final work remains in the git history (to avoid breaking git bisect). Thank you!

Jo-stfc:
the original commits were done on the xrdceph submodule. should I make a PR targetting that and you can merge it back into the current repo?

Jo-stfc:
or i could generate the diffs from the submodule and apply them as separate commits

Jo-stfc:
ok. I'll create a branch in the stfc/xrootd-ceph repo with the changes

amadio:
Thank you!

Jo-stfc:
https://github.com/stfc/xrootd-ceph/tree/merge_ready

Jo-stfc:
specfile/build changes can be skipped, everything else is needed

Jo-stfc:
Hi @amadio , yes that compilation error was due to moving the struct directly into XrdCeph, and the test re-defining it. Taking the changes in https://github.com/stfc/xrootd/commit/6dd8cd1d85b13b709ae31506db3ff08e636e60d6 in **src/XrdCeph/CMakeLists.txt** (except for the version change, which was done to match versions between xrootd and xrdceph) and **‎tests/XrdCeph/XrdCeph.cc** should fix the compilation

amadio:
My rebased branch is here: https://github.com/xrootd/xrootd/compare/master...amadio:xrootd:ceph

I did take the CMake changes, but it seems that there's stuff missing from what you linked above. Do you have your latest branch with the full history, that is, without combining many commits into just a few? I will try to use that one today to merge things back. If I can't, I will probably take the full diff and apply at once, there are some big changes which make it quite hard to resolve. The other option is to rework the changes into smaller chunks and merge them little by little into the main repository.

Jo-stfc:
The changes I flagged above were things that I had to do to make compilation work on merging xrootd (they are in the RALxrdceph branch, but in the 'big commit'). I'll make a fresh branch that gets from merge_ready to full readiness

amadio:
@Jo-stfc Looks like removing `CephFile` struct from tests did it indeed on my branch. I pushed again after rebasing. I will see how far I get with the fine grained branch later and let you know if I manage. Cheers,

amadio:
For the record, we have worked together (Jyothish and myself) to rebase the branch from RAL on top of the current master branch, and the result has just been pushed above. We plan to release this with XRootD 5.8.0.

Jo-stfc:
compiled and deployed on a test server. it's passing integration tests. I think it's ok to merge now.

PR #2344, opened 2024-09-20
Avoid test failures when $RANDOM returns a multiple of 1024

ellert:
Calling `openssl rand` with 0 as the argument is an error:
```
$ openssl rand -base64 0
rand: Use -help for summary.
$ echo $?
1
```


amadio:
Oops, forgot about that. Thanks for the fix. Merging.

PR #2348, opened 2024-09-29
Change cache age logic to only indicate fully cached files

bbockelm:
The cache mtime logic (whose sole use is to have the `Age` header set) was fairly misleading: a `HEAD` (or `stat`) request would cause the cinfo file to be written, and the cache to indicate it had the data.

This was not useful: the object was indicated as cached even when not a single byte of data was present.

This takes the opposite approach: mtime and ctime are only set (and hence the `Age` header is returned) only if the file is complete.  This means a `GET` only returns the `Age` header if there is data available!

@amadio - I think this would be appropriate for backporting.  I think the current behavior is a bug.

### DIFF
--- a/bindings/python/src/PyXRootDCopyProcess.cc
+++ b/bindings/python/src/PyXRootDCopyProcess.cc
@@ -101,15 +101,15 @@ namespace PyXRootD
 
     val = XrdCl::DefaultCPInitTimeout;
     env->GetInt( "CPInitTimeout", val );
-    uint16_t initTimeout = val;
+    time_t initTimeout = val;
 
     val = XrdCl::DefaultCPTPCTimeout;
     env->GetInt( "CPTPCTimeout", val );
-    uint16_t tpcTimeout = val;
+    time_t tpcTimeout = val;
 
     val = XrdCl::DefaultCPTimeout;
     env->GetInt( "CPTimeout", val );
-    uint16_t cpTimeout = val;
+    time_t cpTimeout = val;
 
     if ( !PyArg_ParseTupleAndKeywords( args, kwds, "ss|HbbbbssssbIHHHbHLLLbs:add_job",
          (char**) kwlist,
--- a/bindings/python/src/PyXRootDCopyProgressHandler.cc
+++ b/bindings/python/src/PyXRootDCopyProgressHandler.cc
@@ -32,8 +32,8 @@ namespace PyXRootD
   //----------------------------------------------------------------------------
   // Notify when a new job is about to start
   //----------------------------------------------------------------------------
-  void CopyProgressHandler::BeginJob( uint16_t          jobNum,
-                                      uint16_t          jobTotal,
+  void CopyProgressHandler::BeginJob( uint32_t          jobNum,
+                                      uint32_t          jobTotal,
                                       const XrdCl::URL *source,
                                       const XrdCl::URL *target )
   {
@@ -54,7 +54,7 @@ namespace PyXRootD
   //----------------------------------------------------------------------------
   // Notify when the previous job has finished
   //----------------------------------------------------------------------------
-  void CopyProgressHandler::EndJob( uint16_t                   jobNum,
+  void CopyProgressHandler::EndJob( uint32_t                   jobNum,
                                     const XrdCl::PropertyList *result )
   {
     PyGILState_STATE  state    = PyGILState_Ensure();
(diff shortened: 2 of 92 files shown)


amadio:
Hi @bbockelm, we don't do backports for minor releases. This will be in next patch release and everyone should just move to that. When we have a major release out (i.e. XRootD 6.0), then we will support 5.x for some time still and will have backports, but for compatible versions that's not necessary.

bbockelm:
> This will be in next patch release and everyone should just move to that.

Yup, that's what I meant by backport -- it's useful in whatever-the-next-5.x-release-is.

osschar:
OK, I could do the following:
- set mtime to creation time
- set ctime to the last time data file was written to (mtime of data file)
- set atime to what cache thinks was the last access time of the file (mtime of cinfo)
The last thing is consistently done for what is to go into 5.8.

Then, about Age etc, there is also this function to consider:
https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfc.cc#L1074
Note though, that pss (or posix?) will forward the stat to the origin if the file does not exist in cache. This will for sure need to be modified along the same lines as File::Open().

I'm not quite sure what to think about the requirement that the file needs to be complete. This is clearly applicable to Pelican / OSDF where you guys work with full files ... but it does not apply to the CMS use case where files are often only partially read (and re-read).

stat fields also have file size and n_fs_blocks ... this can be used to deduce if the file is complete. I don't know how this translates to http response ... but something you could use in the xroot-to-http response conversion.

Remember, there are also these two functions in XrdPfc::Cache, the first to decide if direct-local-fs (e.g., lustre) is possible, and the second that is used for the only-if-cached directive.
https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfc.cc#L786
https://github.com/xrootd/xrootd/blob/master/src/XrdPfc/XrdPfc.cc#L914

We have two config vars (m_onlyIfCachedMinSize and m_onlyIfCachedMinFrac) introduced for the second function -- to decide whether to consider the file cached or not.

Then, since you've opened this can of worms (local vs. remote stat) ... I think it's also a good time to start thinking about writable caches, i.e., writing files to local space instead of to the remote, which is what Pss does now. It might require ABI changes ... at least for synchronization, even if we decide to completely separate writable namespaces / osses / or even services.

abh3:
Yes, this really needs to be thought through a bit more. While I agree there is some bugginess in the way this is handled the side-effects of the fix may be worse than the bugginess for certain applications and Pelican is not the only application here. So, in that respoect I do agree with Matevz. As for writable caches, I hae no idea why we neeed to go there at this point. We already have FRM which can be more easily made a writable cache than Pfc. So, I would go down that route before contorting Pfc into som,ething it was not desiged to do. So, let's give all of this more thought.

bbockelm:
@osschar -- I'm a bit ambivalent about what to do when there's partial data and would be fine with reusing the "if cached" variables.  I hope we can agree that setting the `Age` header when there is zero data present is the wrong thing to do!

How about this:
- Set `mtime` to the time when the first data was written (not creation time of the cinfo but really when the first byte arrived).  0 if no data is present (or 0 if below the threshold set by the "if cached" logic).
- Set `ctime` to the last time data file was written to (mtime of data file), 0 if no data is present.
- Set atime to what cache thinks was the last access time of the file (`mtime` of cinfo), 0 if no data is present.
- On cache bypass, since there's no data in the cache, `mtime = ctime = atime = 0`.
- On cache redirect-to-local, since that's the _real_ data we should use the real mtime / ctime / atime.

Thoughts?  Combined with https://github.com/xrootd/xrootd/issues/2349, this would take care of my main annoyance: you right now get an unavoidable "cache hit" for every query of the cache.

abh3:
Based on what we said in response to issue 2349 we need only to discuss the implementation of how age is determined. Since this pull request attempts to solve GET related issues and we submit that your requirement can be met by returning AGE in response to HEAD this pull request no longer applies.

Of all the options given the only one that makes sense in the response to a HEAD request which translates to a stat() call is to set atime to 0 to indicate the file is not in the cache. Hence, in the presence of a file cache, the http plugin should set AGE to 0. Note the AGE header will not be returned if there is no file cache.

If you agree with this approach I will close the pull request.

bbockelm:
@abh3 - I know some discussion is still happening in #2349.  Since we'll overhaul the guts of the patch, we can indeed close this PR for now.

(Let's just not lose the thread on this... it's an important fix to get done)

abh3:
@bbockelm I'd prefer to leave this open until we get the final patch in. That way we are sure not to loose track of this. The alternative was to start a new ticket but those may get lost in he pile we have and we don't have the history there. Anyway, the patch should be here within a few weeks.

bbockelm:
This was partly replaced by #2372.  Closing this in favor of finishing up that work.

Issue #2354, opened 2024-10-10
Race condition during HTTP-TPC requests leading to unintended file deletion


gmgigi96:
During a series of HTTP-TPC requests, we saw a race condition that led to the deletion of a file in EOS.

I will report here all the logs, in both EOS and FTS to show what happened:

- At 00:37:19, FTS initiates an HTTP-TPC request (in PULL mode):
```
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > COPY /eos/cms/xxx.root HTTP/1.1
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Host: eoscms-ns-ip563.cern.ch
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Accept: */*
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Source: https://xxx:1095/xxx.root
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > X-Number-Of-Streams: 1
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Secure-Redirection: 1
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > ClientInfo: file-id=xxx;retry=0
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > TransferHeaderAuthorization: xxx
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Credential: none
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > Authorization: xxx
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > RequireChecksumVerification: false
INFO    Sat, 05 Oct 2024 00:37:19 +0200; Davix: > User-Agent: libdavix/0.8.7 libcurl/7.76.1
```

On EOS we got the request:
```
241005 00:37:19 time=1728081439.869261 func=MatchesPath              level=INFO  logid=static.............................. unit=mgm@eoscms-ns-ip563.cern.ch:1094 tid=00007fbbfb3ff700 source=EosMgmHttpHandler:198          tident= sec=(null) uid=99 gid=99 name=- geo="" verb=COPY path=xxx.root
```

- At 00:47:20, FTS timeout after 10 minutes:
```
INFO    Sat, 05 Oct 2024 00:47:20 +0200; Gfal2: Copy failed with mode 3rd pull: copy timeout of 601s
[...]

PR #2355, opened 2024-10-11
2354 race condition http tpc pull

ccaffy:
As described in #2354, the PULL transfer will fail if the remote size of the file cannot be fetched within a timeout of 60 seconds.
I also removed the second call to `DetermineXferSize()` during the transfer as this is an information that we get during the first call.

Also, this PR contains the removal of the XRD_CHUNK_RESP preprocessing directive. It was always set to true so I removed it to clean the code!


src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -442,28 +439,32 @@ int TPCHandler::DetermineXferSize(CURL *curl, XrdHttpExtReq &req, State &state,
                                   bool &success, TPCLogRecord &rec, bool shouldReturnErrorToClient) {
     success = false;
     curl_easy_setopt(curl, CURLOPT_NOBODY, 1);
+    // Set a custom timeout of 120 seconds (= CONNECT_TIMEOUT for convenience) for the HEAD request
+    curl_easy_setopt(curl, CURLOPT_TIMEOUT, CONNECT_TIMEOUT);
```
abh3, line 443: Is there a reason why CONNECT_TIMEOUT is not configurable? Just wondering.

src/XrdTpc/XrdTpcTPC.cc
```diff
@@ -442,28 +439,32 @@ int TPCHandler::DetermineXferSize(CURL *curl, XrdHttpExtReq &req, State &state,
                                   bool &success, TPCLogRecord &rec, bool shouldReturnErrorToClient) {
     success = false;
     curl_easy_setopt(curl, CURLOPT_NOBODY, 1);
+    // Set a custom timeout of 120 seconds (= CONNECT_TIMEOUT for convenience) for the HEAD request
```
abh3, line 442: Hmmm, as far as I can see CONNECT_TIMEOUT is defined as 60 so why is the comment saying 120?

abh3 (changes requested):
Just a curiosity question and a comment correction which may raise other issues.

src/XrdTpc/XrdTpcTPC.cc
ccaffy: I believe it's because we've set it to 60 seconds (Linux default connection timeout) : https://github.com/xrootd/xrootd/pull/2242#discussion_r1588798785
If there's a need to make it configurable, I can of course, tell me :)

src/XrdTpc/XrdTpcTPC.cc
ccaffy: Yes, typo from me, will fix.

src/XrdTpc/XrdTpcTPC.cc
abh3: No I tdon't think we need to make it configurable at this point. Let's see if we get any complaints.

abh3 (approved):
Thanks!

amadio (approved):

PR #2358, opened 2024-10-17
[PMark] Readjust the definition of src and dest relative to the server.

abh3:
Realign FF packets to correspond to the technical spec as weird as it is.


abh3:
I am closing this pull request as the modification were included in pull request 2359.

https://github.com/xrootd/xrootd/pull/2359/

PR #2362, opened 2024-10-24
[XrdHttp] GET - Do not send a double response to a single request

### LINKED ISSUE #2351: Invalid chunk framing for HTTP
Under heavy load, we are getting an error from our HTTP clients, "invalid byte in chunk length".  This occurs when XrdPfc gets into a bad state and starts spitting out frequent errors.

In chunked mode, after a complete chunk is sent there is a header containing a hexadecimal number encoded as an ASCII string followed by a CRLF.  The client error message suggests there's something in the header that's not an ASCII-encoded digit; I'm thinking there's potentially some sort of server error message bleeding through.

I've not been able to reproduce it reliably but, to me, this suggests there's something in the `GET` error path that is incorrectly returning an error message when in chunked mode.  We should review the error handling around the `GET` path and see if there's an obvious place where we write to the socket _not_ in chunked mode.

ccaffy:
After 200 OK is sent to the client, the server should not send another response in case a failure happens. This breaks the HTTP protocol.

Fixes #2351

I've just put the logic of sending trailer to the client in a function and call that function in case an error happens.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -1230,7 +1230,11 @@ int XrdHttpReq::ProcessHTTPReq() {
             memcpy(xrdreq.close.fhandle, fhandle, 4);
 
             if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
-              prot->SendSimpleResp(404, NULL, NULL, (char *) "Could not run close request.", 0, false);
+              if (m_transfer_encoding_chunked && m_trailer_headers && m_status_trailer) {
+                std::string errMsg {"Could not run close request."};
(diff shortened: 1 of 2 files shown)


bbockelm:
@ccaffy - note I also pushed a fix onto https://github.com/xrootd/xrootd/pull/2300/commits/606f2d8bf5fde0e0364402d76dc1f133462ec6f5 as I was debugging there.

Given there would be huge merge conflicts on that PR if this is merged, probably better to use the other one.

ccaffy:
Hi @bbockelm , Thanks for the fix! Indeed you have done pretty much the same work as me. I only worked on the GET part though. So yours is more complete. Feel free to push your fixes with what you've done ! I'm therefore closing this MR.

Issue #794, opened 2018-08-08
Purpose of NODEBUG macro and whole logic with it


vp1981:
Hello,
I tried to compile the ```xrootd``` project with ```NODEBUG``` macro set and it failed. A quick look at the errors tells me that some code was written as if ```NODEBUG``` is always unset. Then what is the purpose of this macro and all the logic with it?

Some errors:
- from GCC (very intolerant to some code styles):
```
tmp/xrootd-4.8.4/src/XrdSut/XrdSutAux.cc: In function ‘const char* XrdSutHome()’:                                                                                  
tmp/xrootd-4.8.4/src/XrdSut/XrdSutAux.cc:474:54: error: suggest braces around empty body in an ‘if’ statement [-Werror=empty-body]                                 
          DEBUG("Warning: home directory undefined! ");                                                                                                                           
                                                      ^                                                                                                                           
cc1plus: all warnings being treated as errors
...
```
- from CLANG:
```
/home/vladimir/tmp/xrootd-4.8.4/src/XrdOfs/XrdOfs.cc:217:22: error: use of undeclared identifier 'epname'                                                                         
      XrdOfsFS->Emsg(epname, error, EADDRINUSE, "open directory", dir_path);                                                                                                      
                     ^                                                                                                                                                            
/home/vladimir/tmp/xrootd-4.8.4/src/XrdOfs/XrdOfs.cc:221:4: error: use of undeclared identifier 'epname'                                                                          
   AUTHORIZE(client,&Open_Env,AOP_Readdir,"open directory",dir_path,error);
...
```

P.S. I found this while I was researching Issue #790 .

vp1981:
@xrootd-dev (Andy),
may be add a TODO item, say for xrootd 5?

PR #2367, opened 2024-11-03
[Pfc] Reduce verbosity of errors originating from remote block reads.

### LINKED ISSUE #2327: [XrdPfc] Log messages too frequent for cache timeout
When a cache is using virtual placement and the remote site backing it has issues, 
The following lines keep throwing timeout errors and bloat the log files to 90+GB/day. 
https://github.com/xrootd/xrootd/blob/ae3391f39bf3993dc9d767c00bf55611bd20f6a5/src/XrdPfc/XrdPfcFile.cc#L1220 and https://github.com/xrootd/xrootd/blob/ae3391f39bf3993dc9d767c00bf55611bd20f6a5/src/XrdPfc/XrdPfcFile.cc#L1356
Is it possible to throttle the logs to once/file or to a certain frequency? Or are there other possible ways to trim the logs?

osschar:
Report each error only once for each IO object, then report total counts for each error type once the IO object is detached.

Previous repeating errors have been downgraded to Debug level.

This addresses #2327.

Targeting 5.7.2.

### DIFF
--- a/src/XrdPfc/XrdPfcIO.hh
+++ b/src/XrdPfc/XrdPfcIO.hh
@@ -77,6 +77,7 @@ private:
 
    void SetInput(XrdOucCacheIO*);
 
+protected:
    // Variables used by File to store IO-relates state. Managed under
    // File::m_state_cond mutex.
    friend class File;
@@ -85,6 +86,15 @@ private:
    int    m_active_prefetches {0};
    bool   m_allow_prefetching {true};
    bool   m_in_detach         {false};
+
+   int                m_incomplete_count {0};
+   std::map<int, int> m_error_counts;
+   bool register_incomplete_read() {
+      return m_incomplete_count++ == 0;
+   }
+   bool register_block_error(int res) {
+      return m_error_counts[res]++ == 0;
+   }
 };
 }
 
--- a/src/XrdPfc/XrdPfcIO.hh
+++ b/src/XrdPfc/XrdPfcIO.hh
@@ -77,7 +77,6 @@ private:
 
    void SetInput(XrdOucCacheIO*);
 
-protected:
    // Variables used by File to store IO-relates state. Managed under
    // File::m_state_cond mutex.
    friend class File;
@@ -87,6 +86,7 @@ protected:
    bool   m_allow_prefetching {true};
    bool   m_in_detach         {false};
 
+protected:
    int                m_incomplete_count {0};
    std::map<int, int> m_error_counts;
    bool register_incomplete_read() {
--- a/src/XrdPfc/XrdPfcIOFile.cc
+++ b/src/XrdPfc/XrdPfcIOFile.cc
@@ -153,10 +153,29 @@ void IOFile::DetachFinalize()
 {
    // Effectively a destructor.
 
-   TRACE(Info, "DetachFinalize() " << this);
+   TRACE(Debug, "DetachFinalize() " << this);
 
    m_file->RequestSyncOfDetachStats();
    Cache::GetInstance().ReleaseFile(m_file, this);
+
+#ifndef NODEBUG
+   if (( ! m_error_counts.empty() || m_incomplete_count > 0) && XRD_TRACE What >= TRACE_Error) {
+      char info[1024];
+      int pos = 0, cap = 1024;
+      bool truncated = false;
+      for (auto [err, cnt] : m_error_counts) {
+         int len = snprintf(&info[pos], cap, " ( %d : %d)", err, cnt);
+         if (len >= cap) {
+            truncated = true;
+            break;
+         }
+         pos += len;
+         cap -= len;
+      }
+      TRACE(Error, "DetachFinalize() " << this << " n_incomplete_reads=" << m_incomplete_count
+            << ", block (error : count) report:" << info << (truncated ? " - truncated" : ""));
+   }
+#endif
 
    delete this;
 }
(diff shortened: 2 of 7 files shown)


src/XrdPfc/XrdPfcIOFile.cc
```diff
@@ -153,11 +153,30 @@ void IOFile::DetachFinalize()
 {
    // Effectively a destructor.
 
-   TRACE(Info, "DetachFinalize() " << this);
+   TRACE(Debug, "DetachFinalize() " << this);
 
    m_file->RequestSyncOfDetachStats();
    Cache::GetInstance().ReleaseFile(m_file, this);
 
+#ifndef NODEBUG
```
amadio, line 161: Please see https://github.com/xrootd/xrootd/issues/794. We are thinking about removing the `NODEBUG` macro altogether for XRootD 6.0, so we should probably not introduce new uses of it. I suggest using the more standard `NDEBUG` macro which is often used to disable debug-only functionality and extra checks for relase builds. In the process of removing `NODEBUG`, I will probably select some things to keep (gated by `#ifndef NDEBUG`) instead of removing everything.

src/XrdPfc/XrdPfcTrace.hh
```diff
@@ -64,13 +64,19 @@
    if (XRD_TRACE What >= TRACE_ ## act) SYSTRACE(XRD_TRACE, 0, m_traceID, 0, \
        TRACE_STR_ ## act << x << " " << GetLocalPath())
 
+#define TRACEF_INT(act, x) \
+   if (XRD_TRACE What >= act) \
+   {static const char* t_what[]={"","error ","warning ","info ","debug ","dump "};\
```
amadio, line 69: This will create a static variable for each call site of `TRACEF_INT`. I see that another macro already does this above, though. Should we have a simple declaration shared by all macros instead?

src/XrdPfc/XrdPfcIO.hh
```diff
@@ -77,6 +77,7 @@ private:
 
    void SetInput(XrdOucCacheIO*);
 
+protected:
```
amadio, line 80: Could make only the new things protected? This is to avoid exposing previously private parts of the class unnecessarily.

src/XrdPfc/XrdPfcIO.hh
```diff
@@ -85,6 +86,15 @@ private:
    int    m_active_prefetches {0};
    bool   m_allow_prefetching {true};
    bool   m_in_detach         {false};
+
+   int                m_incomplete_count {0};
+   std::map<int, int> m_error_counts;
+   bool register_incomplete_read() {
```
amadio, line 92: I mean to put `protected:` above this line, then only the new interfaces are protected, and the new and pre-existing members can remain private.

osschar:
Thank you @amadio ... I'm just finishing something else, then I'll put in all those changes.

src/XrdPfc/XrdPfcIOFile.cc
osschar: Took out the ifdef, this is an important part of the error stream, reporting summary of encountered errors for a whole IO session.

src/XrdPfc/XrdPfcIO.hh
osschar: Done, moved it down.

src/XrdPfc/XrdPfcTrace.hh
osschar: Agreed, created the array in XrdPfc namespace.

amadio:
@osschar Please don't @mention people in commit messages. These create GitHub notifications everytime anything is done on such commits :-)

I will squash this one into the previous commit upon merging if you don't mind.

amadio:
Thanks. The Alpine "failure" was just a git problem that happens when I rebase using the GitHub web interface. The code is actually fine and will be in 5.7.2.

bbockelm:
@amadio / @osschar -- heads up!

The Pelican CI tests which test against the `master` branch started having failures in the XCache code overnight.  @turetske did some bisecting and it appears the backport of this PR was what broke it.

amadio:
Thank you for letting us know, @bbockelm. Could you please coordinate with @osschar to fix it? This is why I'd like to have tests for XCache in our test suite.

osschar:
Yes, sure ... @bbockelm, @turetske you'll have to tell me some more, you know :) We can also do osg slack.

bbockelm:
@turetske -- can you provide the above info?  You were the one reporting the issue originally...

amadio:
@bbockelm I merged an extra patch from @osschar which may have been reverted inadvertently by git during a rebase. I'd like to relase 5.7.2 soon, but first I need confirmation that the problem you reported above is fixed. Thank you.

turetske:
I’ll both test and link to the failure if it still fails this morning.Sent from my iPhoneOn Nov 25, 2024, at 4:37 AM, Guilherme Amadio ***@***.***> wrote:﻿
@bbockelm I merged an extra patch from @osschar which may have been reverted inadvertently by git during a rebase. I'd like to relase 5.7.2 soon, but first I need confirmation that the problem you reported above is fixed. Thank you.

turetske:
@amadio Everything seems to be fixed.

amadio:
@turetske Great! Thanks for letting me know. I still would like to know exactly what failed, and how to check myself again in the future if everything is fine in Pelican before making a release, if that's possible. Cheers,

turetske:
@amadio So I can't tell what exactly was the failure. I just know it was a an issue with Stat and XrootD. I can link you to the failing macOS test run (ignore the ubuntu failures, that was a separate bug).

In the future, we will keep pinning macOS to the latest xrootd release. However, we rare planning on adding tests to check against the source of xrootd as a matrix tests. Once that is up and running, I'll show you how to access them.

PR #2369, opened 2024-11-11
Define XRootD_DEFINITIONS in XRootDConfig.cmake

ellert:
See discussion in PelicanPlatform/xrootd-s3-http/pull/53


amadio:
Hi @ellert, please rebase rather than merge. That way only your commits will show up. However, I think the current way you are handling the definitions is not optimal. We need to modernize the CMake to be target-based, and get the required flags via the [`INTERFACE_COMPILE_OPTIONS`](https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_COMPILE_OPTIONS.html) property from the targets, by using `target_compile_options(...)` then installing the targets. This needs a major refactoring of parts of the build system, so I my plan is to work on it for v6.0 next year. We can have a stopgap solution in between, but we need to ensure that the build system and `XRootDConfig.cmake` do not get out of sync, which the current patch doesn't do. Cheers,

Issue #2349, opened 2024-09-30
`Age` header should be sent for a `HEAD` request


bbockelm:
The `Age` header is currently only sent when a `GET` is performed against an XCache configuration.

This is a bit of a catch-22: there's no way to query if a cache has a given file without it triggering XCache to download a part of the file (I currently do a byte-range request for the first byte of the file).  Due to #2348, that means XCache effectively always claims to have the file -- making the header less-than-useful.

The `Age` header logic is fairly simple: let's go ahead and also send it for the HTTP `HEAD` verb.

abh3:
So, Matevz, Alja, and I agree the age header should be returned with a HEAD request *if* the server has a file cache. The HTTP plugin can easily determined if a file cache exists. An AGE 0 will imply the file does not exist in the cache though strictly you should be using GET with IF-CACHED to avoid race conditions. We will proceed on this path unless we hear otherwise. Also, please read our responses to pull request 2348 on how we will handle the side effects of GET vs HEAD.

bbockelm:
@abh3 - from my reading of [RFC 9111 ("HTTP Caching")](https://httpwg.org/specs/rfc9111.html#field.age), I think `Age` should be missing if there's the cache miss case (rather than set to 0).

This is not said outright but implied by this sentence:

> The presence of an Age header field implies that the response was not generated or validated by the origin server for this request.

Since, in the cache miss case, we contact the origin server, the `Age` header shouldn't be included.

abh3:
Well, since Mozilla had a big hand in defining he spec, I'd go with their interpretation:
"The Age header is usually close to zero. If it is Age: 0, it was probably fetched from the origin server; otherwise, it was usually calculated as a difference between the proxy's current date and the [Date](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date) general header included in the HTTP response." Note that on a head request, age 0 may be interpreted as not existing in the cache since nothing was actually fetched. On a get it may be interpreted as above. I wouldn't put a huge amount of interpretation on what is a rather loose spec.

bbockelm:
There's still some ambiguity between "this was cached 1ms ago and we rounded to zero" and "this is not cached".  It'd be good to be able to distinguish the two for completeness.

However, I think we can live with things as proposed - might mean we just ignore a quite-new copy.

@osschar - are you and Alja working on this?  Who is holding the potato?

abh3:
[...]

PR #2372, opened 2024-11-20
[Pfc] Stat call: fix behavior and improve performance, part of #2349

osschar:
`PFC` part of addressing #2349.

### Notes:
1. the branch is on xroot.git so it's potentially easier to manipulate by @abh3  and @amadio.
2. the PR is agianst master -- it is targeting 5.7.2.

### Changes

- For most parts, uses stat of the data file for reporting.

- st_size is always set to the full size of the remote file. This is either obtained from cinfo file or from xattr variable created when the file is initially placed into cache (this is a new optimization).

- st_atim is set to 0 when the file is not considered to be fully cached, as determined by the pfc.onlyifcached config settings.

### Open questions:

1. XrdPfc::Cache::Stat() now always returns 0 (success) or an -errno
code. Specification in XrdOucCahce::Stat() says it should return > 0 (ie, 1)
if the operation is to be forwarded to the federation. It is questionable if
this is indeed even desired -- but could be part of an extended protocol or
additional opaques parameters / flags.

2. st_atime is set to 0 if file is to be considered not fully cached. This is
set both for when stat is called on an currently open file and when it is
performed from the disk. Arguably, in the first case, when file currently
active, this setting could (should?) be avoided.

3. Should Stat() call protect the file from immediate purging? This means it
is guranteed to stick around at least for the purge-interval, 5 min or so.
This could be further tuned.

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -31,6 +31,7 @@
 #include "XrdSys/XrdSysPthread.hh"
 #include "XrdSys/XrdSysTimer.hh"
 #include "XrdSys/XrdSysTrace.hh"
+#include "XrdSys/XrdSysXAttr.hh"
 
 #include "XrdXrootd/XrdXrootdGStream.hh"
 
@@ -42,6 +43,8 @@
 #include "XrdPfcInfo.hh"
 #include "XrdPfcIOFile.hh"
 #include "XrdPfcIOFileBlock.hh"
+
+extern XrdSysXAttr *XrdSysXAttrActive;
 
 using namespace XrdPfc;
 
@@ -411,7 +414,7 @@ void Cache::ReleaseRAM(char* buf, long long size)
 
 File* Cache::GetFile(const std::string& path, IO* io, long long off, long long filesize)
 {
-   // Called from virtual IO::Attach
+   // Called from virtual IOFile constructor.
    
    TRACE(Debug, "GetFile " << path << ", io " << io);
 
@@ -447,6 +450,7 @@ File* Cache::GetFile(const std::string& path, IO* io, long long off, long long f
       }
    }
 
+   // This is always true, now that IOFileBlock is unsupported.
    if (filesize == 0)
    {
       struct stat st;
@@ -504,7 +508,7 @@ void Cache::ReleaseFile(File* f, IO* io)
    dec_ref_cnt(f, true);
 }
 
-  
+
 namespace
 {
 
@@ -581,7 +585,8 @@ void Cache::inc_ref_cnt(File* f, bool lock, bool high_debug)
 
 void Cache::dec_ref_cnt(File* f, bool high_debug)
 {
-   // Called from ReleaseFile() or DiskSync callback.
+   // NOT under active lock.
+   // Called from ReleaseFile(), DiskSync callback and stat-like functions.
 
    int tlvl = high_debug ? TRACE_Debug : TRACE_Dump;
    int cnt;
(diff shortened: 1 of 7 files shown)


amadio:
Doesn't compile on macOS: https://my.cdash.org/viewBuildError.php?buildid=2730063. Could you please have a look?

abh3:
Answers to Open Questions:
1) Well, the API specs assume that if you have no knowledge of the fie you must return a positive number and the caller is responsible for figuring it out. It also implies that atime is set to zero as the file cannot be in the cache. I don't see how you can violate that spec as there is no defined interface to reliably forward the stat(). So, how did you manage to ignore the API and make this work?

2) I don't understand the issue here. If the file is currently open is may or may not be "cached". At some point if may be sufficiently cached if it's a new file as, presumably, it will be written via [pre]fetching. Subsequent stat() calls should reflect the current status. What am I missing here?

3) My gut reaction is that it is unnecessary and adds yet more complexity for dubious use cases. So, I'd say no.

amadio:
Same for this one, could we keep this for 5.8 which will come next month? This would give more time for discussions and testing. Thanks!

osschar:
Hmmh, the Stat that is now is quite broken ... and according to Andy's answers what is in this PR is correct (other than figuring out the 1 return from Stat). Pelican folks are eager to get this in (together with the Brian's http stat fixes) so they can move to the next level of testing.

As this will not break anything and only improves Stat() ... could we put it in and then fine-tune it in 5.8, if needed.

osschar:
The last commit addresses case 1. -- given the rest of the stat call processing is implemented, the correct way of doing this for now is to return +1 when the file is not in the cache.

Cases 2. and 3. are implemented correctly for now.

amadio:
Ok, I had seen some merge conflicts with your other changes, but I will attempt to resolve them today. If I can't, I will merge the rest of planned changes and ask you to rebase, then we can do the release early next week. Most likely I will manage and release today, though.

It would be nice to add some XCache tests to our test suite, though. That would make me have an easier time merging changes if the lines are already covered by existing tests.

src/XrdPfc/XrdPfc.cc
```diff
@@ -899,6 +907,78 @@ int Cache::LocalFilePath(const char *curl, char *buff, int blen,
...
+
+   if (file_size <= m_configuration.m_onlyIfCachedMinSize)
+   {
+      if (frac_on_disk >= m_configuration.m_onlyIfCachedMinFrac)
+         return true;
+   }
+   else
+   {
+      if (bytes_on_disk >= m_configuration.m_onlyIfCachedMinSize &&
+          frac_on_disk  >= m_configuration.m_onlyIfCachedMinFrac)
```
amadio, line 976: I wonder if this should be `||` instead of `&&`, but it probably doesn't make much of a difference.

amadio (commented):
I like this refactoring. Makes the code clearer and more maintainable. I'm trusting that this has been tested, so I'm adding it to the next release. However, I'd really like to see some XCache tests added to the test suite in the near future. I will try to add a base setup like we have in `tests/XRootD` as a start, then it can be expanded by others with more commands, like @ccaffy did recently for HTTP in #2375.

src/XrdPfc/XrdPfc.cc
amadio: Ok, great, thanks!

amadio:
No, no, the second commit was not lost, I merged it with the previous commit as described above.

osschar:
Sorry, I meant the third one ... the one that returns +1 from Cache::Stat() (2 cases) if the cache doesn't have a valid file. I'm in touch with Andy to see if he needs to push something else on top.

osschar:
There were 3 ... the original, macos atim fix and the last one that made things consistent with what is currently in posix and above. Will send you a patch relative to current master through other channels.
Sorry about the mess :(

amadio:
No worries, I can wait until we resolve this to release.

amadio:
Just push a branch to the main repository with your patch. I can pick it from there.

osschar:
OK, thanks ... branch: pfc-stat-age-2349-commit-3

PR #2375, opened 2024-11-21
[Tests] Added more HTTP tests for CI

ccaffy:
More HTTP tests including GET, GET with Range-requests, HEAD with multiple checksum. Aim is to gradually add coverage to HTTP use-cases.

### DIFF
--- a/tests/XRootD/http.sh
+++ b/tests/XRootD/http.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 function setup_http() {
-	require_commands davix-{get,put,mkdir,rm} openssl
+	require_commands davix-{get,put,mkdir,rm} openssl curl
 	openssl rand -base64 -out macaroons-secret 64
 }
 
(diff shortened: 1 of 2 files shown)


tests/XRootD/http.sh
```diff
@@ -75,5 +75,66 @@ function test_http() {
 	       assert davix-rm "${HOST}/${TMPDIR}/${i}.ref"
 	done
 
+  # GET range-request
+	## Upload a file with a fixed content string
+	alphabetFile="alphabet.txt"
+	alphabetFilePath="${TMPDIR}/$alphabetFile"
+	echo -n "abcdefghijklmnopqrstuvw987" > $alphabetFilePath
+	assert curl -v -L -H 'Transfer-Encoding: chunked' "${HOST}/$alphabetFilePath" --upload-file $alphabetFilePath
+  outputFilePath=${TMPDIR}/output.txt
```
amadio, line 84: Indentation looks a bit funny here. Also, could you please check the script with `shellcheck`? I think we can also avoid repetition by adding an extra assert function that checks two strings for equality (like ASSERT_EQ from Google Test, but for bash).

tests/XRootD/http.sh
```diff
@@ -75,5 +75,66 @@ function test_http() {
...
+  ## Download the file to a file and sanitize the output (remove '\r')
+  curl -v -L --silent -H 'range: bytes=0-3,24-26' "${HOST}/$alphabetFilePath" --output - | tr -d '\r' > $outputFilePath
+  ## Check the first content range header received
+  contentRange=`cat $outputFilePath | grep -i 'Content-range' | awk 'NR==1'`
```
amadio, line 88: I suggest using `$( commands )` instead of back ticks, it's easier to read and allows nesting.

tests/XRootD/http.sh
ccaffy: very weird, I don't see this indentation issue on my IDE...

ccaffy: Yes, OK for `assert_eq()` :+1:

ccaffy: DOne

PR #2380, opened 2024-12-01
Add type checking to format strings used in the client library logging routines and fix the errors that it triggers

ellert:
The format for some size_t variables were recently changed from %d to %llu. This fixed a problem on 64 bit architectures, since there size_t is 64 bits wide and %d is the format for a 32 bit type and %llu is the format for a 64 bit type. However, this change introduced a regression on 32 bit architectures where size_t is 32 bits wide.

This change caused some of the tests to fail with a segmentation fault on the Debian armhf architecture:

The following tests FAILED:
	108 - XRootD::noauth::test (Failed)
	111 - XRootD::host::test (Failed)
	114 - XRootD::unix::test (Failed)
	117 - XRootD::sss::test (Failed)
	120 - XRootD::http::test (Failed)

This first commit adds a type cast so that the new format works for all architectures.

It is possible to get warnings from the compilers for bad format strings that don't match the types of the variables by using __atribute__((format)). The second commit adds this attribute to the logging routines in the client library.

The third commit fixes the errors from this type checking produced by the compiler for the master branch.

The fourth commit contains additional fixes for the devel branch relared to the "timeout uses time_t" change. This commit should no go to mastar untill that change does.

The first three commits should go to master.

### DIFF
--- a/src/XrdCl/XrdClCopyProcess.cc
+++ b/src/XrdCl/XrdClCopyProcess.cc
@@ -368,7 +368,7 @@ namespace XrdCl
     Log *log = DefaultEnv::GetLog();
     std::vector<PropertyList>::iterator it;
 
-    log->Debug( UtilityMsg, "CopyProcess: %llu jobs to prepare",
+    log->Debug( UtilityMsg, "CopyProcess: %zu jobs to prepare",
                 pImpl->pJobProperties.size() );
 
     std::map<std::string, uint32_t> targetFlags;
--- a/src/XrdCl/XrdClAsyncVectorReader.hh
+++ b/src/XrdCl/XrdClAsyncVectorReader.hh
@@ -135,7 +135,7 @@ namespace XrdCl
               if( !chfound )
               {
                 log->Error( XRootDMsg, "[%s] VectorReader: Impossible to find chunk "
-                            "buffer corresponding to %d bytes at %ld",
+                            "buffer corresponding to %d bytes at %lld",
                             url.GetHostId().c_str(), rdlst.rlen, rdlst.offset );
                 readstage = ReadDiscard;
                 continue;
@@ -181,7 +181,7 @@ namespace XrdCl
               if( !st.IsOK() || st.code == suRetry )
                  return st;
 
-              log->Dump( XRootDMsg, "[%s] VectorReader: read buffer for chunk %d@%ld",
+              log->Dump( XRootDMsg, "[%s] VectorReader: read buffer for chunk %d@%lld",
                          url.GetHostId().c_str(), rdlst.rlen, rdlst.offset );
 
               //----------------------------------------------------------------
--- a/src/XrdCl/XrdClClassicCopyJob.cc
+++ b/src/XrdCl/XrdClClassicCopyJob.cc
@@ -890,8 +890,8 @@ namespace
 
         if( !ch->status.IsOK() )
         {
-          log->Debug( UtilityMsg, "Unable read %d bytes at %ld from %s: %s",
-                      ch->chunk.GetLength(), ch->chunk.GetOffset(),
+          log->Debug( UtilityMsg, "Unable read %d bytes at %llu from %s: %s",
+                      ch->chunk.GetLength(), (unsigned long long) ch->chunk.GetOffset(),
                       pUrl->GetObfuscatedURL().c_str(), ch->status.ToStr().c_str() );
           delete [] (char *)ch->chunk.GetBuffer();
           CleanUpChunks();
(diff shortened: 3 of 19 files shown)


ellert:
The fourth commit (related to changes in the devel branch) was split out to a separate PR #2412.
This PR was then rebased to use the master branchas base.

src/XrdCl/XrdClCopyProcess.cc
```diff
@@ -369,7 +369,7 @@ namespace XrdCl
     std::vector<PropertyList>::iterator it;
 
     log->Debug( UtilityMsg, "CopyProcess: %llu jobs to prepare",
-                pImpl->pJobProperties.size() );
+                (unsigned long long) pImpl->pJobProperties.size() );
```
amadio, line 372: Rather than cast, let's fix with `%zd` instead of `%llu`. It was my mistake to use the wrong one, but accorting to [GCC](https://www.gnu.org/software/libc/manual/html_node/Integer-Conversions.html), this is the better fix.

ellert: In the past I was asked not to use %z in xrootd for portability reasons. When I used it in the past I was asked to change it to use a cast instead. Is %z OK to use now? A grep of the code shows only one single use of the %z in the entire code. I also think %z is a better choice and will gladly change it in this PR, but I just want to check so that I won't be asked to change it back again if i do.

abh3: Technically, one should not use %z, as  you said, as it's not portable. So, the recommended solution is to cast the size_t variable to a unsigned long and use %lu. This addresses size differences for size_t.  I also grepped for use of %z and was surprised to find actually 4 uses:
XrdCl/XrdClLog.cc:    strftime( tz, 8, "%z", &tsNow );
XrdCl/XrdClUtils.cc:    strftime( now, 30, "%Y-%m-%d %H:%M:%S %z", &tsNow );
XrdOuc/XrdOucUtils.cc:      {snprintf(buff, bsz, "%zu", bytes);
XrdXrootd/XrdXrootdTpcMon.cc:"\"Src\":\"%s\",\"Dst\":\"%s\",\"Size\":%zu}";

I suppose I should not have been because I was responsible for two of them, sigh.

ellert: > to find actually 4 uses:

I did the grep in master, so found one less, since one of the matches in your list is only on devel. Two of the matches are for strftime format strings and not for printf/scanf and friends. So we agree on the number.

amadio: Maybe it's time to revisit the portability issue. Before C99, this might indeed have been an issue, but today, all platforms we care about should be fine with it. [MUSL libc](https://github.com/bpowers/musl/blob/94cb2ec2a0ffcb47d24dbf7a30e462505396cf54/src/stdio/vfprintf.c#L79) (i.e. Alpine, Gentoo, Void Linux, etc),  [macOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/printf.3.html),  [FreeBSD](https://man.freebsd.org/cgi/man.cgi?printf(3)), [OpenBSD](https://man.openbsd.org/printf.3#_zu) all support it. Even [Microsoft](https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-140#size-prefixes-for-printf-and-wprintf-format-type-specifiers) supports it these days. Unless there is a platform we care about where it is really an issue, I'd say let's bite the bullet and migrate to it. @abh3 What do you think?

abh3:
I would be fine with it but when you look at the documentation many times it does not list %z making you wonder whether or not it's supported. Given that you really don't need to use %z since the replacement is a trivial type cast,; we can avoid all of this uncertainty here by using that approach.

src/XrdCl/XrdClCopyProcess.cc
ellert: https://cplusplus.com/reference/clibrary/ says:

C++98 includes the C library as described by the 1990 ISO C standard and its amendment #1 (ISO/IEC 9899:1990 and ISO/IEC 9899:1990/DAM 1).

C++11 includes the C library as described by the 1999 ISO C standard and its Technical Corrigenda 1, 2 and 3 (ISO/IEC 9899:1999 and ISO/IEC 9899:1999/Cor.1,2,3), plus [<cuchar>](https://cplusplus.com/%3Ccuchar%3E) (as by ISO/IEC 19769:2004).

The z length modifier is part of the C99 standard, so is therefore available in C++11 and later. The minimum C++ version for xrootd is newer than that:

https://github.com/xrootd/xrootd/blob/b4d994962a132ac81e29302a3fef6978891369bf/cmake/XRootDOSDefs.cmake#L19-L20

I changed the PR to use the z length modifier as suggested by @amadio.

amadio: Thank you, I will include this in 5.7.3 which will be released shortly.

PR #2381, opened 2024-12-01
[XrdSciTokens] Unit test coverage for token-based authorization

bbockelm:
This adds an initial, simple unit test for XrdSciTokens authentication.

The goal isn't to be comprehensive but to lay down a baseline that can be used for future changes.

### DIFF
--- a/tests/XRootD/CMakeLists.txt
+++ b/tests/XRootD/CMakeLists.txt
@@ -9,6 +9,13 @@ if(ENABLE_HTTP_TESTS)
   list(APPEND XROOTD_CONFIGS http)
 endif()
 
+# Note: HAVE_SCITOKEN_CONFIG_SET_STR is required to override the CA file
+# for the scitokens library; otherwise, the test will fail with TLS errors.
+if(BUILD_SCITOKENS AND HAVE_SCITOKEN_CONFIG_SET_STR)
+  list(APPEND XROOTD_CONFIGS scitokens)
+  list(APPEND scitokens_FIXTURES SciTokens)
+endif()
+
 foreach(CONFIG ${XROOTD_CONFIGS})
   add_test(NAME XRootD::${CONFIG}::setup
     COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/test.sh ${CONFIG} setup")
@@ -34,10 +41,19 @@ foreach(CONFIG ${XROOTD_CONFIGS})
   list(APPEND ${CONFIG}_LOGS ${CMAKE_CURRENT_BINARY_DIR}/${CONFIG}/client.log)
   list(APPEND ${CONFIG}_LOGS ${CMAKE_CURRENT_BINARY_DIR}/${CONFIG}/xrootd.log)
 
+  list(APPEND ${CONFIG}_FIXTURES XRootD::${CONFIG})
+
   set_tests_properties(XRootD::${CONFIG}::test
     PROPERTIES
-      FIXTURES_REQUIRED XRootD::${CONFIG}
+      FIXTURES_REQUIRED "${${CONFIG}_FIXTURES}"
       ENVIRONMENT "BINARY_DIR=${CMAKE_BINARY_DIR}"
       ATTACHED_FILES_ON_FAIL "${${CONFIG}_LOGS}"
   )
 endforeach()
+
+if(BUILD_SCITOKENS AND HAVE_SCITOKEN_CONFIG_SET_STR)
+  set_tests_properties(XRootD::scitokens::setup
+    PROPERTIES
+      FIXTURES_REQUIRED SciTokens
+  )
+endif()
--- a/tests/XRootD/scitokens.cfg
+++ b/tests/XRootD/scitokens.cfg
@@ -1,0 +1,26 @@
+set name = scitokens
+set port = 7095
+
+set pwd = $PWD
+set src = $SOURCE_DIR
+
+set scitokensdir = $PWD
+
+xrootd.seclib libXrdSec.so
+xrd.protocol XrdHttp:7095 libXrdHttp.so
+
+xrd.tlsca certfile $scitokensdir/../issuer/tlsca.pem
+xrd.tls $scitokensdir/../issuer/tls.crt $scitokensdir/../issuer/tls.key
+
+ofs.authlib ++ libXrdAccSciTokens.so config=$src/scitokens-module.cfg
+ofs.authorize 1
+acc.authdb $src/scitokens.authdb
+
+http.header2cgi Authorization authz
+http.exthandler xrdtpc libXrdHttpTPC.so
+
+# For root://-protocol test, force usage of a token.
+sec.protocol ztn
+sec.protbind * only ztn
+
+continue $src/common.cfg
--- a/tests/XRootD/scitokens.sh
+++ b/tests/XRootD/scitokens.sh
@@ -26,6 +26,20 @@ function setup_scitokens() {
 	fi
 	chmod 0600 "$OUTPUTDIR/token"
 
+	# Create a create-only token
+	if ! xrdscitokens-create-token issuer_pub_1.pem issuer_key_1.pem test_1 \
+		"https://localhost:7095/issuer/one" storage.create:/subdir > "$OUTPUTDIR/token_create"; then
+		echo "Failed to create 'storage.create' token"
+		exit 1
+	fi
+
+	# Create a modify token
+	if ! xrdscitokens-create-token issuer_pub_1.pem issuer_key_1.pem test_1 \
+		"https://localhost:7095/issuer/one" storage.modify:/subdir_modify > "$OUTPUTDIR/token_modify"; then
+		echo "Failed to create 'storage.modify' token"
+		exit 1
+	fi
+
 	popd || exit 1
 }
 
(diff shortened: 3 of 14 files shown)


amadio:
:+1: For adding tests. However, I don't see them in the coverage report. I think we need to do some special tricks also in plugins to let their coverage be accounted for. I will see if I manage this, so we can see the new coverage in the report when merging. Cheers,

bbockelm:
Note that adding unit tests for write showed the XrdHttp had the wrong response for `PUT` (see [aa60af1](https://github.com/xrootd/xrootd/pull/2381/commits/aa60af1f7fe12b2210e8aadf17020e1c137859f1)).

A full fix is slightly more involved; I figured I'd create a separate issue.

tests/XRootD/CMakeLists.txt
```diff
@@ -34,10 +40,22 @@ foreach(CONFIG ${XROOTD_CONFIGS})
   list(APPEND ${CONFIG}_LOGS ${CMAKE_CURRENT_BINARY_DIR}/${CONFIG}/client.log)
   list(APPEND ${CONFIG}_LOGS ${CMAKE_CURRENT_BINARY_DIR}/${CONFIG}/xrootd.log)
 
+  list(APPEND ${CONFIG}_FIXTURES XRootD::${CONFIG})
+  if(${CONFIG} STREQUAL "scitokens")
+    list(APPEND ${CONFIG}_FIXTURES SciTokens)
+  endif()
```
amadio, line 46: Please do this outside of the `foreach` loop. If statements inside a for loop when they are only true once is not a good pattern.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
+
+function setup_scitokens() {
+	require_commands curl
+	if [ -z "$HOSTNAME" ]; then
+		HOSTNAME=`hostname`
+	fi
```
amadio, line 7: This should be taken care of in [`test.sh`](https://github.com/xrootd/xrootd/blob/8a4ca755c57bd01b807a74a4aa3fc89b3f42116b/tests/XRootD/test.sh#L133). I think it's better to set hostname there for everybody. I set it for `root://` protocol already, I suggest we replace with or add
```bash
: ${HOSTNAME:=localhost}

export HOSTNAME
```
in `test.sh`, then we can reuse it [here](https://github.com/xrootd/xrootd/blob/8a4ca755c57bd01b807a74a4aa3fc89b3f42116b/tests/XRootD/http.sh#L26) as well.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+
+	cat > "$PWD/scitokens/authdb" << EOF
+u * /issuer lr
+EOF
```
amadio, line 11: Files with immutable contents should not be created on the fly, but simply commited to the source and used from there directly as `$src/...` in the XRootD configuration files, like is done for `common.cfg`.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+	cat > "$PWD/scitokens/scitokens.cfg" <<EOF
+[Global]
+audience = https://$HOSTNAME:8095
+
+[Issuer test]
+issuer = https://$HOSTNAME:8095/issuer/one
+base_path = /protected
```
amadio, line 18: This file could also be made constant by using `https://localhost:8095/issuer/one` instead of `$HOSTNAME`.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+EOF
+
+	# Setup the issuer and protected directory
+	ln -sf "$PWD/../issuer/export" "$PWD/scitokens/xrootd/issuer"
+	mkdir -p "$PWD/scitokens/xrootd/protected"
```
amadio, line 23: It's cleaner to remove `$PWD/` from paths and use relative paths if it doesn't break anything. For example:
```sh
mkdir -p scitokens/xrootd/protected
```

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+	echo 'Hello, World' > "$PWD/scitokens/xrootd/protected/hello_world.txt"
+
+	# Override the scitoken cache location; otherwise, contents of prior test runs may be cached
+	export XDG_CACHE_HOME=$(mktemp -d "${PWD}/${NAME}/cache-XXXXXX")
+	mkdir -p "$XDG_CACHE_HOME/scitokens"
+
+	# Create a read-only token
+	$PWD/../scitokens/xrdscitokens-create-token ../issuer/issuer_pub_1.pem ../issuer/issuer_key_1.pem test_1 https://$HOSTNAME:8095/issuer/one storage.read:/ > "$PWD/scitokens/token"
+	if [ "$?" -ne 0 ]; then
```
amadio, line 32: Is there a reason not to use [`assert`](https://github.com/xrootd/xrootd/blob/8a4ca755c57bd01b807a74a4aa3fc89b3f42116b/tests/XRootD/test.sh#L9-L11) command here, which was created for this? Also, the very long lines are not nice. Let's either break them or prepend `$PWD/../scitokens` to `PATH`, `pushd issuer` then run the command, it will be a lot more readable.

You could also `export ISSUER=https://$HOSTNAME:8095/issuer/one` and use that.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+	# Create a modify token
+	$PWD/../scitokens/xrdscitokens-create-token ../issuer/issuer_pub_1.pem ../issuer/issuer_key_1.pem test_1 https://$HOSTNAME:8095/issuer/one storage.modify:/subdir_modify > "$PWD/scitokens/token_modify"
+	if [ "$?" -ne 0 ]; then
+		echo "Failed to create 'storage.modify' token"
+		exit 1
+	fi
+}
+
+function teardown_scitokens() {
+	echo
```
amadio, line 53: You don't have to define a function if it's not doing anything. It's run [conditionally](https://github.com/xrootd/xrootd/blob/8a4ca755c57bd01b807a74a4aa3fc89b3f42116b/tests/XRootD/test.sh#L157-L159), if it exists.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,122 @@
...
+EOF
+
+	# Setup the issuer and protected directory
+	ln -sf "$PWD/../issuer/export" "$PWD/scitokens/xrootd/issuer"
+	mkdir -p "$PWD/scitokens/xrootd/protected"
+	echo 'Hello, World' > "$PWD/scitokens/xrootd/protected/hello_world.txt"
+
+	# Override the scitoken cache location; otherwise, contents of prior test runs may be cached
+	export XDG_CACHE_HOME=$(mktemp -d "${PWD}/${NAME}/cache-XXXXXX")
+	mkdir -p "$XDG_CACHE_HOME/scitokens"
```
amadio, line 28: The teardown procedure should always delete the scitokens directory (or the setup, which ensures a clean start), you can probably just set this as `export XDG_CACHE_HOME=${PWD}/${NAME}/cache`.

amadio (changes requested):
Overall very nice, let's iterate a bit and this should be good to be added to the test suite. I would like to setup a generic SSL CA that could work for GSI tests as well, so I will try to adapt it from here. I've had some trouble creating something that will work and pass all tests from `xrdgsitest` in the past, let's see if I manage this time, then we can reuse that for all TLS testing. Cheers,

tests/XRootD/CMakeLists.txt
bbockelm: Fixed in latest rebase.

tests/XRootD/scitokens.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
bbockelm: Latest rebase runs clean in my version of `shellcheck`.

tests/XRootD/test.sh
```diff
@@ -147,7 +147,7 @@ function teardown() {
 	# Kill all processes that created pid files and are still running
 	for PIDFILE in *.pid; do
 		test -s "${PIDFILE}" || continue
-		PID="$(ps -o pid= "$(cat "${PIDFILE}")")"
+		PID="$(ps -o pid= "$(cat "${PIDFILE}")" || continue)"
```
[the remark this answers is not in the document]
bbockelm: Looks like you fixed this in https://github.com/xrootd/xrootd/commit/c9aaf675bbf87adccc215cc0a29c5e25ee061f31 so I dropped the commit from the latest rebase.

tests/XRootD/scitokens.cfg
```diff
@@ -0,0 +1,24 @@
...
+xrd.protocol XrdHttp:8095 libXrdHttp.so
+
+xrd.tlsca certfile $scitokensdir/../issuer/tlsca.pem
+xrd.tls $scitokensdir/../issuer/tls.crt $scitokensdir/../issuer/tls.key
+
+ofs.authlib ++ libXrdAccSciTokens.so config=$scitokensdir/$name/scitokens.cfg
+ofs.authorize 1
+acc.authdb $scitokensdir/scitokens/authdb
+
+scitokens.trace all
```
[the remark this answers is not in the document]
bbockelm: Done.

tests/XRootD/scitokens.sh
bbockelm: Based on your suggestions below, I took a different route -- using `localhost` across the board.  That allows some of the config files to be constant and not need to be generated at runtime.  Latest rebase removes all the `HOSTNAME` pieces.

tests/XRootD/scitokens.sh
bbockelm: Done.

tests/XRootD/scitokens.sh
bbockelm: Took a quick review through the code and did this in most places (the prior `ln` line needs the absolute version so I left it there).

tests/XRootD/scitokens.sh
bbockelm: Removed in latest rebase.

tests/XRootD/scitokens.sh
bbockelm: Done.

bbockelm:
@amadio - this weekend I finally I found some time to sit down and go through your (thorough and useful!) review.  Just pushed an initial version of the rebase; I think I addressed all the comments.

I'm sure there will be a few more minor revisions to touch up the tests on the different platforms but it should be ready for re-review once the unit tests all pass.

tests/XRootD/scitokens.cfg
```diff
@@ -0,0 +1,24 @@
...
+
+set scitokensdir = $PWD
+
+xrootd.seclib libXrdSec.so
+xrd.protocol XrdHttp:8095 libXrdHttp.so
+
+xrd.tlsca certfile $scitokensdir/../issuer/tlsca.pem
+xrd.tls $scitokensdir/../issuer/tls.crt $scitokensdir/../issuer/tls.key
+
+ofs.authlib ++ libXrdAccSciTokens.so config=$scitokensdir/$name/scitokens.cfg
```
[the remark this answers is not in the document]
bbockelm: Done.  The latest commit in the PR now includes tests over the `root://` protocol.

tests/XRootD/scitokens.sh
bbockelm: Refactored quite a bit around here -- this is now wrapped in a function (which makes it far easier to see what's being tested).

`assert` couldn't be used because I need to redirect the output of `curl` and that's not compatible with the function.  I tried wrapping everything into an invocation of `/bin/sh` with a one-liner but it made things into an unreadable mess of quotes (and wasn't ultimately successful).  You're welcome to try to re-introduce it but I wasn't able to figure out an approach that made things better, not worse.

bbockelm:
Ok -- all the tests are passing in the PR that are passing in the `devel` branch.  The failing tests, upon inspection, are clearly failing for other reasons (being triggered in code that isn't touched by this PR).

amadio:
Hi @bbockelm Yes, we are aware of the failures, they are due to IPV6 changes. Thanks for the update, I will review and merge later. I might merge this one directly into master rather than devel. Cheers,

bbockelm:
Yes -- I was tempted to try to edit the PR to point it at the `master` branch instead.  However, I was worried that such a large change might mess up the GitHub metadata and lose the original review comments so I didn't try it.  Feel free to target whichever branch you see as most appropriate.

bbockelm:
@amadio - any news on merging the tests for SciTokens?  There are two other open PRs waiting on this (including https://github.com/xrootd/xrootd/pull/2417), which would be convenient for others.

amadio:
Could you please retarget onto master? I do intend to pick this up before the next release.

tests/XRootD/scitokens.sh
amadio: Thanks, I will merge it as it is now, we can refine later.

bbockelm:
Thanks @amadio!

Please ping when the `devel` rebase is done and I'll rebase #2397 and #2417 which both build on top of these basic unit tests.

It feels great to have some test coverage here!

amadio:
I plan to rebase `devel` after I release 5.7.3, but I can do it now as well if you want to go on with your pull requests.

bbockelm:
That would be much appreciated, thanks!

We keep a separate xrootd branch with all the PRs in-use by Pelican and rebasing will help me tidy up that branch.

amadio:
Ok, I will push what I have for master and update devel on top of that.

amadio:
Done.

PR #2395, opened 2024-12-20
[Pfc] Fix FD leak when reading file-size from cinfo file in Cache::DetermineFullFileSize()

### LINKED ISSUE #2392: [XrdPfc] Major FD leak in XRootD 5.7.2
It appears that the XrdPfc module in XRootD 5.7.2 leaks a file descriptor for every cache hit.

On first blush, it appears to be due to this code:

https://github.com/xrootd/xrootd/commit/78d1fb7b12a8a136606108de7a2399baf4e7e509#diff-d7d0c4214a23465d01f5980c58761d4f64a9df920d96c159cae525230f261ab4R947-R953

introduced by @osschar.

Looking at a few production caches, all had hundreds of thousands of FDs open and were crashing about daily once they hit the relevant system limits.

osschar:
Also fixes leak of XrdOssDF object on the heap.

Note: this code-path is only taken if file-size can not be determined from the xattrs (a new feature, written when file is first created in the cace) so this expected to happen:
- during transition to this new version where old files do not have the information yet;
- when a FS that does not support xattrs is used.

Fixes #2392

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -944,13 +944,21 @@ long long Cache::DetermineFullFileSize(const std::string &cinfo_fname)
 
    XrdOssDF *infoFile = m_oss->newFile(m_configuration.m_username.c_str());
    XrdOucEnv env;
+   long long ret;
    int res = infoFile->Open(cinfo_fname.c_str(), O_RDONLY, 0600, env);
-   if (res < 0)
-      return res;
-   Info info(m_trace, 0);
-   if ( ! info.Read(infoFile, cinfo_fname.c_str()))
-      return -EBADF;
-   return info.GetFileSize();
+   if (res < 0) {
+      ret = res;
+   } else {
+      Info info(m_trace, 0);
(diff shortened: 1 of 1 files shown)


amadio:
@abh3 @osschar I guess we should create a v5.7.3 with this fix soon. What do you think?

osschar:
Yes please! And humble apologies for screwing this up :( Sigh, doing too many things at the same time.

abh3:
Yes, I vote for a 5.7.3 but I also would like to have @osschar devise an Xcache test that test for file descriptor leakage (it should be pretty simple) and let's make sure that 5.7.3 does have this problem nailed.

amadio:
I can only agree with @abh3 here. It would be very good to extend coverage of the test suite, as it makes it much easier to catch this kind of problem early and ensure there are no regressions.

@osschar Please take a look at the tests/XRootD subdirectory, where we have server tests setup. It might be a good starting point for creating XCache tests.

bbockelm:
FWIW -- I ran this on a cache overnight and, as of this morning, the file descriptor count remains low.  So, the patch itself looks good.

amadio:
Thank you for testing @bbockelm!

PR #2397, opened 2024-12-31
[XrdSciTokens] Handle multiple authorization token setups

bbockelm:
(Note this is built on top of #2381 so those extra commits show up here.  Will need a rebase once that PR is merged)

This adds new functionality to the XrdSciTokens plugin, allowing it to manage multiple tokens.  It introduces the concept of a "_required issuer_" (meaning an verified token *must* be present in the request from that issuer) and an "_acceptable issuer_" (meaning the verified token from that issuer is permitted to result in authorizations).

The use case is for a dual-token setup in Pelican: to access data, one token is required from the cache (issued by the data federation's issuer) and a second one from the user (issued by the data owner's issuer).  In that setup, the cache token is from the "required issuer" but the user's token is from the "acceptable issuer" (the latter may be used, for example, to determine the POSIX user account to read the data as at the origin).

Notes on the contents:
- By default, there's no changed behavior.  No issuers are required and all issuers are acceptable for any operation.
- The required/acceptable labels can be "read", "write", "all", or "none".  This allows an issuer to be only acceptable for "read", for example -- meaning there's now a mechanism to reject create/modify/write tokens from an issuer if the origin would like to be read-only (also a Pelican TODO item).
- `README.md` has been updated to explain the new configuration parameters in the `scitokens.cfg`.
- Unit test coverage is added for the various helper logic functions in XrdSciTokens.  This required a refactoring of the code so it's available in a header (I don't want to build the code twice with potentially different flags).
- To comply with RFC 6750, `access_token` is now an accepted CGI parameter within the URL.  The various sanitization routines have been updated accordingly.
- **IMPORTANT**: When developing unit tests for the `access_token` sanitization, I found `XrdOucEnv` currently fails to sanitize when `authz` is the only environment entry (e.g., `EnvTidy` run on `authz=foo` resulted in `authz=foo`).  This bug was fixed in the rewrite of `XrdOucEnv`.  If need be, we may want to backport the fix to master.
- Integration test has been added for both the HTTPS and root protocols.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -243,7 +243,7 @@ struct MapRule
 
     const std::string match(const std::string &sub,
                             const std::string &username,
-                            const std::string &req_path,
+                            const std::string_view &req_path,
                             const std::vector<std::string> &groups) const
     {
         if (!m_sub.empty() && sub != m_sub) {return "";}
@@ -251,7 +251,7 @@ struct MapRule
         if (!m_username.empty() && username != m_username) {return "";}
 
         if (!m_path_prefix.empty() &&
-            strncmp(req_path.c_str(), m_path_prefix.c_str(), m_path_prefix.size()))
+            strncmp(req_path.data(), m_path_prefix.c_str(), m_path_prefix.size()))
         {
             return "";
         }
@@ -273,6 +273,14 @@ struct MapRule
     std::string m_result;
 };
 
+// Control whether a given issuer is required for the paths it authorizes
+enum class AuthzSetting {
+    None, // Issuer's authorization is not necessary
+    Read, // Authorization from this issuer is necessary for reads.
+    Write, // Authorization from this issuer is necessary for writes.
+    All, // Authorization from this issuer is necessary for all operations.
+};
+
 struct IssuerConfig
 {
     IssuerConfig(const std::string &issuer_name,
@@ -284,8 +292,12 @@ struct IssuerConfig
                  const std::string &default_user,
                  const std::string &username_claim,
                  const std::string &groups_claim,
-                 const std::vector<MapRule> rules)
+                 const std::vector<MapRule> rules,
+                 AuthzSetting acceptable_authz,
+                 AuthzSetting required_authz)
         : m_map_subject(map_subject || !username_claim.empty()),
+          m_acceptable_authz(acceptable_authz),
+          m_required_authz(required_authz),
           m_authz_strategy(authz_strategy),
           m_name(issuer_name),
           m_url(issuer_url),
@@ -298,6 +310,8 @@ struct IssuerConfig
     {}
 
     const bool m_map_subject;
+    const AuthzSetting m_acceptable_authz;
+    const AuthzSetting m_required_authz;
     const uint32_t m_authz_strategy;
     const std::string m_name;
     const std::string m_url;
@@ -308,8 +322,6 @@ struct IssuerConfig
     const std::vector<std::string> m_restricted_paths;
     const std::vector<MapRule> m_map_rules;
 };
-
-}
 
 class OverrideINIReader: public INIReader {
 public:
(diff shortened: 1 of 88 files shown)


bbockelm:
@ccaffy - would you feel comfortable reviewing this?  I don't think you've touched much of the scitokens code but you may be interested in learning.  Though, admittedly, the important part is to make sure there's no changes in default behavior and the new functionality has reasonable unit/integration test coverage.

The reason I thought of you is the pieces on:
- Better RFC 6750 compliance through the `access_token` parameter handling (and corresponding scrubbing)
- The fix of the broken scrubbing code in the `XrdOucEnv`.
- The unit tests covering the scrubbing code.

abh3:
It appears that the pull request now needs to be rebased (perhaps Guilherme can fix that). As for the XrdOucEnv change. It would be much better if the fix for handling the case where authz was the only token on the command line should be a separate pull request as it's a bug fix and not a feature plus it makes it possible to port.

As for this feature, we still don't understand where the requirement for this feature came from beyond Pelican. Even for Pelican we don't quite understand how this is actually used in practice. Would it be possible to get together next week so that we understand the impact this feature would have on performance, reliability, and security and whether it could be better handled with a Pelican specific header?

amadio:
I have added the ARM builds on `master`, rebased `devel` onto it, and rebased this branch onto the new `devel`, which no longer contains the ipv6 changes which caused previous CI failures. The CI should be green and also cover ARM architecture now. Cheers,

amadio:
Please note that after the last push the CI is green on devel, so any failures in CI now are from the changes being introduced here.

abh3:
All the more reason that the "fix" for authz removal should be a separate patch.

amadio:
Just rebased this on top of current `devel` branch. I noticed that your changes kind of undo my change in https://github.com/xrootd/xrootd/commit/81970bbf571dfaf311a8da4693b8f97e9832013c to not have code duplication now that we use the `is_subdirectory()` function is another place (i.e. `xrdfs`, in https://github.com/xrootd/xrootd/commit/d57d992624d6c7d6ffc2aa5642c63ea6d6a37cf8). For now I just resolved the conflicts in the easiest way, I will add a review now and we can then work through this to get it merged.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -243,15 +243,15 @@ struct MapRule
 
     const std::string match(const std::string &sub,
                             const std::string &username,
-                            const std::string &req_path,
+                            const std::string_view &req_path,
```
amadio: I'd prefer to keep taking a reference to string here. One argument is for consistency: why take only this one as `std::string_view` and keep the others as `std::string&`? Ideally, one should also choose between taking either a reference to string, or a `std::string_view` by value, taking a reference to a `std::string_view` doesn't make a lot of sense. Finally, although it's not the case here, `std::string_view` is often [misused](https://github.com/root-project/root/pull/937/commits/7169aa5bb2e1064829d52b35cc4bccfefac3b4aa) in the sense that people take in a `std::string_view` only to then create a `std::string` from it, which does exactly what we'd like to avoid when using `std::string_view` in the first place: copying over the data.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -347,13 +359,67 @@ class OverrideINIReader: public INIReader {
...
+    bool apply(Access_Operation oper, const std::string_view path) const {
+        auto is_subdirectory = [](const std::string_view& dir, const std::string_view& subdir) {
+            if (subdir.size() < dir.size())
+                return false;
+
+            if (subdir.compare(0, dir.size(), dir, 0, dir.size()) != 0)
+                return false;
+
+            return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir == "/";
+        };
```
amadio: This is the function I mentioned, we can take it from `XrdOucPrivateUtils.hh` now (maybe keeping the code for this in the .cc file, as I don't see the advantage of moving it to the headers.

amadio (commented):
This is a bit too big set of non-trivial changes to security sensitive code to be accepted as is. It's quite likely that new bugs could creep in with undesirable consequences. As such, I think that it's better to do this work in smaller chunks. The code here is very hard to review because there are a lot of changes moving code blocks around that get mixed with actual changes in the current code for the new functionality.

The changes migrating things from `std::string&` to `std::string_view` are somewhat orthogonal and just add noise to the patches here, so I think that should be done in a separate commit at the end, or just dropped, since there's not much benefit from using `std::string_view` here in any case.

As for `authz` vs `access_token` and how to treat them in XRootD, I like @abh's suggestion of mapping `access_token` to `authz` to avoid extra complications in the code, but I understand the need for separate cache vs origin tokens for Pelican. The right solution is probably to use the existing `Authorization:` and `Proxy-Authorization:` headers for this, and to follow [`RFC9110`](https://datatracker.ietf.org/doc/html/rfc9110) and [`RFC9111`](https://datatracker.ietf.org/doc/html/rfc9111) as much as possible, so that we improve interoperability with other software.

I think this is a good opportunity stop and think things through so that we can improve XCache not only for the Pelican use cases, but also to allow XCache to be deployed for multiple VOs and users with proper authentication (possibly not just tokens as well).

bbockelm:
> The right solution is probably to use the existing [Authorization:](https://datatracker.ietf.org/doc/html/rfc9110#section-11.6.2) and [Proxy-Authorization:](https://datatracker.ietf.org/doc/html/rfc9110#section-11.7.2) headers for this

We do use `Authorization`; that's where the `authz` comes from typically.

`Proxy-Authorization` is probably the wrong thing here.  The header is used to authorize the client to a HTTP proxy server.  In our case, we are wanting to authorize the back half - from the cache to the origin, not the client to the cache.

amadio:
I'm closing this to reduce clutter. This has been superceded by #2439 as far as I can tell. If needed, feel free to reopen after resolving existing merge conflics.

PR #2403, opened 2025-01-10
[XrdXrootdProtocol] do_WriteSpan() - Add written bytes in file statistics

### LINKED ISSUE #2368: f-stream does not record write bytes for XrdHttp
(Filing this one based on reports from @patrickbrophy)

When files are uploaded via the HTTP protocol, it appears the f-stream monitoring's record at close shows 0 bytes written.

@patrickbrophy reviewed the generated UDP structures with a hex dump utility to determine that, when the `xrootd` protocol is used, there's a non-zero number of bytes indicated as uploaded in the f-stream.  However, when a HTTP PUT is used via `curl`, there's alway zero bytes recorded as written.

ccaffy:
Fixes #2368

### DIFF
--- a/src/XrdXrootd/XrdXrootdXeq.cc
+++ b/src/XrdXrootd/XrdXrootdXeq.cc
@@ -3379,6 +3379,7 @@ int XrdXrootdProtocol::do_WriteSpan()
    if (Monitor.InOut())
       Monitor.Agent->Add_wr(IO.File->Stats.FileID, Request.write.dlen,
                                                   Request.write.offset);
+   IO.File->Stats.wrOps(IO.IOLen); // Optimistically correct
 
 // Trace this entry
 //


ccaffy:
Hi @abh3,

You can see in the issue #2368 how I managed to find the problem causing the f-stream monitoring not to report any written bytes.

However, I saw this code was written by you, so any insights about my fix' correctness is welcome :)

Thanks in advance,

Cheers,
Cedric

abh3:
OK, will check this out. Thanks for looking into it.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -3379,6 +3379,7 @@ int XrdXrootdProtocol::do_WriteSpan()
    if (Monitor.InOut())
       Monitor.Agent->Add_wr(IO.File->Stats.FileID, Request.write.dlen,
                                                   Request.write.offset);
+  IO.File->Stats.wrOps(IO.IOLen); // Optimistically correct
```
abh3, line 3382: This looks correct but the alignment appears one character off with the preceding "if".

abh3 (approved):
Yep, an oversight. Just one comment. Thanks for fixing this.

PR #2405, opened 2025-01-14
Correct forward declaration of XrdSfsFSctl

bbockelm:
Withouth this correction, `-Wall -Werror` builds on Mac fail when importing the headers with the following message:

```
/xrootd/build/release_dir/include/xrootd/XrdSfs/XrdSfsInterface.hh:161:1: error: 'XrdSfsFSctl' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
struct XrdSfsFSctl //!< SFS_FSCTL_PLUGIN/PLUGIO/PLUGXC parms
^
/xrootd/build/release_dir/include/xrootd/XrdOfs/XrdOfsFSctl_PI.hh:44:1: note: did you mean struct here?
```

I think aligning the actual declaration and the forward declaration is a reasonable thing to do here.

### DIFF
--- a/src/XrdOfs/XrdOfsFSctl_PI.hh
+++ b/src/XrdOfs/XrdOfsFSctl_PI.hh
@@ -41,7 +41,7 @@ class XrdOucErrInfo;
 class XrdSecEntity;
 class XrdSfsFile;
 class XrdSfsFileSystem;
-class XrdSfsFSctl;
+struct XrdSfsFSctl;
 class XrdSysError;
   
 /******************************************************************************/


abh3:
I approve of this pull request but why did alpine and coverage fail as this is a trivial one? Aberation?

abh3 (approved):
Correct, thanks Brian.

amadio:
The coverage build moved automatically to Ubuntu 24.04 (it's using the `ubuntu-latest` runner), where there's a minor issue in the coverage generation of `XrdOucJson.hh`. I am moving it back to Ubuntu 22.04 while I sort that out. The Alpine and macOS builds fail because of https://github.com/xrootd/xrootd/commit/7731531aff0c1122d9df4f3052682a25188cd29e and https://github.com/xrootd/xrootd/commit/d54785d9a1ac5d9ac4342c4c280f733e2e01bf31 (causes "Invalid Address" errors). I suspect this is a problem in the client that needs further changes to get fixed.

amadio:
@abh3 See [here](https://my.cdash.org/tests/247165519) a more verbose log which shows a failure on macOS 14. I think we internally have something in the client changing `127.0.0.1` to `[::ffff:127.0.0.1]`, which may be the underlying reason for the failures in `devel` and on macOS (as the network setup for IPv6 is kinda broken there, but also this reminds me of https://github.com/xrootd/xrootd/issues/1779).

PR #2407, opened 2025-01-16
[XrdScheduler] Add control of NPROC limit reset. Fixes #2396

### LINKED ISSUE #2396: Xrootd client import reduces nproc limit
Dear xrootd developers,

It seems like importing xrootd client in python sets up soft nproc limit of 30k. Here is an example:
```
bash-5.1$ cat test_nproc.py 
#!/usr/bin/python3
from resource import getrlimit, RLIMIT_NPROC

if __name__ == '__main__':
    print("Before:", getrlimit(RLIMIT_NPROC))
    from pyxrootd import client
    print("After:", getrlimit(RLIMIT_NPROC))
bash-5.1$ python3 test_nproc.py 
Before: (234027, 234027)
After: (30000, 234027)
bash-5.1$ 
```

I believe, the limit is coming from [here](https://github.com/xrootd/xrootd/blob/master/src/Xrd/XrdScheduler.hh#L43), being set up [here](https://github.com/xrootd/xrootd/blob/master/src/Xrd/XrdScheduler.cc#L191).

This side effect of library importing is unexpected and sometimes annoying (e.g. LHCb pilots are initializing gfal context when running,  and therefore inadvertently setting up this limit, which causes some issues on WNs with many cores). Would it be possible to remove the limit for the xrootd client or at least make it configurable (e.g. via env variable)?

Best Regards,
Alex

smithdh:


smithdh:
To solve #2396 I've tried to do what appears to be the smaller change, move the NPROC rlimit change from an XrdScheduler constructor to a dedicated method, which is only called from the server code. A Scheduler object is still constructed as part of the XrdUtils.so library and in the case of client that will be unused. However rearranging the set of global objects in the 'XrdGlobal' namespace (used by some Xrd/ classes, e.g. XrdLink XrdPoll) seemed to be a larger change.

amadio:
@smithdh Could you please base your branch on master? That will avoid the CI failures on Alpine and macOS. Thanks!

smithdh:
Revised & rebased to master:

The intended behavior this patch should bring now:

(i) By default keep same behavior for servers (i.e. things which construct an XrdConfig object, cmsd & xrootd daemons). That means the nproc soft rlimit is set to 30000 if possible.

(ii) Other things that link or load libXrdUtils.so have a modified behavior: instread of setting the limit to 30000, the limit is raised to 30000 unless it is already less restrictive than that. (i.e. this applies to apps like xrdcp, frm_xfragent etc but could apply to anything linked with XrdCl, including python bindings or c++).

(iii) Introduced XRDLEAVENPROC: If that env variable is set then NPROC is not changed (in either of the cases).

amadio:
@smithdh This is ready for merging, right?

smithdh:
Yes, this is ready to merge. Thanks.

Issue #2212, opened 2024-03-14
XRootD drops poller events on Mac OS X


bbockelm:
We enabled a new unit test in Pelican and have started to see consistent failures in XRootD.  The unit test uploads files in small bursts (i.e., uploads 2 files, pauses, uploads 2 files, etc).

Consistently, on Mac OS X only, the second round of file uploads fails.  This is because the poller never calls back to HTTP so ofs.Write isn't called -- meaning the file is *forever locked* open.  You need to restart xrootd to manipulate the file.

When this occurs, there are always issues reported by the poller.  Below is the full trace log of all activity until the poller fails.  The individual actions printed to the log all look correct.

The issue always occurs after one file descriptor has been removed from poller 0 then re-added.  If I patch the code so poll entries are never cleared (and hence never reused), the issue goes away.

However, I've been completely stumped on what's causing the issue.  I can confirm that the file descriptor value in the table is `-1` when the problem occurs, even though the file descriptor is set to earlier when it is attached.  If I make the detach event write `-2` into the table, the incorrect value `-2` is later in the table -- making me suspect there's a threading issue here, not a memory corruption.  If I instrument the poll table in the debugger, the problem goes away and there's no "unexpected" reads/writes to the memory location, just those you'd expect from reading the source code of `XrdPollPoll.icc`.  Again, adding some latency due to the debugger makes me suspect that we're looking at a race condition.

I'd note throughout the file, there are read/writes to the poll table from multiple threads without any consistency of whether the `PollMutex` is held.

It is full of undefined behavior but correcting this seems non-trivial; for example, the `PollTab` object is passed to the `poll` syscall without any lock on it while `XrdPollPoll::Include` (which takes the `PollMutex` lock and writes to `PollTab`) is invoked in parallel.

```
F4HP7QL65F:pelican bbockelm$ egrep 'XrdLink|Xrd.Poll|XrdPoll' /tmp/baz.4
time="2024-03-14T07:57:14-05:00" level=info msg="240314 07:57:14 259 Xrd_Poll: Starting poller 0" daemon=xrootd
time="2024-03-14T07:57:14-05:00" level=info msg="240314 07:57:14 259 Xrd_Poll: Starting poller 1" daemon=xrootd
time="2024-03-14T07:57:14-05:00" level=info msg="240314 07:57:14 259 Xrd_Poll: Starting poller 2" daemon=xrootd
[...]

PR #2408, opened 2025-01-16
MacOS stress test

bbockelm:
This implements a simple HTTP-based stress test that, on my laptop, fails 100% of the time.  All it does is try an increasing number of concurrent HTTP uploads.

By "stress tests", it tends to fail at 10 concurrent transfers _provided_ this line is set:

```
    if (idx != 2) curl_easy_setopt(handle.get(), CURLOPT_FORBID_REUSE, 1L);
```

If, instead, that line is:

```
    curl_easy_setopt(handle.get(), CURLOPT_FORBID_REUSE, 1L);
```

then I've never gotten the stress test to fail, even with more than 100 concurrent transfers.  Note -- `CURLOPT_FORBID_REUSE`, when set, causes a new TCP socket to be created for each transfer.  By turning it off, there's multiple requests on a single TCP socket.

I believe this is due to the fact that HTTP is such a simple protocol the poller only fires once per socket _unless_ there is socket reuse -- at which point we trigger the race condition mentioned in https://github.com/xrootd/xrootd/issues/2212.

Hopefully the GitHub runner will trigger the failure just as reliably as my laptop!

### DIFF
--- a/src/Xrd/XrdPollPoll.icc
+++ b/src/Xrd/XrdPollPoll.icc
@@ -34,6 +34,8 @@
 #include "Xrd/XrdPollPoll.hh"
 #include "Xrd/XrdScheduler.hh"
 
+#include <vector>
+
 /******************************************************************************/
 /*                             n e w P o l l e r                              */
 /******************************************************************************/
@@ -274,7 +276,15 @@ void XrdPollPoll::Start(XrdSysSemaphore *syncsem, int &retcode)
 
 // Now do the main poll loop
 //
-   do {do {numpolled = poll(PollTab, PollTNum, -1);}
+   std::vector<struct pollfd> PollTabCopy;
+   do {// Duplicate the polling table so we don't need to hold the PollMutex
+       // while we are sleeping in the poll()
+       PollMutex.Lock();
+       PollTabCopy.resize(PollTNum);
+       memcpy(PollTabCopy.data(), PollTab, sizeof(struct pollfd) * PollTNum);
+       PollMutex.UnLock();
+
+       do {numpolled = poll(PollTabCopy.data(), PollTabCopy.size(), -1);}
           while(numpolled < 0 && (errno == EAGAIN || errno == EINTR));
 
        // Check if we had a polling error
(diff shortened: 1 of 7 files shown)


amadio:
This is quite well timed, thanks! We just received a new mac mini that will be used for EOS/XRootD development and CI. I will try this there as well.

amadio:
Yes, I do intend to test on the new Mac mini, but the failures in CI are on Alpine and Ubuntu. Submitted changes should pass the CI before we do further testing on them.

src/Xrd/XrdPollPoll.icc
```diff
@@ -274,7 +276,15 @@ void XrdPollPoll::Start(XrdSysSemaphore *syncsem, int &retcode)
 
 // Now do the main poll loop
 //
-   do {do {numpolled = poll(PollTab, PollTNum, -1);}
+   std::vector<struct pollfd> PollTabCopy;
+   do {// Duplicate the polling table so we don't need to hold the PollMutex
+       // while we are sleeping in the poll()
+       PollMutex.Lock();
+       PollTabCopy.resize(PollTNum);
+       memcpy(PollTabCopy.data(), PollTab, sizeof(struct pollfd) * PollTNum);
```
amadio, line 284: This is going to be really slow. Let's add a commit to fix the poller on macOS, but I suggest having two `PollTab` structs allocated, then swapping them at each call to `poll(PollTab, PollTNum, -1)` instead of copying over each time and calling `PollTabCopy.resize(PollTNum);` inside a loop. Please keep only this change and the addition of the test in this pull request and split unrelated changes (modernize linking of libcurl) into a separate pull request.

bbockelm: I'm not following.

While the poll is running, the other code is making arbitrary changes to the data structure (that's part of the core of what makes the implementation thread-unsafe).  We can't swap here because the copy of the data structure fed to the `poll` function is missing an unknown number of changes.  The only option is to create a copy.

With respect to the resize -- this is typically separate from a reallocation and the "extra cost" is zeroing out any growth in the file descriptor count.

Avoiding the copy is going to require a much more significant surgery to the poller.

bbockelm: Ah, scratch that -- I think I see what you mean a bit lower in the code.  Maybe a swap will work after all; I'll tinker.

amadio:
Alpine uses MUSL instead of glibc. The failure is likely related to that.

bbockelm:
Ah -- actually, I think it was me misreading the test logs from the prior run.  On closer inspection, there's something basic missing on my added new unit test which seems quite fixable.

bbockelm:
@amadio - can you re-run the Mac OS X job a handful of times?  I lack the permission to re-run tests in GitHub.  I want to make sure that the tests are reliably passing and this isn't a fluke where I got lucky.

The larger refactor -- rewriting the logic so only one thread writes to the table at once (which would obviate needing making a copy of the table) -- was unsuccessful, unfortunately.

bbockelm:
@amadio - ok, tests have passed several times in a row so I think I'm happy.

src/Xrd/XrdPollPoll.icc
amadio: Do you plan to move from copying to swapping between two `PollTab`s?

bbockelm: Unfortunately, my tinkering failed -- I wasn't able to find a way to prevent multiple threads from writing into the table while the poll is ongoing.  That means, for now, the copy is required.

I don't think a swap will work here: you need to synchronize the changes so you need to copy the data back and forth.  Luckily, the resize shouldn't mean a reallocate: the only "cost" should be the memcpy.

amadio: Ok, since this is only for macOS in any case, let's fix now and optimize later then.

bbockelm: Yes - that's roughly my thinking.  I'd love to spend more time here but, pragmatically, there are only so many hours over the weekend.

It certainly beats my prior patch we carried for CI purposes: we prevented file descriptors from being deleted from the table, meaning it slowly grew (but otherwise "behaved" correctly) throughout the course of testing.

PR #2412, opened 2025-01-21
Fix format errors found by the compiler

ellert:
Additional fixes related to the use time_t for timeouts change

The PR was split out from #2380. The commit in this PR is related to the change of type for timeouts to time_t that is only in the devel branch. It was split out to a separate PR in order to rebase the original PS to masterl


ellert:
The proposed changes in this PR have been applied independently by @smithdh in commit 653bf57.
Closing.

amadio:
@ellert Apologies, I didn't realize that this and the other were the same.

PR #2417, opened 2025-01-25
[XrdSciTokens] Automatically add WLCG-audiences upon request

### LINKED ISSUE #2416: [XrdScitokens] Automatically determine audience according to WLCG rules
The WLCG has fairly reasonable rules for the audience that will be used for a server; however, the scitokens.cfg configuration requires one to understand and manually implement the rules which is fairly admin-hostile.  Things are simple enough that we could either implement them completely programmatically or with simple inputs.

I propose adding two knobs to the scitokens configuration

```
[Global]
append_wlcg_audiences = [true|false]
wlcg_audience_hostnames = [hostname1] [hostname2] [...]
```

- If `append_wlcg_audiences` is set to true (default false; I'm reluctant to change existing security behavior) then we programmatically generate the WLCG audience values.  These include the hostname and the hostname + scheme (and optionally port if not the IANA-registered default port for the scheme).  So, for `hostname.example.com`, this would inject the following four entries (assuming the server has HTTPS, roots, and root all running on port 1094):
   - `https://wlcg.cern.ch/jwt/v1/any`
   - `https://hostname.example.com:1094`
   - `root://hostname.example.com`
   - `roots://hostname.example.com`
   - `hostname.example.com`
- If `wlcg_audience_hostnames` are set, then these are appended to the accepted hostnames in the audiences.  If `append_wlcg_audiences` is enabled, the hostname that XRootD has calculated is *always* used.

The goal is simple deployments would be able to use just this config:

```
[Global]
append_wlcg_audiences = true
```

where larger deployments (or EOS) would use:

```
[Global]
append_wlcg_audiences = true
wlcg_audience_hostnames = redirector.site.example.com
```

Compare this to what @rptaylor had to calculate for his site in https://github.com/xrootd/xrootd/discussions/2413; I think this approach is far more admin-friendly.

bbockelm:
WLCG-style audiences are now added if the following is set in the `[Global]` section:

```
append_wlcg_audiences = true
```

Further, one can customize the hostnames used for the audiences by adding the following in the same section:
```
wlcg_audience_hostnames = redirector.site.example.com
```

Fixes #2416

### DIFF
--- a/bindings/python/src/PyXRootDCopyProcess.cc
+++ b/bindings/python/src/PyXRootDCopyProcess.cc
@@ -101,15 +101,15 @@ namespace PyXRootD
 
     val = XrdCl::DefaultCPInitTimeout;
     env->GetInt( "CPInitTimeout", val );
-    uint16_t initTimeout = val;
+    time_t initTimeout = val;
 
     val = XrdCl::DefaultCPTPCTimeout;
     env->GetInt( "CPTPCTimeout", val );
-    uint16_t tpcTimeout = val;
+    time_t tpcTimeout = val;
 
     val = XrdCl::DefaultCPTimeout;
     env->GetInt( "CPTimeout", val );
-    uint16_t cpTimeout = val;
+    time_t cpTimeout = val;
 
     if ( !PyArg_ParseTupleAndKeywords( args, kwds, "ss|HbbbbssssbIHHHbHLLLbs:add_job",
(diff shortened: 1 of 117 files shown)


bbockelm:
Going to close this for now (but not going to delete the branch as I want to come back).  This feature is quite useful but I want to land the prerequisite refactorings first.

PR #2419, opened 2025-01-27
[XrdOfs] Export configured FSctl plugin to global configuration

### LINKED ISSUE #2404: [XrdPss] Expose FSctl interface via the PSS layer
I'm writing an `XrdHttpExtHandler` plugin that exposes an API for managing the cache contents (particularly, prestaging and eviction).  The prestaging functionality is straightforward enough so far -- there's the `XrdOss::Read` overload with no buffer provided that accomplishes this -- but I was surprised to see that the `FSctl` interface isn't accessible.

In particular, when developing https://github.com/PelicanPlatform/xrdhttp-pelican/pull/7, I found that the `XrdFSCtl_PC*` environment pointer (e.g., as in calls like `m_fsctl = (XrdOfsFSctl_PI*)xrdEnv->GetPtr("XrdFSCtl_PC*");`) is inaccessible to the `XrdHttpExtHandler` layer.  This is because the FSctl plugin is initialized by the `XrdXrootd` layer and added to an `XrdOucEnv` object allocated on the stack instead of the global `XrdOucEnv`.

I'm fine if this was a conscious decision (and I suspect it was, given how carefully the environment is managed) -- but then would like the API to be accessible via the bridge using the `FSctl` call.  It looks like this may be as simple as implementing an `FSctl` override for the `XrdPssSys` class (there's already use of the FSctl plugin object in the class as part of the "open").

[Note this is fairly distinct from https://github.com/xrootd/xrootd/issues/2402; #2402 has to do with querying open file handles for more information and ensuring this is equally accessible to both client querying the cache and the cache querying the origin (and possibly the cache acting as a pure proxy with the IO objects).  What I'm looking for here is exposing the filesystem-level call.]

bbockelm:
The FSctl plugin is necessary to invoke filesystem-level operations on the PFC.  However, unlike the `XrdOss` and `XrdAccAuthorize` plugins, it is only exported in the xrootd protocol environment, not the process-wide environment.

This provides symmetry between the FSctl and the other plugins. It joins other parts of the PFC available via the process-wide export.

I also investigated invoking the plugin indirectly through some other layer (such as the xrootd protocol bridge) but, unfortunately, the use case for needing to invoke the plugin is a `XrdHttpExtHandler`; the handler framework does not provide access to the protocol bridge.

Fixes #2404

### DIFF
--- a/src/XrdOfs/XrdOfsConfig.cc
+++ b/src/XrdOfs/XrdOfsConfig.cc
@@ -299,7 +299,9 @@ int XrdOfs::Configure(XrdSysError &Eroute, XrdOucEnv *EnvInfo) {
 // the cache-specific FSctl() operation. We check if a plugin was provided.
 //
    if (ossFeatures & XRDOSS_HASCACH)
-      FSctl_PC = (XrdOfsFSctl_PI*)EnvInfo->GetPtr("XrdFSCtl_PC*");
+      {FSctl_PC = (XrdOfsFSctl_PI*)EnvInfo->GetPtr("XrdFSCtl_PC*");
+       if (xrdEnv && FSctl_PC) xrdEnv->PutPtr("XrdFSCtl_PC*", FSctl_PC);
+      }
 
 // Configure third party copy phase 2, but only if we are not a manager.
 //


abh3 (changes requested):
I don't see why you cannot use the already existing XrdOfs plugin as that gives you access to the PFC plugin using the XC operation code. I would rather you go that way as it already is there and avoids having yet another interface exposed that need not be exposed.

bbockelm:
> I don't see why you cannot use the already existing XrdOfs plugin as that gives you access to the PFC plugin using the XC operation code

Tried that -- the `XrdOfs` plugin is not exported to the process configuration and not available to the `XrdHttpExtHandler`. 
 It's possible that statement is incorrect and I just wasn't able to find the right exported plugin -- if so, could you kindly link me to the name in the environment?

> avoids having yet another interface exposed that need not be exposed.

It looks like we'll have to chose between exposing `XrdOfs*` versus `XrdFSCtl_PC*` -- do you have a preference?  When investigating, I noticed `XrdOfs` is in the private headers and `XrdFSCtl_PC` is public.

abh3:
XrdOfs is already exported as "XrdSfsFileSystem*" which is a public header (i.e. XrdSfsInterface.hh). Take a look at XrdXrootdConfig.cc:311 which is: PutPtr("XrdSfsFileSystem*", osFS) so this pull request really is not needed.

bbockelm:
@abh3 - thanks for the hint!  Indeed, I was looking for `XrdOfs` and had overlooked the right exported interface.

I was able to confirm using the XrdSfsFileSystem does what I need -- will close this PR out.

abh3:
Good. Please keep the syntax consistent with what we currently have. So, we use the XC operation code when we encounter "cache" as the first word after "set" in the request. All tokens after that would be routed to cache fsctl plugin. In the current case the next word would be evict or fevict followed by a path. In your case, I would strongly suggest using pevict because that is what you are doing (i.e. priority evict). That would make changes in the plugin pretty trivial. Since you will be calling the fsctl directly you will need to manually handle a) path normalization, and b) the opaque handling. Failure to do that prior to calling fsctl will lead to undefined behaviour (not good). See XrdXrootdXeq.cc:2824 and the methods rpCheck() and Squash(). These methods should not be replicated as they are rather closely tied to various config options. So, that means we would have to figure out how to give you access to them. This is why we really wanted you to use the bridge as that would simplify all of this. Unfortunately, you are correct, ext handlers don't get access to the bridge and I suspect that is beecause ext handlers may be invoked when there is no bridge because it is the first interaction with the client.

bbockelm:
Having access to those functions in one of the utilities classes would be wonderful.

In the meantime, for this plugin, I'm sanitizing the paths using `std::filesystem` and dropping opaque.  This wouldn't be compatible with all configurations -- but enough to keep moving forward.

@amadio - you might also be interested in this project as it's the first time (to the best of my knowledge...) there's been an integration test for XCache: https://github.com/PelicanPlatform/xrdhttp-pelican/blob/main/tests/xrdhttp-setup.sh.

Issue #2420, opened 2025-01-30
[XrdPfc] Segfault under load


bbockelm:
When the caches are under load, we've been noticing a segfault about a 1-3 times a day.  Looking at the kernel message (sadly, we haven't been able to capture a core file), it's the [following line](https://github.com/xrootd/xrootd/blob/v5.7.2/src/XrdPfc/XrdPfc.cc#L580) in v5.7.2:

```
   int rc = f->inc_ref_cnt();
```

From the assembler code, it appears the function `Cache::inc_ref_cnt(File* f, bool lock, bool high_debug)` is called with a `nullptr` for the argument `f`.

Looking through all the places this function is referenced, there are two call sites without any protection to ensure `f` is non-null:

- `Cache::ConsiderCached`
- `Cache::Stat`

For example, from `Cache::Stat`:

```
      if (it != m_active.end()) {
         file = it->second;
         inc_ref_cnt(file, false, false);
      }
```

Elsewhere in the code, it notes:

```
         // Null File* in m_active map means an operation is ongoing, probably
         // Attach() with possible File::Open(). Ask for retry.
```

So, I believe that if `Stat` -- or `ConsiderCached` are invoked while a file-open is ongoing, then either of those will potentially cause a segfault.

This relies on two clients accessing a file, otherwise closed, at approximately the same time.  That roughly matches the case -- we're seeing a user running a large workflow with a typo in the filename.  That's causing many jobs to start simultaneously and asking for a file that results in a file-not-found (meaning it's not attached), running the race over and over -- and loosing a few times a day.

@amadio - in terms of severity, I believe anyone who can generate a 404 should be able to crash the XCache.

amadio:
Fixed by #2421.

PR #2421, opened 2025-01-30
[XrdPfc] Check for a null pointer dereference

bbockelm:
When looking up the map of active files, the returned value is a a `nullptr` if the file-open is ongoing.  Check for this case and only try to use the file object if it is non-null.

Without this check, the XrdPfc module will trigger a segfault when trying to increase the reference count for a null object.

The places affected by the nullptr already have appropriate fallback code in place if the filename isn't on the active list at all -- we now use that code instead of crashing.

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -1014,7 +1014,13 @@ int Cache::ConsiderCached(const char *curl)
       auto it = m_active.find(f_name);
       if (it != m_active.end()) {
          file = it->second;
-         inc_ref_cnt(file, false, false);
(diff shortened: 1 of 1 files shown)


bbockelm:
Fixes #2420

I was not able to come up with a reasonable unit test for this so we are going to test on a few production caches.

osschar:
Thank you @bbockelm ... this is the correct fix. The file=nullptr case is handled correctly with full-path stat below ... and is the code path that gets called when the file is not open by the cache.

osschar (approved):

bbockelm:
After applying this patch, a production cache went from 4 segfaults / day to 0.  So, I think I'm happy with it.

PR #2422, opened 2025-01-31
[Cl] Also recognise IPv4 mapped addresses

smithdh:
This is a draft, aiming to fix some test failures, usually on alpine and mac os.

### DIFF
--- a/src/XrdCl/XrdClURL.cc
+++ b/src/XrdCl/XrdClURL.cc
@@ -26,6 +26,7 @@
 #include "XrdCl/XrdClOptimizers.hh"
 
 #include <cstdlib>
+#include <cctype>
 #include <vector>
 #include <sstream>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


smithdh:
Will remove time_t commit from this pr, and move out this PR out of draft (for review etc) when time_t changes on devel.

smithdh:
I removed the time_t related change (and rebased), and made a small change to the patch. I'll move the patch out of draft; although I see that we've removed the compatability->mapped change from the devel branch, so inface the original reason for this patch has gone. Perhaps we want to abandon this patch.

abh3:
Not at all, this patch has value. We removed he offending patch because we needed to get the CI moving as it was holding up a lot of stuff. We need to review what the appropriate change is here. The removal of compatible IP addresses is not time driven but was to address long term trends.

abh3:
This is destined for 5.8

PR #2424, opened 2025-02-03
Fix XrdPosix headers when called in C context

schwicke:
The current header files of XrdPosix do not properly work when used in a C context. Compiling this reproducer:

```
#include <stdio.h>
#include <XrdPosix/XrdPosix.hh>
int main(){
}
```
With clang this fails because the code tries to include cstdio instead of stdio.h in C context. After fixing this, it works with clang but not with gcc. There are 2 more issues:

- with gcc 11 and newer the code fails to compile when dirent.h is included. This can be fixed by re-arranging the code properly
- Some of the types defined in XrdPosixExtern.hh conflict with the system defaults, e.g. long long should be __off64_t or int functions are actually size_t. Newer compiler versions complain about this. The patch adapts the function definitions accordingly.

### DIFF
--- a/src/XrdPosix/XrdPosix.hh
+++ b/src/XrdPosix/XrdPosix.hh
@@ -36,6 +36,11 @@
 // to accomplish this. So, redefinition is the most portable way of doing this.
 //
 
+// First define the external interfaces (not C++ but OS compatabile)
+#include <dirent.h>
+#include "XrdPosixExtern.hh"
+
+// Then redefine them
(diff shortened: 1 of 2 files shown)


amadio:
There are no ABI compatibility problems with these changes. Merging.

amadio:
For the record, this was needed in order to add support for XRootD into cernlib:
https://gitlab.cern.ch/DPHEP/cernlib/cernlib/-/merge_requests/321

PR #2425, opened 2025-02-04
Xrdceph otf checksums - calculate checksums such as Adler32 when receiving incoming data

ijjorama:
Initial implementation of on-the-fly/single-pass/streamed checksums. Currently, this uses zlib directly and stores the calculated checksum in object extended attribute XrdCks.adler32.

ToDo: 
Change to using XrdOuc XrdCksCalc derived classes, which provide CRC32C in addition to Adler32, etc.
Only store otf checksum as an object attribute if the configuration file specifies this. The default will be to allow client checksum requests to trigger checksum calculation by reading back data.
Log the otf checksum to an external file, separately to


abh3:
This pull request appears to be based on an outdated branch and needs to be rebased before we can  review it. Please do so.

abh3 (changes requested):
Please see comment on getting this pull request synced with the actual repo branch.

ijjorama:
Closing, assigned to wrong repo.

amadio:
No problem. In the end we would like indeed to merge everything into the main repository. It will make both our lives easier to have a single place where all the changes happen.

PR #2428, opened 2025-02-10
[XrdHttp] Add trailing slash to directories in listing

bbockelm:
Apache, when generating a HTML listing of a local directory, will suffix directory names with a trailing slash.  This PR tweaks the XRootD directory listing to do the same and adds a corresponding integration test.

Beyond matching common practice, the HTTP fsspec plugin in the Python ecosystem uses the trailing slash as a heuristic to indicate a directory.  With this change, Python users should be able to recursively list directories in XRootD.  While not strictly required, this makes a significant improvement for the lives of Python users.


ccaffy (approved):
Looks good to me, thanks !

PR #2429, opened 2025-02-11
[XrdTpcTPC] Added oss.task=httptpc opaque query to the file open URL

### LINKED ISSUE #2427: [XrdTpcTPC] Indicates to the oss layer which kind of HTTP TPC operation is performed (open)
In the EOS project, we are interested to know which kind of operation triggered the file `open()`.

I believe the only way to do this is to pass another opaque query parameter alongside with the already existing `oss.asize` telling which kind of HTTP TPC transfer it is.

Something like `oss.httptpc=pull|push`.

I'm opened for suggestion, I'm really bad at naming...

Also, do we want this to be part of a configuration of the XRootD server so that the default behavior is not changed unless a specific config line is set ?

ccaffy:
Fixes #2427

### DIFF
--- a/src/XrdTpc/XrdTpcUtils.cc
+++ b/src/XrdTpc/XrdTpcUtils.cc
@@ -20,13 +20,16 @@
 //------------------------------------------------------------------------------
 #include "XrdTpcUtils.hh"
 #include "XrdOuc/XrdOucTUtils.hh"
+#include "XrdTpc/XrdTpcTPC.hh"
 
 #include <sstream>
 
-std::string XrdTpcUtils::prepareOpenURL(const std::string & reqResource, std::map<std::string,std::string> & reqHeaders, const std::map<std::string,std::string> & hdr2cgimap, bool & hasSetOpaque) {
+std::string XrdTpcUtils::prepareOpenURL(const std::string & reqResource, std::map<std::string,std::string> & reqHeaders, const std::map<std::string,std::string> & hdr2cgimap) {
   auto iter = XrdOucTUtils::caseInsensitiveFind(reqHeaders,"xrd-http-query");
-  bool found_first_header = false;
   std::stringstream opaque;
+
+  // https://github.com/xrootd/xrootd/issues/2427 we tell the oss layer that this transfer is a HTTP TPC one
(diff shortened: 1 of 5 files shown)


src/XrdTpc/XrdTpcUtils.cc
```diff
@@ -20,14 +20,17 @@
...
 #include <sstream>
 
-std::string XrdTpcUtils::prepareOpenURL(const std::string & reqResource, std::map<std::string,std::string> & reqHeaders, const std::map<std::string,std::string> & hdr2cgimap, bool & hasSetOpaque) {
+std::string XrdTpcUtils::prepareOpenURL(const std::string & reqResource, std::map<std::string,std::string> & reqHeaders, const std::map<std::string,std::string> & hdr2cgimap) {
   auto iter = XrdOucTUtils::caseInsensitiveFind(reqHeaders,"xrd-http-query");
-  bool found_first_header = false;
   std::stringstream opaque;
 
+  // https://github.com/xrootd/xrootd/issues/2427 we tell the oss layer that this transfer is a HTTP TPC one
+  opaque << "?" << TPC::TPCHandler::OSS_TASK_OPAQUE;
```
amadio, line 32: Why not do `opaque << "?oss.task=httptpc";` here and get rid of `TPC::TPCHandler::OSS_TASK_OPAQUE`?

ccaffy: HAHA I initially had this duplication everywhere and hesitated to push without putting this variable. I wanted to remove the duplication. As you prefer if you prefer I can duplicate without issue!

amadio: I don't mind it as is, but duplicating makes the code more readable. If you don't want to revert the work for deduplicating, we can merge as is. Just waiting on ack from @abh3.

abh3: Hmm, let's see a) use a declared static, b) use define symbol with a shorter name, or c) use a string constant. What's wrong with each? a) not particularly clear and rather clunky, b) people hate define symbols because of namespace issues, c) easy to make a mistake because of duplication. So, it's a wash; any approach is just as good or bad, depending on how you look at it  So, I'm neutral. Frankly, I'd just keep it as it is.

PR #2430, opened 2025-02-12
[XrdHttp] Quote the resource when handling a redirect

esindril:
The resource string is unquoted when the request is handled but in case of a redirect the resource should be quoted again otherwise the client will be redirected with a different path than the initial one.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -45,6 +45,7 @@ jobs:
           bash \
           cmake \
           ceph-dev \
+          curl \
           curl-dev \
           fuse-dev \
           fuse3-dev \
--- a/docker/Dockerfile.alpine
+++ b/docker/Dockerfile.alpine
@@ -4,6 +4,7 @@ RUN apk add \
 	bash \
 	cmake \
 	ceph-dev \
+	curl \
 	curl-dev \
 	fuse-dev \
 	fuse3-dev \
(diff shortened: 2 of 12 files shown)


amadio:
Any chance we could have a test added for this?

amadio:
Look in `tests/XRootD/http.sh`. It might be just a matter of adding a few commands there.

esindril:
Just a few extra remarks related to this pull request:
* this fixes my particular problem with handling `#` in the file names, but in general the `quote` and `unquote` methods are not consistent/symmetric.
* I have modified the `quote` method to not quote the path separator `/` as this would screw up all the paths when redirecting. It's not clear to me what was the initial use-case for quoting `/`.
* There should be an alignment between what XRootD accepts in terms of valid characters for paths https://xrootd.web.cern.ch/doc/dev6/XRdv520.htm#_Toc173272867 and what (HTTP) URI considers reserved characters https://datatracker.ietf.org/doc/html/rfc3986#page-12
* When configuring the xroot and http support on different ports the redirection when coming with a HTTP request will alwatys wrongly redirect to the XRD port (maybe there is some hidden option to fix this?!). Therefore, in order to have HTTP support at the cluster level the XROOTD and HTTP protocols need to be served on the same port.

amadio:
Something that was added doesn't work as expected: https://my.cdash.org/tests/253509135

esindril:
Yes, curl was missing from the Alpine image.

PR #2439, opened 2025-02-24
[XrdSciTokens] Accept `access_token` CGI parameter

bbockelm:
On reading the definition of the `Bearer` Authentication scheme in RFC 6750, I stumbled upon Section 2.3 ("URI Query Parameter") which specifies the query parameter name that should be used when embedding the token in the URL; it should be `access_token`.

This commit continues to accept `authz` (too much historical baggage to break) but adds in `access_token` alongside it.

This commit includes unit test coverage for the new parameter name, including tests for `XrdOucEnv::TidyEnv` and fixes bugs for `TidyEnv` in the case the tidying resulted in the empty string.

This is the first step of splitting up #2397 into smaller, easier-to-review chunks.  Note the change to `src/XrdSciTokens/XrdSciTokensAccess.cc` is primarily whitespace shifts; it's easier to see the semantic differences from `git diff -w`.

### DIFF
--- a/src/XrdMacaroons/XrdMacaroonsAuthz.cc
+++ b/src/XrdMacaroons/XrdMacaroonsAuthz.cc
@@ -173,7 +173,8 @@ Authz::Access(const XrdSecEntity *Entity, const char *path,
     }
 
     const char *authz = env ? env->Get("authz") : nullptr;
-    if (authz && !strncmp(authz, "Bearer%20", 9))
+    if ((authz && !strncmp(authz, "Bearer%20", 9)) ||
+        (!authz && (authz = env ? env->Get("access_token") : nullptr) && !strncmp(authz, "Bearer%20", 9)))
     {
         authz += 9;
     }
--- a/src/XrdHttpTpc/XrdHttpTpcUtils.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcUtils.cc
@@ -43,6 +43,9 @@ std::string XrdHttpTpcUtils::prepareOpenURL(const std::string & reqResource, std
           reqHeaders["Authorization"] = token.substr(6);
           has_authz_header = true;
         }
+      } else if (!strncmp(token.c_str(), "access_token=", 13) && !has_authz_header) {
+        reqHeaders["Authorization"] = token.substr(13);
+        has_authz_header = true;
       } else {
         opaque << "&" << token;
       }
--- a/src/XrdOuc/XrdOucEnv.cc
+++ b/src/XrdOuc/XrdOucEnv.cc
@@ -102,13 +102,13 @@ char *XrdOucEnv::Delimit(char *value)
 
 void XrdOucEnv::EnvBuildTidy()
 {
-   char *tidyEnv, *authInfo;
-   int aBeg, aEnd;
+   char *tidyEnv, *authInfo, *accessAuthInfo = nullptr, *working=global_env;
+   int aBeg, aEnd, aCount = 6;
 
 // We need to sanitize the current env string by removing auth info. If there
-// is no auth informationn, then we can short cicuit this.
+// is no auth information, then we can short circuit this.
 //
-   if ((authInfo = strstr(global_env, "authz=")) == 0)
+   if ((authInfo = strstr(global_env, "authz=")) == 0 && (accessAuthInfo = strstr(global_env, "access_token=")) == 0)
       {Put(TIDY_ENVVAR, "");
        return;
       }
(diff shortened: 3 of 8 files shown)


src/XrdMacaroons/XrdMacaroonsAuthz.cc
```diff
@@ -173,6 +173,10 @@ Authz::Access(const XrdSecEntity *Entity, const char *path,
 
     const char *authz = env ? env->Get("authz") : nullptr;
     if (authz && !strncmp(authz, "Bearer%20", 9))
+    {
+        authz += 9;
```
amadio, line 177: If and else have the same action. Something doesn't look right.

amadio (commented):
The whitespace changes together with all the rest don't make reviewing that easy. I need to finish CHEP papers, but after that I will come back and review more carefully.

src/XrdMacaroons/XrdMacaroonsAuthz.cc
bbockelm: The desired behavior is correct, the formatting is clumsy.  I'll see if I can't come up with a way to make it readable without having a 150 character conditional.

bbockelm: I'm not sure the formatting is all that much better -- but came up with a revision that's hopefully less confusing.

bbockelm:
@amadio - can I give a kind nudge on this PR?  The functionality of #2397 and #2417 will need to be rebased on top of this and I want to decrease the out-of-tree patches we are carrying.

amadio:
Hi @bbockelm, I do want to have this change. We should try to follow RFCs as much as possible to improve interoperability with other software. However, I would really like to release 5.8.0 quite soon, so I'd prefer to keep this for after 5.8.0 is out, since the patch is not that simple. I think it's better to give some time for testing this within EOS before adding to a release.

abh3:
So, it would appear that access_token provides essentially the same type of authorization as authz. So, effectively they are synonyms of each other. It would seem far simpler to simply replace the access-token with authz and reuse all the code that is already there. You may want to remove duplicates if someone specified both but it really should not matter as the client could have had multiple authz tokens anyway. So, am I missing something or can this be reduced to a single file updated of about 10 lines or so?

abh3 (commented):
See my comment on how this update may be significantly simplified.

bbockelm:
> See my comment on how this update may be significantly simplified.

Indeed, it would significantly simplify things - but also make it miss one of the whole points of doing the change.

Fundamentally, I need a way two pass two distinct credentials up the chain.  By adding support for `access_token` and keeping `authz` present (but separate), I can do all the authorization pieces I need for Pelican with very modest subsequent changes (esp. by doing configuration tricks like loading the SciTokens module twice, once for each credential to analyze).  I think I have a roadmap to having all the other pieces quite generic.

Once this is established, then I can keep moving forward with landing other SciTokens improvements like the refactor to enable unit tests and WLCG-specific functionality improvements.  Otherwise, we're really stuck and I should probably begin forking the module to reduce the cost of managing the overlapping patch sets.

abh3:
OK, but also don't underestimate the power of sending additional authorization outside of authz via the Sec Entity object as this avoids all the gobbledygook code to obfuscate tokens and provide a direct passthrough to the authorization module.

bbockelm:
@amadio - are we OK to merge this one?

This is the core part of the refactor of #2397 into smaller distinct steps as requested.  If we can get this going then I can finish up the 3-4 subsequent PRs for the rest of the functionality.

amadio:
In any case, if we do decide to go forward with what is here in its current form, I'd like to at least have a clear set of rules for how `authz` and `access_token` should be treated in case both are present. However, if nothing breaks current behavior, I don't have any objections.

@abh3 What are your thoughts?

bbockelm:
@amadio - it's not that I don't intend to follow RFC9110/9111 but I was pointing out that they're not applicable here.  The XRootD Proxy File Cache, despite it's name, does not implement any of the HTTP protocol proxying semantics (e.g., `CONNECT` verb and friends).  It's more of a conceptual proxy/cache in the way that CloudFlare provides proxy functionality in its CDN product but isn't actually a HTTP protocol proxy.  It would be horrendously confusing to have a non-proxy start using the proxy headers.

In terms of documentation -- happy to do so!  Right now, `XrdSciTokens/README.md` would be the natural fit to state down the ordering semantics.  Any objections to that?  Even ignoring picking up `access_token` CGI in this PR, there's already undocumented behavior around how ZTN and `authz` CGI interact which would be quite nice to get on paper and out of the code.

amadio:
@abh3, @bbockelm I'd like to take a decision regarding the changes proposed here. As stated in section 2.3 of [RFC6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.3):

> Because of the security weaknesses associated with the URI method, *including the high likelihood that the URL containing the access token will be logged*, it SHOULD NOT be used unless it is impossible to transport the access token in the `Authorization` request header field or the HTTP request entity-body. Resource servers MAY support this method.

In addition to the above, having multiple ways of providing tokens in the query parameters will make URL token redaction more costly, and more confusing for users, as `authz` has been used for a long time now with XRootD.

I would suggest that this should be handled via special headers in HTTP, and moved to cgi with the new `http.header2cgi ... strip-on-redirect` to avoid exposing tokens beyond the server they are meant to. If this is not sufficient, then let's schedule a meeting to discuss possible ways forward.

amadio:
Closing this for now. If this is still needed, a workable solution should be discussed in one our development meetings.

PR #2441, opened 2025-02-25
[Tests] Add third-party copy test script 

dynamic-entropy:


bbockelm:
Notes from Zoom chat with @dynamic-entropy:
- Should cover HTTPS as this is the most common way things are used.  Would have slight preference doing this via `curl` command line so we can more tightly control the headers utilized.
- For realism, we should utilize the actual authentication methods instead of leaving auth wide open.  This means leveraging the [SciTokens](https://github.com/xrootd/xrootd/blob/master/tests/scitokens/CMakeLists.txt#L12) fixture to make tokens and host certificates available.  Luckily, since that fixture already exists, pulling this configuration in should require minimal additional work.  @amadio -- if we are pulling in authentication, is it better to improve the cluster test or create a new fixture for the configuration?
- Also for realism, we should have the test leverage TLS (e.g., HTTPS) instead of an unauthenticated protocol.
- Currently, there's no "negative test".  We should cover the case where we try to transfer a source file that is not found and a file that exists but where there's no permission to read.

bbockelm:
One more addendum:
- Please run the script through the `shellcheck` command to help identify common bash pitfalls.

dynamic-entropy:
I need to either do some CMake magic or create two nodes with the same cluster or an entirely different setup which only runs these tests on platforms that have SciTokens available. 

We cannot do HTTP TPC without SSL enabled, and I am relying on the SciTokens fixture to create the necessary certs and CAs.

amadio:
Please rebase on `master` again. I just pushed a small fix to get rid of PollerTest.FunctionTest failures. Cheers,

amadio:
@dynamic-entropy Please check that at each commit point, things still work. On the first commit, I get this:
```
-- Found Python: /usr/bin/python3 (found version "3.12.9") found components: Interpreter Development.Module
CMake Error: File /home/amadio/src/xrootd/tests/TPCTests/srv1.cfg does not exist.
CMake Error at tests/TPCTests/CMakeLists.txt:12 (configure_file):
  configure_file Problem configuring file


CMake Error: File /home/amadio/src/xrootd/tests/TPCTests/srv2.cfg does not exist.
CMake Error at tests/TPCTests/CMakeLists.txt:12 (configure_file):
  configure_file Problem configuring file


CMake Error: File /home/amadio/src/xrootd/tests/TPCTests/common.cfg does not exist.
CMake Error at tests/TPCTests/CMakeLists.txt:12 (configure_file):
  configure_file Problem configuring file
...
-- Configuring incomplete, errors occurred!
Command exited with the value: 1
Error(s) when configuring the project
CMake Error at /home/amadio/src/xrootd/test.cmake:213 (message):
  Configuration failed
```
This means we cannot git-bisect through your changes. You can simply merge some of the commits to fix this, I think (or we can merge with squash, although that's suboptimal).

dynamic-entropy:
Hello @amadio 
Thanks for taking a look. I have edited some and squashed relevant commits. 
Cheers

amadio:
Ok, thanks. I will try again.

amadio:
It seems that, when run alone, the test may pass, but when I run it along with the other tests, it always fails as shown below:
```
 58/131 Test #107: XRootD::tpc::test .........................................................***Failed    0.01 sec
++ uname
+ [[ Linux == \D\a\r\w\i\n ]]
+ : /home/amadio/src/xrootd/build/src/xrdadler32
+ : /home/amadio/src/xrootd/build/src/xrdcrc32c
+ : /home/amadio/src/xrootd/build/src/XrdCl/xrdcp
+ : /home/amadio/src/xrootd/build/src/XrdCl/xrdfs
++ command -v openssl
+ : /usr/bin/openssl
++ command -v curl
+ : /usr/bin/curl
+ check_commands /home/amadio/src/xrootd/build/src/xrdadler32 /home/amadio/src/xrootd/build/src/xrdcrc32c /home/amadio/src/xrootd/build/src/XrdCl/xrdcp /home/amadio/src/xrootd/build/src/XrdCl/xrdfs /usr/bin/openssl /usr/bin/curl
+ missing=()
+ local missing
+ for PROG in "$@"
+ [[ ! -x /home/amadio/src/xrootd/build/src/xrdadler32 ]]
+ for PROG in "$@"
+ [[ ! -x /home/amadio/src/xrootd/build/src/xrdcrc32c ]]
+ for PROG in "$@"
+ [[ ! -x /home/amadio/src/xrootd/build/src/XrdCl/xrdcp ]]
+ for PROG in "$@"
+ [[ ! -x /home/amadio/src/xrootd/build/src/XrdCl/xrdfs ]]
+ for PROG in "$@"
+ [[ ! -x /usr/bin/openssl ]]
+ for PROG in "$@"
+ [[ ! -x /usr/bin/curl ]]
+ [[ 0 -gt 0 ]]
+ hosts=(['srv1']='root://localhost:10951' ['srv2']='root://localhost:10952')
+ declare -A hosts
+ hosts_http=(['srv1']='https://localhost:10951' ['srv2']='https://localhost:10952')
+ declare -A hosts_http
+ trap cleanup ERR
+ RMTDATADIR=/srvdata/tpc
+ LCLDATADIR=/home/amadio/src/xrootd/build/tests/TPCTests/localdata/tpc
+ mkdir -p /home/amadio/src/xrootd/build/tests/TPCTests/localdata/tpc
+ mkdir -p /home/amadio/src/xrootd/build/tests/TPCTests/generated_tokens
+ mkdir -p /home/amadio/src/xrootd/build/tests/TPCTests/.cache/scitokens
+ rm -rf '/home/amadio/src/xrootd/build/tests/TPCTests/.cache/scitokens/*'
+ setup_scitokens
+ /home/amadio/src/xrootd/build/tests/scitokens/xrdscitokens-create-token /home/amadio/src/xrootd/build/tests/issuer/issuer_pub_1.pem /home/amadio/src/xrootd/build/tests/issuer/issuer_key_1.pem test_1 https://localhost:7095/issuer/one 'storage.modify:/ storage.create:/ storage.read:/'
Failed to open /home/amadio/src/xrootd/build/tests/issuer/issuer_pub_1.pem: No such file or directory
+ echo 'Failed to create token'
Failed to create token
+ exit 1
```

tests/TPCTests/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+  add_test(NAME XRootD::tpc::start
+    COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/setup.sh start")
+  set_tests_properties(XRootD::tpc::start
+    PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_SETUP XRootD_TPC
+               FIXTURES_REQUIRED SciTokens)
+
+  add_test(NAME XRootD::tpc::test
+    COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/test.sh")
+  set_tests_properties(XRootD::tpc::test
+    PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_REQUIRED "XRootD_TPC;SciTokens")
```
amadio: Looks like this is not being respected by CMake:
```
 51/131 Test #109: SciTokens::setup ..........................................................   Passed    1.16 sec
        Start 126: XRootD::scitokens::setup
 52/131 Test #126: XRootD::scitokens::setup ..................................................   Passed    0.07 sec
        Start 106: XRootD::tpc::start
        Start 128: XRootD::scitokens::test
 53/131 Test #128: XRootD::scitokens::test ...................................................   Passed    0.53 sec
        Start 110: SciTokens::teardown
 54/131 Test #110: SciTokens::teardown .......................................................   Passed    0.01 sec
 55/131 Test #125: XRootD::http::test ........................................................   Passed    2.08 sec
        Start 124: XRootD::http::teardown
 56/131 Test #124: XRootD::http::teardown ....................................................   Passed    0.07 sec
 57/131 Test #106: XRootD::tpc::start ........................................................   Passed    2.06 sec
        Start 107: XRootD::tpc::test
 58/131 Test #107: XRootD::tpc::test .........................................................***Failed    0.01 sec
```

amadio: I think I figured it out, you need `FIXTURES_REQUIRED "XRootD_TPC;XRootD::scitokens;SciTokens"` here to make CMake not destroy things before you are done with them.

dynamic-entropy: Hmm, interesting that this is happening. 
At this commit, there is no involvement of the `XRootD::scitokens`. 

Here, the `SciTokens` fixture is enough to generate the required PKI for TLS.

That is added in later commits when we need the tokens.

tests/TPCTests/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+  add_test(NAME XRootD::tpc::start
+    COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/setup.sh start")
+  set_tests_properties(XRootD::tpc::start
+    PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_SETUP XRootD_TPC
+               FIXTURES_REQUIRED XRootD::scitokens)
+
+  add_test(NAME XRootD::tpc::test
+    COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/test.sh")
+  set_tests_properties(XRootD::tpc::test
+    PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_REQUIRED "XRootD_TPC;XRootD::scitokens")
```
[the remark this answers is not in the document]
dynamic-entropy: Again, this does not seem to happen in my tests ever.

dynamic-entropy:
> It seems that, when run alone, the test may pass, but when I run it along with the other tests, it always fails as shown below:

It seems that the tests take more than a minute to run everything that requires a token. 
Should I add a commit to increase the lifetime of the token?

tests/TPCTests/CMakeLists.txt
dynamic-entropy: When I checkout to the first commit, CMake does respect the order.

```
        Start 109: SciTokens::setup
107/131 Test #109: SciTokens::setup ..........................................................   Passed    0.78 sec
        Start 106: XRootD::tpc::start
108/131 Test #106: XRootD::tpc::start ........................................................   Passed    2.02 sec
        Start 107: XRootD::tpc::test
109/131 Test #107: XRootD::tpc::test .........................................................   Passed    1.99 sec
        Start 108: XRootD::tpc::teardown
110/131 Test #108: XRootD::tpc::teardown .....................................................   Passed    0.02 sec
        Start 126: XRootD::scitokens::setup
111/131 Test #126: XRootD::scitokens::setup ..................................................   Passed    0.02 sec
        Start 128: XRootD::scitokens::test
112/131 Test #128: XRootD::scitokens::test ...................................................   Passed    0.16 sec
        Start 110: SciTokens::teardown
113/131 Test #110: SciTokens::teardown
```

I fail to understand why is it different in your case.

dynamic-entropy:
@amadio I have added a commit to request a token with an extended (2 min) lifetime for the TPC tests. 
Cheers

tests/TPCTests/CMakeLists.txt
amadio: This is not with the first commit, but after you replace `SciTokens` fixture with `XRootD::scitokens`.

tests/TPCTests/test.sh
```diff
@@ -64,7 +64,7 @@ declare -A hosts_http=(
 
 setup_scitokens() {
 	if ! ${XRDSCITOKENS_CREATE_TOKEN} "${XRDSCITOKENS_ISSUER_DIR}"/issuer_pub_1.pem "${XRDSCITOKENS_ISSUER_DIR}"/issuer_key_1.pem test_1 \
-		"https://localhost:7095/issuer/one" "storage.modify:/ storage.create:/ storage.read:/"> "${PWD}/generated_tokens/token"; then
+		"https://localhost:7095/issuer/one" "storage.modify:/ storage.create:/ storage.read:/" 120 > "${PWD}/generated_tokens/token"; then
```
amadio, line 67: Can we make the default 5min? I suspect on slow systems 2min might still not be enough, especially if other tests run between the token being created and the actual TPC tests. If the default is 5min, then you also don't need to pass it in here. Later you could add proper command line argument parsing with [`getopt(3)`](https://www.man7.org/linux/man-pages/man3/getopt.3.html) to this command (i.e. parse `--issuer`, `--validity`, etc).

tests/TPCTests/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+  add_test(NAME XRootD::tpc::teardown
+    COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/setup.sh teardown")
+  set_tests_properties(XRootD::tpc::teardown
+    PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_CLEANUP XRootD_TPC)
+
+endif()
```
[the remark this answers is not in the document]
dynamic-entropy: Done

tests/TPCTests/CMakeLists.txt
dynamic-entropy: Very weird actually, I will give this a thought on what's going on. 
I have amended the commit and added both the fixtures as requirement for the test.

tests/TPCTests/test.sh
dynamic-entropy: I did not want to change the default so as not to disturb the existing tests that might depend on it being a 1 minute.
However, I can change the token lifetime here to 5 mins if you prefer?

tests/TPCTests/CMakeLists.txt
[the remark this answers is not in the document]
amadio: It seems to have happened on the CI, and it definitely happens almost every time on my machine, though.

tests/TPCTests/test.sh
amadio: Ok, sounds reasonable. I think it looks ready to merge now.

amadio (approved):
I think this is good to merge now.

amadio:
Actual test failures: https://my.cdash.org/tests/260335655 and https://my.cdash.org/tests/260335657. Failed due to
```
250317 10:03:14 30735 scitokens_Validate: Failed to deserialize SciToken: token expired
250317 10:03:14 30735 XrootdXeq: User authentication failed; token expired
250317 10:03:14 30735 XrdTLS: runner.31565:24@localhost TLS error rc=0 ec=6 (zero_return) errno=0.
250317 10:03:14 30735 XrootdXeq: runner.31565:24@localhost disc 0:00:00
```

amadio:
The timeout should be enough to cover the slowest running test suite, otherwise tests will run between the creation and usage of the token and make the TPC test fail. Since the expiration doesn't matter much, let's make it bigger, like 30min to ensure on slow systems the test doesn't fail because the token expires (or you have to generate and use the token within the same test).

dynamic-entropy:
Okay, I can reproduce this on Ubuntu.
This does not look like a shortcoming of the test though. 

The transfers seem to stall in Ubuntu and fail with "Internal server error". 
I will give the server logs a look again in the afternoon. For now, I failed to pinpoint the problem.

bbockelm:
@amadio - do you think the prior failures could be related to https://github.com/xrootd/xrootd/issues/2464?

That is, if there's a problem with the IPv6 code for HTTP redirects as reported b @ellert, then it might be what causes the periodic test failure (and, depending on the runner you land on for retries, may explain why the failures are not consistent -- latest rebase passes just fine).  So, it's possible it's not really the scitokens/TPC-related pieces but the underlying HTTP tests for multi-server setups.

bbockelm (approved):
@dynamic-entropy and I went over the tests in Zoom.  All my initial suggestions have been addressed and this is starting to look quite nice from my point of view.

amadio:
@bbockelm Indeed, the test is looking good. However, before merging, we need to understand and fix what is causing it to fail. I'd like to avoid merging tests that fail into master/devel, because then other pull requests will have failures as well. It could be that the reasons are the same as #2464. If yes, then we will need to adapt what runs to the network environment we have (or document that a working ipv4/ipv6 environment is a requirement to run the test suite).

amadio:
There is an actual failure to transfer, but the return code is still 201, which doesn't make sense to me (from https://my.cdash.org/tests/260410204):
```
+++ /usr/bin/curl -X COPY -L -s -o /dev/fd/63 -w '%{http_code}' -H 'Destination: https://localhost:10951//srvdata/tpc/srv2_to_srv1.ref_http_push' -H 'Authorization: Bearer eyJhbGciOiJFUzI1NiIsImtpZCI6InRlc3RfMSJ9.eyJhdWQiOiJodHRwczpcL1wvd2xjZy5jZXJuLmNoXC9qd3RcL3YxXC9hbnkiLCJleHAiOjE3NDIyNDExMzYsImlhdCI6MTc0MjIzOTMzNiwiaXNzIjoiaHR0cHM6XC9cL2xvY2FsaG9zdDo3MDk1XC9pc3N1ZXJcL29uZSIsImp0aSI6IjRjY2Q0MGI5LTI3YzAtNDU2NS1iMTY2LTk1MTUzMzQ5OTNlNSIsIm5iZiI6MTc0MjIzOTMzNiwic2NvcGUiOiJzdG9yYWdlLm1vZGlmeTpcLyBzdG9yYWdlLmNyZWF0ZTpcLyBzdG9yYWdlLnJlYWQ6XC8iLCJzdWIiOiJ0ZXN0Iiwid2xjZy52ZXIiOiIxLjAifQ.OUmqXthbJlkDlxW5m98oDp3ale79EoMNjsKhg29fdPyDoAhLEonuIqCTO4VOORyeduxBHoYC2Af2BwoaMtLX_g' -H 'TransferHeaderAuthorization: Bearer eyJhbGciOiJFUzI1NiIsImtpZCI6InRlc3RfMSJ9.eyJhdWQiOiJodHRwczpcL1wvd2xjZy5jZXJuLmNoXC9qd3RcL3YxXC9hbnkiLCJleHAiOjE3NDIyNDExMzYsImlhdCI6MTc0MjIzOTMzNiwiaXNzIjoiaHR0cHM6XC9cL2xvY2FsaG9zdDo3MDk1XC9pc3N1ZXJcL29uZSIsImp0aSI6IjRjY2Q0MGI5LTI3YzAtNDU2NS1iMTY2LTk1MTUzMzQ5OTNlNSIsIm5iZiI6MTc0MjIzOTMzNiwic2NvcGUiOiJzdG9yYWdlLm1vZGlmeTpcLyBzdG9yYWdlLmNyZWF0ZTpcLyBzdG9yYWdlLnJlYWQ6XC8iLCJzdWIiOiJ0ZXN0Iiwid2xjZy52ZXIiOiIxLjAifQ.OUmqXthbJlkDlxW5m98oDp3ale79EoMNjsKhg29fdPyDoAhLEonuIqCTO4VOORyeduxBHoYC2Af2BwoaMtLX_g' --cacert /home/runner/work/xrootd/xrootd/build/tests/issuer/tlsca.pem https://localhost:10952//srvdata/tpc/srv2.ref
...
Perf Marker
Timestamp: 1742239457
Stripe Index: 0
Stripe Bytes Transferred: 0
Total Stripe Count: 1
RemoteConnections: tcp:[::1]:10951
End
failure: Internal transfer failure, local=/srvdata/tpc/srv2.ref, remote=https://localhost:10951//srvdata/tpc/srv2_to_srv1.ref_http_push, HTTP library failure=Timeout was reached++ http_code=201
++ echo 201
++ return 0
+ assert_eq 201 201 'HTTP TPC push failed'
+ [[ 201 == \2\0\1 ]]
```

amadio:
The error comes from this place in the code:

https://github.com/xrootd/xrootd/blob/53474d9dc1b3d562ea2931705135b76e34b60f64/src/XrdTpc/XrdTpcTPC.cc#L757-L763

amadio:
Looks like the current changes are still not enough to fix the issue, the test still failed on [Alma 10](https://my.cdash.org/tests/266145421).

dynamic-entropy:
Hello @amadio 

Brian just fixed the content length check. Could you please trigger the jobs to run? Seems like GitHub did not run them on my recent push.

PR #2442, opened 2025-02-28
[Xrootd] Fix gstream configuration processing.

osschar:
This part was missing from commit 3ad5ad5584bb63c76110b8a1073e46dc13e99b3e in #2279. This caused the wrong index of `gXrd` to apply to everything after the new `oss` entry.
In particular, 'pfc' monitor driver was placed in the wrong environment and could not be retrieved in the `Pfc` code, causing pfc gstream to stop working.


abh3 (approved):
Looks good, thanks Matevz!

PR #2443, opened 2025-03-01
[XrdHttp] Fix HTTP protocol errors on failure

bbockelm:
When using `X-Transfer-Status`, there were three subtle HTTP protocol
errors:
- The trailer contained a `\n` which is not a permitted character. This caused clients to view the last CRLF as leftover data on
  an idle TCP channel.
- A trailer was sent but no `Trailer` header was sent in the response.
- When the server said that the connection would be kept alive but an error occurred, it closed the TCP connection unexpectedly on the client.  The server now keeps the TCP connection open if keepalive is enabled, provided the HTTP response was sent cleanly.

To prevent regressions, this adds the first unit test that injects read failures mid-transfer.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -524,7 +524,10 @@ bool XrdHttpReq::Error(XrdXrootd::Bridge::Context &info, //!< the result context
     free(s);
   }
 
-  if (PostProcessHTTPReq()) reset();
+  auto rc = PostProcessHTTPReq();
+  if (rc) {
+    reset();
+  }
 
   // If we are servicing a GET on a directory, it'll generate an error for the default
   // OSS (we don't assume this is always true).  Catch and suppress the error so we can instead
@@ -532,7 +535,7 @@ bool XrdHttpReq::Error(XrdXrootd::Bridge::Context &info, //!< the result context
   if ((request == rtGET) && (xrdreq.header.requestid == ntohs(kXR_open)) && (xrderrcode == kXR_isDirectory))
     return true;
   
-  return false;
+  return rc == 0;
 };
 
 bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
@@ -1133,8 +1136,7 @@ int XrdHttpReq::ProcessHTTPReq() {
 
             if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
               mapXrdErrorToHttpStatus();
-              sendFooterError("Could not run close request on the bridge");
-              return -1;
+              return sendFooterError("Could not run close request on the bridge");
             }
             return 0;
           } else {
@@ -1175,8 +1177,7 @@ int XrdHttpReq::ProcessHTTPReq() {
             if (!prot->Bridge->Run((char *) &xrdreq, (char *) res.c_str(), l)) {
               mapXrdErrorToHttpStatus();
               prot->SendSimpleResp(httpStatusCode, NULL, NULL, httpStatusText.c_str(), httpStatusText.length(), false);
-              sendFooterError("Could not run listing request on the bridge");
-              return -1;
+              return sendFooterError("Could not run listing request on the bridge");
             }
 
             // We don't want to be invoked again after this request is finished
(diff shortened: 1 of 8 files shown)


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1188,8 +1191,7 @@ int XrdHttpReq::ProcessHTTPReq() {
 
             if (!prot->Bridge->Run((char *) &xrdreq, 0, 0)) {
               mapXrdErrorToHttpStatus();
-              sendFooterError("Could not run close request on the bridge");
-              return -1;
+              return sendFooterError("Could not run close request on the bridge");
```
amadio, line 1194: I think it we already discussed that it's better not to hide the return value in `sendFooterError`, but return the error explicitly afterwards (which also allow returning something different than `-1` if needed.

bbockelm: I'm not sure we discussed it (apologies if I just forgot).

What are you suggesting?  Doing something like:

```
if (sendFooterError("blah") == -1) {
    return -1;
} else {
    return 1;
}
```

This seems redundant so I must be missing some aspect of the suggestion.

If we ever want to make it so `sendFooterError` returns different values than `ProcessHTTPReq` we can certainly refactor then.

tests/XRootD/http.sh
```diff
@@ -195,7 +195,7 @@ function test_http() {
   ## Download fails on a read failure
   # Default HTTP request: TCP socket abruptly closes
   curl -v --raw "${HOST}/${TMPDIR}/fail_read.txt" 2>&1 | sed 's/blah//g' >"$outputFilePath"
-  assert_eq "1" "$(grep -c -E '\* transfer closed with [0-9]+ bytes remaining to read' "$outputFilePath")" "Download did not fail as expected: '$(cat "$outputFilePath")'"
+  assert_eq "1" "$(grep -c -E '\* (end of response with [0-9]+ bytes missing|transfer closed with [0-9]+ bytes remaining to read)' "$outputFilePath")" "Download did not fail as expected: '$(cat "$outputFilePath")'"
```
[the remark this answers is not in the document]
bbockelm: ICYMI, the latest patch revision takes a different approach.  Instead of looking at the error message, we count the bytes sent in response and make sure they are fewer than expected (which is required here as the OSS layer is forced to throw an error).

src/XrdHttp/XrdHttpReq.cc
bbockelm: @amadio - doing a pass through all my open PRs.  Kind reminder about the above question.  I'm not sure I understand your original question so I don't quite know how to address your concern.

Thanks!

amadio: The previous discussion was in #2300 when I [requested changes](https://github.com/xrootd/xrootd/pull/2300#pullrequestreview-2404575191). In particular, I had requested to make the function void here: https://github.com/xrootd/xrootd/pull/2300/files#r1822452730

amadio: Ok, we have discussed in person and agree that it's easier to just let this function return something.

tests/XrdOssTests/xrdoss_tests.cc
dynamic-entropy, line 1: Could you rename this to `XrdOssTests.cc` to keep it consistent with the structure in other tests?

dynamic-entropy, line 1: Could you rename this to `XrdOssTests.cc` to keep it consistent with the structure in other tests?

dynamic-entropy (commented):
Planning to build on the XrdOssTests to improve the webdav error messages. Left a request to rename the test file.

bbockelm:
@amadio - I realized this PR was still open when rebasing branches.  Certainly protocol errors are a bug -- was there a reason it didn't make it into 5.8.2?  Just an oversight?

amadio:
Yes, oversight, sorry. There were still unresolved conversations, so I wasn't sure this was ready or not. We can add this to the next patch release in ~2 weeks. I will tag it so I don't forget.

bbockelm:
No worries!  We carry this out-of-tree so it can wait until the next patch release.  Just wanted to make sure there wasn't a deeper issue.

amadio:
@bbockelm In order for us to be able to merge this, please rebase it on top of the current master branch. If CI is red I cannot merge. Thanks.

PR #2444, opened 2025-03-03
[XrdTpc] Modernize libcurl linkage

bbockelm:
Small improvement I noticed for XrdTpc while cleaning up other sources:

We define a minimum CMake version such that we're guaranteed to have the `CURL::libcurl` target; use that to match CMake's current philosophy.


amadio:
The failed build is a known intermittent failure. I'm rerunning and will merge. Thanks.

Issue #2432, opened 2025-02-13
Define XRootD_BIN_DIR in XRootDConfig.cmake


bbockelm:
As I've been modernizing my CMake setups, I've been changing tests to not rely on "whatever was set in `$PATH`" but instead use the XRootD install as discovered by CMake.

Awkwardly, there's no way to discover from the installed `XRootDConfig.cmake` where the binaries live.  The best I came up with was something like https://github.com/PelicanPlatform/xrootd-s3-http/blob/main/test/CMakeLists.txt#L61:

```
    ENVIRONMENT XROOTD_BINDIR=${XRootD_DATA_DIR}/../bin"
```

It's obviously not pretty and definitely not foolproof.

Can this information be exported?

(Should we have a "build" label for CMake-related issues?)

Issue #2415, opened 2025-01-24
Provide CMake targets in next release?


hageboeck:
In the next release, could you consider providing the following CMake targets with proper link and include paths?
- XRootD::XrdCl
- XRootD::XrdUtils
- [...]

PR #2453, opened 2025-03-15
CMake build system cleanups and restructuring

amadio:
In preparation to address issues #2415 and #2432, I started to perform a general restructuring of the build system for XRootD 5.8, to use more idiomatic and modern CMake constructs. This pull request is my current draft of it, which I would like you to review.

Before doing anything, I had to set the output directory for binaries and libraries to be `<BUILD_DIR>/bin` and `<BUILD_DIR>/lib`, otherwise paths to binaries such as `xrdcp`, etc, change depending on which CMake directory they were build. Overall, I think this makes it much simpler to manage `PATH` in tests, etc, since you only need `<BUILD_DIR>/bin` and not `src/` for server binaries and `src/XrdCl/` for client binaries, etc. Libraries and plugins are automatically loaded from the build directory when specified only by name, so in configuration files, if you have full paths, you can just use the basename without the directory part in the future.

I am moving most `src/Xrd*.cmake` scripts to `src/Xrd*/CMakeLists.txt` and moving build logic as close as possible to where it's needed, such as calls to `find_package`, etc. I would also like to use this opportunity (since I will need to touch almost all build system files) to update the CMake style we use to be more consistent across the whole repository, maybe something similar to what's shown [here](https://community.kde.org/Policies/CMake_Coding_Style) and [here](https://gitlab.kitware.com/cmake/community/-/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file). Please let me know your opinions, I'd prefer to adjust style before merging this.

I still have to deal with `XrdDaemons.cmake`, `XrdHeaders.cmake`, `XrdServer.cmake`, and `XrdUtils.cmake`. I'd like to hear your thoughts on how to move their bits down to subdirectories. My plan is to create the libraries and binaries in `src/CMakeLists.txt` and use `target_sources` in subdirectories to add the files to the targets as needed.

After this restructuring is over, I will work on improving the exported files (i.e. `XRootDConfig.cmake`) and consider creating functions to manage libraries and plugins in a more central way in case global changes are needed. For now, I'm just using plain CMake commands everywhere.

Thanks for reviewing!

Update: I've finished most of the remaining work. Now only `XrdHeaders.cmake` needs to be dealt with. I've dealt with #2415 and #2432 as the last thing in the pull request, but it's still not over yet, I might add a few more changes before adding to `master` and into 5.8.0.


bbockelm:
@amadio - the improvements look great!  Much more idiomatic from the CMake side.

Two thoughts, both for the "next round" I think:
- It might be worthwhile to reach out to CMake-based dependencies (such as [SciTokens](https://github.com/scitokens/scitokens-cpp/)) and ask them to modernize and install CMake targets as part of their build process.
- For hard dependencies, we can use `FetchContent_MakeAvailable` ([here's](https://github.com/PelicanPlatform/xrootd-s3-http/blob/main/CMakeLists.txt#L47-L65) an example from a different project) to pull things if they're not present.  That's allowed me to get rid of in-tree vendor dependencies such as `tinyxml`.  That makes it easier for folks getting started: if done for the entire dependency tree, you can start with nothing but a build chain and have a pretty functional xrootd build.  If you take this approach, coordinate with packagers such as @ellert as most distros have a strict no-network-connectivity-during-build and there needs to be a straightforward mechanism to skip this feature.

amadio:
> * It might be worthwhile to reach out to CMake-based dependencies (such as [SciTokens](https://github.com/scitokens/scitokens-cpp/)) and ask them to modernize and install CMake targets as part of their build process.

Yes, I'm coming for you, haha. I recently realized that scitokens-cpp does not export a version anywhere, so it's not possible to detect what version is installed. It would be nice to at least export something in the headers or a `ScitokensConfigVersion.cmake` and `ScitokensConfig.cmake`, like most packages.

As for bundling dependencies, well, I'm not a big fan of that. When I was working in ROOT I was always fighting against it, but wasn't heard. I think bundled dependencies can easily lead to security vulnerabilities and suffer from code rot, since they are rarely updated. In XRootD, I've removed the bundled `isa-l` already, and hope to remove the rest instead of adding more. It's not a big deal to install the dependencies on most OSs and Linux distros, since XRootD is already officially packaged in most places. In most distros it's a single command using either the `xrootd.spec` or `debian/control` file.

amadio:
I'm going to split this into two parts, one which I will merge soon to master and which contains part of the changes here, plus a few extras. The other, which I will keep in this pull request, is to do the refactoring for moving `Xrd*.cmake` to `Xrd*/CMakeLists.txt`, which I am still working on and will merge into a future release.

PR #2454, opened 2025-03-15
[XrdPss] Implement the auto-stat protocol for XrdPss

bbockelm:
This fixes failures in directory listings through `XrdHttp` for XCache.

Since the `XrdPss` did not implement the auto-stat protocol (i.e., implemented the `StatRet` virtual function in `XrdOssDF`), the xrootd protocol handler does a directory listing followed by a `stat` for each entry.  This would cause failures for `XrdPfc` since stat'ing a subdirectory entry fails with:

```
Unable to locate <subdir name>; is a directory
```

if a file is cached in that subdirectory.  This results in a strange user experience where

```
curl http://cache.example.com/dir
```

succeeds at first but, if you download an object in the subdirectory,

```
curl http://cache.example.com/dir/subdir/foo.txt
```

then the subsequent directory listing of `/dir` fails because the stat of `subdir` fails when it's on the local disk.

Luckily, since XrdCl returns sufficient information to always do the auto-stat, this is not only an optimization in the proxy but also serves as a bugfix.


src/XrdPosix/XrdPosixDir.hh
```diff
@@ -69,6 +70,11 @@ static int         dirNo(DIR *dirP)  {return *(int *)dirP;}
 
        DIR        *Open();
 
+       // Return the stat info corresponding to the current directory entry
+       // On error, returns -errno; otherwise, returns 0 and stores a reference
```
abh3, line 74: Hmm, it stores a pointer not a reference to the buffer. Please correct the comment.

src/XrdPosix/XrdPosixDir.hh
```diff
@@ -40,6 +40,7 @@
 #endif
 
 #include <unistd.h>
+#include <sys/stat.h>
```
abh3, line 43: Generally, if we don't make use of an object we simply forward declare it and the definition of the object (i.e. the include file) is placed  inthe cc file where it is used. That way we try to avoid implied dependencies on include files which happen from time to time.

src/XrdPosix/XrdPosixMap.cc
```diff
@@ -80,6 +80,41 @@ mode_t XrdPosixMap::Flags2Mode(dev_t *rdv, uint32_t flags)
...
+/*                             S t a t F i l l                                */
+/******************************************************************************/
+
+int XrdPosixMap::Entry2Buf(const XrdCl::DirectoryList::ListEntry &dirEnt, struct stat &buf)
+{
+   auto statInfo = dirEnt.GetStatInfo();
+   if (!statInfo) return EIO;
+
+   memset(&buf, '\0', sizeof(buf));
+   buf.st_mode   = XrdPosixMap::Flags2Mode(nullptr, statInfo->GetFlags());
```
abh3, line 94: Why specifically state the class method you are calling in the same class where the call occurs? It makes it more difficult to read.

src/XrdPosix/XrdPosixMap.cc
```diff
@@ -80,6 +80,41 @@ mode_t XrdPosixMap::Flags2Mode(dev_t *rdv, uint32_t flags)
...
+   // Since the UID/GID isn't known by the client, when these are translated by
+   // XrdXrootdProtocol::StatGen back to xrootd protocol flags, they will get zero'd
```
abh3, line 96: xroot is the protocol and xrootd is the implementation of the protocol. Please correct.

src/XrdPosix/XrdPosixMap.cc
```diff
@@ -80,6 +80,41 @@ mode_t XrdPosixMap::Flags2Mode(dev_t *rdv, uint32_t flags)
 
    return newflags;
 }
+
+/******************************************************************************/
+/*                             S t a t F i l l                                */
+/******************************************************************************/
+
+int XrdPosixMap::Entry2Buf(const XrdCl::DirectoryList::ListEntry &dirEnt, struct stat &buf)
```
abh3, line 88: As I mentioned where this is called, this method should be allowed to return the reason for the failure as a text message as well. EIO doesn't capture much here. Of course, that assumes the we may have more than one error return and more than one failure condition could occur. I'm not insisting on this as it's unlikely to occur.

src/XrdPosix/XrdPosixMap.hh
```diff
@@ -35,6 +35,7 @@
 #include "XrdCl/XrdClXRootDResponses.hh"
 
 #include <cstdint>
+#include <sys/stat.h>
```
abh3, line 38: Again, please forward declare struct stat and move the include to the cc file where the object is used.

src/XrdPosix/XrdPosixXrootd.cc
```diff
@@ -1060,6 +1060,22 @@ int XrdPosixXrootd::Readdir64_r(DIR *dirp, struct dirent64  *entry,
    return rc;
 }
 
+int XrdPosixXrootd::StatRet(DIR *dirp, struct stat *buf)
```
abh3, line 1063: Normally, methods are placed in this file are in alphabetic order so please place this after stafs().

abh3 (changes requested):
No drastic issues but we do need some changes largely for consistency and alignment  with R6.

src/XrdPosix/XrdPosixXrootd.cc
bbockelm: Done.

src/XrdPosix/XrdPosixMap.hh
bbockelm: Done.

src/XrdPosix/XrdPosixMap.cc
bbockelm: I refactored this to set the error message object directly, giving a (somewhat) more descriptive error message.

src/XrdPosix/XrdPosixMap.cc
bbockelm: Agreed.  This function was originally in `XrdPosixDir` but I refactored it to be in the same `XrdPosixMap` class.

Fixed.

src/XrdPosix/XrdPosixDir.hh
bbockelm: Done.

src/XrdPosix/XrdPosixDir.hh
bbockelm: :) I think I was being overly pedantic - a pointer is a type of reference (unfortunately, a C++ reference is also a reference...).  Happy to change it.

src/XrdPosix/XrdPosixDir.cc
```diff
@@ -83,10 +83,27 @@ dirent64 *XrdPosixDir::nextEntry(dirent64 *dp)
    dp->d_reclen = d_nlen + dirhdrln;
    strncpy(dp->d_name, d_name, d_nlen);
    dp->d_name[d_nlen] = '\0';
+
+   // Note we fail if the stat info is needed but not available
+   int rc;
+   if (myBuf && (rc = XrdPosixMap::Entry2Buf(*dirEnt, *myBuf)))
+      {eNum = rc;
+       ecMsg.SetErrno(eNum);
```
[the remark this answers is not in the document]
bbockelm: Good idea!  Fixed in the latest commit.

bbockelm:
@abh3 / @amadio - what's the latest here?  Last I heard, about a month ago, Guilherme said that he wanted Andy to take one last look.

amadio:
Hi @bbockelm, I think this should go into 5.9 since it adds new functionality. The next feature release should come sometime in the Summer (i.e. before the next FTS/XRootD workshop).

bbockelm:
How's it a new feature?

If you do a directory listing on an empty cache, it succeeds.

If you download a file in the subdirectory of that directory, then the directory listing stops working.

With this, the directory listing works regardless of cache state.

amadio:
Fair enough, I meant that it adds new symbols to existing classes, which normally should go into a feature release. However, I will check ABI compatibility now and if nothing breaks I can put this into the patch release tomorrow.

PR #2457, opened 2025-03-18
[XrdPss] Fix proxy when using tokens and crc32c

bbockelm:
When debugging the proxy with crc32c enabled and using token auth, I hit two issues fixed by this PR:
1.  `crc32c` is not in the hardcoded list of known algorithms.  Probably due to the fact `crc32c` support is newer than the PSS checksum support?
2. XrdPss drops the authorization token when it creates the forwarded request.  Unlike other calls to construct a URL, it disabled the appending of the environment CGI.
   - A side-effect of this was it also dropped the `cks.type` parameter for non-xrootd URLs (e.g., `https://` or `pelican://`) because `cks.type` was only explicitly forwarded for xroot-protocol-based origins.

On the upside, with this, I'm able to get checksums to pass-through from the cache to a Pelican-based origin!

### DIFF
--- a/src/XrdPss/XrdPssCks.hh
+++ b/src/XrdPss/XrdPssCks.hh
@@ -78,7 +78,7 @@ struct csInfo
 
 csInfo *Find(const char *Name);
 
-static const int csMax = 4;
+static const int csMax = 8;
 csInfo           csTab[csMax];
 int              csLast;
 };
--- a/src/XrdPss/XrdPssCks.cc
+++ b/src/XrdPss/XrdPssCks.cc
@@ -170,7 +170,7 @@ int XrdPssCks::Init(const char *ConfigFN, const char *DfltCalc)
 // See if we need to set the default calculation
 //
    if (DfltCalc)
-      {for (i = 0; i < csLast; i++) if (!strcmp(csTab[i].Name, DfltCalc)) break;
+      {for (i = 0; i <= csLast; i++) if (!strcmp(csTab[i].Name, DfltCalc)) break;
        if (i >= csMax)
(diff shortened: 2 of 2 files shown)


src/XrdPss/XrdPssCks.hh
```diff
@@ -78,7 +78,7 @@ struct csInfo
 
 csInfo *Find(const char *Name);
 
-static const int csMax = 4;
+static const int csMax = 5;
```
amadio, line 81: Is this not one more than it should? The last entry is `csTab[3]`, no? Is there a sentinel at the end?

amadio: Yes, I agree with the analysis above.

@osschar Could you please check this?

abh3: The following two lines should have implied that csMax is the size of the table caTab[].
static const int csMax = 8;
csInfo           csTab[csMax];

So, we have a fixed sized table of 8 elements. The use of csLast relative to looking/manipulating csTab should imply that the table is not completely populated so, in fact, csLast points to the last valid entry in the table. 

This is the case here. The code supports up to 8 different checksums. The table is pre-populated with native checksums but one is allowed to add additional checksums beyond that and those will be added to the table in the standard way one does that during configuration processing.

abh3: Yes, if you look at my edited response, while don't explicitly say so, I would increase to be the same between the two versions. Thanks!

abh3:
@amadio People are now always submitting to master and retargeting to devel doesn't produce good results. We need to straighten this out. So, I am not merging this for now. Otherwise it's ready for merging.

PR #2459, opened 2025-03-22
Fix builds where xrootd is embedded CMake dependency

bbockelm:
When doing builds of XRootD that are embedded in other projects, `${CMAKE_SOURCE_DIR}` refers to the top-level build directory (which is not the xrootd source code but the parent project).  Remove use of `${CMAKE_SOURCE_DIR}` where not needed and switch to `${CMAKE_CURRENT_SOURCE_DIR}` (the absolute path to the directory where the current cmake file resides) for things like build flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,7 @@ configure_file(src/XrdVersion.hh.in src/XrdVersion.hh)
 # Build in subdirectories
 #-------------------------------------------------------------------------------
 
-include_directories(BEFORE ${CMAKE_SOURCE_DIR}/src ${CMAKE_BINARY_DIR}/src)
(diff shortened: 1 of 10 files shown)


amadio:
Hi @bbockelm I think this change is good, but this will create a lot of rebase problems for my ongoing restructuring of the build system in #2453. I think I'd rather just take care of this issue in that pull request if you don't mind.

bbockelm:
When do you plan on landing the larger refactor?  I was hoping to get this in a patch release to simplify a build.

amadio:
Sorry for the long delay, things were busy with the EOS workshop and this fell through the cracks. I plan to release my CMake refactor as soon as possible, since no interfaces are changed, it's just a modernization of the build system. It will likely come with the next patch release.

Issue #2433, opened 2025-02-14
Use `sysusers.d` config file for creating the `xrootd` user


amadio:
Let's follow downstream (Fedora and EPEL), and use sysusers.d config file for creating the xrootd user in the spec file. For more information, please see https://src.fedoraproject.org/rpms/xrootd/pull-request/2# and https://fedoraproject.org/wiki/Changes/RPMSuportForSystemdSysusers.

bbockelm:
Yes!

Is it possible to also request a fixed, assigned UID for the `xrootd` user on those platforms as there is for `condor`?

Just this week we hit an issue where the system UID for `xrootd` did not match the container UID for `xrootd`, causing failures in having xrootd access SMB/CIFS-based filesystems (this is because the kernel invokes [cifs-upcall](https://linux.die.net/man/8/cifs.upcall) in the system namespace while the access occurred from the container).

amadio:
I guess an assigned UID will have to be requested downstream, in the distributions.

@ellert What is your plan regarding the pull request linked above? What about the assigned uid for xrootd? It would be nice to know before putting this in our release. Thanks!

ellert:
In order to get a static rather than a dynamic uid you need to file a request and motivate why there is a need for it:

https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_allocation_strategies
https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_soft_static_allocation

So it should not be impossible to get one with a good motivation.

ellert:
https://pagure.io/packaging-committee/issue/1444

amadio:
Thank you very much, @ellert! Is there a similar procedure we could follow for Debian/Ubuntu (hopefully using the same id assignment if possible)?

ellert:
Yes, I have now also found this:

https://www.debian.org/doc/debian-policy/ch-opersys.html#uid-and-gid-classes

And here:

https://salsa.debian.org/debian/base-passwd

At the bottom of the page:

> You *may not* use any uids or gids in the 60000-64999 range without *first* requesting an allocation from base-passwd@packages.debian.org and waiting for confirmation that the allocation has been granted.

Since here the request is handled by e-mail, I can't give link to a ticket. I have sent a request using much the same wording as for teh Fedora/EPEL case. Since Fedora ad Debian uses different uid ranges for these allocation the uid can not be the same for the two.

amadio:
[...]

PR #2460, opened 2025-03-23
Workaround for %pre being evaluated before sources are unpacked in %prep

ellert:
See the last paragraph in this comment:

https://github.com/xrootd/xrootd/issues/2433#issuecomment-2741703886

With this tweak using the file in the tar ball works.

### DIFF
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -453,7 +453,7 @@ install -m 644 -p packaging/common/%{name}.pp \
 %ldconfig_scriptlets server-libs
 
 %pre server
-%sysusers_create_compat packaging/rhel/%{name}-sysusers.conf
+%sysusers_create_compat %(tar -z -x -f %SOURCE0 %{name}%{!?_with_git:-%{version}}/packaging/rhel/xrootd-sysusers.conf -O > /tmp/xrootd-sysusers.conf && echo /tmp/xrootd-sysusers.conf)
 
 %post server
 %tmpfiles_create %{_tmpfilesdir}/%{name}.conf
--- a/xrootd.spec
+++ b/xrootd.spec
@@ -453,7 +453,7 @@ install -m 644 -p packaging/common/%{name}.pp \
 %ldconfig_scriptlets server-libs
 
 %pre server
-%sysusers_create_compat %(tar -z -x -f %SOURCE0 %{name}%{!?_with_git:-%{version}}/packaging/rhel/xrootd-sysusers.conf -O > /tmp/xrootd-sysusers.conf && echo /tmp/xrootd-sysusers.conf)
+%sysusers_create_compat %(tar -z -x -f %{SOURCE0} --no-anchored xrootd-sysusers.conf -O > /tmp/xrootd-sysusers.conf && echo /tmp/xrootd-sysusers.conf)
 
 %post server
 %tmpfiles_create %{_tmpfilesdir}/%{name}.conf


amadio:
Is this really needed? I'm fine with the current solution of just installing the extra dependency on Alma 8. I wish I could depend on EPEL and friends in the spec file, but it seems this is a limitation of the system that I cannot really depend on it.

ellert:
This is not about the installation of the sysusers file. This is about the pre installation scriptlet.

With the spec as is before the proposed change:
```
$ rpm -q --scripts -p xrootd-server-5.8.0-1.el9.x86_64.rpm 
preinstall program: /bin/sh
```
With the proposed change:
```
$ rpm -q --scripts -p xrootd-server-5.8.0-1.el9.x86_64.rpm 
preinstall scriptlet (using /bin/sh):

# generated from xrootd-sysusers.conf
getent group 'xrootd' >/dev/null || groupadd -r 'xrootd' || :
getent passwd 'xrootd' >/dev/null || \
    useradd -r -g 'xrootd' -d '/var/spool/xrootd' -s '/usr/sbin/nologin' -c 'XRootD runtime user' 'xrootd' || :
```
I.e. with the original version the scriptlet is not created because at the time the %pre section is evaluated, the source has not been unpacked yet and the information in it can not be used to create the scriptlet.

In Fedora 43 it is empty in both cases, because there the  RPM version i super new, and knows how to create users and groups from the provides, but in older release, including RHEL, this is not the case and the scriptlet is needed.

Provides in EPEL 9 case:
```
$ rpm -q --provides -p xrootd-server-5.8.0-1.el9.x86_64.rpm | grep -E '(user|group)'
group(xrootd)
user(xrootd)
```
Provides in Fedora 43 case:
```
$ rpm -q --provides -p xrootd-server-5.8.0-1.fc43.x86_64.rpm | grep -E '(user|group)'
group(xrootd)
user(xrootd) = dSB4cm9vdGQgLSAiWFJvb3REIHJ1bnRpbWUgdXNlciIgL3Zhci9zcG9vbC94cm9vdGQgLQAA

$ echo dSB4cm9vdGQgLSAiWFJvb3REIHJ1bnRpbWUgdXNlciIgL3Zhci9zcG9vbC94cm9vdGQgLQAA | base64 -d
u xrootd - "XRootD runtime user" /var/spool/xrootd -
```
I.e. the provides contains the line from the sysusers file and RPM can create the user from the information in the provides and the scriptlet is not needed. But not for old versions of RPM like the one in RHEL 9. There the scriptlet is needed.

xrootd.spec
```diff
@@ -453,7 +453,7 @@ install -m 644 -p packaging/common/%{name}.pp \
 %ldconfig_scriptlets server-libs
 
 %pre server
-%sysusers_create_compat packaging/rhel/%{name}-sysusers.conf
+%sysusers_create_compat %(tar -z -x -f %SOURCE0 %{name}%{!?_with_git:-%{version}}/packaging/rhel/xrootd-sysusers.conf -O > /tmp/xrootd-sysusers.conf && echo /tmp/xrootd-sysusers.conf)
```
amadio, line 456: Here's a suggestion to simplify things a bit:
```suggestion
%sysusers_create_compat %(tar -z -x -f %{SOURCE0} --no-anchored xrootd-sysusers.conf -O > /tmp/xrootd-sysusers.conf && echo /tmp/xrootd-sysusers.conf)
```

amadio:
Thank you for the explanation. I understand why the change is necessary. I suggested a small improvement to not depend on `_with_git` for the extraction of the file, or its path inside the archive. Could apply that? Then I can merge. Thank you!

PR #2472, opened 2025-04-06
[XrdPfc] Enable write-through mode for cache

bbockelm:
Allow the PFC to proxy ("write-through") writes when configured to do so (default is current behavior of disabling writes).

The implementation is simple -- when writing is enabled, don't attach the PFC to the cache object.

Adds a new parameter:
```
pfc.writemode [writethrough | off]
```
where the default write mode is `off`.

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -180,7 +180,11 @@ XrdOucCacheIO *Cache::Attach(XrdOucCacheIO *io, int Options)
 {
    const char* tpfx = "Attach() ";
 
-   if (Cache::GetInstance().Decide(io))
+   if (Options & XrdOucCache::optRW)
+   {
+      TRACE(Info, tpfx << "passing through write operation" << io->Path());
+   }
+   else if (Cache::GetInstance().Decide(io))
    {
       TRACE(Info, tpfx << obfuscateAuth(io->Path()));
 
@@ -1071,6 +1075,10 @@ int Cache::Prepare(const char *curl, int oflags, mode_t mode)
    // Do not allow write access.
    if (oflags & (O_WRONLY | O_RDWR | O_APPEND | O_CREAT))
    {
+      if (Cache::GetInstance().RefConfiguration().m_write_through)
+      {
+         return 0;
+      }
       TRACE(Warning, "Prepare write access requested on file " << f_name << ". Denying access.");
       return -EROFS;
    }
--- a/src/XrdPfc/XrdPfcConfiguration.cc
+++ b/src/XrdPfc/XrdPfcConfiguration.cc
@@ -31,6 +31,7 @@ using namespace XrdPfc;
 XrdVERSIONINFO(XrdOucGetCache, XrdPfc);
 
 Configuration::Configuration() :
+   m_write_through(false),
    m_hdfsmode(false),
    m_allow_xrdpfc_command(false),
    m_data_space("public"),
@@ -608,6 +609,21 @@ bool Cache::Config(const char *config_filename, const char *parameters)
       }
    }
 
+   // set the write mode
+   if ( ! tmpc.m_writemodeRaw.empty())
+   {
+      if (tmpc.m_writemodeRaw == "writethrough")
+      {
+         m_configuration.m_write_through = true;
+      }
+      else if (tmpc.m_writemodeRaw != "off")
+      {
+         m_log.Emsg("ConfigParameters", "Unknown value for pfc.writemode (valid values are `writethrough` or `off`): %s",
+                    tmpc.m_writemodeRaw.c_str());
+         return false;
+      }
+   }
+
    // get number of available RAM blocks after process configuration
    if (m_configuration.m_RamAbsAvailable == 0)
    {
@@ -690,6 +706,7 @@ bool Cache::Config(const char *config_filename, const char *parameters)
       {
          loff += snprintf(buff + loff, sizeof(buff) - loff, "       pfc.hdfsmode hdfsbsize %lld\n", m_configuration.m_hdfsbsize);
       }
+      loff += snprintf(buff + loff, sizeof(buff) - loff, "       pfc.writemode %s\n", m_configuration.m_write_through ? "writethrough" : "off");
 
       if (m_configuration.m_username.empty())
       {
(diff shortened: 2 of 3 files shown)


bbockelm:
Note: Mac failure is due to:
```
Error message was: Failed to connect to my.cdash.org port 443 after 1089 ms: Couldn't connect to server
```
I think we can trigger a retry on that to clear it up.

amadio:
I think the macOS errors are real. Also, it would be good to check with @Jo-stfc about their implementation as well (see [slides](https://indico.cern.ch/event/1510817/contributions/6358257/attachments/3018163/5323801/XRootDColab0225.pdf#page=4) from collaboration meeting).

Jo-stfc:
thanks for the reminder, @amadio.
I've added the PR for the RAL implementation
https://github.com/xrootd/xrootd/pull/2473
This has been tested and working for 2 weeks now for our workernode XCaches at RAL for LHCb.

Jo-stfc:
If the origin is walled off, am I understanding correctly that the behaviour you want from the XCache is to write to the local cache space instead? you could get a similar behaviour by setting up a second, local xrootd server and redirecting the writes to that instead?

Jo-stfc:
or if the proxy can access the origin, you could have a forwarding proxy and redirect writes to it?

bbockelm:
I mean, yes, I could double the number of servers we need ... but why?  We have code that already does this, it's just not currently accessible.

Jo-stfc:
Hi Brian, my concern is that we may have a conflicting use case for that scenario in the way we use xcaches at RAL. 
It's possible to disentangle them, but they may cause some confusion if not properly differentiated.

in the RAL use case the redirect was designed for, we're setting up an Xcache and an origin server on each node to increase the available throughput, while preventing vector-read IOps from overloading our storage. The storage in our case is co-located with the compute, so the redirect does not add much network overhead. As far as the user is concerned they talk to the common endpoint alias and the redirect enables it to work as a single read/write endpoint to the storage directly
i.e, how should the site advertise that a R/W node is write to local space or write back to remote? 

I'm unsure of the expected behaviour under these scenarios:
1) if the cache is a forwarding cache to a single storage - client requests xcache:/path/to/local
    is the file expected to be written to the cache or remote? how will that file be kept track of later? 
2) the cache is for a full caching site (virtual placement) - client requests xcache:/remote/origin/path/
     remote may or may not be read only? 
     
 The situation I want to avoid is a user writing a file and not finding it where they expect to when they look for it later.

bbockelm:
As you say, these are pretty distinct use cases - I don't think we should aim for a single configuration that tackles both.

With this change and the config enabled, the cache will behave as if the PFC module isn't loaded when writes occur (that's what it does - bypasses the PFC and using the pure PSS layer instead of returning an error).  There's no caching of the file on the local host so a subsequent read pulls data in as if it's the first time the cache has seen it.

src/XrdPfc/XrdPfc.cc
```diff
@@ -1071,6 +1075,10 @@ int Cache::Prepare(const char *curl, int oflags, mode_t mode)
    // Do not allow write access.
    if (oflags & (O_WRONLY | O_RDWR | O_APPEND | O_CREAT))
    {
+      if (Cache::GetInstance().RefConfiguration().m_write_through)
+      {
+         return 0;
```
abh3, line 1080: Returning a 0 will cause a SEGV. You need to return the XrdOucCacheIO object that was passed to you via the args so that you pass through the I/O.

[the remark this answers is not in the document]
bbockelm: This is not the attach function but `Prepare`.  The return of 0 indicates that the open should proceed.  This will cause the `Attach` function to be invoked and, if this mode is enabled, the PFC will simply decline to interpose and return the original IO object.

In fact, that's what I like about this -- we're not adding new functionality here, we're leveraging existing functionality and just "declining to interpose".

bbockelm: (This appears to be a duplicate comment so I am just resolving the conversation)

src/XrdPfc/XrdPfc.cc
```diff
@@ -180,7 +180,11 @@ XrdOucCacheIO *Cache::Attach(XrdOucCacheIO *io, int Options)
 {
    const char* tpfx = "Attach() ";
 
-   if (Cache::GetInstance().Decide(io))
+   if (Options & XrdOucCache::optRW)
+   {
+      TRACE(Info, tpfx << "passing through write operation" << io->Path());
```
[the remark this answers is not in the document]
bbockelm: Done in the latest rebase.

bbockelm:
@abh3 - the requested cosmetic changes were made but I think the deeper request is incorrect (the function returns an integer, not a pointer).  Can you take another look?

abh3 (approved):

src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -31,6 +31,7 @@ using namespace XrdPfc;
 XrdVERSIONINFO(XrdOucGetCache, XrdPfc);
 
 Configuration::Configuration() :
+   m_write_through(false),
```
amadio, line 34: Do we envision more modes? For example, the mode wanted by RAL which redirects? Since they are not compatible, the config should make the modes mutually exclusive, such that both features can be merged. A boolean here is not enough to represent the modes if we want to have more than just `writethrough` and `off` as write modes. If we will only have "on" and "off", I'd rather name the option `pfc.writethrough` and make it work similar to `ofs.authorize`.

amadio:
@osschar Could you please review this pull request? Thank you.

amadio (commented):
Before merging, I'd like to make sure that this and RAL's write mode by redirect can be merged together (even if they cannot be used together). The `writemode` probably needs to be something other than a boolean to accomodate other modes in the future as well.

src/XrdPfc/XrdPfcConfiguration.cc
osschar: Hmmh, you mean to have per-prefix (or, eventually, regexp) action defined in file with lines like this:
```
/store/user deny | this-one | ral-one | write-here
```

Do we have the RAL mode implemented / available already? I am fine with:
- leaving it as is and changing later; or
- changing the bool to either an enum or just leaving it as std::string.

osschar (approved):
Other than the mode / name / flag type that we are still discussing ... I'm fine with this.

src/XrdPfc/XrdPfcConfiguration.cc
bbockelm: To be clear, the "RAL mode" has nothing to do with caches.  It works at the protocol layer, implementing redirects on write operations.  The queries never get the OFS/OSS layer (and hence are irrelevant to the PSS/PFC).

I do envision other modes in the future -- I've been prototyping writeback caching separately.

abh3:
Frankly, this is somewhat of a syntactic sugar argument. If it makes sense to do on/off for consistency or future compatibility then I would just do it. I don't think we really need permission here for something like this because it doesn't change anything algorithmically. It simply changes how the configuration option should be structured. So, the choice here is to have @bbockelm do the change or do it ourselves. I would take the path of least résistance given the history of this pull request.

abh3:
Do you need a reminder?

Issue #2470, opened 2025-04-04
TPC COPY Failure for files < 1 MB on curl > 8


dynamic-entropy:
Transfers for files smaller than 1MB fail with a curl version greater than at least 8.12 for the TPC "push" mode.

We have been able to reproduce these errors on the following platforms. 
```
AlmaLinux Kitten 10 | curl 8.9.1
Ubuntu 24.04 | curl 8.5.0
Gentoo | curl 8.12 | davix 0.8.10
```

In case of Ubuntu, the transfer stalls, only to fail with following message sent to the client.
```
failure: Internal transfer failure, local=/srvdata/tpc/srv2.ref, remote=https://localhost:10951//srvdata/tpc/srv2_to_srv1.ref_http_push, HTTP library failure=Timeout was reached
```

In Alma10 the transfer fails immediately with following error sent to the client
```
failure: Internal transfer failure, local=/srvdata/tpc/srv2.ref, remote=https://localhost:10951//srvdata/tpc/srv2_to_srv1.ref_http_push, HTTP library failure=Failed to open/read local data from file/application
```

Here is a diff of working and failing logs for the client on gentoo (courtesy of @amadio )
```
gentoo ~ $ diff -u fails.log works.log 
--- fails.log    2025-04-02 15:24:42.872141763 +0200
+++ works.log    2025-04-02 15:24:04.872086965 +0200
@@ -3,8 +3,8 @@
 DAVIX(core): HTTP/SSL Session caching ENABLED
 DAVIX(core): Redirection Session caching ENABLED
 DAVIX(core): libdavix path /usr/lib64/libdavix.so.0, version: 0.8.10
-DAVIX(grid): Hop: https://gentoo.cern.ch:8081/file.raw
-DAVIX(http): Create HttpRequest for https://gentoo.cern.ch:8081/file.raw
+DAVIX(grid): Hop: https://gentoo.cern.ch:8081/file8M.b64
+DAVIX(http): Create HttpRequest for https://gentoo.cern.ch:8081/file8M.b64
 DAVIX(http):  -> negotiateRequest
 DAVIX(http): NEON start internal request
 DAVIX(http): no cached ne_session, create a new one 
@@ -17,7 +17,7 @@
 DAVIX(http): enable client cert authentication by callback 
 DAVIX(ssl): ah_create, for WWW-Authenticate
 DAVIX(socket): Running pre_send hooks
-> COPY /file.raw HTTP/1.1
[...]

PR #2475, opened 2025-04-09
[XrdTpc] Force use of `Expect: 100-continue` header

bbockelm:
On older versions of libcurl (at least 7.x), PUTs unconditionally triggered the use of `Expect: 100-continue`.

We've observed that newer versions of libcurl no longer force this behavior -- but also have bugs when it's not present!  Upstream issue is reported at https://github.com/curl/curl/issues/17004.

Pragmatically, until libcurl is fixed -- and we can assume the fix is deployed everywhere -- we will need to go back to the older behavior as seen on AlmaLinux 9 and similar.

Credit to @dynamic-entropy for uncovering the issue, narrowing it down, and coming up with the approach for the fix.  I just added some comments :)

Fixes: https://github.com/xrootd/xrootd/issues/2470

### DIFF
--- a/src/XrdTpc/XrdTpcState.cc
+++ b/src/XrdTpc/XrdTpcState.cc
@@ -33,6 +33,7 @@ void State::Move(State &other)
     m_start_offset = other.m_start_offset;
     m_status_code = other.m_status_code;
     m_content_length = other.m_content_length;
+    m_push_length = other.m_push_length;
     m_stream = other.m_stream;
     m_curl = other.m_curl;
     m_headers = other.m_headers;
@@ -66,6 +67,7 @@ bool State::InstallHandlers(CURL *curl) {
             curl_easy_setopt(curl, CURLOPT_READDATA, this);
             struct stat buf;
             if (SFS_OK == m_stream->Stat(&buf)) {
+                m_push_length = buf.st_size;
                 curl_easy_setopt(curl, CURLOPT_INFILESIZE_LARGE, buf.st_size);
             }
         } else {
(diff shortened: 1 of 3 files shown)


src/XrdTpc/XrdTpcState.cc
```diff
@@ -90,9 +90,13 @@ bool State::InstallHandlers(CURL *curl) {
...
 /**
- * Handle the 'Copy-Headers' feature
+ * Setup any headers necessary for the GET/PUT operation
+ *
+ * Currently includes:
+ * - Handle the 'Copy-Headers' feature
+ * - Adding `Expect: 100-continue` to get around a libcurl bug on uploads.
  */
-void State::CopyHeaders(XrdHttpExtReq &req) {
+void State::ConfigureHeaders(XrdHttpExtReq &req) {
```
amadio, line 99: How about `SetupHeaders` instead of `ConfigureHeaders`? I think it makes more sense and in any case that's what you call it in your comment above.

bbockelm: Done!

ccaffy:
Thanks for the fix @bbockelm ,

I did not manage to see how you get the value `m_content_length` for HTTP TPC PUSH in SetupHeaders.
I tried it on my box and I only get '-1' instead of a value there, even if I manually set the Content-Length header on the COPY PUSH curl request.

The only place where m_content_length is set is in `State::SetTransferParameters(...)`  which is called by runCurlWithStreams() called by PULL transfers.

Maybe I missed something though ;)

Thanks

PR #2476, opened 2025-04-10
[XrdTpc] Remove use of newer bash constructs in test

bbockelm:
Mac OS X's bash does not support associative arrays.  By going to more primitive indexed arrays (which are admittedly less pretty), we can enable the TPC test on Mac OS X, broadening the overall test coverage.


bbockelm:
@amadio - while this functions just fine on my laptop, it looks like the SciTokens dep isn't built on Mac OS X, meaning this doesn't result in improved CI coverage.

PR #2477, opened 2025-04-10
[XrdPfc] Add simple integration test for the PFC

bbockelm:
This adds a simple integration test that downloads a file from a XCache server through another fixture as an origin, then verifies the file is unchanged.

@amadio - the approach is as what we discussed earlier today ... added explicit dependency between fixtures but did not rework how the CMakeLists.txt is working.

### DIFF
--- a/tests/XRootD/cache.cfg
+++ b/tests/XRootD/cache.cfg
@@ -1,0 +1,24 @@
+set name = cache
+set port = 7096
+
+set src = $SOURCE_DIR
+set pwd = $PWD
+
+xrootd.trace all
+pfc.trace info
+
+xrootd.seclib libXrdSec.so
+xrd.protocol XrdHttp:$port libXrdHttp.so
+
+http.desthttps false
+http.selfhttps2http false
+
+xrootd.chksum max 2 md5 adler32 crc32 crc32c
+
+ofs.osslib libXrdPss.so
+pss.cachelib libXrdPfc.so
+
+pss.setopt DebugLevel 4
+pss.origin localhost:5094
+
+continue $src/common.cfg
--- a/tests/XRootD/cache.sh
+++ b/tests/XRootD/cache.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+export XrdSecPROTOCOL=host
+
+function setup_cache() {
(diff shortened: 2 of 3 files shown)


tests/XRootD/cache.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+export XrdSecPROTOCOL=host
+
+function setup_cache() {
+	require_commands diff
+	mkdir "${NAME}/xrootd/cache"
```
amadio, line 7: You don't need the mkdir command, it's already done here:

https://github.com/xrootd/xrootd/blob/6579e066a25401ae6e6e3c235d494ef7884fb173/tests/XRootD/test.sh#L119

tests/XRootD/cache.cfg
```diff
@@ -0,0 +1,25 @@
...
+xrootd.seclib libXrdSec.so
+xrd.protocol XrdHttp:$port libXrdHttp.so
+
+http.desthttps false
+http.selfhttps2http false
+
+xrootd.chksum max 2 md5 adler32 crc32 crc32c
+
+ofs.osslib libXrdPss.so
+#ofs.ckslib * libXrdPss.so
```
amadio, line 19: Why add a commented out line?

bbockelm: I can drop it for now.  I had wanted to enable checksums but the server was failing to start -- we can keep it simpler and revisit later why it's crashing.

tests/XRootD/cache.sh
bbockelm: Agreed.  Removed in the latest commit.

amadio:
There seems to be a race condition between the host and cache test setups. I will fix.

bbockelm:
@amadio - indeed, it looks like the setup.sh script does not verify that XRootD is functioning and listening on a port before proceeding.  I'm surprised this doesn't lead to other failures?

This is the (ugly) shell code I use to force my unit tests to pause until XRootD is ready:
https://github.com/PelicanPlatform/xrootd-s3-http/blob/main/test/s3-setup.sh#L302-L322

If you know of better ways, I'll be quite interested!

PR #2482, opened 2025-04-12
[XrdPosix] Map operation timeouts to ETIME

bbockelm:
Elsewhere, we use ETIME to indicate a server timeout (particularly, XRootD uses it for "504 Gateway Timeout").  Many users have found it confusing that timeouts result in "stale NFS file handle". Beyond giving the correct HTTP status code, this will cause the following message to be provided:

```
sTREAM ioctl timeout
```

It's not the greatest -- and the capitalization is strange -- but at least it includes the word "timeout".  For the xroot protocol, one receives 3035, `kXR_TimerExpired`.


amadio:
Looks good to me, let's hear from @abh3 before merging, though, since you asked for his review as well.

bbockelm:
Yup!  Mostly wrote this down as a discussion point... if we agree it's an improvement, the patch itself is obviously trivial.

bbockelm:
@abh3 - this one is still waiting for your input.  I understand @amadio would like to do a patch release soon and this would be a good opportunity.

PR #2485, opened 2025-04-14
Increase DISCOVERY_TIMEOUT

eabdullin1:
While building xrootd for AlmaLinux Kitten 10 (x86_64_v2), I faced with a timeout issue during tests:
```
CMake Error at /usr/share/cmake/Modules/GoogleTestAddTests.cmake:112 (message):
  Error running test executable.

    Path: '/builddir/build/BUILD/xrootd-5.8.0/redhat-linux-build/tests/XrdCl/XrdClFileTest.cc-tests'
    Working directory: '/builddir/build/BUILD/xrootd-5.8.0/redhat-linux-build/tests/XrdCl'
    Result: Process terminated due to timeout
    Output:
      

Call Stack (most recent call first):
  /usr/share/cmake/Modules/GoogleTestAddTests.cmake:226 (gtest_discover_tests_impl)
```
To fix this, I increased the test discovery timeout from the default 5 seconds to 10 seconds by setting the DISCOVERY_TIMEOUT property. This change resolves the issue and allows me to build it on AlmaLinux x86_64_v2.


amadio:
Is this really needed in all of the tests? Most tests have a short list, only the XrdCl tests have many subtests in the same binary. Could you please check which ones actually would fail and increase only those? Thank you!

eabdullin1:
> Is this really needed in all of the tests? Most tests have a short list, only the XrdCl tests have many subtests in the same binary. Could you please check which ones actually would fail and increase only those? Thank you!

I tried increasing the timeout only for **XrdCl** tests, but unfortunately, it didn’t help. Each next test failed with the same timeout issue until I increased the timeout for all of them.

PR #2492, opened 2025-04-17
[XrdHttp/TPC] URL encoding/decoding handles '=' sign

### LINKED ISSUE #2491: [XrdHttp] URL encoding/decoding not symmetrical
The following unit test fails:
```cpp
  {
    char input [] = "zteos64%3AMDAF5PGJ4Wa12g%3D";
    std::string unquoted(unquote(input));
    std::string quoted(quote(unquoted.c_str()));
    EXPECT_EQ(input,quoted);
  }
```

It outputs:
```
Expected equality of these values:
  input
    Which is: "zteos64%3AMDAF5PGJ4Wa12g%3D"
  quoted
    Which is: "zteos64%3AMDAF5PGJ4Wa12g="
[  FAILED  ] XrdHttpTests.urlUnquoteQuoteTest (0 ms)

```

The decoding the string and encoding the decoded string should give the original string.
I'll try to use curl_easy_escape and curl_easy_unescape

ccaffy:
Fixes #2491 

Please do not merge as it is not tested fully yet. But first-sight comments are welcome!

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -585,14 +585,12 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
     redirdest += buf;
   }
 
-  redirdest += quote(resource.c_str());
+  redirdest += encode_str(resource.c_str()).c_str();
   
   // Here we put back the opaque info, if any
   if (vardata) {
-    char *newvardata = quote(vardata);
     redirdest += "?&";
-    redirdest += newvardata;
-    free(newvardata);
+    redirdest += encode_opaque(vardata).c_str();
   }
   
   // Shall we put also the opaque data of the request? Maybe not
@@ -643,24 +641,14 @@ void XrdHttpReq::appendOpaque(XrdOucString &s, XrdSecEntity *secent, char *hash,
   // this works in most cases, except if the url already contains the xrdhttp tokens
   s = s + "?";
   if (!hdr2cgistr.empty()) {
-    char *s1 = quote(hdr2cgistr.c_str());
-    if (s1) {
-        s += s1;
-        free(s1);
-    }
+    s += encode_opaque(hdr2cgistr).c_str();
   }
   if (p && (l > 1)) {
-    char *s1 = quote(p+1);
-    if (s1) {
-      if (!hdr2cgistr.empty()) {
-        s = s + "&";
-      }
-      s = s + s1;
-      free(s1);
+    if (!hdr2cgistr.empty()) {
+      s = s + "&";
     }
+    s = s + encode_opaque(p + 1).c_str();
   }
-
-
 
   if (hash) {
     if (l > 1) s += "&";
@@ -675,76 +663,45 @@ void XrdHttpReq::appendOpaque(XrdOucString &s, XrdSecEntity *secent, char *hash,
     if (secent) {
       if (secent->name) {
         s += "&xrdhttpname=";
-        char *s1 = quote(secent->name);
-        if (s1) {
-          s += s1;
-          free(s1);
+        s += encode_str(secent->name).c_str();
         }
       }
 
       if (secent->vorg) {
         s += "&xrdhttpvorg=";
-        char *s1 = quote(secent->vorg);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->vorg).c_str();
       }
 
       if (secent->host) {
         s += "&xrdhttphost=";
-        char *s1 = quote(secent->host);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->host).c_str();
       }
       
       if (secent->moninfo) {
         s += "&xrdhttpdn=";
-        char *s1 = quote(secent->moninfo);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->moninfo).c_str();
       }
 
       if (secent->role) {
         s += "&xrdhttprole=";
-        char *s1 = quote(secent->role);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->role).c_str();
       }
       
       if (secent->grps) {
         s += "&xrdhttpgrps=";
-        char *s1 = quote(secent->grps);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->grps).c_str();
       }
       
       if (secent->endorsements) {
         s += "&xrdhttpendorsements=";
-        char *s1 = quote(secent->endorsements);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(secent->endorsements).c_str();
       }
       
       if (secent->credslen) {
         s += "&xrdhttpcredslen=";
         char buf[16];
         sprintf(buf, "%d", secent->credslen);
-        char *s1 = quote(buf);
-        if (s1) {
-          s += s1;
-          free(s1);
-        }
+        s += encode_str(buf).c_str();
       }
       
       if (secent->credslen) {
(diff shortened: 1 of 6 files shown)


ccaffy:
Funnily enough, adding the `=` sign in the encoding function makes the authentication not working.

Provided that a server has `http.header2cgi Authorization authz` configuration, the resulting CGI created out of a `curl -H 'Authorization: Bearer token'` will result in `authz%3DBearer%20eyJhbGciOiJFUzI1NiI`. This CGI had the `=` sign replaced!

This is because we have:
```cpp
    std::string hdr2cgistrEncoded = encode_str<std::string>(hdr2cgistr.c_str());
    resourceplusopaque.append(hdr2cgistrEncoded.c_str());
```

One should only encode what is after the `=` sign!

Nice catch for the CI tests with authentication!

ccaffy:
Actually, this full code does not work anymore when adding the `=` sign to the encoding:
```cpp
void XrdHttpReq::appendOpaque(XrdOucString &s, XrdSecEntity *secent, char *hash, time_t tnow) {

  int l = 0;
  char * p = 0;
  if (opaque)
    p = opaque->Env(l);

  if (hdr2cgistr.empty() && (l < 2) && !hash) return;

  // this works in most cases, except if the url already contains the xrdhttp tokens
  s = s + "?";
  if (!hdr2cgistr.empty()) {
    s += encode_str<std::string>(hdr2cgistr.c_str()).c_str();
  }
  if (p && (l > 1)) {
    if (!hdr2cgistr.empty()) {
      s = s + "&";
    }
    s = s + encode_str<std::string>(p + 1).c_str();
  }
[...]
```

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -585,14 +585,12 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
     redirdest += buf;
   }
 
-  redirdest += quote(resource.c_str());
+  redirdest += encode_str<std::string>(resource.c_str()).c_str();
```
amadio, line 588: This looks fishy, `encode_str<std::string>`, but with `resource.c_str()`, which is a `char*`, then `.c_str()` at the end again to get from `std::string` to `char*`. Maybe some optimization is possible here?

ccaffy: It's because redirdest is a XrdOucString, it allows me not to have to free the raw pointer afterwards...

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -858,11 +805,10 @@ void XrdHttpReq::parseResource(char *res) {
   resourceplusopaque = resource;
   // Whatever comes after is opaque data to be parsed
   if (strlen(p) > 1) {
-    buf = unquote(p + 1);
+    const char * buf = decode_str(p + 1).c_str();
```
amadio, line 808: There's a [warning](https://my.cdash.org/viewBuildError.php?type=1&buildid=2913086) here that needs to be fixed. It's a dangling pointer to a temporary, which could cause crashes at runtime.

src/XrdHttp/XrdHttpUtils.hh
```diff
@@ -79,15 +83,97 @@ bool Fromhexdigest(const unsigned char *input, int length, unsigned char *out);
...
+ * !!! IT MUST BE FREED AFTER USAGE USING free(...) !!!
+ */
+inline char * decode_raw(const char * str) {
+  if(!str) {
+    return nullptr;
+  }
+  size_t strLength = strlen(str);
+  // uniquely own the temporary copy
+  std::unique_ptr<char[]> buf(new char[strLength + 1]);
+  std::memcpy(buf.get(), str, strLength + 1);
```
amadio, line 106: I wonder why you preferred not to use a `std::string` here.

src/XrdHttp/XrdHttpUtils.hh
```diff
@@ -79,15 +83,97 @@ bool Fromhexdigest(const unsigned char *input, int length, unsigned char *out);
...
+  return quote(str);
+}
+
+/**
+ * Encodes the URL passed in parameter (converts all letters consider illegal in URLs to their
+ * %XX versions).
+ * Calls quote()
+ * Returns a std::string containing the encoded string
+ */
+inline std::string encode_str(const char * str) {
```
amadio, line 130: I'd take a reference to a string here as input, to avoid having to deal with `nullptr`. Same for the other functions.

src/XrdHttp/XrdHttpUtils.hh
ccaffy: It's because I cannot (look at the `unquote()` function, it does modify the char * passed in parameter).

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -858,11 +805,10 @@ void XrdHttpReq::parseResource(char *res) {
   resourceplusopaque = resource;
   // Whatever comes after is opaque data to be parsed
   if (strlen(p) > 1) {
-    buf = unquote(p + 1);
-    opaque = new XrdOucEnv(buf);
+    std::string decoded = decode_str(p + 1);
+    opaque = new XrdOucEnv(decoded.c_str());
     resourceplusopaque.append('?');
     resourceplusopaque.append(p + 1);
```
amadio, line 811: Is this correct? Shouldn't `opaque` be used to append?

ccaffy: Nice catch!

ccaffy: NOPE :D

ccaffy: The resourceplusopaque string should contain the encoded string (I checked out master and did the following breakpoint):
```
(gdb) n
862	    opaque = new XrdOucEnv(buf);
(gdb) p buf
$4 = 0x7fffe8012e10 "test=test["
(gdb) n
863	    resourceplusopaque.append('?');
(gdb) n
250507 10:56:04 2371660 daemon.2451780:28@ccaffy-eosdev-al9 Xrootd_Protocol: f702 request timeout; read 0 of 24 bytes
250507 10:56:04 2371660 Xrd_Poll: Poller 0 enabled daemon.2451780:28@ccaffy-eosdev-al9
864	    resourceplusopaque.append(p + 1);
(gdb) n
865	    free(buf);
(gdb) p resourceplusopaque
$5 = {_vptr.XrdOucString = 0x41cc10 <vtable for XrdOucString+16>, str = 0x7fffe8004f20 "/tmp/fic.txt_dest?test=test%5B", len = 30, siz = 31, 
  static blksize = -1}
(gdb) 
```

src/XrdHttp/XrdHttpUtils.hh
ccaffy: I'm modifying it but this means that the user should check if the char * passed is nullptr or not (one cannot construct a string with a nullptr).

ccaffy: It's already the case in the code so I modified it!

ccaffy:
> Hi @ccaffy, @abh3 says we should test the encoding/decoding of `%%` vs `;`. I hope we can do that tomorrow, then we can add this to the release.

- The `%%` does not work. It basically removes what is after: `authz%3Dzteos64%%test` becomes `authz=zteos64`.
- ';' is not handled by the `quote()` code.

So either I add those cases, or we wait 5.9 in which I plan to change quote and unquote with libcurl as I initially did ;)

tests/XrdHttpTests/XrdHttpTests.cc
```diff
@@ -574,3 +574,78 @@ TEST(XrdHttpTests, xrdHttpReadRangeHandlerMultiChunksTwoRanges) {
     ASSERT_EQ(false, static_cast<bool>(error));
   }
 }
+
+TEST(XrdHttpTests, strDecodeEncodeTest) {
+  std::string input = "zteos64%3AMDAF5PGJ4Wa12g%3D";
+  {
+    std::string decoded(decode_str(input.c_str()));
+    std::string encoded(encode_str(decoded.c_str()));
+    ASSERT_EQ(input, encoded);
```
amadio, line 583: How about having the long list of encoded/decoded strings, and using a simple loop over all cases in the tests?

```cpp
const std::pair<std::string, std::string> input_pairs = { ... };

for(auto [encoded, decoded] : input_pairs) {
    ASSERT_EQ(decode_str(encoded), decoded);
    ASSERT_EQ(encoded, encode_str(decoded));
    ASSERT_EQ(encode_str(decode_str(encoded)), encoded);
    ASSERT_EQ(decode_str(encode_str(decoded)), decoded);
}
```

ccaffy: Yes I did it but did not push yet ;)

Issue #2261, opened 2024-05-15
WLCG DOMA, WebDAV Error Message Improvement Project


ccaffy:
We should try as much as possible to align the HTTP TPC error messages with what's recommended by DOMA: https://twiki.cern.ch/twiki/bin/view/LCG/WebdavErrorImprovement

One example among many others: perform a HTTP TPC PULL where the source is offline gives the following:

```
< HTTP/1.1 201 Created
< Connection: Keep-Alive
< Server: XrootD/v5.5.5-361-g1c80707
< Content-Type: text/plain
< Transfer-Encoding: chunked
< 
Perf Marker
Timestamp: 1715757277
Stripe Index: 0
Stripe Bytes Transferred: 0
Total Stripe Count: 1
End
* Connection #0 to host xrootd.cern.ch.cern.ch left intact
failure: Internal transfer failure, local=/tmp/BIGFILE_5GB, remote=https://xrootd2.cern.ch:2001/tmp/BIGFILE_5GB_COPY, HTTP library failure=Couldn't connect to server
```

The document says it should return 404 as the source is unreachable.

PR #2496, opened 2025-04-25
Add webdav error codes for file path and permission errors

dynamic-entropy:
Related Issue: #2261 

This PR introduces a new `XrdHttpError` class to standardise HTTP error handling and integrates it into the HTTP request processing workflow.

* Add the `XrdHttpError` class in `src/XrdHttp/XrdHttpError.hh` to encapsulate HTTP error responses. 
* Build on top of the `XrdOssTests` module to simulate filesystem errors (e.g., no space, quota exceeded) for testing purposes.
* Expanded the HTTP test suite (`tests/XRootD/http.sh`) with new test cases to handle errors for destination file/path category as per the webdav error improvement document.
https://twiki.cern.ch/twiki/bin/view/LCG/WebdavErrorImprovement


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2369,7 +2370,11 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
 
 
         // If we are here it's too late to send a proper error message...
-        if (xrdresp == kXR_error) return -1;
+        if (xrdresp == kXR_error) {
+          XrdHttpError errResp = XrdHttpError::Create(xrdresp, xrderrcode, XrdHttpReq::rtPUT, (XRequestTypes)ntohs(xrdreq.header.requestid), etext, NULL, NULL);
+          errResp.Send(prot, keepalive);
```
amadio, line 2375: Unused return value. It should probably be checked (I know it's not checked before, but if the return value doesn't matter, we can make it `void` instead).

src/XrdHttp/XrdHttpError.hh
```diff
@@ -0,0 +1,79 @@
...
+                    }
+                    else {
+                        code = 500;
+                        errCode = "Unknown";
+                    }
+                    break;
+                default:
+                    code = 500;
+                    errCode = "Internal Server Error";
+                    break;
```
amadio, line 69: This is confusing. Why have two distinct cases for 500? Can we let all 500 use `"Internal Server Error"`?

src/XrdHttp/XrdHttpError.hh
```diff
@@ -0,0 +1,79 @@
...
+            }
+            statusText = formatStatusText(errCode, etext);
+            return XrdHttpError(code, errCode, statusText, desc, header_to_add);
+        }
+
+        int Send(XrdHttpProtocol *prot, bool keepalive){
+            return prot->SendSimpleResp(httpStatusCode, desc, header_to_add,
+                httpStatusText.c_str(), httpStatusText.length(), keepalive);
```
amadio, line 77: The usage of this class seems to be always to create the error, call this, then destroy the object. I think the idea of just using a function for the combined operation taking all parameters and returning in one go might be cleaner.

amadio (changes requested):
This is a good start, but I think that a class is not necessary, we can have a table with all the errors in it, a function to index into this table to get the actual error properties, and another function to format the string and call `prot->SendSimpleResp(...)`. We should also check the return value of `SendSimpleResp` or consider making it void if it really doesn't matter.

@abh3 and @ccaffy may have different opinions, so let's hear from them as well.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2369,7 +2370,11 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
 
 
         // If we are here it's too late to send a proper error message...
-        if (xrdresp == kXR_error) return -1;
```
bbockelm, line 2372: We can't do this, unfortunately.

Per the comment above, we have already started the response and HTTP rules are that you can't send a second response to the same request.  This change breaks the HTTP protocol framing.

For `GET`, we have a small extension where the client can opt-in to a trailer (which is part of the HTTP protocol though rarely used) that indicates whether an error occurred while sending the `GET` response.  We could do a similar thing for `PUT` -- however, for now, we should keep the behavior of just abruptly closing the socket and save the refactor of the `PUT` response for later.

tests/XRootD/http.sh
```diff
@@ -189,4 +189,62 @@ function test_http() {
   receivedHeader=$(grep -i 'Test:' "$outputFilePath")
   assert_eq "1" "$(echo "$receivedHeader" | wc -l | sed 's/^ *//')" "Incorrect number of 'Test' header values"
   assert_eq "$expectedHeader" "$receivedHeader" "HEAD is missing statically-defined Test header"
+
+  set -ex
```
bbockelm, line 193: Let's avoid setting `e` here at least; we should explicitly check for failures where needed (`e` can have bad interactions with traps).

`x` I'm indifferent on but likely should be cleaned up to have the unit test be less verbose.

tests/XRootD/http.sh
```diff
@@ -189,4 +189,62 @@ function test_http() {
...
+    local http_code
+
+    # Run the curl command, capture HTTP code and body
+    http_code=$(curl -s -L -w "%{http_code}" -o "$body_file" "$@")
+    local body
+    body=$(< "$body_file")
+    rm -f "$body_file"
+
+    # Print body to stderr for debug visibility
+    echo "$body" >&2
```
bbockelm, line 211: Delay printing out the body until an error has occurred.  That'll keep the tests less chatty for successes.

tests/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ if(NOT ENABLE_SERVER_TESTS)
   return()
 endif()
 
+add_subdirectory( XrdOssTests )
```
bbockelm, line 31: @amadio - do we have to worry about race conditions between the build and the running of the tests?

Here, `XRootD::http::test` will require the built library.  I assume that `ctest` is guaranteed to only run after a full build so we don't need any explicit dependency in CMake.

Is that assumption correct?

tests/XRootD/http.sh
amadio: I agree, let's try not to make tests too verbose. A clear message on failure is more than enough. Tests that pass don't need to create any output.

tests/CMakeLists.txt
amadio: Yes, `ctest` runs only after everything is already built.

abh3:
I must admit this comment got buried on all of the traffic. First, I am a firm believer that the simplest solution is usually the best one. In this particular case, I can only fine a quibble of whether to use a table or a map. If the number of error codes is small and will not substantially grow, then a table is fine. Otherwise, I would be in favor of a map. I would also be in favor of a map if the growth of error codes is ill-defined and may actually substantially increase given the current trends. I also would add that if a map leads to a simpler implementation, then I would be in favor of a map. I think the problem here is that map and table initialization or not exactly the same. One occurs at compile time and the other at run time, though it's a trivial difference especially since it occurs only once. All things being equal, I'd go with the map.

Anyway, those are my thoughts.

Andy

amadio:
@dynamic-entropy Could you please rebase this on top of the current master branch? I will prepare a 5.8.3 release soon (waiting for some fixes that need to go into EOS), and I'd like to include the changes from this pull request. Thank you.

amadio (commented):
After looking more thoroughly, maybe parts of this are better for 5.9 (like adding the new `WebdavErrorMessage` function). However, the other changes would be ok for 5.8.3.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1631,19 +1643,17 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc, const char *hea
...
-    ss << header_to_add << crlf;
+  if (header_to_add && (header_to_add[0] != '\0')) ss << header_to_add << crlf;
 
   ss << crlf;
 
   const std::string &outhdr = ss.str();
   TRACEI(RSP, "Sending resp: " << code << " header len:" << outhdr.size());
-  if (SendData(outhdr.c_str(), outhdr.size()))
-    return -1;
+  if (SendData(outhdr.c_str(), outhdr.size())) return -1;
```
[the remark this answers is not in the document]
dynamic-entropy: Sure!

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -861,46 +861,46 @@ void XrdHttpReq::sendWebdavErrorMessage(
...
-
-    prot->SendSimpleResp(httpStatusCode, desc, header_to_add,
-                         httpStatusText.c_str(), httpStatusText.length(),
-                         keepalive);
+    httpErrorBody = "ERROR: " + errCode + ": " + etext + "\n";
   }
+
+  prot->SendSimpleResp(httpStatusCode, desc, header_to_add,
+                       httpErrorBody.c_str(), httpErrorBody.length(),
+                       keepalive);
```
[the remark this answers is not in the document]
dynamic-entropy: Seems like an oversign when rebasing. Will fix it. Thanks for cathing it.

dynamic-entropy (commented):
Made the requested changes. Thanks for taking a look.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2369,7 +2370,11 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
 
 
         // If we are here it's too late to send a proper error message...
-        if (xrdresp == kXR_error) return -1;
+        if (xrdresp == kXR_error) {
+          XrdHttpError errResp = XrdHttpError::Create(xrdresp, xrderrcode, XrdHttpReq::rtPUT, (XRequestTypes)ntohs(xrdreq.header.requestid), etext, NULL, NULL);
```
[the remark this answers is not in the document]
dynamic-entropy: Changed back from the use of the factory pattern.

PR #2506, opened 2025-05-14
Correct concurrency behavior around bridge `Redrive`.

bbockelm:


src/XrdXrootd/XrdXrootdTransit.cc
```diff
@@ -477,11 +464,7 @@ void XrdXrootdTransit::Redrive()
           TRACEP(REQ, "Bridge redrive callback rc="<<rc
                       <<" runStatus="<<runStatus);
         }
-        if (runStatus)
-           {AtomicBeg(runMutex);
-            AtomicZAP(runStatus);
-            AtomicEnd(runMutex);
-           }
+        runStatus.store(0, std::memory_order_release);
```
abh3, line 467: I am really reary about clearing runStatus here. This change was introduced by commit 0588dca
 and I understand what the comments say it wants to fix. However, the logic here does not follow the logic in Process() where runStatus is only cleared if not in runWait. Here it can leave runStatus in an inconsistent state. I still am trying to figure out the implication of that.

abh3 (changes requested):
Agreed, that runStatus should be an acquire/release atomic. However, while it uses acquire/release semantics operationally, it uses memory_order_seq_cst in all debugging statements which can be a significant change by merely turning on debugging. So, please use memory_order_acquire when it's referenced in a debug statement for consistency.

It's also likely that runWait should be a relaxed atomic. So, it might as well be declared an XrdSys::RAtomic<int> which does not require any changes in usage (i.e. memory ordering is automatically applied to all references).

Finally, we have an outstanding issue about clearing runStatus in Redrive() in a way that may cause problems. But let's cross that bridge after we see whether these changes actually fix the problem.

src/XrdXrootd/XrdXrootdTransit.cc
bbockelm: Happy to make the change but I wanted to point out that this statement, as written, doesn't appear to be true:

> Process() where runStatus is only cleared if not in runWait

From `Process()`:

```
                rc = (Resume ? XrdXrootdProtocol::Process(lp) : Process2());
                if (rc >= 0)
                   {if (runWait) rc = -EINPROGRESS;
                    if (!runDone) return rc;
                    runStatus.store(0, std::memory_order_release);
                   }
```

So `runStatus` is set to 0 if `rc >= 0` and `runDone` is true.  Otherwise, it is not changed.

Now, in this line of code, `runStatus` is unconditionally cleared (regardless of the value of `runDone` or `rc`).  I do not understand the differing logic so I'll defer to you here.

abh3:
I believe that @bbockelm wants to run with the patch fr a while to see if any crashes occur before signing off on this. I figure no crashes so far but that may also be due to a fix in Xcache that was creating a situation that prompted the crash, so maybe we no longer have an environment that will settle the question. Brian?

bbockelm:
Hit a few bumps on the rollout completely unrelated to XRootD.  So, need to wait until tomorrow to get this back into play.

Unfortunately, most of the crashes today ended deep in the PFC code so I suspect this isn't the end of the bugs...

bbockelm:
This, along with https://github.com/xrootd/xrootd/commit/dd72712e67b57ed93b83425d354c5c6887cceb91, sat on the UW-Madison cache over the past 5 days and there have been zero crashes.  Given that, previously, there were 0-3 crashes / day, I think this has made a positive improvement and I'm marking it as no-longer-draft.

amadio:
@bbockelm I pushed @abh3's fix to master, now merging this on top and soon will release 5.8.3 with both included. Thank you!

abh3:
@bbockelm Just for completeness here. Did this fix stop crashes after we fixed the previous problem?

bbockelm:
Yes -- I've not seen any more crashes from this lately.  I think we're good!

PR #2509, opened 2025-05-15
Drop unnecessary change of PATH

ellert:
The test binaries are in ${CMAKE_BINARY_DIR}/bin which is already in PATH.


amadio:
Indeed, the path if the binaries was changed and this is not needed anymore.

PR #2365, opened 2024-10-29
[Cl] Allow out-of-order message-sent confirmation and reply reception

smithdh:
This is needed in order to allow a planned change in the way substreams are assigned to network pollers.

PR #2516, opened 2025-05-20
[Cl] Reset state when resending a message. Fixes: 3a957083

smithdh:
Fixes lost response when an error occurs before the resend of a requeued request. e.g. a requeued message can happen when the client has been redirected or told to wait and resend a request later by kxr_wait. An error before resend could be that the stream connect is lost or the request times out before the resend. Introduced in #2365.

### DIFF
--- a/src/XrdCl/XrdClXRootDMsgHandler.cc
+++ b/src/XrdCl/XrdClXRootDMsgHandler.cc
@@ -914,7 +914,9 @@ namespace XrdCl
 
   //----------------------------------------------------------------------------
   // We're here when we requested sending something over the wire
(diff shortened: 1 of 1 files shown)


amadio:
@smithdh Normally, I add a `Fixes: commithash` when a commit fixes another one, maybe you could add it too in your commit message (same for the description in the pull request, it would be nice to add it to the commit). Example: b0e427bf13100c1a00ad4f0b3181b0c1831b78be. Thank you!

amadio:
@smithdh :+1: Thanks for the updated commit message. Please put out of draft when you feel it's ready.

Issue #2517, opened 2025-05-21
[XrdPosix] Race conditions in callback handler triggers segfault


bbockelm:
We had about a dozen segfaults overnight with most of them ending in the following trace back:

```
#0  0x00007faca74d953c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007faca748c686 in raise () from /lib64/libc.so.6
#2  0x00007faca7476833 in abort () from /lib64/libc.so.6
#3  0x00007faca7477170 in __libc_message.cold () from /lib64/libc.so.6
#4  0x00007faca74e35e7 in malloc_printerr () from /lib64/libc.so.6
#5  0x00007faca74e4e23 in _int_free () from /lib64/libc.so.6
#6  0x00007faca74e7925 in free () from /lib64/libc.so.6
#7  0x00007faca7894d0f in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_assign(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /lib64/libstdc++.so.6
#8  0x00007fabc54709b1 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign (__str="Range Not Satisfiable\r (read operation at offset 655360)", this=0x7fa3cc002f70)
    at /usr/include/c++/11/bits/basic_string.h:1387
#9  std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator= (__str="Range Not Satisfiable\r (read operation at offset 655360)", this=0x7fa3cc002f70)
    at /usr/include/c++/11/bits/basic_string.h:681
#10 XrdOucECMsg::operator= (rhs="Range Not Satisfiable\r (read operation at offset 655360)", this=0x7fa3cc002f68) at /usr/src/debug/xrootd-5.8.2-1.3.osg24.el9.x86_64/xrootd-5.8.2/src/XrdOuc/XrdOucECMsg.hh:160
#11 XrdPosixMap::Result (Status=..., ecMsg=..., retneg1=retneg1@entry=false) at /usr/src/debug/xrootd-5.8.2-1.3.osg24.el9.x86_64/xrootd-5.8.2/src/XrdPosix/XrdPosixMap.cc:200
#12 0x00007fabc5470082 in XrdPosixFileRH::HandleResponse (this=0x7fa928010e70, status=0x7fab981b1670, response=0x0)
[...]

PR #2518, opened 2025-05-21
Fix race conditions around use of `ecMsg`

bbockelm:
There is one `ecMsg` object per file while there can be an arbitrary number of outstanding operations per file (and hence, multiple threads writing to the object's memory location concurrently).  We've observed crashes caused by two simultaneous overwrites leading to a double-delete of the underlying object.

This commit takes a fairly simple approach: wrap all read and writes of `ecMsg` with a new mutex to prevent race conditions.

@abh3 - this is one approach for fixing https://github.com/xrootd/xrootd/issues/2517.  I don't find it particularly elegant - it steps on a lot of code in a lot of places - but I wanted to get something out to the ops team as there were hundreds of segfaults across the infrastructure by the end of the day.

### DIFF
--- a/src/XrdFfs/XrdFfsMisc.cc
+++ b/src/XrdFfs/XrdFfsMisc.cc
@@ -77,7 +77,7 @@ char XrdFfsMisc_get_current_url(const char *oldurl, char *newurl)
     }
 
     XrdOucECMsg   ecMsg;
-    XrdPosixAdmin adm(oldurl,ecMsg);
+    XrdPosixAdmin adm(oldurl,ecMsg,nullptr);
     if (adm.isOK() && adm.Stat())
     {
 // We might have been redirected to a destination server. Better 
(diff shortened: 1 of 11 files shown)


abh3:
Well, I went in a different direction to minimize the number of changes and avoid future problems. The fix I pushed makes sure that uses of ECMsg are serialized. It should have the same effect as your patch except that it can't prevent messages from trouncing other messages. We can prevent that by defauling the behavior to always append, though that won't necessarily prevent interleaving messages unless you use one of the multi-line API's. Anyway, the patch I pushed affects only three files. It also assumes that destructors are the last thing to be done when returning from a method.

abh3:
Oh yes, one more thing @bbockelm - if you could test this in one of your Xcaches that keep crashing that would be super good. It seems that Pelican is the one that drives the cache to rather extreme conditions, right?

amadio:
@bbockelm Please let me know if the commit fixes the crashes in Pelican. I already rebased it onto master and will push updated master/devel branches with this change once I receive confirmation that the crashes are fixed. Cheers,

PR #2525, opened 2025-06-03
Tpc worker pool

### LINKED ISSUE #2401: [XrdTpc] Refactor to TPC code to utilize curl multi-handle objects
Currently, the XrdTpc code creates all the interactions with libcurl from scratch for each individual transfer (see [`TPCHandler::RunCurlWithUpdates`](https://github.com/xrootd/xrootd/blob/master/src/XrdTpc/XrdTpcTPC.cc#L572)).  Things were originally designed like this to be immensely conservative -- if nothing is shared between any of the transfers, then there's less opportunity for bad state to propagate between multiple transfer attempts.

However, after about 8 years experience with the code, I think we can be a bit more confident. A drawback of the current design is the interaction with the remote HTTP server is torn down at the end of `TPCHandler::RunCurlWithUpdates` (the TLS connection is shut down, TCP sockets are closed, and the DNS resolution cache is cleared) which has obvious performance implications.  The TCP windows are restarted for each transfer and there are multiple RTTs of startup time for each file.

A more modern design is would be something like we do in [xrdcl-pelican](https://github.com/PelicanPlatform/xrdcl-pelican).  Each operation is captured as a request object that is entered into a [queue](https://github.com/PelicanPlatform/xrdcl-pelican/blob/main/src/CurlUtil.hh#L162) and the calling thread waits on the request object to complete with a condition variable.  A separate thread pool of workers picks up the transfer from the queue and executes it in a given multi-handle (see [`CurlWorker::Run`](https://github.com/PelicanPlatform/xrdcl-pelican/blob/main/src/CurlUtil.cc#L778)), invoking the appropriate thread upon completion.

While `xrdcl-pelican` (and the similar [`xrootd-s3-http`](https://github.com/PelicanPlatform/xrootd-s3-http)) code scales well with a single thread pool, there'd be an extra twist for XrdTpc: packet marking.  Packet marking relies on the ability to segregate transfers by TCP stream; this implies, for example, there needs to be multiple transfer queues based on a packet marking key (or, at minimum, the `org` field of the `XrdSecEntity`).  Since there's no longer a fixed cost of a single thread pool, we would also want each thread pool to launch on demand and shut down when idle.

dynamic-entropy:
Fix https://github.com/xrootd/xrootd/issues/2401


bbockelm:
@dynamic-entropy and I discussed this PR a bit over Zoom.  Beyond the code review itself, items we want to work on:
- Get this PR into a patch build and deploy at a site that actively does packet marking to look for regressions "at scale".
- Implement the `Cancel()` method and write a unit test to ensure it works.  (The unit test should start a COPY, disconnect, and look to ensure the transfer does not continue to completion).
- We don't recall whether a slow/stalled transfer would ever automatically shutdown (or if this is always triggered by FTS).  Test with and without this patch, by hand, to make sure there is no change in behavior (we might desire a change in behavior later but that should be a separate PR).
- Look at the performance markers and ensure the information is the same before & after this patch.  We don't think there's any unit test coverage for the performance marker contents.  Particularly, does getting the [remote connection list](https://github.com/xrootd/xrootd/blob/346858c2ce7813e20c7d10b17a4ac583469ba200/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L425) still work?

src/XrdHttpTpc/XrdHttpTpcPMarkManager.cc
```diff
@@ -32,7 +32,20 @@ PMarkManager::SocketInfo::SocketInfo(int fd, const struct sockaddr * sockP) {
...
+      mSciTag(req.mSciTag),
+      mResource(req.resource.c_str()),
+      mTransferWillStart(false),
+      mTpcType(tpcType) {}
+
+PMarkManager::PMarkManager(XrdNetPMark *pmark, int sciTag,
+                           const TPC::TpcType tpcType)
+    : mPmark(pmark),
+      mSciTag(sciTag),
+	  mResource(NULL),
```
bbockelm: Small nitpick - please align the whitespace consistently here.

src/XrdHttpTpc/XrdHttpTpcPool.hh
```diff
@@ -0,0 +1,164 @@
...
+       std::string GetRemoteConnDesc();
+       void SetActive();
+       void SetDone(int status, const std::string &msg);
+       bool IsActive() const; 
+       void Cancel();
+       void UpdateRemoteConnDesc();
+
+
+      private:
+        std::atomic<bool> m_active{false};
```
bbockelm, line 57: Please fix alignment issues.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -0,0 +1,479 @@
...
+    request.SetActive();
+
+    CURLcode res = static_cast<CURLcode>(-1);
+    int running_handles = 1;
+	const int update_interval{1};
+	time_t now = time(NULL);
+	time_t last_update = now - update_interval; // Inorder to always fetch on first pass 
+
+	auto fail_and_exit = [&](int code, const std::string& msg) -> bool {
+      curl_multi_remove_handle(multi_handle, curl);
```
bbockelm: Fix the inconsistent whitespace mapping here.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -0,0 +1,479 @@
...
+    // get the lock and remove itself from the map).
+    while (true) {
+        m_mutex.lock_shared();
+        std::lock_guard<std::shared_mutex> guard{m_mutex, std::adopt_lock};
+        auto iter = m_pool_map.find(handler.GetIdentifier());
+        if (iter != m_pool_map.end()) {
+            if (!iter->second->IsDone()){
+            	queue = iter->second;
+                break;
+			}
```
bbockelm: Another small whitespace issue.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -1004,7 +843,7 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
     State state(0, stream, curl, false, req.tpcForwardCreds);
     state.SetupHeaders(req);
     state.SetContentLength(sourceFileContentLength);
-
+
```
bbockelm: avoid unnecessary whitespace changes.

src/XrdHttpTpc/XrdHttpTpcTPC.hh
```diff
@@ -117,6 +112,7 @@ private:
     int SendPerfMarker(XrdHttpExtReq &req, TPCLogRecord &rec, TPC::State &state);
     int SendPerfMarker(XrdHttpExtReq &req, TPCLogRecord &rec, std::vector<State*> &state,
         off_t bytes_transferred);
+	int SendPerfMarker(XrdHttpExtReq &req, TPCLogRecord &rec, TPC::State &state, std::string desc);
```
bbockelm: Fix whitespace alignment.

bbockelm (commented):
Getting quite close!

- Please squash a few of the debugging commits.
- Review through the code and make sure no unnecessary developer debugging messages have snuck through.
- Some errant whitespace issues have crept up (lines not aligning, switching between tab & spaces).  Try setting your IDE to have tab stop be at 8 spaces to make these accidental changes more visible as you're fixing them.
- It's great that we can now limit the number of concurrent transfers!  Let's make these configurable, `tpc.max_active_transfers_per_queue` and `tpc.max_waiting_transfers_per_queue`?  Let's set the `max_active` to 50 by default and keep `max_waiting` where it is.
- For selection of queues, we should include the experiment name.  This way, if no scitag is enabled, we still partition between different experiments.

tests/TPCTests/test.sh
```diff
@@ -28,12 +24,12 @@ check_commands() {
 }
 
 function error() {
-	echo "error: $*" >&2; exit 1;
+    echo "error: $*" >&2; exit 1;
```
dynamic-entropy, line 27: I would like to keep these whitespace changes here; they are changes from the use of tabs to spaces for indentation. 
The rest of the file uses spaces, but for some lines, my editor seems to have used spaces. These are corrections to that.

dynamic-entropy (commented):
Squashed the debugging commits and cleaned some changes. Still kept some whitespace changes - they correct a previous mistake of mixing the use of tabs and spaces for indentation in the test bash scripts.

src/XrdHttpTpc/XrdHttpTpcConfigure.cc
```diff
@@ -96,6 +96,22 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
             } else {
                 m_first_timeout = 2*m_timeout;
             }
+        } else if (!strcmp("tpc.max_active_transfers_per_queue", val)) {
+            if (!(val = Config.GetWord())) {
+                Config.Close();
+                m_log.Emsg("Config", "tpc.max_active_transfers_per_queue value not specified.");
+                return false;
+            }
+            unsigned int max_active = std::stoi(val);
```
bbockelm: I think this can switch to `XrdOuca2x::a2i`.  It's undocumented but here's the signature:

```
int XrdOuca2x::a2i(XrdSysError &Eroute, const char *emsg, const char *item,
                                             int *val, int minv, int maxv)
```

The `minv` and `maxv` arguments are "minimum value" and "maximum value" which is likely useful here.  It also takes care of logging of error messages, which may shorten the code a bit.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -550,171 +501,68 @@ int TPCHandler::SendPerfMarker(XrdHttpExtReq &req, TPCLogRecord &rec, std::vecto
...
-        logTransferEvent(LogMask::Error, rec, "CURL_INIT_FAIL",
-            "Failed to initialize a libcurl multi-handle");
-        std::stringstream ss;
-        ss << "Failed to initialize internal server memory";
-        return req.SendSimpleResp(rec.status, NULL, NULL, generateClientErr(ss, rec).c_str(), 0);
-    }
 
-    //curl_easy_setopt(curl, CURLOPT_BUFFERSIZE, 128*1024);
+int TPCHandler::RunCurlWithUpdates(CURL *curl, XrdHttpExtReq &req, State &state, TPCLogRecord &rec) {
+    TPCRequestManager::TPCRequest request("tpc", req.mSciTag, curl);
```
bbockelm: You can extract the security entity from the `req` object via the `GetSecEntity()` method.  From there, the `vorg` attribute should contain the experiment name (note - may be `nullptr`).  You can use this to partition the queues via the experiment even in cases where scitags are not configured.

.gitignore
```diff
@@ -5,3 +5,5 @@ docker/xrootd.tar.gz
 # Python build artifacts
 dist/
 *.egg-info
+.clang-format
```
bbockelm: Let's avoid changing `.gitignore` in this PR.

src/XrdHttpTpc/XrdHttpTpcConfigure.cc
dynamic-entropy: Thankyou! Switched to `XrdOuca2x::a2i`. Though I still needed the temporary variables for type casting.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
dynamic-entropy: Done, the new queue identifier is of the form `tpc_<vorg>_<scitag>`. Both vorg and scitag may or may not be set.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -0,0 +1,444 @@
...
+
+CURL *TPCRequestManager::TPCRequest::GetHandle() const { return m_curl; }
+
+int TPCRequestManager::TPCRequest::GetScitag() const { return m_scitag; }
+
+bool TPCRequestManager::TPCRequest::IsActive() const { return m_active.load(std::memory_order_relaxed); }
+
+std::string TPCRequestManager::TPCRequest::GetLabel() const { return m_label; }
+
+std::string TPCRequestManager::TPCRequest::GenerateIdentifier(const char *label, char *vorg, int scitag) {
```
bbockelm:
```suggestion
std::string TPCRequestManager::TPCRequest::GenerateIdentifier(const std::string &label, const char *vorg, int scitag) {
```

- If we assume `label` is always present, then we should enforce that through the static typing mechanism (make it a reference).
- If we're not changing `vorg` at all, we should declare it `const`.

bbockelm:
FYI - this has been running at the CMS T2 in Nebraska for a little under 2 weeks, including a mini-DC.  No issues were reported by the Nebraska team.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -325,9 +323,19 @@ int TPCRequestManager::TPCRequest::GetScitag() const { return m_scitag; }
...
-    ss << m_ident << "_" << m_scitag;
+    ss << label;  // always present
+
+    if (vorg && *vorg) {
+        ss << "_" << vorg;
+    }
+
+    if (scitag != -1) {
+        ss << "_" << scitag;
+    }
```
amadio, line 338: I'm not a big fan of all the renaming in APIs and variable names. Unless there's a good reason to rename things, they should keep the current names. This ties the `TPCRequestManager` too strongly to worker pools defined by vorg/scitag, which may not always even be available.

I'd prefer to keep the `TPCRequestManager` oblivious to how we label them, to keep the possibility of using other ways to have worker pools than by vorg and scitag. The point at which we can customize is by passing a label to the constructor, so this whole commit should likely be dropped.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -18,8 +18,8 @@ decltype(TPCRequestManager::m_pool_map) TPCRequestManager::m_pool_map;
 decltype(TPCRequestManager::m_init_once) TPCRequestManager::m_init_once;
 decltype(TPCRequestManager::m_mutex) TPCRequestManager::m_mutex;
 decltype(TPCRequestManager::m_idle_timeout) TPCRequestManager::m_idle_timeout = std::chrono::minutes(1);
-unsigned TPCRequestManager::m_max_pending_ops = 20;
-unsigned TPCRequestManager::m_max_workers = 20;
+unsigned TPCRequestManager::m_max_pending_ops = 20;  // default max_pending_transfers_per_queue
+unsigned TPCRequestManager::m_max_workers = 50;      // default mac_active_transfers_per_queue
```
amadio, line 22: Let's keep the default with the same value as before, at 20, or if the default is to be changed, it should be acompanied by a reason why it's being changed.

src/XrdHttpTpc/XrdHttpTpcPool.cc
```diff
@@ -80,33 +91,19 @@ bool TPCRequestManager::TPCQueue::TPCWorker::RunCurl(CURLM *multi_handle, TPCReq
...
 
-        // Check for cancellation
         if (!request.IsActive()) {
-            curl_multi_remove_handle(multi_handle, curl);
-            std::stringstream ss;
-            ss << "Transfer cancelled";
-            m_queue.m_parent.m_log.Log(LogMask::Info, "TPCWorker", ss.str().c_str());
-            request.SetDone(499, ss.str());
-            return false;
+            return fail_and_exit(499, "Transfer cancelled");
```
amadio, line 96: This is not a standards conforming HTTP status code, it's a code specific to nginx, so we should probably stop using it here and return something else. Unfortunately, the specification is not clear about what we should do either: https://twiki.cern.ch/twiki/bin/view/LCG/HttpTpcTechnical#Transfer_Cancellation_and_Failur

src/XrdSys/XrdSysRAtomic.hh
```diff
@@ -37,9 +37,31 @@ T   operator=(T v) volatile noexcept
...
+        return *this;
+    }
+
+    RAtomic& operator=(const RAtomic& other) volatile noexcept {
+        if (this != &other) {
+            _m.store(other._m.load(std::memory_order_relaxed), std::memory_order_relaxed);
+        }
+        return *this;
+    }
+
```
amadio, line 64: I'm not sure that we want to add all these new operators here. Atomic operations are subtle, and we should keep things visible to avoid introducing bugs inadvertently.

@abh3 Can you please comment on your thoughts here?

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -407,11 +407,30 @@ int TPCHandler::GetContentLengthTPCPull(CURL *curl, XrdHttpExtReq &req, uint64_t
...
+    ss << "Timestamp: " << time(NULL) << crlf;
+    ss << "Stripe Index: 0" << crlf;
+    ss << "Stripe Bytes Transferred: " << state.BytesTransferred() << crlf;
+    ss << "Total Stripe Count: 1" << crlf;
+    if (!desc.empty()) ss << "RemoteConnections: " << desc << crlf;
+    ss << "End" << crlf;
+    rec.bytes_transferred = state.BytesTransferred();
+    logTransferEvent(LogMask::Debug, rec, "PERF_MARKER");
+    return req.ChunkResp(ss.str().c_str(), 0);
+}
```
amadio, line 428: This is adding code duplication, we already have two `TPCHandler::SendPerfMarker` methods. We should not add a third. I think it's better to think how to consolidate them into a single one instead.

Also, this commit is mixing up a lot more changes than what it says it's doing (`Update remote connection list from curl`). Please try to structure changes as one logical change per commit.

tests/TPCTests/test.sh
```diff
@@ -195,25 +195,37 @@ perform_http_tpc() {
...
     local dst_file_http="${hosts_http[$dst_idx]}/${RMTDATADIR}/${src}_to_${dst}${file_suffix}.ref_http"
     local http_code
+    local result_line
+    local body_file
+    body_file=$(mktemp)
 
     if [[ "$mode" == "push" ]]; then
         dst_file_http="${dst_file_http}_push"
-        http_code=$(${CURL} -X COPY -L -s -o >(cat >&2) -w "%{http_code}" \
+        http_code=$(${CURL} -X COPY -L -s -o "$body_file" -w "%{http_code}" \
```
amadio, line 223: This commit has a very long commit message in a single line. Please keep the first line to about 50 characters, and optionally add a blank line and longer description later. Long commits in one single line break `git log --oneline` output. See also: https://github.com/xrootd/xrootd/blob/master/docs/CONTRIBUTING.md#describe-your-changes-well

amadio: Note also that we will change "201 Created" in the future (#2625).

tests/TPCTests/test.sh
amadio: One commit per logical change, please.

https://github.com/xrootd/xrootd/blob/master/docs/CONTRIBUTING.md#make-separate-commits-for-logically-separate-changes

src/XrdHttpTpc/XrdHttpTpcTPC.cc
amadio: I don't think that vorg and scitag are a good way to decide on worker pools. We should probably think a bit more about how to create the different partitions before refactoring many classes tying them to taking scitags in their constructors, etc. In principle, the `TPCRequestManager` should be agnostic about these things, otherwise if we need to change strategy later (say, to allow worker pools based on who submitted it) it will be very hard.

amadio (changes requested):
I'd like to understand better the motivation for tying the worker pools to scitags and vorgs, as these are not always going to be available. It's not clear to me that this strategy will get us better results. I think we should look at real world cases to have a more informed decision on how to group transfers into the worker pools.

amadio:
@abh3 Please let us know your thoughts here too. Thank you!

abh3 (commented):
I have also discussed this with @amadio and I too am mystified on why the queues are based on scitags. It definitely complicates things doing it that way. I think we really need to have another discussion on what the ultimate goal is and consider simpler ways of achieving it.

amadio:
Due to the problems reported in the users list linked above, merge conflicts, and lack of consensus, I'm closing this for now. Let's discuss this in our development meetings to have a fresh replacement which is not tied to scitags and that is based on XRootD 6.0.

PR #2530, opened 2025-06-05
[XrdOfs] Coalesce results for all open requests

bbockelm:
Change the OFS so all threads waiting on the file handle lock during an in-progress open will get the same result.

Previously, if there were N threads trying to open a file that did not exist, then they would each grab the lock, try the open, and then error out.  In total, there would be N open requests.

With this, only the first thread to get the lock will trigger an open request; when the first thread fails, the N-1 waiting threads will all either get the successful handle or get the same error message.

When file-open requests take 100ms to fail, there's an approximately 3x speedup when there are 20 or 50 client threads.  I was not able to come up with any case where this slowed down the file-open throughput (tested the combination of 1, 20, and 50 client threads; fast file opens and slow file opens; file existing and file not existing).

### DIFF
--- a/src/XrdOfs/XrdOfs.cc
+++ b/src/XrdOfs/XrdOfs.cc
@@ -485,9 +485,11 @@ int XrdOfsFile::open(const char          *path,      // In
 
    struct OpenHelper
          {const char   *Path;
+          bool          isOpening;
           XrdOfsHandle *hP;
           XrdOssDF     *fP;
           int           poscNum;
+          int           retc{-EIO};
 
           int           OK() {hP = 0; fP = 0; poscNum = 0; return SFS_OK;}
 
@@ -495,8 +497,10 @@ int XrdOfsFile::open(const char          *path,      // In
                        : Path(path), hP(0), fP(0), poscNum(0) {}
 
                        ~OpenHelper()
-                       {int retc;
-                        if (hP) hP->Retire(retc);
+                       {if (hP)
+                           {hP->FinishOpen(retc);
+                            hP->Retire(retc);
+                           }
                         if (fP) delete fP;
                         if (poscNum > 0) XrdOfsFS->poscQ->Del(Path, poscNum, 1);
                        }
@@ -504,7 +508,7 @@ int XrdOfsFile::open(const char          *path,      // In
 
    mode_t theMode = (Mode | XrdOfsFS->fMask[0]) & XrdOfsFS->fMask[1];
    const char *tpcKey;
-   int retc, isPosc = 0, crOpts = 0, isRW = 0, open_flag = 0;
+   int isPosc = 0, crOpts = 0, isRW = 0, open_flag = 0;
    int find_flag = open_mode & (SFS_O_NOWAIT | SFS_O_RESET | SFS_O_MULTIW);
    XrdOucEnv Open_Env(info,0,client);
 
@@ -575,9 +579,9 @@ int XrdOfsFile::open(const char          *path,      // In
 // If we have a finder object, use it to direct the client. The final
 // destination will apply the security that is needed
 //
-   if (XrdOfsFS->Finder && (retc = XrdOfsFS->Finder->Locate(error, path,
+   if (XrdOfsFS->Finder && (oP.retc = XrdOfsFS->Finder->Locate(error, path,
                                                    find_flag, &Open_Env)))
-      return XrdOfsFS->fsError(error, retc);
+      return XrdOfsFS->fsError(error, oP.retc);
 
 // Preset TPC handling and if not allowed, complain
 //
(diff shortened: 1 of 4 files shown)


amadio:
There's still a [compiler warning](https://my.cdash.org/viewBuildError.php?type=1&buildid=2948931) that needs to be fixed for this PR to be ready.

abh3:
Yes, and I still need to review it.

src/XrdOfs/XrdOfs.cc
```diff
@@ -485,26 +485,30 @@ int XrdOfsFile::open(const char          *path,      // In
 
    struct OpenHelper
          {const char   *Path;
+          bool          isOpening;
           XrdOfsHandle *hP;
           XrdOssDF     *fP;
           int           poscNum;
+          int           retc{-EIO};
```
amadio, line 492: What is the reason to move `retc` to a wider scope? Can't it be left where it was before? If this change is desired, it should be implemented as a separate commit to follow the rule of one logical change per commit.

src/XrdOfs/XrdOfsHandle.cc
```diff
@@ -271,18 +281,49 @@ int XrdOfsHandle::Alloc(const char *thePath, int Opts, XrdOfsHandle **Handle)
...
+       lock.unlock();
        myMutex.Lock(); hP->Path.Links--; myMutex.UnLock();
        return nolokDelay;
       }
 
-// Get a new handle
+// If we are here, then there is no existing handle for this path; we need to
+// allocate a new handle and tell the caller it is responsible for opening it.
 //
+   //OfsEroute.Emsg("Alloc", "Creating a new file handle");
```
amadio, line 321: I'd prefer we do not add dead code in new commits.

src/XrdOfs/XrdOfs.cc
```diff
@@ -664,9 +668,31 @@ int XrdOfsFile::open(const char          *path,      // In
...
+           return XrdOfsFS->fsError(error, SFS_STARTED);
+          }
+       if (openRC == -ETXTBSY) return XrdOfsFS->Stall(error, -1, path);
+       if (openRC == -EDESTADDRREQ)
+          {char *url = Open_Env.Get("FileURL");
+           if (url) {error.setErrInfo(-1, url); return SFS_REDIRECT;}
+          }
+       if (XrdOfsFS->Balancer && openRC == -ENOENT)
+          XrdOfsFS->Balancer->Removed(path);
+       return XrdOfsFS->Emsg(epname, error, openRC, "open", path);
```
amadio, line 695: It seems this part is the real bug fix, most of the remaining changes are only style changes. This should therefore be a separate commit from the rest, and come before anything else in the series.

src/XrdOfs/XrdOfs.cc
```diff
@@ -727,38 +756,40 @@ int XrdOfsFile::open(const char          *path,      // In
...
+          return XrdOfsFS->Emsg(epname, error, oP.retc, "fchmod", path);
        XrdOfsFS->poscQ->Commit(path, oP.poscNum);
       }
 
 // Set compression values and activate the handle
 //
+   oP.hP->Lock();
    if (oP.fP->isCompressed() > 0)
-      {oP.hP->isCompressed = 1;
+      {oP.hP->setCompressed(true);
```
amadio, line 787: I don't see the advantage of using a method for this. Likely the changes adding new interfaces can be skipped for this bug fix.

src/XrdOfs/XrdOfsHandle.hh
```diff
@@ -188,11 +217,43 @@ static XrdOfsHanTab  rwTable;    // File Handles open r/w
...
+            int                    *openRC{nullptr};     // Return code for open from other thread
+       };
+       XrdOfsHandleOpenWaiter *FirstWaiter{nullptr}; // Next handle waiting for open
+
+       XrdOssDF *ssi;                      // Storage System Interface
+
+       XrdOfsHanPsc *Posc;       // -> Info for posc-type files
        XrdOfsHandle *Next;
        XrdOfsHanKey  Path;       // Path for this handle
-       XrdOfsHanPsc *Posc;       // -> Info for posc-type files
```
amadio, line 195: This substantially changes the layout of the class, which means that as it is it should probably only go into XRootD 6.0.

amadio (changes requested):
This commit is quite big to be made in one go. I think it makes sense to split the work into one logical change per commit by separating style changes from functional changes at least, and having the bug fix come before the style changes. Some of the added interfaces do make sense, but because the class layout is also changed, this requires at least recompilation, so as it is I would only suggest merging this into XRootD 6.0.

src/XrdOfs/XrdOfsHandle.hh
bbockelm: It changes the layout of an internal class that's not part of the ABI. Why wait for a major release for an internal change?

src/XrdOfs/XrdOfs.cc
bbockelm: Ah - it comes from trying to prevent the data structure from growing another 8 bytes.  Adding another `char` will create a hole due to alignment issues.

By collapsing all the things used as bitmasks (and are separately synchronized) into a single bitmask, we free up enough space to have a flag dedicated to indicate the open is in-progress, which is needed for the condition variable.

(That said, cleaning up the undefined behavior due to race conditions and eliminating a mutex lock on every call is indeed a separate fix I can pull out in the next revision of the patch)

src/XrdOfs/XrdOfsHandle.hh
amadio: It may be ok, but I have to run the ABI check build to know this for sure. The header is installed in the private include directory, so it could be that after this change, upgrading XRootD may cause EOS to crash. If the header is really not used, then we should also stop installing it and this change is then fine to make.

amadio: This is what I see when I run the ABI check for `libXrdServer.so`:

screenshot

bbockelm: Ah, indeed, if the XrdOfs is part of the public ABI this is going to only be doable in V6.  There's just too much of the internals that are then exposed to make this fix.

bbockelm:
@abh3 - to make sure I understand your comment, you'd be OK with the increase in the struct size that's (partially) avoided by compressing things to a bitmask?

I mostly did the bitmask work as I thought you'd strongly want it in the first place.

src/XrdOfs/XrdOfsHandle.hh
amadio: Yes, unfortunately any changes in size of structs changes also offsets and makes updating XRootD break EOS. So I think that unless an alternative way of doing this that doesn't involve breaking ABI is possible, let's leave this for XRootD 6.0. Since we've now accumulated enough changes to warrant a major release (file cloning, client fixes for timeouts, this change, and other changes currently sitting in `devel`), the plan is to do it this year (i.e. sometime in either November or December), so the wait won't be so long anymore.

abh3:
Yes, I am OK with increasing the size as it would simplify the whole PR. Of course, as @amadio points out, it breaks EOS, so we will need it to go into a major release.

amadio:
@bbockelm Are you still working on this pull request?

PR #2532, opened 2025-06-06
Make HTTP's maximum open delay configurable

bbockelm:
By default, the bridge will retry operations internally until the maximum delay is hit and then return an error.  By default, the maximum delay is 3600 seconds; this is well beyond the timeouts of a typical HTTP client.

This makes the delay configurable for HTTP with the following:

```
http.maxdelay <seconds>
```

as in

```
http.maxdelay 10
```

It also tunes down the default delay to 30s.

Note: the response time can still be greater than 30s; this just limits how long the bridge will process "wait" responses for.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -26,6 +26,7 @@
 #include "Xrd/XrdBuffer.hh"
 #include "Xrd/XrdLink.hh"
 #include "XProtocol/XProtocol.hh"
+#include "XrdOuc/XrdOuca2x.hh"
 #include "XrdOuc/XrdOucStream.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 #include "XrdOuc/XrdOucGMap.hh"
@@ -45,6 +46,7 @@
 #include "XrdOuc/XrdOucUtils.hh"
 #include "XrdOuc/XrdOucPrivateUtils.hh"
 
+#include <charconv>
 #include <openssl/err.h>
 #include <openssl/ssl.h>
 #include <vector>
@@ -84,6 +86,7 @@ char *XrdHttpProtocol::listredir = 0;
 bool XrdHttpProtocol::listdeny = false;
 bool XrdHttpProtocol::embeddedstatic = true;
 char *XrdHttpProtocol::staticredir = 0;
+int XrdHttpProtocol::m_maxdelay = -1;
 XrdOucHash<XrdHttpProtocol::StaticPreloadInfo> *XrdHttpProtocol::staticpreload = 0;
 
 kXR_int32 XrdHttpProtocol::myRole = kXR_isManager;
@@ -899,6 +902,7 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
       TRACEI(REQ, " Authorization failed.");
       return -1;
     }
+    if (m_maxdelay > 0) Bridge->SetWait(m_maxdelay, false);
 
     // Let the bridge process the login, and then reinvoke us
     DoingLogin = true;
@@ -1084,6 +1088,7 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
       else if TS_Xeq("tlsreuse", xtlsreuse);
       else if TS_Xeq("auth", xauth);
       else if TS_Xeq("tlsclientauth", xtlsclientauth);
+      else if TS_Xeq("maxdelay", xmaxdelay);
       else {
         eDest.Say("Config warning: ignoring unknown directive '", var, "'.");
         Config.Echo();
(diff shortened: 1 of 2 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2985,6 +2989,27 @@ int XrdHttpProtocol::xauth(XrdOucStream &Config) {
   return 0;
 }
 
+int XrdHttpProtocol::xmaxdelay(XrdOucStream &Config) {
+  char *val = Config.GetWord();
+  if(val) {
+    auto len = strlen(val);
```
abh3: We actually have a way to deal with this kind of directive parameter that is far simpler by using the class in XrdOucXrdOuca2x.hh which provdes consistentcy. In this case a simple call to a2tm() would handle all of the error cases and give you seconds using a standard xrootd time interval format of n[s|m|h]. I'd prefer that be used here.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -84,6 +85,7 @@ char *XrdHttpProtocol::listredir = 0;
 bool XrdHttpProtocol::listdeny = false;
 bool XrdHttpProtocol::embeddedstatic = true;
 char *XrdHttpProtocol::staticredir = 0;
+unsigned XrdHttpProtocol::m_maxdelay = 30;
```
abh3: SetWait() takes an int so this really should be an int (later comments show how this would not be a problem).

Also, I am rather uncomfortable with reducing it to 30 seconds as the default, given the previous default was effectively no limit (3600 seconds). As the code is written this now becomes the enforced default even if the directive was not specified at all. That's a significant behaviour change. This would primarily affect anyone who uses tape systems as the wait time would be too short to bring a file online and those sites would stop working because of this default. Frankly, other than unlimited there is no good default for everyone (see suggestion below).

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -898,6 +900,7 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
       TRACEI(REQ, " Authorization failed.");
       return -1;
     }
+    Bridge->SetWait(m_maxdelay, false);
```
abh3: Given that there is no good default, one can override the default by specifying the new directive, If not specified, you skip issuing the SetWait(). This keeps the old behaviour but lets you establish a new behaviour for cases where it makes sense for a short maximum wait time. For instance, m_maxdelay = -1; and here if (m_maxdelay > 0) Bridge->SetWait(m_maxdelay, false);

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -426,6 +427,10 @@ protected:
   // Url to redirect to in the case a /static is requested
   static char *staticredir;
 
+  // Maximum amount of time an operation on the bridge can be
+  // delayed
+  static unsigned m_maxdelay;
```
abh3: As noted above, this should be an int.

abh3 (changes requested):
OK, this is all about avoiding  behaviour changes and using consistent handling of directives. It also winds up being less code.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Thanks, I didn't know about `a2tm`.  If possible, it'd be nice to have some docstring on the function and how to use it -- in the latest revision, I just went from other examples and the source code.

Item resolved in the latest revision.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: I've got serious questions about setups relying on HTTP clients waiting for dozens of minutes for a server to respond.  But, I suppose, stranger things have happened!

In the latest push, I took your advice and made this an `int` and kept the default unchanged.

src/XrdHttp/XrdHttpProtocol.cc
bbockelm: Done in the latest push.

bbockelm:
@abh3 - I believe all your review comments have been addressed in the latest branch revision and requested a new review in the GitHub interface.  Thanks!

smithdh:
Hi @bbockelm . I happened to also see some not desirable behvior around the wait retry, while testing fix(es) for #2602. We do not cancel or break out of the redrive incase the http client has gone away (the redrive job will wait for the delay period, reissue its XrdXrootdProtocol reqeust, and if the result is still kxr_wait another redrive is scheduled etc.). If there was a situation where say, Open was returning kxr_wait(some small value) for a period of  e.g. 5 minutes, and there was a e.g. a population of 100 http clients reqeusting a file with a client-side timeout of 10 seconds with immediate retry, we chould have 3000 redrive jobs scheduled (and network fd's open).

I haven't found a proposal for properly detecting that the http client has gone. We could detect, e.g. POLLRDHUP from poll to indicate client gone, but that seems not a safe assuption, as an http client could have shutdown(fd,SHUT_WR) its connection, but still be reading for the request result on the half open conneciton.

PR #2535, opened 2025-06-08
Avoid reference beyond end of table

### LINKED ISSUE #2534: [XrdXrootd] Server segfaults while computing checksums under load
When querying for many checksums under load via HTTP, I've been able to reliably get the server to crash within the XrdXrootd framework.

Log attached; haven't picked through it yet.

[origin.txt](https://github.com/user-attachments/files/20641380/origin.txt)

bbockelm:
Previously, when the last entry in the table was deleted, there was a check of the `N+1` entry.  This could potentially cause a segfault if the end of the table is the end of a page.

A segfault was triggered when running under ASAN with a high concurrency of jobs requesting checksums.

Fixes #2534

### DIFF
--- a/src/XrdOuc/XrdOucTable.hh
+++ b/src/XrdOuc/XrdOucTable.hh
@@ -137,7 +137,9 @@ T  *Remove(int Tnum)
            Table[Tnum].Fnum = avlnum;
            avlnum = Tnum;
            if (Tnum == (curnum-1))
-              while(curnum && Table[curnum].Item == 0) curnum--;
+              {if (curnum == maxnum) curnum--;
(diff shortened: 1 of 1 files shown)


amadio:
Although the fix seems good to me, since [coverage of XrdOucTable](https://my.cdash.org/viewCoverageFile.php?buildid=2950452&fileid=22160023) is not complete, and this is a simple and independent class, would you consider adding some unit tests based on Google test to complete coverage of this class? ChatGPT can probably generate you a good starting template for it. Thank you.

amadio:
That's ok, I probably will release 5.8.4 by the end of next week or end of the month at the latest. If the test is not ready by then, I will just merge this and we can add the test later.

PR #2536, opened 2025-06-08
[XrdPfc] Ensure a reference to the file is kept when scheduling a sync

bbockelm:
As soon as the block is freed, it may cause the file to be marked as "inactive" (and thus eligible for deletion).

If a file sync is desired, then we must delay decreasing the block's reference count until after the sync is scheduled to ensure it is still alive when the sync scheduling increases the reference count on the file.

This change eliminates a use-after-free segfault observed on my local stress test which would trigger about 50% of the time (10 clients reading unique 3-byte files in a caching setup would crash after 20-30 seconds).


osschar (approved):
Yup, looks good. Strange we haven't banged into this sooner, it's been like this for quite some time.

Issue #2541, opened 2025-06-12
[XrdHttp] Add CORS Origin header handling


ccaffy:
### Feature Description

It may happen that an XRootD HTTP server is accessed via web browsers with Javascript tools (JSROOT for example).

On of the security feature of the web browsers are the CORS verification and any resource access where CORS verification cannot be done will fail.

With the current version of XRootD, we can have static headers returned for each HTTP request and this can be configured by the server admin using `http.staticheader  Access-Control-Allow-Origin https://myorigin.cern.ch`

The problem is that it does not fully fulfill the needs of my users.

Ideally, this functionality would work like the following:

The XRootD server administrator sets a list of allowed origins:
```
http.corsorigin https://myorigin1.cern.ch
http.corsorigin https://myorigin2.cern.ch
...
http.corsorigin https://myoriginN.cern.ch
```

The client will send an "Origin" header with their GET request:

```
curl -X GET -H 'Origin: https://myorigin2.cern.ch' https://xrootd.cern.ch/my/file.txt
```

The server will check that the Origin header provided by the user is part of the allowed origins. If it is part of it, the server will return the following header with the GET response:

```
Access-Control-Allow-Origin: https://myorigin2.cern.ch
```

If it is not part of the list of allowed origins, then the server will simply not return any `Access-Control-Allow-Origin`.

Suggestions and comments are of course welcome :)

For more reference:
- https://datatracker.ietf.org/doc/html/rfc6454
- https://fetch.spec.whatwg.org/#http-cors-protocol
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Access-Control-Allow-Origin

ccaffy:
Think about taking into account the "Vary" header too

ccaffy:
@abh3, yes indeed.

abh3:
I agree with Brian. This is especially true as XRooD is primarily a plug-in architecture. This would appear to be a prime candidate for a plug-in implementation.

ccaffy:
I haven't seen any plugin for nginx, the CORS stuff is configured on the main config file: 
- https://enable-cors.org/server_nginx.html
- https://www.baeldung.com/linux/nginx-cross-origin-policy-headers

However I don't mind creating a new plugin for that. How would you see the configuration of that functionality in XRootD look like then?

Thanks

ccaffy:
[...]

PR #2542, opened 2025-06-12
[XrdHttp] Added header utils class for header parsing

ccaffy:
This makes header parsing located in one place in addition to allow unit testing for header parsing.
As an example, the Repr-Digest header parsing has been implemented.

This is a preparation for future header parsing refactoring and new header parsing addition (https://github.com/xrootd/xrootd/issues/2541). It would be nice to have it in master so that I can refer to this code on my future developments.

### DIFF
--- a/src/XrdHttp/CMakeLists.txt
+++ b/src/XrdHttp/CMakeLists.txt
@@ -17,6 +17,7 @@ add_library(XrdHttpUtils SHARED
                              XrdHttpStatic.hh
                              XrdHttpTrace.hh
   XrdHttpUtils.cc            XrdHttpUtils.hh
+  XrdHttpHeaderUtils.cc      XrdHttpHeaderUtils.hh
 )
 
 set_target_properties(XrdHttpUtils
--- a/src/XrdHttp/XrdHttpHeaderUtils.cc
+++ b/src/XrdHttp/XrdHttpHeaderUtils.cc
@@ -1,0 +1,59 @@
+//------------------------------------------------------------------------------
+// This file is part of XrdHTTP: A pragmatic implementation of the
+// HTTP/WebDAV protocol for the Xrootd framework
+//
+// Copyright (c) 2025 by European Organization for Nuclear Research (CERN)
+// Author: Cedric Caffy <ccaffy@cern.ch>
+// File Date: Jun 2025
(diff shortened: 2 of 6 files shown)


src/XrdHttp/XrdHttpHeaderUtils.cc
```diff
@@ -0,0 +1,58 @@
...
+  // Expected format per entry: <cksumType>=:<digestValue>:
+  std::vector<std::string> digestNameValuePairs;
+  XrdOucTUtils::splitString(digestNameValuePairs, header, ",");
+
+  for (const auto &digestNameValue : digestNameValuePairs) {
+    auto equalPos = digestNameValue.find('=');
+    if (equalPos == std::string::npos || equalPos >= digestNameValue.size() - 1)
+      continue;
+
+    std::string cksumType = digestNameValue.substr(0, equalPos);
```
bbockelm: Can we avoid a bunch of allocation and string copies by switching to string views?  Just did a quick read through but it looks like we're not modifying the string but just trimming out the pieces we want.

ccaffy: Yup, very nice suggestion, done!

src/XrdHttp/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_library(XrdHttpUtils SHARED
                              XrdHttpStatic.hh
                              XrdHttpTrace.hh
   XrdHttpUtils.cc            XrdHttpUtils.hh
+  XrdHttpHeaderUtils.cc      XrdHttpHeaderUtils.hh
```
abh3, line 20: I think it would be better if we keep all the utilities in the utils class. Of course, if yu are going to say "but this is only the start and I will be adding a dozen more", then separating out utilities that specifically deal with headers is ok. If it's only or two or so additional functions it would be better placed in XrdHttpUtils.

ccaffy: Hi Andy,

Thanks for your feedback. Yes, it's just the beginning. Ideally I'd like all header parsing code to go in this class so they are all unit tested without the need of XrdHttp!

That's a basis for my future work in refactoring the XrdHttp code.

Cheers,
Cedric

PR #2548, opened 2025-06-18
[XrdPfc] Avoid using I/O object after it has been destroyed

### LINKED ISSUE #2547: [XrdPfc] Crash in the block error response code
### Operating System

AlmaLinux 9.5

### XRootD Version

xrootd-5.8.2-1.5.osg24.el9.x86_64

### Description

Observed the following crash in a Pelican cahe yesterday:

```
#6  XrdPfc::IO::register_block_error (res=-18, this=<optimized out>) at /usr/src/debug/xrootd-5.8.2-1.5.osg24.el9.x86_64/xrootd-5.8.2/src/XrdPfc/XrdPfcIO.hh:94
#7  XrdPfc::File::ProcessBlockResponse (this=0x7fb828219e20, b=0x7fbe14843e20, res=-18) at /usr/src/debug/xrootd-5.8.2-1.5.osg24.el9.x86_64/xrootd-5.8.2/src/XrdPfc/XrdPfcFile.cc:1468
#8  0x00007fc9925d4c05 in XrdPfc::BlockResponseHandler::Done (this=0x7fb2d00e8640, res=<optimized out>) at /usr/src/debug/xrootd-5.8.2-1.5.osg24.el9.x86_64/xrootd-5.8.2/src/XrdPfc/XrdPfcFile.cc:1680
#9  0x00007fc9a2c80cb0 in XrdPosixFileRH::DoIt() () from /lib64/libXrdPosix.so.3
#10 0x00007fca851d8f6a in XrdScheduler::Run (this=0x414780 <XrdGlobal::Sched>) at /usr/src/debug/xrootd-5.8.2-1.5.osg24.el9.x86_64/xrootd-5.8.2/src/Xrd/XrdScheduler.cc:365
```

Still picking apart the 24GB core dump.  It doesn't appear to the same source as #2533, however.

_No response_

### Additional Information

_No response_

bbockelm:
If the block prefetch fails after the corresponding I/O object has been detached/freed, then previously the code would try to retry the block download (which required a valid I/O object) which would be guaranteed to segfault.

Now, if the block fails after the I/O object is detached, then we simply clean it up.  This mirrors the logic when the prefetch fails while the I/O is still attached.

Fixes #2547


src/XrdPfc/XrdPfcFile.cc
```diff
@@ -1409,7 +1409,8 @@ void File::ProcessBlockResponse(Block *b, int res)
    {
       IO     *io = b->get_io();
       IoSet_i mi = m_io_set.find(io);
-      if (mi != m_io_set.end())
+      auto found_io = mi != m_io_set.end();
```
osschar: please change `auto` to `bool` :)

src/XrdPfc/XrdPfcFile.cc
```diff
@@ -1428,20 +1429,22 @@ void File::ProcessBlockResponse(Block *b, int res)
...
       }
       else
       {
          TRACEF(Error, tpfx << "io " << b->get_io() << " not found in IoSet.");
       }
+
+      // If nothing is subscribed to this block -- or if the originating I/O object
+      // is gone -- and the block read failed/incomplete or the file is in shutdown,
+      // then we do not want to re-issue the read later.  Immediately free and return.
+      if ((b->m_refcnt == 0 || !found_io) && (res != b->get_size() || m_in_shutdown))
```
osschar: please change `res != b->get_size()` to `res < 0` (as it was in the original conditional). the read size mismatch case is handled at the top, at this stage `res < 0` is the only thing it matters.

osschar (approved):
Thank you for this fix! Sorry about nitpicking with the changes but this multi-io-async stuff is pretty hard to follow as it is already.

src/XrdPfc/XrdPfcFile.cc
bbockelm: Addressed in the latest push.

src/XrdPfc/XrdPfcFile.cc
bbockelm: Addressed in the latest push.

osschar:
Yeah ... something else is rotten ... the io shouldn't have been detached before this code is entered. So, while this fix might be a band-aid, it's just hiding the real issue.

amadio:
> Yeah ... something else is rotten ... the io shouldn't have been detached before this code is entered. So, while this fix might be a band-aid, it's just hiding the real issue.

In that case we should fix the symptom, but the root cause, no?

bbockelm:
It can be two things!

The current PR does fix incorrect logic that is otherwise guaranteed to segfault.

Just there appears to be a second issue here.

amadio:
@osschar What is the plan here? Do we merge the current change or is there going to be an alternative patch to fix the root cause of this issue? I'd like to release XRootD 5.8.4 before the end of this month with a fix for this included if possible. Thank you!

osschar:
I'm on vacation until the second week of July. Given the other symptoms (N_prefetch blocks < 0) I was suspecting XrdClPelican might be invoking some response handlers multiple times ... @bbockelm said he'll cross check on his side. This code has been running happily for 5+ years until now.

I wouldn't commit this ... if anything we can add an early check and exit with assertion or exception to make it easier to debug in the wild.

amadio:
@osschar Thanks! Sorry to bother you, enjoy your vacations!

osschar:
Thanks! How do we go about this then? We know there is something wrong with the number of blocks being prefetched for a given file ... as it dips below 0. We could add a trap at the point where it hits -1 ... at this point we will be in the handler that is "close" to the one that caused this to happen (but it is likely not the XrdCl poller thread that handled the event that caused the callback).

One more option -- are there some curl internal retries that could invoke the error callback and then, after a retry, a success one (or the same callback with a different exit code)?

Where there further occurrences out there in the wild? Are you using changes from this PR for Pelican now?

amadio:
@bbockelm What is the path forward here? Could you pleanse answer @osschar's questions? I'd like to get a clear agreement on whether the changes introduced here are needed and whether they address the right problem or just hide a symptom of a deeper problem as suggested before.

PR #2552, opened 2025-06-24
2541 cors plugin implementation

ccaffy:
This is draft and not properly tested. Despite having unit tests, I plan to add system test as well!

Please have a look so I can get early feedback and correct things once I'm back.

Thanks,
Cheers,
Cedric

This is the config file:
```
  http.cors libXrdCors.so
  cors.origin https://test.cern.ch
# AND ONE CAN PUT EVERYTHING IN ONE LINE IF NEEDED
  cors.origin https://test2.cern.ch https://test.ch
```

This is the effect on a PUT request:

```
> PUT /tmp/fic.txt_dest? HTTP/1.1
> Host: xrootd.cern.ch:1096
> User-Agent: curl/7.76.1
> Accept: */*
> Repr-Digest: adler=:RXJyb3IK=:, sha256=:test:
> Origin: https://test.cern.ch
> Content-Length: 5242880
> Expect: 100-continue
> 

< HTTP/1.1 100 Continue
< Connection: Close
< Server: XrootD/v5.7.1-22-g82a3fc0e8
< Access-Control-Allow-Origin: https://test.cern.ch

```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -109,6 +109,7 @@ if( NOT XRDCL_ONLY )
 
   add_subdirectory( XrdMacaroons )
   add_subdirectory( XrdVoms )
+  add_subdirectory( XrdHttpCors )
   add_subdirectory( XrdCeph )
   add_subdirectory( XrdSciTokens )
 endif()
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -45,6 +45,7 @@
 #include "XrdTls/XrdTlsContext.hh"
 #include "XrdOuc/XrdOucUtils.hh"
 #include "XrdOuc/XrdOucPrivateUtils.hh"
+#include "XrdHttpCors/XrdHttpCors.hh"
 
 #include <charconv>
 #include <openssl/err.h>
@@ -103,6 +104,8 @@ BIO *XrdHttpProtocol::sslbio_err = 0;
 XrdHttpSecXtractor *XrdHttpProtocol::secxtractor = 0;
 bool XrdHttpProtocol::isRequiredXtractor = false;
 struct XrdHttpProtocol::XrdHttpExtHandlerInfo XrdHttpProtocol::exthandler[MAX_XRDHTTPEXTHANDLERS];
+std::string XrdHttpProtocol::xrdcorsLibPath;
+XrdHttpCors * XrdHttpProtocol::xrdcors = nullptr;
 int XrdHttpProtocol::exthandlercnt = 0;
 std::map< std::string, std::string > XrdHttpProtocol::hdr2cgimap; 
 
@@ -1075,6 +1078,7 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
       else if TS_Xeq("secretkey", xsecretkey);
       else if TS_Xeq("desthttps", xdesthttps);
       else if TS_Xeq("secxtractor", xsecxtractor);
+      else if TS_Xeq("cors", xcors);
       else if TS_Xeq3("exthandler", xexthandler);
       else if TS_Xeq("selfhttps2http", xselfhttps2http);
       else if TS_Xeq("embeddedstatic", xembeddedstatic);
@@ -1141,6 +1145,16 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
 // Test if this is a caching server
 //
    if (myEnv->Get("XrdCache")) hasCache = true;
+
+   // Load CORS plugin if configured
+   if(xrdcorsLibPath.size()) {
+     if(LoadCorsHandler(&eDest, xrdcorsLibPath.c_str()) != 0) {
+       return 1;
+     }
+     if (xrdcors->Configure(ConfigFN, &eDest) != 0) {
+       return 1;
+     }
+   }
 
 // If https was disabled, then issue a warning message if xrdtls configured
 // of it's disabled because httpsmode was auto and xrdtls was not configured.
@@ -1649,6 +1663,13 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc,
     ss << iter->second;
   } else {
     ss << m_staticheaders[""];
+  }
+
+  if(xrdcors) {
+    auto corsAllowOrigin = xrdcors->getCORSAllowOriginHeader(CurrentReq.m_origin);
+    if(corsAllowOrigin) {
+      ss << *corsAllowOrigin << crlf;
+    }
   }
 
   if ((bodylen >= 0) && (code != 100))
@@ -2760,6 +2781,18 @@ int XrdHttpProtocol::xsecxtractor(XrdOucStream& Config) {
   return 0;
 }
 
+int XrdHttpProtocol::xcors(XrdOucStream& Config) {
+  char * val;
+  // Get the path
+  val = Config.GetWord();
+  if (!val || !val[0]) {
+    eDest.Emsg("Config", "No CORS plugin specified.");
+    return 1;
+  }
+  xrdcorsLibPath = val;
+  return 0;
+}
+
 /******************************************************************************/
 /*                            x e x t h a n d l e r                           */
 /******************************************************************************/
@@ -3227,6 +3260,15 @@ int XrdHttpProtocol::LoadExtHandler(XrdSysError *myeDest, const char *libName,
 }
 
 
+int XrdHttpProtocol::LoadCorsHandler(XrdSysError *eDest, const char *libname) {
+  if(xrdcors) return 1;
+  XrdOucPinLoader corsLib(eDest, &compiledVer, "corslib",libname);
+  XrdHttpCors *(*ep)(XrdHttpCorsGetHandlerArgs);
+  ep = (XrdHttpCors *(*)(XrdHttpCorsGetHandlerArgs))(corsLib.Resolve("XrdHttpCorsGetHandler"));
+  if(ep && (xrdcors = ep())) return 0;
+  corsLib.Unload();
+  return 1;
+}
 
 // Tells if we have already loaded a certain exthandler. Try to
 // privilege speed, as this func may be invoked pretty often
--- a/src/XrdHttp/XrdHttpProtocol.hh
+++ b/src/XrdHttp/XrdHttpProtocol.hh
@@ -49,6 +49,7 @@
 #include "XrdHttpChecksumHandler.hh"
 #include "XrdHttpReadRangeHandler.hh"
 #include "XrdNet/XrdNetPMark.hh"
+#include "XrdHttpCors/XrdHttpCors.hh"
 
 #include <openssl/ssl.h>
 
@@ -206,6 +207,7 @@ private:
   static int xsslcert(XrdOucStream &Config);
   static int xsslkey(XrdOucStream &Config);
   static int xsecxtractor(XrdOucStream &Config);
+  static int xcors(XrdOucStream &Config);
   static int xexthandler(XrdOucStream & Config, std::vector<extHInfo> &hiVec);
   static int xsslcadir(XrdOucStream &Config);
   static int xsslcipherfilter(XrdOucStream &Config);
@@ -256,6 +258,7 @@ private:
                             const char *configFN, const char *libParms,
                             XrdOucEnv *myEnv, const char *instName);
 
+  static int LoadCorsHandler(XrdSysError *eDest, const char * libname);
   // Determines whether one of the loaded ExtHandlers are interested in
   // handling a given request.
   //
@@ -376,7 +379,9 @@ protected:
   /// This also can process HTTP/DAV stuff
   XrdHttpReq CurrentReq;
 
+  static std::string xrdcorsLibPath;
 
+  static XrdHttpCors * xrdcors;
   //
   // Processing configuration values
   //
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -57,6 +57,7 @@
 #include "XrdOuc/XrdOucTUtils.hh"
 #include "XrdOuc/XrdOucUtils.hh"
 #include "XrdOuc/XrdOucPrivateUtils.hh"
+#include "XrdHttp/XrdHttpHeaderUtils.hh"
 
 #include "XrdHttpUtils.hh"
 
@@ -206,6 +207,9 @@ int XrdHttpReq::parseLine(char *line, int len) {
     } else if (!strcasecmp(key, "user-agent")) {
       m_user_agent = val;
       trim(m_user_agent);
+    } else if (!strcasecmp(key,"origin")) {
+      m_origin = val;
+      trim(m_origin);
     } else {
       // Some headers need to be translated into "local" cgi info.
       auto it = std::find_if(prot->hdr2cgimap.begin(), prot->hdr2cgimap.end(),[key](const auto & item) {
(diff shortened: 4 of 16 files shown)


src/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ if( NOT XRDCL_ONLY )
 
   add_subdirectory( XrdMacaroons )
   add_subdirectory( XrdVoms )
+  add_subdirectory( XrdCors )
```
abh3: We are trying to group anything specific to http in directories that start with XrdHttp (notice that XrdTpc was changed to XrdHttpTpc). So, it would follow that this should be named XrdHttpCors.

abh3 (changes requested):
Some round I changes needed here. Not too bad at all.

src/CMakeLists.txt
ccaffy: That's absolutely correct, I will perform the change :)

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2750,6 +2765,25 @@ int XrdHttpProtocol::xsecxtractor(XrdOucStream& Config) {
...
+  //
+  val = Config.GetWord();
+  if (!val || !val[0]) {
+    eDest.Emsg("Config", "No CORS plugin specified.");
+    return 1;
+  } else {
+    char libName[4096];
+    strlcpy(libName, val, sizeof(libName));
+    libName[sizeof(libName) - 1] = '\0';
+    if (LoadCorsHandler(&eDest, libName)) {
```
abh3: Please move the load to the cors configuration method called after parsing the config file as it eleongs there not here.

abh3 (changes requested):
Thank you for renaming the files! There are still a couple of changes outstanding from the last review. I have noted them in this review.

src/XrdHttp/XrdHttpProtocol.cc
ccaffy: OK I inspired from the xsecxtractor which does this load:
```cpp
// Try to load the plugin (if available) that extracts info from the
    // user cert/proxy
    if (LoadSecXtractor(&eDest, libName, libParms)) {
      return 1;
    }
```

It avoided to have to declare a string to store the libName. However no problem, will do :)

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -374,7 +377,9 @@ protected:
   /// This also can process HTTP/DAV stuff
   XrdHttpReq CurrentReq;
 
+  static char xrdcorsLibPath[4096];
```
abh3: This is particularly wasteful. It's unlikely any pah will ever be more than a couple of hundred bytes at most. I suggest making this a pointer initialized to null. See my other comments on what to do about it. You can also use std::string if that seems better.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1135,6 +1139,15 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
 //
    if (myEnv->Get("XrdCache")) hasCache = true;
 
+   if(strlen(xrdcorsLibPath) && LoadCorsHandler(&eDest, xrdcorsLibPath)) {
+     return 1;
+   }
+
+  // If xrdcors was loaded, configure it
+  if(xrdcors) {
```
abh3: This section can be combined with the previous one. They are both codependent and either both should occur or none should occur. So, if LoadCorsHandler() succeeds then one should continue to config it. Separate tests are unnecessary.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1135,6 +1139,15 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
 //
    if (myEnv->Get("XrdCache")) hasCache = true;
 
+   if(strlen(xrdcorsLibPath) && LoadCorsHandler(&eDest, xrdcorsLibPath)) {
```
abh3: Assuming you chose a pointe, then simply test for null, orif using  std::string, simply test for size() > 0.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -2750,6 +2770,18 @@ int XrdHttpProtocol::xsecxtractor(XrdOucStream& Config) {
...
 
+int XrdHttpProtocol::xcors(XrdOucStream& Config) {
+  char * val;
+  // Get the path
+  val = Config.GetWord();
+  if (!val || !val[0]) {
+    eDest.Emsg("Config", "No CORS plugin specified.");
+    return 1;
+  }
+  strlcpy(xrdcorsLibPath, val, sizeof(xrdcorsLibPath));
```
abh3: Assuming the destination is a pointer, all you need to do is a) if not null, free it, b) do a simple strdup to set the pointer. If you prefer to use std::string then just assign it.

src/XrdHttp/XrdHttpProtocol.hh
ccaffy: I agree, I use std::string instead then. Again inspired from the xrdsec config above ;)

abh3 (approved):
Looks OK, thanks!

tests/XrdHttpTests/XrdHttpTests.cc
```diff
@@ -628,4 +629,25 @@ TEST(XrdHttpTests, parseReprDigest) {
     XrdHttpHeaderUtils::parseReprDigest(input,output);
     ASSERT_EQ(expectedMap, output);
   }
+}
+
+TEST(XrdHttpTests, getCORSAllowOriginHeader) {
+  std::unordered_set<std::string> allowedOrigins = {
+    "https://helloworld.cern.ch/"
+    "https://anotherorigins.cern.ch/"
```
chrisburr: By construction I don't think the origin header can have the trailing slash, the allowed syntax is:

```
Origin: null
Origin: <scheme>://<hostname>
Origin: <scheme>://<hostname>:<port>
```

chrisburr: Similarly I think the example in the PR description is invalid.

src/CMakeLists.txt
```diff
@@ -109,6 +109,7 @@ if( NOT XRDCL_ONLY )
 
   add_subdirectory( XrdMacaroons )
   add_subdirectory( XrdVoms )
+  add_subdirectory(XrdHttpCors)
```
amadio: I'm not a big fan of the spaces around the names here, but for style consistency, could you please add it? At one point we can discuss CMake style and change everything, but having a mix of styles is not pleasant.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -206,6 +207,8 @@ int XrdHttpReq::parseLine(char *line, int len) {
     } else if (!strcasecmp(key, "user-agent")) {
       m_user_agent = val;
       trim(m_user_agent);
+    } else if (!strcasecmp(key,"origin")) {
+      XrdHttpHeaderUtils::parseSimpleValue(val, m_origin);
```
amadio: This helper is used only here, I'd skip adding it. See the lines right above, where `trim()` is used directly already. I think that's clearer to just do the same here as well.

src/XrdHttpCors/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+set(XrdHttpCors       XrdHttpCors-${PLUGIN_VERSION})
+
+add_library(${XrdHttpCors} MODULE
+        XrdHttpCorsHandler.cc
+        )
+target_link_libraries(${XrdHttpCors} XrdUtils )
```
amadio: Could you please format this similar to `src/XrdSecunix/CMakeLists.txt`? Thanks.

src/XrdHttpCors/XrdHttpCors.hh
```diff
@@ -0,0 +1,75 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+//------------------------------------------------------------------------------
+
+#ifndef XRD_HTTP_CORS_HEADERS
+#define XRD_HTTP_CORS_HEADERS
```
amadio: The code in general is quite inconsistent about header guard format, however, a lot of it follows the following scheme:
```
src/XrdPfc/XrdPfc.hh:#ifndef __XRDPFC_CACHE_HH__
```
We could follow this for this and the other headers introduced in this PR.

src/XrdHttpCors/XrdHttpCors.hh
```diff
@@ -0,0 +1,75 @@
...
+#include <string_view>
+#include <optional>
+
+
+class XrdOucEnv;
+class XrdOucErrInfo;
+class XrdSysError;
+
+/**
+ * Base class for CORS plugin implementation
```
amadio, line 36: Is it expected to have multiple implementations of this (i.e. a different one in EOS than the one provided by XRootD? I'm trying to understand why this extra abstraction layer is added here.

src/XrdHttpCors/XrdHttpCors.hh
```diff
@@ -0,0 +1,75 @@
...
+   * @param configFN the server configuration file path
+   * @param errP the pointer to the error object
+   * @return 0 if the configuration is successful, 1 otherwise
+   */
+  virtual int Configure(const char * configFN, XrdSysError *errP) = 0;
+  /**
+   * Add trusted origins to the CORS plugin
+   * @param origin the trusted origin to add
+   */
+  virtual void addAllowedOrigin(std::string_view origin) = 0;
```
amadio: Normally `string_view` is used to avoid copies, but here you pretty much must make at least one copy of it to add into `m_origins` in the handler. Given how much `string_view` is misused, I'm more in favor of using `const std::string&` in its place here and in other places in XRootD. `string_view` is actually used very little in the code, so it's probably best to remove current uses and avoid it in general (in favor of using const refs to string):
```sh
xrootd $ git grep -c string_view
src/XrdHttp/XrdHttpHeaderUtils.cc:4
src/XrdHttp/XrdHttpProtocol.cc:1
src/XrdHttp/XrdHttpReq.cc:1
src/XrdHttpTpc/XrdHttpTpcTPC.hh:1
src/XrdOuc/XrdOucJson.hh:2
src/XrdOuc/XrdOucUtils.cc:1
src/XrdOuc/XrdOucUtils.hh:1
tests/scitokens/XrdScitokensCreateJwks.cc:2
tests/scitokens/XrdScitokensCreateToken.cc:1
```

src/XrdHttpCors/XrdHttpCorsHandler.cc
```diff
@@ -0,0 +1,71 @@
...
+}
+
+void XrdHttpCorsHandler::addAllowedOrigin(std::string_view origin) {
+  XrdOucUtils::trim(origin);
+  if(!origin.empty()) {
+    m_origins.emplace(origin);
+  }
+}
+
+XrdVERSIONINFO(XrdHttpCorsget,XrdHttpCorsHandler>);
```
amadio: Looks like there's a typo above (i.e. `>` at the end).

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1644,6 +1658,13 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc,
     ss << m_staticheaders[""];
   }
 
+  if(xrdcors) {
+    auto corsAllowOrigin = xrdcors->getCORSAllowOriginHeader(CurrentReq.m_origin);
+    if(corsAllowOrigin) {
+      ss << *corsAllowOrigin << crlf;
+    }
```
amadio, line 1672: This might be easier to implement if you pass in a reference to `ss` (i.e. the string stream) to the function, which then just adds the origin directly if needed/allowed.

src/CMakeLists.txt
ccaffy: Sure, will do!

src/XrdHttp/XrdHttpReq.cc
ccaffy: It will be used later on for replacing all the code that does:
```cpp
myval = val;
trim(myval);
```

I can however remove it and add it later when I do a commit specifically to refactor that part of the code

src/XrdHttpCors/CMakeLists.txt
ccaffy: Yes, my IDE did it automatically...

src/XrdHttpCors/XrdHttpCors.hh
ccaffy: Well, why not? This opens the code for extension instead of modification

src/XrdHttp/XrdHttpReq.cc
amadio: > I can however remove it and add it later when I do a commit specifically to refactor that part of the code

Yes, this makes a lot more sense, although I think the code is simple enough not to warrant a helper in this case.

src/XrdHttpCors/XrdHttpCorsHandler.cc
ccaffy: Well spotted!

src/XrdHttp/XrdHttpProtocol.cc
ccaffy: It depends, this gives flexibility for the caller of this function. If they don't want to put it in a stringstream but in some other stuff, they can do so?

src/XrdHttpCors/XrdHttpCors.hh
ccaffy: Yes, for this particular function it is not necessary I agree. I'll modify. But for the others?

src/XrdHttpCors/XrdHttpCors.hh
ccaffy: Yes, the guards were generated by my IDE, hence the consistency.
I'll modify the guards to match what is in Pfc ;)

src/XrdHttpCors/XrdHttpCors.hh
amadio: For the others it's up to you, but consider the caveats we've discussed offline.

src/XrdHttp/XrdHttpProtocol.cc
amadio: Yes, if that's the intention, then ok. Feel free to skip this suggestion.

src/XrdHttpCors/XrdHttpCors.hh
ccaffy: I removed the string_view from the signature of the CORS things. For the rest it looks OK!

src/XrdHttp/XrdHttpReq.cc
ccaffy: OK done :)

src/XrdHttpCors/XrdHttpCors.hh
ccaffy: I tried to remove that base class. Unfortunately, this prevents the XrdHttpProtocol to be loaded because it cannot find the plugin's symbols if the server's admin did not configure it.

```
Plugin /lib64/libXrdHttpUtils.so.2: undefined symbol: _ZN11XrdHttpCors9ConfigureEPKcP11XrdSysError protocol libXrdHttp-5.so
```

amadio: Ok, in that case, let's do it in the same way as the other plugins, we can review this for XRootD 6.0 later on.

amadio:
Yes, better to target master, if you plan to have it in XRootD 5.9, as I already picked from devel the things that should go into that version.

ccaffy:
OK I'll target master as I need this for 5.9 definitely

Issue #2564, opened 2025-07-22
[XrdHttp] HTTP protocol violation when PUT request results in an error


bbockelm:
### Operating System

Ubuntu 24.04

### XRootD Version

v5.8.4

### Description

When a file-open fails for a `PUT` request, the XRootD HTTP protocol implementation immediately sends an appropriate error response and then closes the connection.

I believe this, unfortunately, is a violation of the HTTP protocol framing as the client's TCP stack may get a connection reset prior to getting the acknowledgement that the headers were sent.

The latest HTTP 1.1 RFC, RFC 9112 has the following advice in Section 9.6:

> If a server performs an immediate close of a TCP connection, there is a significant risk that the client will not be able to read the last HTTP response. If the server receives additional data from the client on a fully closed connection, such as another request sent by the client before receiving the server's response, the server's TCP stack will send a reset packet to the client; unfortunately, the reset packet might erase the client's unacknowledged input buffers before they can be read and interpreted by the client's HTTP parser.
> 
> To avoid the TCP reset problem, servers typically close a connection in stages. First, the server performs a half-close by closing only the write side of the read/write connection. The server then continues to read from the connection until it receives a corresponding close by the client, or until the server is reasonably certain that its own TCP stack has received the client's acknowledgement of the packet(s) containing the server's last response. Finally, the server fully closes the connection.

Put another way, the client writing a request body can trigger a TCP reset packet to be sent by the server.  If the reset packet is received before the client reads the original server response, the client OS zaps the server response; the client misses the response and just sees a connection reset.

Unfortunately, this is not a theoretical situation.  I stumbled across this while debugging a flaky Pelican unit test; see https://github.com/PelicanPlatform/pelican/issues/2515.

I think the RFC's recommendation is reasonable behavior.  It's not clear, however, if it can easily be implemented within the current framework.

abh3:
Actually, if that is the desired behaviour in all cases, we can change the socket Close() logic to do just that.

abh3:
I don't consider this scarry. Since XRootD protocol doesn't much care how the shutdown is done, i propose we roll it into the  XrdLink so it becomes default behaviour. Notice it took over 15 years to detect this problem so it's not like a deal breaker here. Anyway, the last thing we need is socket code sprinkled throughout the stack.

bbockelm:
> Since XRootD protocol doesn't much care how the shutdown is done, i propose we roll it into the XrdLink so it becomes default behaviour.

Are you sure?

[...]

PR #2565, opened 2025-07-22
Do a clean TLS shutdown for HTTPS

bbockelm:
Unlike XrdTlsSocket, the HTTP protocol did a one-sided TLS shutdown. This saved network a round-trip but at the cost of correctness: if the server shut down the connection after its response while the client was still sending data then the client may recieve a TCP reset prior to reading out the response.

This exact behavior was observed in unit tests and the correct approach is outlined in latest HTTP 1.1 RFC:

https://datatracker.ietf.org/doc/html/rfc9112#name-tls-connection-closure

Basically, we now do the same as `XrdTlsSocket` and perform a bidirectional TLS shutdown.

This takes care of the HTTPS portion of https://github.com/xrootd/xrootd/issues/2564.

### DIFF
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -1963,23 +1963,26 @@ void XrdHttpProtocol::Cleanup() {
 
   if (ssl) {
     // Shutdown the SSL/TLS connection
-    // https://www.openssl.org/docs/man1.0.2/man3/SSL_shutdown.html
-    // We don't need a bidirectional shutdown as
-    // when we are here, the connection will not be re-used.
-    // In the case SSL_shutdown returns 0,
-    // "the output of SSL_get_error(3) may be misleading, as an erroneous SSL_ERROR_SYSCALL may be flagged even though no error occurred."
-    // we will then just flush the thread's queue.
-    // In the case an error really happened, we print the error that happened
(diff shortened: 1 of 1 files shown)


src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1932,23 +1932,26 @@ void XrdHttpProtocol::Cleanup() {
...
-                ERR_remove_thread_state(nullptr);
-            #endif
+            // ret == 0, the unidirectional shutdown was successful; wait for the acknowledgement.
+            ret = SSL_shutdown(ssl);
+            if (ret != 1) {
+                TRACE(ALL, "SSL server failed to receive the SSL shutdown message from the client");
+                ERR_print_errors(sslbio_err);
+            }
         } else {
             //ret < 0, an error really happened.
```
amadio, line 1984: For non-blocking BIO, data may need to be read/written before the shutdown can happen. Please see the [documentation](https://docs.openssl.org/master/man3/SSL_shutdown/#return-values) on return values. Since the code in XrdTlsSocket handles this situation, I think we should handle it here too. For reference, the code I'm referring to is below:

https://github.com/xrootd/xrootd/blob/9f8571358008579b7ba769a9032aca48a54379b9/src/XrdTls/XrdTlsSocket.cc#L725-L765

abh3:
The whole point is to consolidate the code not to replicate things. What would be your suggestion to have this handled in one central place?

PR #2568, opened 2025-07-27
[XrdHttp] Set the sequential I/O flag for simple HTTP requests

bbockelm:
From the HTTP headers, we know exactly whether there will be any non-sequential requests (i.e., vector reads).  If there aren't, provide the hint to the filesystem that we will be doing exclusively sequential reads.

In the original issue (https://github.com/xrootd/xrootd/issues/2155), properly setting this flag resulted in a (frankly shocking!) ~2x improvement in throughput for a shared filesystem.  In the original issue, the plumbing was done at the XRootD layer; this simply pushes through to the HTTP protocol.


abh3 (approved):
Pretty straightforward.

Issue #1795, opened 2022-10-04
Add reflink capability


smithdh:
Add support for reflink via the kXR_dup/kXR_samefs options of kXR_open, e.g. with ioctl_ficlonerange.

smithdh:
I'm updating the ticket to make this useful to track the work on this.

Added a milestone.

At the moment there is an existing branch in this (xrootd) repository: [R6-ficlone](https://github.com/xrootd/xrootd/tree/R6-ficlone) with part of the server side cloning infrastructure.

I had an earlier development here: [xrd_first_ficlone](https://github.com/smithdh/xrootd/tree/xrd_first_ficlone)

PR #2569, opened 2025-07-30
File and range cloning support with ficlone or ficlonerange

smithdh:
Original feature/issue ticket is #1795. I'll add some details here to aid in review shortly.

### DIFF
--- a/src/XrdOss/XrdOss.hh
+++ b/src/XrdOss/XrdOss.hh
@@ -599,8 +599,8 @@ virtual void      Connect(XrdOucEnv &env);
 //! @param  mode   - The new file mode setting.
 //! @param  env    - Reference to environmental information.
 //! @param  opts   - Create options:
-//!                  XRDOSS_coloc  - Colocate file using the placement
-//!                                  information in env "oss.coloc"
+//!                  XRDOSS_coloc  - Colocate file using the URL encoded
+//!                                  path in env "oss.coloc"
 //!                  XRDOSS_mkpath - create dir path if it does not exist.
 //!                  XRDOSS_new    - the file must not already exist.
 //!                  oflags<<8     - open flags shifted 8 bits to the left/
--- a/src/XrdOuc/XrdOucUtils.cc
+++ b/src/XrdOuc/XrdOucUtils.cc
@@ -1620,4 +1620,72 @@ std::string obfuscateAuth(const std::string& input)
   return redacted.append(text + offset);
 }
 
+static bool is_rfc3986_unreserved(unsigned char c)
+{
+  return std::isalnum(c) || c == '-' || c == '_' || c == '.' || c == '~';
+}
+
+/**
+ * This function takes a string and returns the URL encoded string
+ * (rfc3986).
+ *
+ * @param input the string to be encoded
+ * @return the string URL encoded
+ */
+std::string XrdOucUtils::UrlEncode(const std::string &input)
+{
+  static const char hex[] = "0123456789ABCDEF";
+
+  std::string out;
+  out.reserve(input.size() * 3);
+
+  for (unsigned char c: input) {
+    if (is_rfc3986_unreserved(c)) {
+      out.push_back(c);
+    } else {
+      out.push_back('%');
+      out.push_back(hex[c >> 4]);
+      out.push_back(hex[c & 0x0f]);
+    }
+  }
+  return out;
+}
+
+static int from_hex(char c)
+{
+  if (c >= '0' && c <= '9') return c - '0';
+  if (c >= 'A' && c <= 'F') return c - 'A' + 10;
+  if (c >= 'a' && c <= 'f') return c - 'a' + 10;
+  return -1;
+}
+
+/**
+ * This function takes a string and returns the URL decoded string
+ * (rfc3986), also decoding + as ' ' (common in HTML form encoding).
+ *
+ * @param input the URL encoded string
+ * @return the string URL decoded
+ */
+std::string XrdOucUtils::UrlDecode(const std::string &input)
+{
+  std::string out;
+  out.reserve(input.size());
+
+  for (size_t i = 0; i < input.size(); ++i) {
+    if (input[i] == '%' && i + 2 < input.size() &&
+        std::isxdigit(input[i + 1]) &&
+        std::isxdigit(input[i + 2])) {
+      const int hi = from_hex(input[i + 1]);
+      const int lo = from_hex(input[i + 2]);
+      out.push_back(static_cast<char>((hi << 4) | lo));
+      i += 2;
+    } else if (input[i] == '+') {
+      out.push_back(' ');
+    } else {
+      out.push_back(input[i]);
+    }
+  }
+  return out;
+}
+
 #endif
--- a/src/XrdSfs/XrdSfsInterface.hh
+++ b/src/XrdSfs/XrdSfsInterface.hh
@@ -408,8 +408,8 @@ public:
 //! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, SFS_STALL, or SFS_STARTED
 //!
 //! @note When SFS_O_CREATAT is specified, the CGI should contain an element
-//!       oss.coloc=<path> where <path> determines the filesystem in which
-//!       the new file should be created.
+//!       oss.coloc=<path> where <path> is URL encoded and determines the
+//!       filesystem in which the new file should be created.
 //-----------------------------------------------------------------------------
 
 virtual int            open(const char                *fileName,
--- a/src/XrdXrootd/XrdXrootdXeq.cc
+++ b/src/XrdXrootd/XrdXrootdXeq.cc
@@ -1647,7 +1647,7 @@ int XrdXrootdProtocol::do_Open()
    std::string oinfo(opaque ? opaque : "");
    if ((openopts & SFS_O_CREATAT) == SFS_O_CREATAT)
       {std::string coloc = sameFS->XrdSfsp->FName();
-       coloc = "oss.coloc=" + coloc;
+       coloc = "oss.coloc=" + XrdOucUtils::UrlEncode(coloc);
        oinfo += (!oinfo.empty() ? "&" : "") + coloc;
       }
 
--- a/src/XrdOss/XrdOssCreate.cc
+++ b/src/XrdOss/XrdOssCreate.cc
@@ -149,10 +149,12 @@ int XrdOssSys::Create(const char *tident, const char *path, mode_t access_mode,
    bool forcecg = false;
    if (Opts & XRDOSS_coloc)
       {if (crInfo.pOpts & XRDEXP_NOFICL) return -EPERM;
-       const char *coloc = env.Get("oss.coloc");
-       char  coloc_local[MAXPATHLEN+1];
+       const char *colocenv = env.Get("oss.coloc");
+       std::string coloc(colocenv ? colocenv : "");
+       coloc = XrdOucUtils::UrlDecode(coloc);
+       char coloc_local[MAXPATHLEN+1];
        coloc_local[0] = '\0';
-       if (coloc && (retc = GenLocalPath(coloc, coloc_local))) return retc;
+       if ((retc = GenLocalPath(coloc.c_str(), coloc_local))) return retc;
        XrdOssCache_FS *cp = XrdOssCache::Find(coloc_local);
        std::string gpath;
        if (cp) gpath = cp->group + std::string(":") + cp->path;
--- a/src/XrdOuc/XrdOucUtils.hh
+++ b/src/XrdOuc/XrdOucUtils.hh
@@ -153,6 +153,10 @@ static void trim(std::string_view & sv);
 
 static uint8_t touint8_t(const std::string_view sv);
 
+static std::string UrlEncode(const std::string &input);
+
+static std::string UrlDecode(const std::string &input);
+
     XrdOucUtils() {}
     ~XrdOucUtils() {}
 


smithdh:
Hi, after quesiton from @bbockelm I reaslised there's an ordering problem for any potential XrdCl plugin that would like to impliment the cloning functionality. I'll fix when I rebase & fix any merge conflicts.

My idea is to keep the plugin initalisation at Open time. But split the XrdCl::File::SetFileTemplate function at the plugin (and FileStateHandler) level into an "ExportTemplate" and "ImportTemplate" step. i.e. ExportTemplate from source File when user calls SetFileTemplate. But delay ImportTemplate on the target File until just before Open. This was already partially what the PR does; having the ExportTemplate step ensures something reasonable can be done if something strange happens, such as the source File object being destroyed between SetFileTemplate and Open.

src/XrdOfs/XrdOfs.cc
```diff
@@ -821,6 +826,59 @@ int XrdOfsFile::open(const char          *path,      // In
...
+}
+  
+/******************************************************************************/
+
+int XrdOfsFile::Clone(XrdOucCloneSeg cVec[], int n)
+{
+   EPNAME("Clone");
+   int i, j = 0;
+
+do{XrdOfsFile& curFile = *cVec[j].src.ofsFile;
```
amadio: The indentation looks off starting here, could you please fix?

src/XrdOfs/XrdOfs.cc
```diff
@@ -821,6 +826,59 @@ int XrdOfsFile::open(const char          *path,      // In
...
+   int rc = curFile.oh->Select().Clone(&cVec[j], k);
+
+   if (rc < 0)
+      {char etxt[4096];
+       snprintf(etxt,sizeof(etxt),"%s from %s",oh->Name(),curFile.oh->Name()); 
+       return XrdOfsFS->Emsg(epname, error, rc, "clone", etxt);
+      }
+
+   j = i;
+  } while(j < n);
```
amadio: This loop is somewhat confusing to understand. From what I understood you are trying to process multiple elements having the same source file together. However, it seems that if you have 3 elements with same source, then 2 for another, then another 4 for the first one, this code will do three calls to clone, when it could do 2 if `cVec` was sorted based on the source (I didn't check if it is already sorted).

src/XrdOfs/XrdOfs.hh
```diff
@@ -124,6 +125,10 @@ public:
         int            checkpoint(XrdSfsFile::cpAct act,
                                   struct iov *range=0, int n=0);
 
+        int            Clone(XrdSfsFile& srcFile);
+
+        int            Clone(XrdOucCloneSeg cVec[], int n);
```
amadio: I wonder if we shouldn't use `std::vector<XrdOucCloneSeg>` here instead of a raw pointer. Would it hurt performance too much?

src/XrdOuc/XrdOucCloneSeg.hh
```diff
@@ -0,0 +1,57 @@
...
+class XrdOfsFile;
+class XrdOssDF;
+
+struct XrdOucCloneSeg
+{
+union {XrdSfsFile* sfsFile; // Pointer to source file at Sfs level
+       XrdOfsFile* ofsFile; // Pointer to source file at Ofs level
+       XrdOssDF*   ossFile; // Pointer to source file at Oss level
+       int64_t     unxFD;   // The source file descriptor
+      } src;
```
amadio: @abh3 How will we know which of these to use? This looks like it could be easily misused.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -513,6 +515,79 @@ int XrdXrootdProtocol::do_CKsum(char *algT, const char *Path, char *Opaque)
...
+// a limit on it's size. We do this to be able to reuse the data buffer to 
+// prevent cross-cpu memory cache synchronization.
+//
+   if (clVecNum > XrdProto::maxClonesz)
+      return Response.Send(kXR_ArgTooLong, "Clone vector is too long");
+
+// Allocate a new cloe vector
+//
+   std::unique_ptr<XrdOucCloneSeg[]> u_clVecP(new XrdOucCloneSeg[clVecNum]);
+   XrdOucCloneSeg* clVec = u_clVecP.get();
```
amadio: Looks like `std::vector` would fit better here too (passing down a reference to it in other places).

src/XrdOfs/XrdOfs.cc
```diff
@@ -859,13 +869,23 @@ int XrdOfsFile::Clone(XrdOucCloneSeg cVec[], int n)
...
+   XrdOfsFile& curFile = *cVec[j].src.ofsFile;
    XrdOssDF*   ossFile = &curFile.oh->Select();
+   cVec[j].src.ossFile = ossFile;
 
-   for (i = j+1; i < n && cVec[i].src.ofsFile == &curFile; i++)
+   for (i = j+1; i < n && cVec[i].src.sfsFile == sf; i++)
        {cVec[i].src.ossFile = ossFile;}
    int k = i - j;
-   int rc = curFile.oh->Select().Clone(&cVec[j], k);
+   int rc = oh->Select().Clone(&cVec[j], k);
```
amadio: Should/could this not be `ossFile->Clone()`?

src/XrdCl/XrdClFileStateHandler.cc
```diff
@@ -3035,8 +3097,16 @@ namespace XrdCl
 
     req->requestid = kXR_open;
     req->mode      = self->pOpenMode;
-    req->options   = self->pOpenFlags;
+    req->options   = (self->pOpenFlags & 0xffff);
```
amadio, line 3103: I wonder why this change was necessary.

src/XrdCl/XrdClFileStateHandler.hh
```diff
@@ -128,11 +144,34 @@ namespace XrdCl
       //------------------------------------------------------------------------
       static XRootDStatus Open( std::shared_ptr<FileStateHandler> &self,
                                 const std::string                 &url,
-                                uint16_t                           flags,
+                                OpenFlags::Flags                   flags,
```
amadio, line 147: This is a pretty big change, is it not possible to stay with the previous `uint16_t`? Otherwise, we may as well change `uint16_t mode` to `mode_t mode` since we are breaking this.

src/XrdCl/XrdClFileSystem.hh
```diff
@@ -104,8 +105,10 @@ namespace XrdCl
       SeqIO    = kXR_seqio,         //!< File will be read or written sequentially
       PrefName = kXR_prefname,      //!< Hostname response is prefered, applies
                                     //!< only to FileSystem::Locate
-      IntentDirList = kXR_4dirlist  //!< Make sure the server knows we are doing
+      IntentDirList = kXR_4dirlist, //!< Make sure the server knows we are doing
                                     //!< locate in context of a dir list operation
+      Dup      = kXR_dup<<16,       //!< Open file duplicating content from another
+      Samefs   = kXR_samefs<<16     //!< Open file on the same filesystem as another
```
amadio, line 114: This is not very pretty, since only two flags out of several are actually shifted, and as commented at some other point, append is not listed. I wonder now if changing the type on `Open` will break more than we realize.

src/XrdCl/XrdClPlugInInterface.hh
```diff
@@ -275,6 +289,30 @@ namespace XrdCl
...
         return false;
       }
+
+      //------------------------------------------------------------------------
+      //! Obtains an object that contains information regarding the current
+      //! file that another instance of the plugin can use via
+      //! OpenUsingTemplatet or Clone. The returned object should be of a
+      //! class derived from ExportedFileTemplate, the exact type of which could
+      //! be used to ensure the source plugin is the same the destination. In
+      //! the case of Clone() it sould be considered that the source file object
```
amadio: There is a typo: "sould"

src/XrdCl/XrdClUtils.hh
```diff
@@ -261,6 +261,18 @@ namespace XrdCl
         return protver >= kXR_PROTXATTVERSION;
       }
 
+      //------------------------------------------------------------------------
+      //! Check if given server supports kXR_clone and kXR_samefs
+      //------------------------------------------------------------------------
+      inline static bool HasKSameFS( const XrdCl::URL &url )
+      {
+        if( url.IsLocalFile() ) return true;
```
amadio, line 269: This is not true, not all local filesystems support file cloning. This should probably perform an actual check for it.

src/XrdCl/XrdClXRootDTransport.cc
```diff
@@ -1199,6 +1200,19 @@ namespace XrdCl
         break;
       }
 
+      case kXR_clone:
+      {
+        uint32_t numChunks  = (req->clone.dlen)/sizeof(XrdProto::clone_list);
+         XrdProto::clone_list *dataChunk = ( XrdProto::clone_list*)( msg + 24 );
```
amadio: If `24` is `sizeof(something)`, better use that here instead of the raw value.

src/XrdCl/XrdClFileStateHandler.hh
smithdh: hi @amadio thanks. indeed perhaps this could cause some source incompatability between R5 and R6 for no real reason: we now have 2 more open flags than can be expressed by uint16_t, but neither of the new ones are relevant to this Open. How about I revert this to uint16_t, can keep the other type for the new OpenUsingTemplate() (not withstanding any changes for other comments). If we get more flags in the future we can add a new Open() (these aren't virtual functions, so we should just be able to add more if needed).

src/XrdCl/XrdClFileSystem.hh
smithdh: There's now 2 16 bit fields in the protocol for various flags. So I shifted them to fit in a 32 bit flags set (hence the shifting on the 2 new ones). What could you see as alternative? Maybe have a 'Flags2' type and a second flags argument on OpenWithTemplate()?

src/XrdCl/XrdClUtils.hh
smithdh: This test is to check whether it is sending to something that undestands the flag (the "same fs" flag to open), as if it doesn't understand the flag it would silently ignore it, the open could work but not doing what was intended.

src/XrdCl/XrdClFileSystem.hh
amadio: At least for `FICLONE` instead of `FICLONERANGE`, it seems to me that if we could have just `Clone` taking the template file as parameter would allow to not have these flags at `Open`. Then we could have a `CloneRanges`, taking the file to be cloned and the ranges together as well, what do you think? That is, we'd have something like this:
```cpp
XrdCl::File f;
f.Open(...);
auto new_f = f.Clone();
```
or
```
auto new_f = f.CloneRanges(ranges);
```
with `ranges` being a `std::vector<XrdOucCloneSec>`. Although I can see that in `f.Clone()` itself we'd need the flag to open in the same filesystem, so maybe the point is moot.

src/XrdCl/XrdClUtils.hh
amadio: What I meant is that if a file is local, it could still not support cloning, even if the flag is understood. This could happen in case the file is on a filesystem which doesn't support copy on write. However, I see that there is a check for this case in the code, so it's fine. When we call `ioctl`, we may get `ENOTSUPP`, which will get returned to the caller, so I think we can resolve this one.

src/XrdCl/XrdClFileStateHandler.hh
smithdh: hi. sorry, on second check the uint16_t to OpenFlags change is mostly concerning FileStateHandler which is internal to XrdCl. The XrdFile::Open (what the user calls) was already using OpenFlags as the type - so I'd expect that (the intent is) the change should not affect the user. There is a visible change to a member of OpenInfo from XrdClMonitor.hh. Perhaps this would be better as an additional member (e.g. uint16_t oFlags2).

src/XrdCl/XrdClFileStateHandler.cc
smithdh: hi. Previosly this was a uint16_t assigned to a kXR_uint16, now the source is wider. I don't think there's any need to do the masking, it will get narrowed. But I thought to emphasise in this assignment that it is only assigning some of the flags. Do you think it's better without?

src/XrdCl/XrdClFileStateHandler.hh
amadio: Then only the other question remains, should we move to use `mode_t` for the mode?

src/XrdCl/XrdClFileSystem.hh
smithdh: hi. thanks. I hadn't specially pursued that (i.e. allowing a Clone() method in XrdCl to create and return a new File in one step, where we'd automatically include the right flags in the request), altough it might save exposing new flags. However at the protocol level, 5.2.0 documents that kXR_clone is something is done on an already open file, and the kXR_dup, kXR_samefs are options for kXR_open requests. So it seemed best to reflect that in XrdCl; this also allows a user to express Open new file on same filesystem as an existing one, but without any particular reference to cloning.

src/XrdOfs/XrdOfs.cc
smithdh: I believe this is correct as in the diff: The Clone(cVec[],k) is called on the Oss for the target file i.e. the one that has the kXR_clone directed at it and is the one which will be modified. The ossFile member(s) in the cVec[] are the XrdOssDFs of the source file(s).

src/XrdOfs/XrdOfs.cc
smithdh: looks consistent with some other do's in some other components, e.g. in XrdOfsTPCAuth::RunTTL().

src/XrdCl/XrdClFileStateHandler.hh
smithdh: I guess you're thinking of the uint16_t to mode_t (actually probably Access::Mode? - this too should concern FileStateHandler), rather than the Access::Mode of XrdCl::File? The former should not be visible. Needs to change storage for pOpenMode in FileStateHandler (and will narrowed when wirtten into the kXR_open request). I could note that as another PR for this week (for R6) if you would like.

src/XrdCl/XrdClPlugInInterface.hh
smithdh: fixed, pushed.

src/XrdCl/XrdClXRootDTransport.cc
smithdh: fixed, pushed.

src/XrdCl/XrdClFileStateHandler.hh
smithdh: pushed change to revert type of OpenInfo::oFlags, introduced uint16_t OpenInfo::oFlags2 for upper 16 bits.

src/XrdCl/XrdClFileSystem.hh
amadio: I see, thanks for clarifying.

src/XrdOfs/XrdOfs.cc
amadio: Ok, let's keep it as it is then.

smithdh:
Hi Guilherme, I've added a test as an extra commit. (Should indicate success either if the cloning gives not supported, or it is successful). I also noticed a not-apparent problem with one of the new commits that we're rebased on; missing a name for kxr_clone request in the XProtocol::reqName function. So I added another commit with a fix for that (tried to also add some protection there, for missed names in the future).

I made a small change in XrdOssSys::Features too (this is squashed in, it's inside one of the existing commits). That concerned Apple builds; where I didn't implement cloning (no FICLONE ioctl). However it can still honor the kxr_samefs option, so to allow that to work I set XRDOSS_HASFICL for the Oss independent of the availability of the ioctl. (The clone() methods will appropriately return not supported). [Mac OS does have clonefile, and variants, but I didn't think to try adding it. It works slightly differently to the ioctl, but could probably be made to work for whole file, but not range cloning].

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -399,8 +402,13 @@ public:
 //! @param  createMode - The file's mode if it will be created.
 //! @param  client     - Client's identify (see common description).
 //! @param  opaque     - path's CGI information (see common description).
+//! @param  file       - path's CGI information (see common description).
```
abh3: Why is this here duplicating the preceding line? There is no parameter "file".

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -513,6 +515,79 @@ int XrdXrootdProtocol::do_CKsum(char *algT, const char *Path, char *Opaque)
...
+
+// Compute number of elements in the clone vector and make sure we have no
+// partial elements.
+//
+   clVecNum = clVecLen / sizeof(XrdProto::clone_list);
+   if ( (clVecNum <= 0) ||
+         (clVecNum*(int)sizeof(XrdProto::clone_list) != clVecLen) )
+      return Response.Send(kXR_ArgInvalid, "Clone vector is invalid");
+
+// Make sure that we can copy the read vector to our local stack. We must impose
```
abh3: Copy/paste issue. "read vector" is actually "clone vector".

src/XrdXrootd/XrdXrootdXeq.cc
abh3: Technically, yes but I'm not a stickler on this since it's handled correctly.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -1516,20 +1612,42 @@ int XrdXrootdProtocol::do_Open()
...
       openopts|= SFS_O_NOTPC;
 
+// If needed add the colocation pointer. We add it to the end of the opaque
+// string and rely on later use of XrdOucEnv to keep this value, ignoring
+// any possible user supplied key/value pair with our key.
+//
+   std::string oinfo(opaque ? opaque : "");
+   if ((openopts & SFS_O_CREATAT) == SFS_O_CREATAT)
+      {XrdOucEnv te;
+       te.PutPtr("sfs.coloc*", sameFS->XrdSfsp);
```
abh3: Well I am very queasy about passing pointers around this way especially via  CGI. It also conflicts with the documentation for open since it stated "oss.coloc=<path>" and this is not a path. I would prefer that you get the path name from the Sfs object using saneFS->XrdSfsp->FName() and passing that along.

src/XrdOfs/XrdOfs.cc
```diff
@@ -638,10 +639,24 @@ int XrdOfsFile::open(const char          *path,      // In
                      "+ofs_open: failed to enter file into posc queue");
           }
 
+       std::string cinfo = info;
+       // If placement information is present provide a hint to the oss plugin
+       //
+
+       XrdOucEnv Create_Env(cinfo.c_str(),0,client);
+       if ((open_mode & ~SFS_O_CREAT) & SFS_O_CREATAT)
+          {XrdSfsFile *sfp = (XrdSfsFile*)Open_Env.GetPtr("sfs.coloc*");
```
abh3: Obviously this will need to change if you pass a path. But why is Ofs worried about this anyway?

src/XrdOfs/XrdOfs.cc
```diff
@@ -821,6 +836,69 @@ int XrdOfsFile::open(const char          *path,      // In
...
+   if (!cVec[j].src.ofsFile)
+      {char etxt[4096];
+       snprintf(etxt,sizeof(etxt),"%s",oh->Name());
+       return XrdOfsFS->Emsg(epname, error, -EINVAL, "clone", etxt);
+      }
+   XrdOfsFile& curFile = *cVec[j].src.ofsFile;
+   XrdOssDF*   ossFile = &curFile.oh->Select();
+   cVec[j].src.ossFile = ossFile;
+
+   for (i = j+1; i < n && cVec[i].src.sfsFile == sf; i++)
```
abh3: I think I see what you are trying to do here which is about optimizing the clone process. But, frankly, it's not worth it. This version of clone will be very lightly used and likely with very few elements, I would prefer an unoptimized version that simply does one element at a time with yes, repeated calls to get values that you already have (i.e. file pointer). It would demystify this code and certainly make it shorter. All good for maintainability. The overhead is immaterial here.

src/XrdOfs/XrdOfs.cc
abh3: I don't think it matters, I'd prefer the dumb solution here.

src/XrdOss/XrdOss.hh
```diff
@@ -111,6 +112,28 @@ virtual int     StatRet(struct stat *buff) {return -ENOTSUP;}
...
+//-----------------------------------------------------------------------------
+//! Clone contents of a file from one or more oher files.
+//!
+//! @param  cVec  - A vector of struct XrdOucCloneSeg describing the action.
+//! @param  n     - The number of elements in cVec.
+//!
+//! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+//-----------------------------------------------------------------------------
+
+virtual int     Clone(XrdOucCloneSeg cVec[], int n) {return -ENOTSUP;}
```
abh3: Adopting the dumb simple approach will allow you to eliminate 'n' as this method would do one clone segment as a time.

src/XrdOss/XrdOssApi.cc
```diff
@@ -254,6 +259,81 @@ int XrdOssSys::GenRemotePath(const char *oldp, char *newp)
...
+  if (ioctl(fd, FICLONE, srcFile.getFD())==-1)
+    return -errno;
+
+  return XrdOssOK;
+#else
+  return -ENOTSUP;
+#endif
+}
+
+int XrdOssFile::Clone(XrdOucCloneSeg cVec[], int n)
```
abh3: Again, adopting the dumb approach would greatly simplify this method. No loop no worrying what your indexing, low maintainability.

src/XrdOss/XrdOssConfig.cc
```diff
@@ -2166,6 +2166,7 @@ void XrdOssSys::List_Path(const char *pfx, const char *pname,
          ss += (flags & XRDEXP_PURGE    ? " purge"   : " nopurge");
          ss += (flags & XRDEXP_RCREATE  ? " rcreate" : " norcreate");
          ss += (flags & XRDEXP_STAGE    ? " stage"   : " nostage");
+         ss += (flags & XRDEXP_NOFICL   ? " noficl"  : " ficl");
```
abh3, line 2169: Could you write up any new configurable. Nothing fancy, just what you added. and either include it the pull request or send it to me or both.

src/XrdOssCsi/XrdOssCsi.hh
```diff
@@ -76,32 +76,32 @@ friend class XrdOssCsiFileAio;
 friend class XrdOssCsiFileAioJob;
 public:
 
-virtual int     Close(long long *retsz=0) /* override */;
-virtual int     Open(const char *, int, mode_t, XrdOucEnv &) /* override */;
+virtual int     Close(long long *retsz=0) override;
```
abh3: I know it was a natural thing to do but this change should have been a separate pull request. I won't quibble unless @amadio does.

src/XrdOuc/XrdOucCloneSeg.hh
```diff
@@ -0,0 +1,57 @@
...
+class XrdSfsFile;
+class XrdOfsFile;
+class XrdOssDF;
+
+struct XrdOucCloneSeg
+{
+union {XrdSfsFile* sfsFile; // Pointer to source file at Sfs level
+       XrdOfsFile* ofsFile; // Pointer to source file at Ofs level
+       XrdOssDF*   ossFile; // Pointer to source file at Oss level
+       int64_t     unxFD;   // The source file descriptor
```
abh3: You may wish to use std::variant here instead of a contextual union. Alternatively you really don't need a union at all since the only thing that matter is the unxFD in the end. That could have been supplied way back in XrdXrootdXeq::do_Clone() since you could have simply called XrdSfsFile::fctl(SFS_FCTL_GETFD,...) which return the file descriptor. This would have been the simplest solution and the whole calling suquence would have been far shorter. I didn't see any real dependency on having any intermediate objects handy as one made a path to getting the final file descriptor. So, I would strongly encourage adopting this simplification which eliminates about 30% of the code.

src/XrdOuc/XrdOucCloneSeg.hh
abh3: Well you don't as it's all context driven. I came up with two alternatives, std::variant or the grand simplification that just requires an fd entry, no unions needed.

src/XrdOss/XrdOssConfig.cc
```diff
@@ -567,7 +567,7 @@ void XrdOssSys::ConfigCache(XrdSysError &Eroute, bool pass2)
 // Handle default settings
 //
    if (DirFlags & XRDEXP_PFCACHE)
-      {DirFlags |=  XRDEXP_READONLY;
+      {DirFlags |=  XRDEXP_READONLY | XRDEXP_NOFICL;
```
abh3, line 570: I suppose that's the safest default.

src/XrdOssCsi/XrdOssCsi.hh
amadio: @abh3 This does look unrelated to `FICLONE`, but a separate commit is ok, it doesn't have to be a full separate pull request. The guideline is [one logical change per commit](https://github.com/xrootd/xrootd/blob/master/docs/CONTRIBUTING.md#make-separate-commits-for-logically-separate-changes).

smithdh: i'll leave out the uncommenting of override; can be for other pr(s).

smithdh:
Hi. Pushed; rebase to latest master, fix few conficts during that. Thanks for latest review. I hope I've fixed all Andy's first round chage request/suggestions. (Going for 'grand simplification' route). Altough still missing is requested doc change concerning new config options (for space/export). Could you remind me where the doc source I should write or diff against is?

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -515,6 +517,94 @@ int XrdXrootdProtocol::do_CKsum(char *algT, const char *Path, char *Opaque)
...
+        if (!srcFile || currFH != fh.handle)
+           {currFH = fh.handle;
+            if (!(fP = FTab->Get(currFH)))
+               return Response.Send(kXR_FileNotOpen,
+                                    "clone does not refer to an open src file");
+            srcFile = fP->XrdSfsp;
+           }
+
+        int fdNum;
+        if (srcFile->fctl(SFS_FCTL_GETFD, 0, myError) != SFS_OK)
```
abh3, line 582: Just for future reference, for proxy server we will need a different fctl command as proxy servers refuse to return ephemeral FD's since they cannot be used directly. However, if it knows you want to proxy a request using an ephemeral FD it would could return it. So, the upshot is that we will not support proxying ficlone() until sometime later. This is OK, let's see how this works first.

src/XrdOuc/XrdOucCloneSeg.hh
```diff
@@ -0,0 +1,49 @@
...
+//! XrdOucCloneSeg
+//!
+//! The struct defined here is a generic data structure that is used whenever
+//! we need to pass a vector of file segments that need to be cloned into
+//! anothe file.
+//-----------------------------------------------------------------------------
+
+struct XrdOucCloneSeg
+{
+      int64_t      srcFD;    // The source file descriptor
```
abh3: Since file descripts are defined as int32 you should change this to int32 followed by a int32 reserved filed to keep alignment. Since we can then use this extra field for possible future enhancements, add a constructor that zeroes it out.

src/XrdOfs/XrdOfs.cc
```diff
@@ -821,6 +826,58 @@ int XrdOfsFile::open(const char          *path,      // In
...
+}
+  
+/******************************************************************************/
+
+int XrdOfsFile::Clone(XrdOucCloneSeg cVec[], int n)
+{
+   EPNAME("Clone");
+   int i, j = 0;
+
+do{
```
abh3: I think this should be a passthrough. Let the Oss optimize this as needed. So, pass the whole vector to the Oss. We will need to do that in any case for proxy servers as they need to see the whole vector. As it stands now, this prevents a reasonable implementation of proxy cloning.

src/XrdOfs/XrdOfs.hh
abh3: I doubt it would matter much given the other overheads. I'm neutral on it.

src/XrdOss/XrdOssApi.cc
```diff
@@ -254,6 +259,77 @@ int XrdOssSys::GenRemotePath(const char *oldp, char *newp)
...
+     return -EBADF;
+
+  if (n == 0)
+    return XrdOssOK;
+
+   int i, j=0;
+  do{
+   int sFd = cVec[j].srcFD;
+
+   for (i = j+1; i < n && cVec[i].srcFD == sFd; i++);
```
abh3: Given that the ioctl() can only do one element at a time, I don't think you need anything more that a simple loop that goes through the coloc vector and issues the ioctl() for each element. So, this can be dramatically simplified by using a single simple iterative loop.

src/XrdOss/XrdOssApi.cc
```diff
@@ -254,6 +259,77 @@ int XrdOssSys::GenRemotePath(const char *oldp, char *newp)
...
+
+   for(int k=j;k<i;k++) 
+      {struct file_clone_range fr;
+       fr.src_fd = cVec[k].srcFD;
+       fr.src_offset = cVec[k].srcOffs;
+       fr.src_length = cVec[k].srcLen;
+       fr.dest_offset = cVec[k].dstOffs;
+       int rc = ioctl(fd, FICLONERANGE, &fr);
+       if (rc < 0) return -errno;
```
abh3: If you are worried that you can't issue a reasonable message upon the error, use the new getErrMsg() API associated with the target file to relay context. Now that I said that, I see that was only fully implemented for XrdPss but with file cloning we need to implement it for XrdOss so that error text can be returned as simple errno's don't cut it any more. I'd suggest doing the simplification and delay the full getErrMsg() extension for now.

smithdh:
I've reaslied that the current code for passing XrdSfsp->FName() via "oss.coloc" CGI will fail if the filename has an '&' in it, the name will be effectively truncated. I had a look around but I don't think we've got any established practice of encoding or escaping the value to avoid this in the Xrootd partions of our code, but I could take the XrdHttp approach (i.e. quote() from XrdHttpUtils.cc). But I'd like to check if there's a alternative suggesiton of how to handle this. Do you see any?

smithdh:
Hi Andy. I think I've addressed everything from the previous reviews; but sorry if I've missed something. Do you think you could re-review to see if there's more?

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1620,4 +1620,72 @@ std::string obfuscateAuth(const std::string& input)
...
+  std::string out;
+  out.reserve(input.size() * 3);
+
+  for (unsigned char c: input) {
+    if (is_rfc3986_unreserved(c)) {
+      out.push_back(c);
+    } else {
+      out.push_back('%');
+      out.push_back(hex[c >> 4]);
+      out.push_back(hex[c & 0x0f]);
```
amadio, line 1648: What happens with `UrlDecode(UrlEncode("Genève"))`? Would it work? I assume something like `ł` (used in Polish) would not at least, so I wonder if we should think about how to support these things in the future, since they may be used in filenames be people.

smithdh: Hi. Thanks for looking at this: I think we can assume the string (i.e. the filename in the std::string passed to the encode/decode) is an ASCII-compatible encoding. i.e. that the fundamental unit of encoding is a byte, with ASCII (0x00-0x7f) encode/decoding identically. I believe that's a requirement of a posix filename, altough I don't think we discuss this in the xrootd docs. UTF-8 is a type of ascii-encoding (as are many other encodings too). So if I do something like:

```
xrdcp /etc/hosts root://myserver.cern.ch//Genèveł.hosts
```

The filename will be encoded somehow and passed to xrdcp; on my alma9 this is UTF-8. Which is

```
00000000  47 65 6e c3 a8 76 65 c5  82 2e 68 6f 73 74 73     |Gen..ve...hosts|
```

this does mean that the std::string of above has length() 15, altough there are only 13 code-points. I can copy the file out ok, specifying the source filename in the excepted way. And xrdfs ls output was correctly rendered. (This of course doesn't relate directly to these UrlEncode/UrlDecode functions).

The UrlEncode/UrlDecode will bytewise encode/decode, giving the %c3%a8 for the bytes at [3,4] above etc. Decode reverses that so it should be ok, I believe.

amadio: Great! Thanks for trying it!

abh3 (approved):
Looks like all my concerns have been addressed.

PR #2570, opened 2025-08-06
Rename after redirection

### LINKED ISSUE #2550: File rename over davs/https fails when using token authentication and redirection.
### Operating System

Rocky Linux 8 (server), Alma Linux 9 (Client)

### XRootD Version

5.8.3

### Description

When trying to use token based authentication on our xroot 5.8.3 storage, consisting of a redirector and 6 hosts, we found that file renames over https/davs failed with 403 permission denied errors. All other processes (write, read, delete etc) worked correctly, and renames with the root:// protocol also work.

Furthermore, if we bypass the redirector and instead try a rename by connecting directly to one of the xroot servers it works without issue. This led to some issues with debugging against other "direct" hosts where we couldn't replicate what we were seeing.

We have tested with both the gfal tools and raw curl, and also tested against 5.8.2 and 5.7.3 hosts at RAL (thanks to Jyothish for providing them) which saw the same issue.

The final error from the curl attempts matches lines from the xroot server logs:
`Unable to renaming to (no overwrite) <filename>; permission denied`

Which stands out due to the odd grammar.

Thanks,
Matt

\#create token
export BEARER_TOKEN=`oidc-token dteam1 --scope="storage.create:/ storage.read:/ storage.modify:/"`
\#curl rename against an xrootd redirector host
curl --capath /etc/grid-security/certificates -L  -vvv -H "Authorization: Bearer ${BEARER_TOKEN}" -XMOVE -H 'Destination:https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename4' https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename3
\# Get 403 error

### Additional Information

\# verbose output of the curl rename with token redacted
$ curl --capath /etc/grid-security/certificates -L  -vvv -H "Authorization: Bearer ${BEARER_TOKEN}" -XMOVE -H 'Destination:https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename3' https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename2
*   Trying 2001:630:80:2fd::aa22:a765:1094...
* Connected to xgate.hec.lancs.ac.uk (2001:630:80:2fd::aa22:a765) port 1094 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
*  CApath: /etc/grid-security/certificates
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Request CERT (13):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Unknown (23):
* TLSv1.3 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS header, Unknown (23):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=UK; O=eScience; OU=Lancaster; L=Physics; CN=xgate.hec.lancs.ac.uk
*  start date: Feb 14 13:03:38 2025 GMT
*  expire date: Mar 16 13:03:38 2026 GMT
*  subjectAltName: host "xgate.hec.lancs.ac.uk" matched cert's "xgate.hec.lancs.ac.uk"
*  issuer: C=UK; O=eScienceCA; OU=Authority; CN=UK e-Science CA 2B
*  SSL certificate verify ok.
* TLSv1.2 (OUT), TLS header, Unknown (23):
> MOVE /cephfs/grid/dteam/matttest/test1M.rename2 HTTP/1.1
> Host: xgate.hec.lancs.ac.uk:1094
> User-Agent: curl/7.76.1
> Accept: */*
> Authorization: Bearer REDACTED 
> Destination:https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename3
> 
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* TLSv1.2 (IN), TLS header, Unknown (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 307 Temporary Redirect
< Connection: Keep-Alive
< Server: XrootD/v5.8.3
< Content-Length: 0
< Location: https://stor004.hec.lancs.ac.uk:1095/cephfs/grid/dteam/matttest/test1M.rename2?authz=Bearer%20TOKENREDACTED
< 
* Connection #0 to host xgate.hec.lancs.ac.uk left intact
* Clear auth, redirects to port from 1094 to 1095Issue another request to this URL: 'https://stor004.hec.lancs.ac.uk:1095/cephfs/grid/dteam/matttest/test1M.rename2?authz=Bearer%20
*   Trying 2001:630:80:2fd::aaf1:4cd6:1095...
* Connected to stor004.hec.lancs.ac.uk (2001:630:80:2fd::aaf1:4cd6) port 1095 (#1)
* ALPN, offering h2
* ALPN, offering http/1.1
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
*  CApath: /etc/grid-security/certificates
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Request CERT (13):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Unknown (23):
* TLSv1.3 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS header, Unknown (23):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: C=UK; O=eScience; OU=Lancaster; L=Physics; CN=stor004.hec.lancs.ac.uk
*  start date: Apr 29 14:41:06 2025 GMT
*  expire date: May 29 14:41:06 2026 GMT
*  subjectAltName: host "stor004.hec.lancs.ac.uk" matched cert's "stor004.hec.lancs.ac.uk"
*  issuer: C=UK; O=eScienceCA; OU=Authority; CN=UK e-Science CA 2B
*  SSL certificate verify ok.
* TLSv1.2 (OUT), TLS header, Unknown (23):
> MOVE /cephfs/grid/dteam/matttest/test1M.rename2?authz=Bearer%20TOKENREDACTED HTTP/1.1
> Host: stor004.hec.lancs.ac.uk:1095
> User-Agent: curl/7.76.1
> Accept: */*
> Destination:https://xgate.hec.lancs.ac.uk:1094/cephfs/grid/dteam/matttest/test1M.rename3
> 
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* TLSv1.2 (IN), TLS header, Unknown (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 403 Forbidden
< Connection: Close
< Server: XrootD/v5.8.3
< Content-Length: 98
< 
* TLSv1.2 (IN), TLS header, Unknown (23):
* Closing connection 1
* TLSv1.2 (OUT), TLS header, Unknown (23):
* TLSv1.3 (OUT), TLS alert, close notify (256):
Unable to renaming to (no overwrite) /cephfs/grid/dteam/matttest/test1M.rename3; permission denied

dynamic-entropy:
Fix #2550 

In https://github.com/xrootd/xrootd/pull/2570/commits/8ce935f370c1b3b4c786585051b2077d6c71ed3c, the desired goal is to add the CGI parameter `authz` to the "Destination" of the `MOVE` operation before it is evaluated for access permissions. 

After a redirection. the authorization header takes the form of a CGI parameter and is no longer present in the redirected request. 

Optionally, one can instruct curl to send the authorization header on redirect with [--location-trusted](https://curl.se/docs/manpage.html#--location-trusted); however, this is not the default behaviour.  
Note that the way things are implemented currently, the use of  `--location-trusted` will not work either after redirection, (as tested here https://github.com/xrootd/xrootd/issues/2550#issuecomment-3144096476 since the presence of `authz` in the resource url will prevent its copying into the `header2cgi` string.  https://github.com/xrootd/xrootd/blob/8ac19b1d2b74521acff9ed0200052a2e373092cc/src/XrdHttp/XrdHttpReq.cc#L211-L218


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1722,6 +1719,19 @@ int XrdHttpReq::ProcessHTTPReq() {
     }
     case XrdHttpReq::rtMOVE:
     {
+        // For move operation query string parameters in resource should also
+        // apply to url in destination header
+        if (opaque) {
+            const char *authz_val = opaque->Get("authz");
```
dynamic-entropy: We only operate on the `authz` parameter to avoid any side effects - i.e., I am not sure if we should or need to copy other CGI parameters too.

tests/cluster/test.sh
```diff
@@ -190,6 +190,36 @@ for host in "${!hosts[@]}"; do
        done
 done
 
+# Renaming operation on meta manager fail with 501
+# Not sure why this is the expected behaviour
+# https://github.com/xrootd/xrootd/blob/8ac19b1d2b74521acff9ed0200052a2e373092cc/src/XrdHttp/XrdHttpReq.cc#L1746-L1752
```
bbockelm, line 195: From discussion:
- There is a check in the code to see if the hostname of the server and the server in the hostname of the `Destination` header URL (the target of the `MOVE` operation) are the same.
- If the server is a manager, then the check is enforced.  After the redirect to a server, the `Destination` still points to the original manager's hostname, meaning the hostnames no longer match.
- Unfortunately, the code only checks for a manager.  In the case of a meta-manager, there's two layers of redirection -- from the meta-manager to the manager then from manager to server.  After the first redirect, the code check no longer passes and is required since the target of the first redirect is a manager.

So, what to do?  Options:
1. Drop the check completely.  Is it even necessary?  We don't enforce it for servers due to the issue above.  The danger of dropping the check is that you could do the rename locally even though that's not what the client had intended.
2. Require the client to "do the right thing" and rewrite the `Destination` header on redirect.  This is effectively what we do now.
3. On redirect, add a `destination` query parameter to the URL which overrides the `Destination` header.  This would be in the control of the redirector (manager or metamanager), allow us to keep the check of the hostname (allowing us to enforce it for servers as well), and solve the problem for multiple layers of redirection.

I think I lean toward (3) myself.  However, it sure seems like this is outside the purview of the original issue.  Maybe we open a new one?

src/XrdHttp/XrdHttpReq.cc
bbockelm: I think it's OK to just work on the `authz` parameter.  However, below we just look for the substring `authz=` inside the URL (which may be in multiple places, such as a query parameter named `authauthauthz=`).  Let's have a helper function that checks for the parameter using URL parsing rules.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -749,6 +749,15 @@ void XrdHttpReq::addCgi(const std::string &key, const std::string &value) {
   hdr2cgistr.append(value);
 }
 
+std::unique_ptr<XrdOucEnv> XrdHttpReq::parseDestination() {
```
dynamic-entropy: @bbockelm Any suggestions for this parse function? 
 - Should there be a member `destination_opaque` in XrdHttpReq as opposed to returning something - my idea is to not bloat the object unless we need the parsed data again. 
 - I opted to keep it as a member of XrdHttpReq as opposed to something within XrdHttpUtils

dynamic-entropy (commented):
@bbockelm made the requested changes for parsing. 

Will defer to @amadio and @ccaffy for comments on options for handling multiple redirects through a manager node.

src/XrdHttp/XrdHttpReq.cc
bbockelm: This is too heavy a hammer for my taste.

Fundamentally, you really just need to know if `authz` is present in the URL.  Parsing this to a hash map of strings-to-strings is an expensive operation with multiple allocations and copies.

[I'd prefer an approach like this one](https://github.com/PelicanPlatform/xrdcl-pelican/blob/main/src/XrdClS3/XrdClS3Factory.cc#L291-L318) where you just step through the string piece by piece and return true/false depending on the presence of `authz`.

dynamic-entropy: Thankyou! 
Changed from parsing destination url to only finding the "authz" param, copying logic from https://github.com/PelicanPlatform/xrdcl-pelican/blob/main/src/XrdClS3/XrdClS3Factory.cc#L291-L318

ccaffy (changes requested):
Is it actually OK that this goes to XrdHttpUtils so that unit tests are created with different cases? Including tests like `authauthauthz=` for example

tests/XrdHttpTests/XrdHttpTests.cc
```diff
@@ -606,4 +606,50 @@ TEST(XrdHttpTests,encodeOpaqueTest) {
   for(auto [decoded,encoded]: decodedEncodedOpaque) {
     ASSERT_EQ(encoded,encode_opaque(decoded));
   }
-}
\ No newline at end of file
+}
+
+TEST(XrdHttpTests, hasCGIParamAuthz) {
```
dynamic-entropy: @ccaffy Added the requested unit test for checking against various conditions, including the presence of false positives e.g. `authauthz`

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -813,9 +813,31 @@ void XrdHttpReq::parseResource(char *res) {
...
+        return false;
+
+    auto query = url.substr(qpos + 1);
+    while (!query.empty())
+    {
+        auto amp = query.find('&');
+        auto param = (amp == std::string::npos) ? query : query.substr(0, amp);
+
+        auto eqpos = param.find('=');
+        auto key = (eqpos == std::string::npos) ? param : param.substr(0, eqpos);
```
ccaffy: Should this be `std::string_view::npos` ?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -813,9 +813,31 @@ void XrdHttpReq::parseResource(char *res) {
...
+{
+    auto qpos = url.find('?');
+    if (qpos == std::string::npos)
+        return false;
+
+    auto query = url.substr(qpos + 1);
+    while (!query.empty())
+    {
+        auto amp = query.find('&');
+        auto param = (amp == std::string::npos) ? query : query.substr(0, amp);
```
ccaffy: Should this be `std::string_view::npos` ?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -813,9 +813,31 @@ void XrdHttpReq::parseResource(char *res) {
...
+
+        auto eqpos = param.find('=');
+        auto key = (eqpos == std::string::npos) ? param : param.substr(0, eqpos);
+
+        if (key == paramName)
+            return true;
+        if (amp == std::string::npos)
```
ccaffy: Should this be `std::string_view::npos` ?

dynamic-entropy: Essentially, the difference lies in one being a constexpr and the other a simple constant. Changed it for clarity since we are using string_view throughout. They are the same value otherwise. 
https://en.cppreference.com/w/cpp/string/basic_string_view/npos.html
https://en.cppreference.com/w/cpp/string/basic_string/npos.html
Thanks

ccaffy (approved):
All good, think about squashing your commits please before merging :)

dynamic-entropy:
Also rebased on master.

ccaffy:
OK thanks, maybe rename your commit including between square brackets the component it refers to. Like this:
image

bbockelm (approved):
Looks good to me!

Thanks for including both unit and integration test coverage.

I think we can discuss with @ccaffy about the more thorough fix for the host check issue outlined in one of my comments -- but that seems clearly out of scope for this one.

tests/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ endif()
 add_subdirectory( XrdOssTests )
 add_subdirectory( XRootD )
 add_subdirectory( cluster )
+add_subdirectory( authenticated_cluster)
```
amadio, line 35: For the sake of getting this in sooner, can we split the addition of this new cluster setup into a separate pull request? As I've discussed with @dynamic-entropy before, the current cluster tests have the annoying issue that it often leaves running processes behind, which then make tests fail the next time you run them. I'd like to improve that and some other things, like using variables for binaries, which can be avoided (like it's done in `tests/XRootD` tests). So ideally we can work together on this and cover also extending the test suite to do tests with Kerberos5 and GSI as well. What do you think?

src/XrdHttp/XrdHttpReq.hh
```diff
@@ -252,6 +252,9 @@ public:
 
   void addCgi(const std::string & key, const std::string & value);
 
+  // Check if the URL has a cgi parameter
+  static bool hasCGIParam(std::string_view url, std::string_view paramName);
```
amadio: As per my [other](https://github.com/xrootd/xrootd/pull/2552/files#r2276593099) [comments](https://github.com/xrootd/xrootd/pull/2397#discussion_r2036767848), `std::string_view` is often misused, and it's used very little in XRootD currently. I'd strongly prefer to continue to use `const std::string&` instead and stay away from `string_view`. Sorry to be annoying about this, but I've had this position for a long time :-) See https://github.com/root-project/root/pull/2608#discussion_r218036622

@abh3 What's your opinion on this (i.e. `string_view` vs `const string&`)?

dynamic-entropy:
Reworded commits and opened the issue from Brian's comments: https://github.com/xrootd/xrootd/issues/2572

tests/cluster/test.sh
amadio: I'd like to hear @abh3's opinion here. From note number 2 in the [docs](https://xrootd.web.cern.ch/doc/dev6/XRdv520.pdf#view=Fit&page=91), moves across filesystems are not supported, in which case the move can use the CGI from the source for the destination as well, which won't require copying it. From the second bullet point above, the destination doesn't match after redirect, so it seems to me that the redirect should be fixed to make the destination match, since moves can only work within the same server.

src/XrdHttp/XrdHttpReq.hh
bbockelm: I was borderline about this myself.

`std::string_view` makes the implementation far more readable, especially given the number of off-by-one mistakes when using the C string parsing functions.

However, for the interface into the function, I feel like `const std::string &` does provide lots of consistency with the rest of feel of the library.

dynamic-entropy: Okay, I will change the implementation to use `std::string&`  for consistency. However, I also note that in this case, none of the arguments in the referred links hold.  But only the fact that we want to maintain the feel of the library?

As per our discussion about wanting to move towards a world where we do not have `authz` as a query parameter in redirect URLs, the question then changes to - do we want this PR or do we want to solve it with one big change of moving towards not having `authz` in redirected URLs.  
And if I should wait till we address that before rewriting this. @amadio @bbockelm

tests/CMakeLists.txt
dynamic-entropy: I agree that we should extend the tests to other auth mechanisms too. 

For the issue of failing to run incase of a rerun, the current tests run a cleanup before start, hence helping with the case. One simple way to ensure cleanup is that the script provided to cleanup_fixture does not exit on first error.

As discussed in person, we can sit together and architect what we want going forward. I can simply remove the commit that adds the `authenticated_cluster` to the tests.

src/XrdHttp/XrdHttpReq.hh
bbockelm: I don't think we should change any major semantics or approaches as part of a bugfix.

(Especially as the proposals to make those changes requires breaking effectively all HTTP clients by default and making it so WLCG-specific clients are needed to use things.)

dynamic-entropy: Yes, however, @amadio argues that he wishes to do that as part of 5.9, and if we do that, we will not require the bug fix proposed in this PR.

amadio: > I don't think we should change any major semantics or approaches as part of a bugfix.

That's not what we want here, but in XRootD 6.0, moving tokens to the query parameters of URLs will have to go away, as this is a security issue that needs to be addressed.

> Especially as the proposals to make those changes requires breaking effectively all HTTP clients by default and making it so WLCG-specific clients are needed to use things.

That's not true. Dropping `Authorization` happens [_by design_](https://datatracker.ietf.org/doc/html/rfc9110#section-15.4-6.3.1), for security reasons. In the future, we want to improve security by not bypassing this mechanism that is implemented in most clients. Using `curl --location-trusted` is a small price to pay for improving security.

amadio:
I discussed this with @abh3 yesterday. The conclusion is that the simplest fix should be applied here, i.e., just move the lines added by commit https://github.com/xrootd/xrootd/commit/6e01c0d22bc037fac184ff4b55d860c8f9a32fa5 to where moves are processed (shown below):

https://github.com/xrootd/xrootd/blob/d9af49d8a21d9ea86078582491669390660f4654/src/XrdHttp/XrdHttpReq.cc#L1723-L1754

Since we plan to support `access_token` as a synonym of `authz` in the future, the extra opaque from the source should be appended verbatim to the destination, as it's currently done, otherwise it will not work for `access_token` later.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1722,6 +1717,13 @@ int XrdHttpReq::ProcessHTTPReq() {
     }
     case XrdHttpReq::rtMOVE:
     {
+      // Incase of a move cgi parameters present in the CGI str
+      // are appended to the destination in case of a MOVE.
+      if (resourceplusopaque != "") {
+          strchr(destination.c_str(), '?') ? destination.append("&") : destination.append("?");
```
amadio: Can we use find here too?
```
destination.append((destination.find("?") == std::string::npos) ? "?" : "&");
```

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -1722,6 +1717,13 @@ int XrdHttpReq::ProcessHTTPReq() {
...
+          int npos = resourceplusopaque.find("?");
+          destination.append(resourceplusopaque.c_str(), npos + 1, resourceplusopaque.length() - npos - 1);
```
amadio: Can `resourceplusopaque` be just a resource, without `opaque` info (worth adding a test with a URL without an `authz` in it)? If yes, then `npos` needs to be checked before appending. You can also do just this:
```cpp
destination.append(resourceopaque.c_str() + npos + 1);
```
There's no need to specify the length.

src/XrdHttp/XrdHttpReq.cc
dynamic-entropy: Done thankyou!

src/XrdHttp/XrdHttpReq.cc
dynamic-entropy: Added a check for `resourceplusopaque` without `opaque`.

src/XrdHttp/XrdHttpExtHandler.cc
```diff
@@ -28,24 +28,25 @@
 #include "XrdOuc/XrdOucEnv.hh"
 
 /// Sends a basic response. If the length is < 0 then it is calculated internally
-int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long int bodylen)
+int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long bodylen)
```
amadio: This is a new commit which was not here before. Can we just fix the rename after redirection problem in this pull request? Ok, we can keep the tests, but this commit is unrelated and should come in a separate pull request.

dynamic-entropy: Apologies! and thanks for catching. 

That commit was in my master and rebase picked that up too. Dropped it from the PR. 
That commit is for this other PR: https://github.com/xrootd/xrootd/pull/2585

amadio: Thanks!

PR #2575, opened 2025-08-15
New tests for authenticated cluster

dynamic-entropy:
This is a rough copy of the existing cluster tests with added config and files to enable scitoken auth. 
The idea is to use this branch to develop, together with @amadio, more configurable tests.


amadio:
This is not needed anymore, since the tests will get merged as part of #2570 in the end.

PR #2577, opened 2025-08-17
Workaround Mac OS X pecularity with RLIMIT_NOFILE

bbockelm:
After ignoring error messages along these lines for years:

```
25: 250817 12:36:02 259 oss_Open_ufs: Unable to reloc FD /tmp/foo; invalid argument
```

on Mac OS X, I dug into the underlying cause.  It appears that `setrlimit` for `RLIMIT_NOFILE` can return successfully (and `getrlimit` returns the value just set) but calls to `fcntl(fd, F_DUPFD, arg)` have an additional limit -- the `arg` must be below `getdtablesize()`.

On Linux, this is a no-op: `getdtablesize()` is implemented via a call to `getrlimit`.  However, on Mac OS X, it appears these are two potentially-distinct limits!

Use the smaller of the two enforced limits.  With this, the relocation errors go away.

While the `ifdef` here is not strictly necessary (the extra logic should be a no-op on Linux), it's such a Mac OS X-specific curiosity that I wanted things clearly partitioned so it can hopefully be removed at some point in the far future.

### DIFF
--- a/src/XrdOss/XrdOssConfig.cc
+++ b/src/XrdOss/XrdOssConfig.cc
@@ -274,7 +274,28 @@ int XrdOssSys::Configure(const char *configfn, XrdSysError &Eroute,
            if (setfd)
               {if (setrlimit(RLIMIT_NOFILE, &rlim))
                   Eroute.Emsg("Config", errno, "set fd limit");
-                  else FDLimit = rlim.rlim_cur;
+                  else
+                     {
+#ifdef __APPLE__
+//
+// As of macOS Sequoia 15.6, setrlimit above will be successful but the limit on the
+// file descriptor table size may still be lower than RLIMIT_NOFILE.  Calls to
+// `fcntl(fd, F_DUPFD, arg)` as is done in `XrdSysFD_Dup1` will fail if the provided
+// `arg` is greater than `getdtablesize()` per the online man pages.  After testing,
+// it seems the two limits exist independently.
+//
+// One can verify this by (a) testing `getrlimit()` again to see the larger limit and
+// then doing a `XrdSysFD_Dup1` that's below `rlim.rlim_cur` and above `getdtablesize()`;
+// it will fail with "Invalid Argument".
+//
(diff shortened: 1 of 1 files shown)


src/XrdOss/XrdOssConfig.cc
```diff
@@ -274,7 +274,28 @@ int XrdOssSys::Configure(const char *configfn, XrdSysError &Eroute,
...
+// it seems the two limits exist independently.
+//
+// One can verify this by (a) testing `getrlimit()` again to see the larger limit and
+// then doing a `XrdSysFD_Dup1` that's below `rlim.rlim_cur` and above `getdtablesize()`;
+// it will fail with "Invalid Argument".
+//
+// On Linux, the man pages note that `getdtablesize()` is implemented via the equivalent
+// `getrlimit` call; hence, there's no need for this extra check.
+//
+                      FDLimit = static_cast<rlim_t>(getdtablesize()) < rlim.rlim_cur ? getdtablesize() : rlim.rlim_cur;
```
amadio, line 294: There's a [warning](https://my.cdash.org/viewBuildError.php?type=1&buildid=3044934) here. `getdtablesize()` returns an `int`, not `rlim_t`. The function that returns `rlim_t` is `getrlimit()`.

bbockelm: @amadio - I think you might be looking at the results from before my force-push to 0545fb4?  That force-push was to fix the precise warning you quote.  Looking at the latest run, I don't see the warning.

amadio: No, the version I see on GitHub now still has the static cast to `rlim_t`. Did you push the right version to GitHub?

bbockelm: The static cast to `rlim_t` is my preferred version.  That prevents the issue in the warning (signed vs unsigned comparison).

I'd rather have the comparison be between two `rlim_t` (as opposed to narrowing the value of `rlim.rlim_cur`) because the Mac OS X manpage says the actual minimum value of `getdtablesize` is 20.

PR #2582, opened 2025-08-21
[Cl] Avoid poller callbacks blocking each other. Fixes #2578

### LINKED ISSUE #2578: Deadlock around idle timeout & close of stream with multiple substreams
### Operating System

Linux david-dev-al9.cern.ch 5.14.0-503.14.1.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Nov 19 21:25:22 EST 2024 x86_64 x86_64 x86_64 GNU/Linux

### XRootD Version

v5.8.4

### Description

Process can hang after using XrdCl with multiple substreams.

Using the follwing python program (but targeting an actual running server):

```
#!/usr/bin/python3

from XRootD import client as xclient
from pyxrootd import client
import time

f1 = xclient.File()
f1.open("root://xroot-data-server.example//nonexisting-file")

time.sleep(3)
```

Run, e.g. with

XRD_LOGLEVEL=Dump XRD_TIMEOUTRESOLUTION=1 XRD_DATASERVERTTL=1 XRD_SUBSTREAMSPERCHANNEL=16 ./testpy-openwait.py

I find that often (but not always) the program never exits.

### Additional Information

_No response_

smithdh:


src/XrdCl/XrdClChannel.hh
```diff
@@ -158,7 +158,7 @@ namespace XrdCl
 
     private:
 
-      URL                    pUrl;
+      std::shared_ptr<URL>   pUrl;
```
amadio, line 161: This is an ABI breaking change.

abh3:
How is that? It obviously breaks internal linkage but I could not find any public dependence on ClChannel remaining unchanged.

src/XrdCl/XrdClChannel.hh
amadio: @abh3 is right, I ran the ABI check and indeed, this is not exposed in public APIs, so the change is fine, sorry for the noise.

amadio:
@abh3 You are indeed right, the Channel implementation is not exposed, so the change passed the ABI checks. This can safely go into 5.9.

smithdh:
I found this still needs some work; I'll have to keep in in draft for a while longer.

amadio:
Nevermind, I see this is based on `devel`... Could you please rebase on master?

PR #2585, opened 2025-08-25
Add StartSimpleResp and SendData to External header API

dynamic-entropy:
For XRootd 6 - Breaks ABI but cleans up the code as previously envisioned for XRootD 5

### DIFF
--- a/src/XrdHttp/XrdHttpExtHandler.cc
+++ b/src/XrdHttp/XrdHttpExtHandler.cc
@@ -27,25 +27,25 @@
 #include "XrdHttpProtocol.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 
-/// Sends a basic response. If the length is < 0 then it is calculated internally
-int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long int bodylen)
+int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long bodylen)
 {
   if (!prot) return -1;
 
-  // @FIXME
-  // - need this to circumvent missing API calls and keep ABI compatibility
-  // - when large files are returned we cannot return them in a single buffer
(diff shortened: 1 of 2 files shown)


src/XrdHttp/XrdHttpExtHandler.cc
```diff
@@ -27,25 +27,25 @@
 #include "XrdHttpProtocol.hh"
 #include "XrdOuc/XrdOucEnv.hh"
 
-/// Sends a basic response. If the length is < 0 then it is calculated internally
-int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long int bodylen)
+int XrdHttpExtReq::SendSimpleResp(int code, const char* desc, const char* header_to_add, const char* body, long long bodylen)
 {
   if (!prot) return -1;
 
-  // @FIXME
```
dynamic-entropy, line 35: @amadio 
It simply allows direct calls to the additional send response APIs in XrdHttpProtocol (namely `StartSimpleResp` and `SendData`, without the hack.
As you can see in the comments, this could have gone in Xrootd 5.

src/XrdHttp/XrdHttpExtHandler.hh
```diff
@@ -75,6 +75,12 @@ public:
   /// Sends a basic response. If the length is < 0 then it is calculated internally
   int SendSimpleResp(int code, const char *desc, const char *header_to_add, const char *body, long long bodylen);
 
+  // Start response to the client; often followed by data in multiple chunks
+  int StartSimpleResp(int code, const char *desc, const char *header_to_add, long long bodylen, bool keepalive);
```
dynamic-entropy, line 79: @amadio Wasn't the ABI CI supposed to break here?

amadio: No, because `XrdHttp` has no `SONAME`. However, this change does break ExtHandler plugins, so I may have to adjust the build to include more libraries.

PR #2562, opened 2025-07-21
[XrdPfc] Introduce validation of http cache control parameters  

alja:
### Description 
Query and validate http cache control directives: ETag, must-revalidate, and max-age.

The cache control parameters are saved in xattr of the *cinfo file at the time of *cinfo file creation in function XrdPfc::File::Open(). The parameter values are retrieved with  XrdCl::File::Fcntl() query.

The cache control parameters are tested in Pfc::Cache::Prepare() where the values are compared from the save xattr attributes from *cinfo file and the one queried with XrdCl::FileSystem.::Query() The comparison is made only if must-revalidate is true or expired.

The File::Fcntl() and FileSystem::Query() use new query code XrdCl::QueryCode::Head.

### Test
The improvement has been tested together with the pelican client.
See pending PR https://github.com/PelicanPlatform/xrdcl-pelican/pull/68

#### Workflow
* Use proxy forward to apache server to retrieve cache control attributes. 
* For  example curl -I will trieve the parameters from apache server
```

alja@green:~/xrd-dev$ curl -I http://xrd-cache-1.t2.ucsd.edu/alja/sap.png
HTTP/1.1 200 OK
Date: Mon, 21 Jul 2025 17:46:55 GMT
Server: Apache/2.4.62 (AlmaLinux) OpenSSL/3.2.2 mod_perl/2.0.12 Perl/v5.32.1
Last-Modified: Mon, 21 Jul 2025 17:24:44 GMT
ETag: "3cef6b-63a73c069139b"
Accept-Ranges: bytes
Content-Length: 3993451
Cache-Control: max-age=2592000, must-revalidate, public
Expires: Wed, 20 Aug 2025 17:46:55 GMT
Content-Type: image/png

```
* Access file and save this prameters on the first access in the cinfo file. 
* Continued example shows xattr in file cache
```

curl -O  http://localhost:7777//http://xrd-cache-1.t2.ucsd.edu/alja/sap.png

alja@green:~/xrd-dev$ getfattr  -d -m - data/alja/*cinfo
# file: data/alja/sap.png.cinfo
security.selinux="unconfined_u:object_r:user_home_t:s0"
user.pfc.cache-control="{\"ETag\":\"3cef6b-63a73c069139b\",\"expire\":1753710701,\"max-age\":592000,\"revalidate\":true}"
user.pfc.fsize=0sa+88AAAAAAA=
```

PR #2586, opened 2025-08-25
Etag cache validation using new XrdCl::File::Fcntl interface 

alja:
This is a remake of https://github.com/xrootd/xrootd/pull/2562.

Here, the new client Fcnt interface is added to propagate the query codes.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -2811,7 +2811,7 @@ void XrdHttpReq::addAgeHeader(std::string &headers) {
 }
 
 void XrdHttpReq::addETagHeader(std::string &headers) {
-  headers += std::string("Etag: ") + std::to_string(etagval);
+  headers += std::string("Etag: \"") + std::to_string(etagval) + "\"";
 }
 
 void XrdHttpReq::reset() {


src/XrdOuc/XrdOucCache.hh
```diff
@@ -37,6 +37,8 @@
 
 #include "XrdOuc/XrdOucCacheStats.hh"
 #include "XrdOuc/XrdOucIOVec.hh"
+#include "XrdCl/XrdClBuffer.hh"
```
abh3, line 40: It is disallowed to include XrdCl header files in XrdOuc files. In any case, the change is moot as I already made the appropriate changes.

src/XrdOuc/XrdOucCache.hh
```diff
@@ -165,6 +167,28 @@ long long    FSize() = 0;
 
 virtual int  Fstat(struct stat &sbuff) {(void)sbuff; return 1;}
 
+//------------------------------------------------------------------------------
```
abh3, line 170: This change is not need as the XrdOucCache as I already have done the appropriate changes.

src/XrdPosix/XrdPosixFile.cc
```diff
@@ -386,6 +386,12 @@ int XrdPosixFile::Fstat(struct stat &buf)
    buf.st_mode   = myMode;
    return 0;
 }
+
```
abh3, line 389: No, I opeted to do something different. So, this code is not needed.

src/XrdPosix/XrdPosixFile.hh
```diff
@@ -100,6 +100,8 @@ static void          DelayedDestroy(XrdPosixFile *fp);
 
        int           Fstat(struct stat &buf) override;
 
+       int           Fcntl(XrdCl::QueryCode::Code queryCode, const XrdCl::Buffer& args, XrdCl::Buffer*& res) override;
```
abh3, line 103: No, this code is not needed.

src/XrdPosix/XrdPosixPrepIO.hh
```diff
@@ -48,6 +48,8 @@ long long   FSize() {return (Init() ? fileP->FSize() : openRC);}
 int         Fstat(struct stat &buf)
                  {return (Init() ? fileP->Fstat(buf) : openRC);}
 
+int         Fcntl(XrdCl::QueryCode::Code queryCode, const XrdCl::Buffer& args, XrdCl::Buffer*& res) { return (Init() ? fileP->Fcntl(queryCode, args, res) : openRC); }
```
abh3, line 51: I have implemented a different solution, so this does not apply.

src/XrdCl/XrdClFileStateHandler.cc
```diff
@@ -1858,6 +1858,52 @@ namespace XrdCl
     return SendOrQueue( self, *self->pDataServer, msg, stHandler, params );
   }
 
+
+  //----------------------------------------------------------------------------
+  // Performs a custom operation on an open file, server implementation
+  // dependent - async
+  //----------------------------------------------------------------------------
+  XRootDStatus FileStateHandler::Fcntl( std::shared_ptr<FileStateHandler> &self,
```
abh3, line 1866: I don't understand this change. Technically, it should have been a change to two lines in the existing FileStateHandler::Fcntl()
1) The method definitions and
2) Replacing line XrdClFileStateHandler.cc:1887 to use the query code passed via the argument list.
That's a minor change and this looks way too much of a change than was needed.

src/XrdCl/XrdClFileStateHandler.hh
```diff
@@ -476,6 +476,25 @@ namespace XrdCl
                                  ResponseHandler                   *handler,
                                  time_t                             timeout = 0 );
 
+      //------------------------------------------------------------------------
```
abh3, line 479: This should have simply changed the existing method and not introduced a whole new method. The actual change that was required was pretty trivial as opposed to adding a whole new definition.

abh3 (changes requested):
I think we somehow miscommunicated. The XrdCl changes are more extensive than we discussed and should be redone for simplicity's sake. The XrdPosix changes were supposed to be part of the overall control flow and I did those. The ones here are quite different than the simpler approach I took. So, I would prefer to go with my changes. That includes the XrdOucCache.hh change that were part of the XrdPosix changes that I made. We should discuss this.

PR #2588, opened 2025-08-27
[XrdCl] Additional Fcntl interface to propagate query code parameter.

alja:
XrdCl related changes.


alja:
@abh @osschar 
Please check QueryCode::Code in https://github.com/xrootd/xrootd/blob/6fc2f0e1d409945d5dfc8817c5560f3a224821a6/src/XrdCl/XrdClFileSystem.hh#L59

Perhaps there are better ways of naming the opaque parameters.

abh3 (approved):
Looks good, thanks!

PR #2589, opened 2025-08-29
[XrdPfc] Fix XrdOssStats when running under the PFC

bbockelm:
XrdOssStats currently doesn't work when loaded by the PFC:
- The PFC passes an empty environment which excludes the `oss.gStream*` (and any other object pointers in the environment; just for the stats plugin the gStream is critical).
- XrdOssStats ignores the `runmode` causing the different statistics to be sent with the same header.

### DIFF
--- a/src/XrdPfc/XrdPfcConfiguration.cc
+++ b/src/XrdPfc/XrdPfcConfiguration.cc
@@ -405,7 +405,7 @@ bool Cache::test_oss_basics_and_features()
    Purpose: To parse configuration file and configure Cache instance.
    Output:  true upon success or false upon failure.
  */
-bool Cache::Config(const char *config_filename, const char *parameters)
+bool Cache::Config(const char *config_filename, const char *parameters, XrdOucEnv *env)
 {
    // Indicate whether or not we are a client instance
    const char *theINS = getenv("XRDINSTANCE");
@@ -414,8 +414,10 @@ bool Cache::Config(const char *config_filename, const char *parameters)
    // Tell everyone else we are a caching proxy
    XrdOucEnv::Export("XRDPFC", 1);
 
-   XrdOucEnv myEnv;
-   XrdOucStream Config(&m_log, theINS, &myEnv, "=====> ");
+   XrdOucEnv emptyEnv;
+   XrdOucEnv *myEnv = env ? env : &emptyEnv;
+
+   XrdOucStream Config(&m_log, theINS, myEnv, "=====> ");
 
    if (! config_filename || ! *config_filename)
    {
@@ -496,7 +498,8 @@ bool Cache::Config(const char *config_filename, const char *parameters)
    Config.Close();
 
    // Load OSS plugin.
-   myEnv.Put("oss.runmode", "pfc");
+   auto orig_runmode = myEnv->Get("oss.runmode");
+   myEnv->Put("oss.runmode", "pfc");
    if (m_configuration.is_cschk_cache())
    {
       char csi_conf[128];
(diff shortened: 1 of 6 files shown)


amadio:
Could you please also add documentation to `XrdOssStats`? Thank you.

bbockelm:
@amadio - happy to add documentation but where exactly does documentation go?

I'm happy to write a `README.md` for the plugin as I certainly know how that works.

src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -517,6 +520,8 @@ bool Cache::Config(const char *config_filename, const char *parameters)
       TRACE(Error, "Config() Unable to create an OSS object");
       return false;
    }
+   if (orig_runmode) myEnv->Put("oss.runmode", orig_runmode);
```
abh3, line 523: Out of curiosity, why add the restoration of the previous runmode? In practice, this is actually an endpoint and there will be no additional oss launches. I'm not adverse to this; just wondering about the rational.

abh3 (approved):
Just a minor question, otherwise I see nothing wrong here,

src/XrdPfc/XrdPfcConfiguration.cc
bbockelm: The `myEnv` variable is a pointer to a `XrdOucEnv` that is used for initializing all the OSS objects, not just the PSS/PFC.

If we don't unset the `oss.runmode` then that setting "leaks" to subsequent OSS object creation outside the PSS stack.

src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -414,8 +414,10 @@ bool Cache::Config(const char *config_filename, const char *parameters)
    // Tell everyone else we are a caching proxy
    XrdOucEnv::Export("XRDPFC", 1);
 
-   XrdOucEnv myEnv;
-   XrdOucStream Config(&m_log, theINS, &myEnv, "=====> ");
+   XrdOucEnv emptyEnv;
+   XrdOucEnv *myEnv = env ? env : &emptyEnv;
```
amadio, line 418: You are setting a pointer to a value on the stack here, then passing it along below in `ofsCfg->Load(XrdOfsConfigPI::theOssLib, myEnv)`. Are you sure this is safe to do?

src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -517,6 +520,8 @@ bool Cache::Config(const char *config_filename, const char *parameters)
       TRACE(Error, "Config() Unable to create an OSS object");
       return false;
    }
+   if (orig_runmode) myEnv->Put("oss.runmode", orig_runmode);
+   else myEnv->Put("oss.runmode", "");
```
amadio, line 524: Unless `orig_runmode` is a `nullptr` (cannot tell due to using `auto` above), then you don't need the if/else here, just `myEnv->Put("oss.runmode", orig_runmode);` should work in both cases.

src/XrdPfc/XrdPfcConfiguration.cc
bbockelm: It's worked out in practice ... I don't think the lifetime is well-defined but I can't make any promises about OSS plugins both inside and out of the code base.

(Given the lack of a copy constructor or enough public methods to duplicate the environment object - the underlying reason I have to do this in the first place! - I'm assuming most just take what they need from the environment on load)

src/XrdPfc/XrdPfcConfiguration.cc
bbockelm: Yes, `orig_runmode` can be a `nullptr` (`Get()` returns `char *`) if the mode is not set.

If you call `Put` with a `nullptr`, the process will abort.  Hence, we need the conditional.

Issue #2592, opened 2025-09-09
xrdcp from an xcache does not work with streams option


smithdh:
### Operating System

Linux david-dev-al9.cern.ch 5.14.0-503.14.1.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Nov 19 21:25:22 EST 2024 x86_64 x86_64 x86_64 GNU/Linux

### XRootD Version

v5.8.4

### Description

Noticed that an xrdcp from an xcache of a file already cached does not transfer the file when used with the -S option (i.e. sub streams).

### Steps to Reproduce the Behavior

1. have an xcache with file already cached. (perhaps by using xrdcp to fetch a file). My file was 100MB and could transfer in a very short time.
2. attempt to fetch again with, e.g.  -S 10 option on the xrdcp command line. What I saw was that the xrdcp command appears to hang and does not transfer the file. (I only left it a minute or two, perhaps it will eventally give an error or some other result).

smithdh:
Have done some investigation, so will assign this to me for now and open a PR with attempted fix. I believe the problem is server side, so will give Andy chance to review the PR.

abh3:
Thanks David, I will be eagerly looking forward to the fix.

PR #2593, opened 2025-09-09
[Xrootd] Copy Response with correct stream ID to XrdXrootdAioTask xrootd#2592

smithdh:


smithdh:
This is a fix for #2592. The issue appeared to me to be incorrect stream ID placed in responses sent on bound (i.e. sub) Links, when the aio system is being used. (This happens for xcache, but not regular xrootd server, when the client is trying to use sub-streams).

I think the problem is that a read/pgread command is read from XrdXrootdProtocol object for the main (login) Link, and that Protocol has the correct stream ID set in its Response (the one the client set in the the read/pgread command). The protocol object for the link of the correspoding return pathID is found ('pP'), and a XrdXrootdAioTask is allocated. But the AioTask is prepared with the stream ID in the protocol object for the bound link, which at this point may not be correct. The result is that response is eventually sent on the correct Link, but possibly with the write streamId; thus the client can not match the response to the request, discards it and keeps waiting for a response.

PR #2595, opened 2025-09-09
2531 wlcg data integrity check

ccaffy:
Still under testing with EOS but things look good so far. In case you want to have a first look!

The aim is to implement the WLCG HTTP TPC data integrity check that is in https://www.github.com/labkode/wlcg-tpc and https://indico.cern.ch/event/1538346/contributions/6474197/attachments/3053366/5397622/DOMA%20BDT%20TPC%20DATA%20INTEGRITY%20(1).pdf

Currently will only work with EOS as EOS supports on-the-fly-checksum validation

I added and refactored several things in this PR:
- XrdHttp --> Added Want-Repr-Digest header parsing and Repr-Digest header parsing. Those parsing added two maps to the XrdHttpExtReq object so that EOS can take them without the need for re-parsing those headers.
- Want-Repr-Digest is now handled in native XrdHttp code, it re-uses some part of the `Digest` header handling.
- For HTTP TPC PUSH, the Repr-Digest header is forwarded to the passive server which returns an error in case the checksum of the received file is wrong (this error is displayed to the user)
- For HTTP TPC PULL, the Want-Repr-Digest header is added by the active server to the HEAD request that is done before the file is opened so that the passive server can return the file size and Repr-Digest
    - In case the checksum returned by the passive server and the client-provided one match in terms of type but mismatch in terms of value, the error 412 PRECONDITION_FAILED is returned to the user
    - In case there is no match between the checksum type returned by the passive server and the client-provided one, we will anyway open the file
    - Two extra opaque were created in order to pass the expected checksumtype and checksum value to the OSS layer. `oss.cksumtype=<cksymtype>` and `oss.cksumvalue=<cksumvalue>`. This would allow EOS to open the file and tell the storage node to expect the checksum <cksumtype> of the file written to match the `<cksumvalue>`


bbockelm:
This conflicts with @dynamic-entropy's refactor of the TPC work, no? I think it's strategic to rebase on that to avoid duplicate work.

There are several OSS plugins that  implement on-the-fly checksumming; does this work with those?

abh3:
I did not check if this conflicts with the @dynamic-entropy refactor but it is something to look at. It appears that both of these efforts were taking parallel paths. Yes, duplication should be avoided. As for oss implementation of dynamic checksums, there are only two that I know of and only one that is officially in the repo (Ceph) but not fully distributed for various reasons. I would prefer not to have a checksum implementation in the oss as it would  architecturally conflict with how checksums are calculated now. It would also mean that every oss supplier would need to reimplement it. The plan is to have it available via the XrdCks checksum package. We just haven't yet nailed down what that would look like but it is scheduled for early FY26.

bbockelm:
@abh3 - I'm aware of two checksum implementations, the HDFS plugin (dating back about 15 years) and the multiuser plugin.  They look like they wouldn't conflict with this but I haven't done a thorough look.  It'll be great if the implementation is moved into core instead!

That said, this looks to create custom oss CGI parameters that are currently EOS specific; it'll be useful to have them documented if they are to be standardized in the future.

AFAICT the integrity approach still needs some improvement: if the operation succeeds, you don't know if it's because the remote server succeeded with the integrity check or if it's because the server didn't understand the new request in the first place and ignore it.  Hence, you'd need to do a second check - which somewhat defeats the purpose.  The server's response really should affirmatively indicate it did the check and the check passed.

The protocol issue also looks reflected in the implementation: if this isn't using the EOS OFS, how does the code detect that the feature request was ignored (instead of resulting in a successful check)?  At first read, it seems that it'll silently corrupt for non-EOS unless a subsequent checksum request is performed.

bbockelm:
> As for moving it into the core, that's the whole idea to tamp down the number of different implementations

Indeed, that's why it's wonderful!

@dynamic-entropy - can you try rebasing @ccaffy's change onto your branch to see what the potential conflicts might be?  If they're extensive, perhaps we can split out the nice changes to header parsing from the TPC changes?

@ccaffy - are you coordinating with @labkode on the protocol-level issue that will make this impractical for FTS to rely on this functionality?

ccaffy:
Yes that's correct as I added two fields indeed...

labkode:
I think these changes are fine to be included in 6 as far as EOS devs will be swift to take the new major release.

ccaffy:
Alright, it's done :) Feel free to review more thoroughly should you have time.

Thanks

amadio:
Alright, we have agreed to merge this for XRootD 6.0, so I'm replacing the milestone tag to match.

src/XrdHttpTpc/XrdHttpTpcState.cc
```diff
@@ -159,13 +162,51 @@ void State::SetupHeaders(XrdHttpExtReq &req) {
     }
 }
 
+void State::SetupHeadersForHEAD(XrdHttpExtReq &req) {
+  struct curl_slist *list = NULL;
+  for (std::map<std::string, std::string>::const_iterator hdr_iter = req.headers.begin();
+       hdr_iter != req.headers.end();
+       hdr_iter++) {
+    if (!strcasecmp(hdr_iter->first.c_str(),"copy-header")) {
```
amadio: Can we use `strncasecmp` here too?

src/XrdHttpTpc/XrdHttpTpcState.cc
```diff
@@ -159,13 +162,51 @@ void State::SetupHeaders(XrdHttpExtReq &req) {
     }
 }
 
+void State::SetupHeadersForHEAD(XrdHttpExtReq &req) {
+  struct curl_slist *list = NULL;
+  for (std::map<std::string, std::string>::const_iterator hdr_iter = req.headers.begin();
+       hdr_iter != req.headers.end();
+       hdr_iter++) {
```
amadio:
```cpp
for (auto& [header, value] : req.headers) {
...
}
```

src/XrdHttpTpc/XrdHttpTpcState.cc
```diff
@@ -159,13 +162,51 @@ void State::SetupHeaders(XrdHttpExtReq &req) {
...
+      ss << kv.first << '=' << 5;
+      if(cpt < reprDigestSize) {
+        ss << ',';
+      }
+      cpt++;
+    }
+    list = curl_slist_append(list, ss.str().c_str());
+  }
+
+  if (list != NULL) {
```
amadio: Minor C++ style, not needed to change, but this is better as `if (list)` or `if (list != nullptr)`.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -164,6 +164,31 @@ std::string TPCHandler::prepareURL(XrdHttpExtReq &req) {
...
+            auto clientDigestMatch = req.mReprDigest.find(digestName);
+            if (clientDigestMatch != req.mReprDigest.end()) {
+                // We found a checksum type match between the client-provided one and the source server-provided one
+                if (clientDigestMatch->second != digestValue) {
+                    // The checksum value does not match, return an error to the client 412 PRECONDITION_FAILED
+                    std::stringstream errMsg;
+                    errMsg << "The client-provided checksum"
+                           << " (" << clientDigestMatch->first << "=" <<  clientDigestMatch->second << ")"
+                           << " mismatches the one coming from the remote server"
+                           << " (" << digestName << "=" << digestValue << ")";
```
amadio: In discussing the WLCG error improvement project, we reached the conclusion that all error messages should have a fixed part (which FTS could use for grouping "kinds of erros"), and a variable part, with more specific information. This error message is probably one of the kinds that we'd like to put in this format, so I suggest something like this (but you can reword as you prefer):
```
Mismatch between client-provided and remote server checksums: client = %s, server = %s
```
Then FTS can cut it at the `:` and group all mismatches together in their web UI.

@mpatrascoiu, @abh3 Are you in agreement here?

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -966,15 +999,19 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
...
             full_url += "&oss.asize=" + std::to_string(sourceFileContentLength);
-        } else {
-          // In the case the GetContentLength is not successful, an error will be returned to the client
-          // just exit here so we don't open the file!
+            mismatchDigests = mismatchReprDigest(sourceFileReprDigest,req,rec);
+        }
+        if(!success || mismatchDigests) {
+          // We could not get remote file information, or the checksum provided by the client
+          // does not match the source file one, we fail here and don't open the file
           return 0;
```
amadio: Returning `0` means failure? That's unintuitive on a function that returns `int`. Maybe it should return `bool` if the return value doesn't matter?

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -966,15 +999,19 @@ int TPCHandler::ProcessPullReq(const std::string &resource, XrdHttpExtReq &req)
     {
         //Get the content-length of the source file and pass it to the OSS layer
         //during the open
-        bool success;
-        GetContentLengthTPCPull(curl, req, sourceFileContentLength, success, rec);
+        bool success = false;
+        bool mismatchDigests = false;
+        std::map<std::string,std::string> sourceFileReprDigest;
+        GetRemoteFileInfoTPCPull(curl, req, sourceFileContentLength, sourceFileReprDigest, success, rec);
```
amadio, line 1004: This function should probably return `bool` instead of taking `success` as input, what do you think? Then you'd pass a variable in to get the size (which should always be used, I see here it's not being used from the return value).

src/XrdHttpTpc/XrdHttpTpcUtils.hh
```diff
@@ -32,6 +32,19 @@ namespace TPC {
...
+                         std::map<std::string,std::string> & reqHeadersParm,
+                         const std::map<std::string,std::string> & hdr2cgimapParm,
+                         const std::map<std::string,std::string> & reprDigestParm):
+      reqResource(reqResourceParm),reqHeaders(reqHeadersParm),hdr2cgimap(hdr2cgimapParm),reprDigest(reprDigestParm){}
+
+    const std::string & reqResource;
+    std::map<std::string,std::string> & reqHeaders;
+    const std::map<std::string,std::string> & hdr2cgimap;
+    const std::map<std::string,std::string> & reprDigest;
+  };
```
amadio, line 46: If this struct is part of the class and `public`, then `prepareOpenURL` should probably not be a `static` member. I'd either put the struct outside the class and only in the `.cc` file (or in a private header), and use a free `static` function there for `prepareOpenURL` (to hide it from the public interface), or I'd make the struct and function private (or protected) within `XrdHttpUtils`. Of course, if this makes testing too hard, then we may need to find another way to not put it into the public interface. Anyway, don't try too hard, as I don't think this is an area where ABI needs to be stable across releases.

src/XrdHttp/XrdHttpHeaderUtils.cc
```diff
@@ -51,9 +52,53 @@ void XrdHttpHeaderUtils::parseReprDigest(const std::string &header, std::map<std
...
+    std::transform(key_lower.begin(),key_lower.end(),key_lower.begin(),::tolower);
+
+    try {
+      uint8_t preference_us = XrdOucUtils::touint8_t(preference);
+      // Max allowed value for Repr-Digest is 10
+      preference_us = std::min(preference_us,(uint8_t)10);
+      output[key_lower] = preference_us;
+    } catch (...) {
+      // discard invalid values
+    }
```
amadio, line 102: Can we do this in a way that doesn't need exceptions? I'm afraid it could cost us in terms of performance if exceptions are thrown frequently here.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -149,32 +150,105 @@ void Tobase64(const unsigned char *input, int length, char *out) {
   return;
 }
 
+void Tobase64(const std::vector<uint8_t> & input, std::string & base64Output) {
```
amadio, line 151: I'd prefer to not add duplicate code if possible. It's probably better to change the existing function and migrate code using it to not use pointers than add this one.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -149,32 +150,105 @@ void Tobase64(const unsigned char *input, int length, char *out) {
...
+void base64DecodeHex(const std::string & base64, std::string & hexOutput) {
+  std::vector<uint8_t> bytes;
+  base64ToBytes(base64,bytes);
+  bytesToHex(bytes, hexOutput);
+}
+
 
 static int
-char_to_int(int c)
+char_to_int(char ch)
```
amadio: This should stay as `int`, see https://en.cppreference.com/w/cpp/string/byte/isdigit.html

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -50,6 +50,7 @@
 #include <memory>
 #include <vector>
 #include <algorithm>
+#include <optional>
```
amadio: This seems to be unused?

src/XrdHttp/XrdHttpUtils.hh
```diff
@@ -44,6 +44,8 @@
 #include <vector>
 #include <memory>
 #include <sstream>
+#include <optional>
```
amadio: Same here, seems unused (maybe used in a later commit)?

src/XrdOuc/XrdOucTUtils.hh
```diff
@@ -48,7 +48,7 @@ public:
 //! Split a string
 //------------------------------------------------------------------------
 template<class Container>
-static void splitString( Container &result,  const std::string &input, const std::string &delimiter ) {
+static void splitString( Container &result,  std::string_view input, const std::string &delimiter ) {
```
amadio: Let's not change the interface here, as although this header is private, it's installed into `/usr/include`.

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1459,6 +1460,21 @@ void XrdOucUtils::trim(std::string_view & sv) {
...
+
+  if (ec == std::errc::invalid_argument) {
+    throw std::invalid_argument("Invalid number format");
+  }
+  if (ec == std::errc::result_out_of_range || temp > std::numeric_limits<uint8_t>::max()) {
+    throw std::out_of_range("Value out of range for unsigned short");
+  }
+
+  return static_cast<unsigned short>(temp);
+}
```
amadio, line 1521: Is there a specific reason not to use one of the standard functions to convert? You can convert to a wider type and check for `n < 10` before casting, which should be fine. For example, you could use [`std::strtol`](https://en.cppreference.com/w/cpp/string/byte/strtol.html). This would also avoid exceptions.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
ccaffy: OK, waiting for the agreement and then I'll modify, no problem :)

src/XrdHttpTpc/XrdHttpTpcTPC.cc
ccaffy: OK it's the comment that is misleading :D
Returning 0 is successful yes. I changed the comment --> we return an error to the client but the logic is correct!

src/XrdHttp/XrdHttpUtils.cc
ccaffy: Yes true, removed.

src/XrdHttp/XrdHttpUtils.hh
ccaffy: yup

src/XrdOuc/XrdOucTUtils.hh
ccaffy: Yes, good point, reverting!

src/XrdOuc/XrdOucUtils.cc
ccaffy: it's just more c++ ;)
and as we are using a string_view in parameter, from_chars does not need null-terminated string to work...

Really again, I'm opened to modification if you think it is necessary ;)

src/XrdOuc/XrdOucTUtils.hh
ccaffy: Done

src/XrdHttp/XrdHttpUtils.cc
ccaffy: Yes, indeed thanks!

src/XrdHttpTpc/XrdHttpTpcTPC.cc
amadio: Ok, since there was no reply, let's just consider that we have an agreement here. The only important thing is to put messages in a format like this:
```
Message with a fixed text: message with variable text (i.e. with specific filenames, etc)
```
We can easily change things later if needed (i.e, the particular wording used). Feel free to mark as resolved after changing it.

ccaffy: Yes, done according to your previous comment ;)

PR #2596, opened 2025-09-10
[XrdHttpTpc] Correctly map linux error codes to HTTP error codes

### LINKED ISSUE #2591: Improve error message mapping for TPC transfers
### Operating System

Alma Linux 9

### XRootD Version

5.8.4

### Description

When a third-party copy transfer is launched by a user to copy to an area where they have exceeded their quota, the error shown in the logs is HTTP 400 rather than the expected HTTP 507. We need to improve the error translation to avoid having misleading messages in future releases.

Sample log line as shown in EOS (with HTTP status 400 and clear "out of quota" as error):
```
/var/log/eos/mgm/xrdlog.mgm-2025-09-03-1756864801.gz:250903 03:05:27 3163098 TPC_PullRequest: event=OPEN_FAIL, local=/eos/cms/store/group/cmst3/group/l1tr/elfontan/INFP_151XnoNNVtx/MinBias_TuneCP5_14TeV-pythia8/NuGunAllEta_PU200_151XnoNNVtx/250902_094953/0000/inputs140X_3-164.root, remote=https://eoscms-ns-ip700.cern.ch:443/eos/cms/store/temp/user/elfontan.0d672c30282a7c187c5fce3a8973a0187111469d/cmst3/group/l1tr/elfontan/INFP_151XnoNNVtx/MinBias_TuneCP5_14TeV-pythia8/NuGunAllEta_PU200_151XnoNNVtx/250902_094953/0000/inputs140X_3-164.root, user=elfontan, status=400; Unable to get quota space - quota not defined or exhausted /eos/cms/store/cmst3/group/l1tr/elfontan/INFP_151XnoNNVtx/MinBias_TuneCP5_14TeV-pythia8/NuGunAllEta_PU200_151XnoNNVtx/250902_094953/0000/inputs140X_3-164.root; Disk quota exceeded
```

For more information, see [INC4647375](https://cern.service-now.com/service-portal?id=ticket&table=incident&n=INC4647375).

Launch third-party copy transfer into an area where user has no quota left.

### Additional Information

This is where the bad mapping of the error happens:

https://github.com/xrootd/xrootd/blob/20d203c0b549bf00ed7c8000008e4708ef5f5872/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L834-L846

dynamic-entropy:
Fix #2591 

We had two oppurtunities to map linux error codes to their HTTP counterparts

This PR provides an enum of= list of HTTP error codes defined in the RFC. The use if "Named" enum error is preferred for readability over numbers for e.g. `HTTP_INSUFFICIENT_STORAGE` over `http_507`.

This PR also makes and attempt to map error codes defined in asm-generic/errno-base.h to their HTTP counterparts.

Looking forward we also need a mapping between XrdError and HTTP status codes - this is being done as part of the WebDAV error improvement project.

### DIFF
--- a/src/XrdHttp/XrdHttpUtils.cc
+++ b/src/XrdHttp/XrdHttpUtils.cc
@@ -37,6 +37,7 @@
  */
 
 
+#include "XrdHttpUtils.hh"
 
 #include <cstring>
 #include <openssl/hmac.h>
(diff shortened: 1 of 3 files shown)


src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,209 @@ char *escapeXML(const char *str) {
...
 }
 
+int mapErrNoToHttp(int err) {
+  HttpStatus statusCode;
+
+  switch (err) {
+    case EACCES:
+    case EPERM:
+      statusCode = HTTP_FORBIDDEN;
+      break;
```
amadio: I'm not a fan of using an `enum`, then returning an `int`, which needs a cast. I think it's better to return the `enum` or simply use `int`s directly and forego the `enum` altogether.

```suggestion
HttpStatus mapErrNoToHttp(int err) {
  switch (err) {
    case EACCES:
    case EPERM:
      return HTTP_FORBIDDEN;
    ...
}
```

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,209 @@ char *escapeXML(const char *str) {
...
+  switch (status) {
+    // 1xx Informational
+    case HTTP_CONTINUE:
+      return "Continue";
+    case HTTP_SWITCHING_PROTOCOLS:
+      return "Switching Protocols";
+    case HTTP_PROCESSING:
+      return "Processing";
+    case HTTP_EARLY_HINTS:
+      return "Early Hints";
```
amadio: These would look much better with one case per line with the `return` in front, like in haproxy [here](https://github.com/haproxy/haproxy/blob/master/src/http.c#L424-L505). I also think that using numbers in each case makes it more readable, since the lenghts are all the same. The text is right next to the number in any case.

@abh3 and @ccaffy Do you prefer names or numbers, like it's done in haproxy? Also, do you like more `HTTP_CONTINUE` or some variation of `HTTP_STATUS_100` (see e.g. [here](https://github.com/haproxy/haproxy/blob/07c10ec2f124801b1d58454f13c331c5a0b77b0e/src/http.c#L159-L181))? Is `HTTP_` as prefix maybe prone to clashes later on (e.g. when used together with curl)?

src/XrdHttp/XrdHttpUtils.hh
```diff
@@ -167,12 +167,90 @@ inline std::string encode_opaque(const std::string & opaque) {
...
+  HTTP_NETWORK_AUTHENTICATION_REQUIRED = 511
+};
+
+
 // Escape a string and return a new one
 char *escapeXML(const char *str);
 
+int mapErrNoToHttp(int err);
+
+std::string httpStatusToString(HttpStatus status);
```
amadio: Another inconsistency, you take in an `enum HttpStatus` in `std::string httpStatusToString(HttpStatus status);`, but return `int` in `int mapErrNoToHttp(int err);`

src/XrdHttp/XrdHttpUtils.cc
dynamic-entropy: Agree with the return on the same line as the case. This is my formatter putting it back to two lines instead. Having said that, would appreciate it if we could define a clang-format for the project. 
A case-by-case basis isn't the best idea either.

Regarding the use of a number as an element of enums instead of names, I would prefer not to have an enum at all over that.
I think readability should be valued in places where logic is written, as opposed to having a table-like case matching "numbers with prefix" to "numbers", unless that's the only thing available where the properties of enums are required.

So even though using numbers makes it look good in terms of humans preferring symmetry, no one will actually read that code independently.
By using names, we wish to improve the readability of code where logic is written: for e.g. 
``` 
//  instead of
if verb==not defined: return 405 ;
// we have
if verb==not defined: return HTTP_METHOD_NOT_ALLOWED ;
```

Curl uses `CURLCode` and does not define enums with `HTTP_` prefix. It doesn't even maintain such a status code list.

dynamic-entropy (commented):
Those are not inconsistencies, but a reflection of how the current code is structured. 

All http req code for the time being uses `int` to communicate `httpstatuscodes` - we can continue to do that and not have an enum of codes. 

Or we can slowly (or in one shot) move towards using the enums defined here.
I do not have a strong preference, but the use of enums helps with understanding what a specific code means.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,144 @@ char *escapeXML(const char *str) {
...
+    case EHOSTUNREACH:
+      return HTTP_BAD_GATEWAY;
+
+    default:
+      return HTTP_INTERNAL_SERVER_ERROR;
+  }
+}
+
+
+std::string httpStatusToString( int status) {
```
amadio: Formatting (space only on one side).

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,144 @@ char *escapeXML(const char *str) {
...
+    case 506: return "Variant Also Negotiates";
+    case 507: return "Insufficient Storage";
+    case 508: return "Loop Detected";
+    case 510: return "Not Extended";
+    case 511: return "Network Authentication Required";
+
+    default:
+      switch (status) {
+        default:
+          switch (status) {
```
amadio: Do you really need two extra nested switch statements here? Maybe it was a copy-paste error?

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,140 @@ char *escapeXML(const char *str) {
   return r;
 }
 
+int mapErrNoToHttp(int err) {
```
ccaffy: why not renaming this variable err into errNo?

amadio: To be complete, let's attempt to map all errors which can be returned by `open`, `read`, `write`, `close`, and `seek` at least. For example, we should add at least the following adjustments:

- `EACCESS` &rarr; 403 (Forbidden)
- `EPERM` &rarr; 401 (Unauthorized)
- `ENAMETOOLONG` &rarr; 414 (URI too long)
- `ETIMEDOUT` &rarr; 504 (Gateway timeout)

and the following additions:

- `ELOOP` &rarr; 508 (Loop Detected)
- `EFBIG` &rarr; 413 (File contents too big on write)
- `ENXIO` &rarr; 400 (No such device or address, for e.g. seek beyond file length)
- `EOPNOTSUPP` &rarr; 501 (Operation not supported &rarr; Not Implemented)
- `EROFS` &rarr; 401 or 403 (Read-only filesystem and write was requested)
- `ENOMEM` &rarr; 503

amadio: I think it's ok as it is. It's only used in the switch in any case.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,164 @@ char *escapeXML(const char *str) {
...
+      return HTTP_LOOP_DETECTED;
+
+    case ENOSPC:
+    case EDQUOT:
+      return HTTP_INSUFFICIENT_STORAGE;
+
+    case EFBIG:
+      return HTTP_PAYLOAD_TOO_LARGE;
+
+    case ETXTBSY:
```
amadio: This one should be in the same case as `EEXIST`, which also returns the same error. But actually, this should probably be translated the same as `EAGAIN`/`EBUSY`.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,164 @@ char *escapeXML(const char *str) {
...
+
+    case ECONNREFUSED:
+    case ECONNRESET:
+    case ENETDOWN:
+    case ENETUNREACH:
+    case EHOSTUNREACH:
+    case EPIPE:
+      return HTTP_BAD_GATEWAY;
+
+    case EIO:
```
amadio: You can let this one fallback into `default`, there's no reason to have a separate case to return the same value just before.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,163 @@ char *escapeXML(const char *str) {
...
+
+    case ENOSPC:
+    case EDQUOT:
+      return HTTP_INSUFFICIENT_STORAGE;
+
+    case EFBIG:
+      return HTTP_PAYLOAD_TOO_LARGE;
+
+    case ENOTEMPTY:
+      return HTTP_CONFLICT;
```
amadio: Please let's not have multiple case for the same return value. Thanks.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -496,5 +497,163 @@ char *escapeXML(const char *str) {
...
+
+    case ECONNREFUSED:
+    case ECONNRESET:
+    case ENETDOWN:
+    case ENETUNREACH:
+    case EHOSTUNREACH:
+    case EPIPE:
+      return HTTP_BAD_GATEWAY;
+
+    default: //EIO
```
amadio: That's not just `EIO`, that's all other errors. This comment is misleading, please remove.

amadio:
Alright, good to go now. Thanks a lot!

Issue #2526, opened 2025-06-03
[XrdHttp] Return error instead of no reply if xrootd server is readonly


ccaffy:
If we set the following configuration on the xrootd server:
`all.export /eos/ readonly`

Trying to upload a file to the server will not return an error, but an empty reply:

```
$ curl --upload-file /etc/passwd https://xrootdserver.cern.ch:8080/eos/dteam001/passwd
curl: (52) Empty reply from server

```

From the server logs:
```
250603 16:58:47 2554036 unknown.3:60@aiadm02 http_Protocol: Process is exiting rc:1
250603 16:58:47 2554036 unknown.3:60@aiadm02 Xrootd_Protocol: 0000 open odm /eos/dteam001/passwd2
250603 16:58:47 2554036 unknown.3:60@aiadm02 ofs_open: 200-40664 (664) fn=/eos/dteam001/passwd2
250603 16:58:47 2554036 ofs_open: unknown.3:60@aiadm02 Unable to open /eos/dteam001/passwd2; read-only file system
250603 16:58:47 2554036 unknown.3:60@aiadm02 Xrootd_Response: 0000 sending err 3025: Unable to open /eos/dteam001/passwd2; read-only file system
250603 16:58:47 2554036 http_Req:  XrdHttpReq::Error
250603 16:58:47 2554036 unknown.3:60@aiadm02 http_Req: PostProcessHTTPReq req: 4 reqstate: 0 final_:False
250603 16:58:47 2554036 unknown.3:60@aiadm02 http_Req: PostProcessHTTPReq mapping Xrd error [3025] to status code [500]
250603 16:58:47 2554036 http_Req:  XrdHttpReq request ended.
250603 16:58:47 2554036 unknown.3:60@aiadm02 ofs_close: use=0 fn=dummy
250603 16:58:47 2554036 http_Protocol:  Cleanup
250603 16:58:47 2554036 http_Protocol:  Reset
250603 16:58:47 2554036 http_Req:  XrdHttpReq request ended.
250603 16:58:47 2554036 XrootdXeq: unknown.3:60@aiadm02 disc 0:00:01 (send failure)
```

dynamic-entropy:
[...]

PR #2598, opened 2025-09-16
Webdav error improvement

dynamic-entropy:
This PR 
1. As per the plan to add error codes in small instalments, adds tests and mapping for `File not found` errors

2. Addresses the initial bug reported in https://github.com/xrootd/xrootd/issues/2526 - The server would erroneously skip calling `SendSimpleResp` if a valid mapping was missing. 

3. Introduces a consistent and exhaustive mapping between `xrderr codes` to `http status codes` via a common `errno` translation. 

4. ~Proposes a replacement for the existing (initially deemed temporary) switch case mapping to a static table lookup for WebDAV errno mapping.~

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -817,124 +817,23 @@ void XrdHttpReq::parseResource(char *res) {
     resourceplusopaque.append('?');
     resourceplusopaque.append(p + 1);
   }
-  
-  
-  
 }
 
-void XrdHttpReq::sendWebdavErrorMessage(
-    XResponseType xrdresp, XErrorCode xrderrcode, XrdHttpReq::ReqType httpVerb,
-    XRequestTypes xrdOperation, std::string etext, const char *desc,
-    const char *header_to_add, bool keepalive) {
-  int code{0};
-  std::string errCode{"Unknown"};
-  std::string statusText;
+void XrdHttpReq::generateWebdavErrMsg() {
 
-  switch (httpVerb) {
-    case XrdHttpReq::rtPUT:
-      if (xrdOperation == kXR_open) {
-        if (xrderrcode == kXR_isDirectory) {
-          code = 409;
-          errCode = "8.1";
-        } else if (xrderrcode == kXR_NoSpace) {
-          code = 507;
-          errCode = "8.3.1";
-        } else if (xrderrcode == kXR_overQuota) {
-          code = 507;
-          errCode = "8.3.2";
-        } else if (xrderrcode == kXR_NotAuthorized) {
-          code = 403;
-          errCode = "9.3";
-        }
-      } else if (xrdOperation == kXR_write) {
-        if (xrderrcode == kXR_NoSpace) {
-          code = 507;
-          errCode = "8.4.1";
-        } else if (xrderrcode == kXR_overQuota) {
-          code = 507;
-          errCode = "8.4.2";
-        }
-      }
-      break;
-    default:
-      break;
+  // This block is only used when sending an "X-Transfer-Status" trailer response.
+  // We set the body to "OK" so that the trailer becomes "X-Transfer-Status: 200 OK",
+  // indicating a successful transfer.
+  if (xrdresp == kXR_ok) {
+    httpStatusCode = 200;
+    httpErrorBody = "OK";
+    return;
   }
 
-  // Remove the if at the end of project completion
-  // Till then status text defaults to as set by mapXrdResponseToHttpStatus
-  if (code != 0) {
-    httpStatusCode = code;
-    httpErrorCode = errCode;
-    httpErrorBody = "ERROR: " + errCode + ": " + etext + "\n";
+  // default error
+  httpStatusCode = mapXrdErrToHttp(xrderrcode);
+  httpErrorBody = etext + "\n";
 
-    prot->SendSimpleResp(httpStatusCode, desc, header_to_add,
-                         httpErrorBody.c_str(), httpErrorBody.length(),
-                         keepalive);
-  }
-}
-
-// Map an XRootD error code to an appropriate HTTP status code and message
-// The variables httpStatusCode and httpErrorBody will be populated
-
-void XrdHttpReq::mapXrdErrorToHttpStatus() {
-  // Set default HTTP status values for an error case
-  httpStatusCode = 500;
-  httpErrorBody = "Unrecognized error";
-
-  // Do error mapping
-  if (xrdresp == kXR_error) {
-    switch (xrderrcode) {
-      case kXR_AuthFailed:
-        httpStatusCode = 401; httpErrorBody = "Unauthorized";
-        break;
-      case kXR_NotAuthorized:
-        httpStatusCode = 403; httpErrorBody = "Operation not permitted";
-        break;
-      case kXR_NotFound:
-        httpStatusCode = 404; httpErrorBody = "File not found";
-        break;
-      case kXR_Unsupported:
-        httpStatusCode = 405; httpErrorBody = "Operation not supported";
-        break;
-      case kXR_FileLocked:
-        httpStatusCode = 423; httpErrorBody = "Resource is a locked";
-        break;
-      case kXR_isDirectory:
-        httpStatusCode = 409; httpErrorBody = "Resource is a directory";
-        break;
-      case kXR_ItExists:
-        if(request != ReqType::rtDELETE) {
-          httpStatusCode = 409; httpErrorBody = "File already exists";
-        } else {
-          // In the case the XRootD layer returns a kXR_ItExists after a deletion
-          // was submitted, we return a 405 status code with the error message set by
-          // the XRootD layer
-          httpStatusCode = 405;
-        }
-        break;
-      case kXR_InvalidRequest:
-        httpStatusCode = 405; httpErrorBody = "Method is not allowed";
-        break;
-      case kXR_noserver:
-        httpStatusCode = 502; httpErrorBody = "Bad Gateway";
-        break;
-      case kXR_TimerExpired:
-        httpStatusCode = 504; httpErrorBody = "Gateway timeout";
-        break;
-      default:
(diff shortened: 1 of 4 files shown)


src/XrdHttp/XrdHttpReq.cc
```diff
@@ -856,8 +859,29 @@ void XrdHttpReq::sendWebdavErrorMessage(
...
       break;
+    }
+    case XrdHttpReq::rtPROPFIND:{}
+    case XrdHttpReq::rtDELETE:{}
+    case XrdHttpReq::rtPOST:{}
+    case XrdHttpReq::rtOPTIONS:{}
+    case XrdHttpReq::rtPATCH:{}
+    case XrdHttpReq::rtMKCOL:{}
+    case XrdHttpReq::rtMOVE:{}
+    case XrdHttpReq::rtHEAD: {}
```
abh3: I can't say this kind of syntax is any clearer than just leaving nothing after the colon and falling into a break at the end. Plus, I'm not comfortable that there is no default. Yes, it's unlikely that another rtxxxx code will be defined in the future but it's not impossible, As it sits, a code that isn't recognized is simply ignored, further making it odd that there is a defined list for ones to be ignored.

abh3 (commented):
Just pointed out some not often used syntax in this particular context. Otherwise, I didn't see anything pop out. Of course, we need to find out why all the tests mysteriously failed.

src/XrdHttp/XrdHttpReq.cc
dynamic-entropy: Hello Andy
This PR was a "stash" - usually for me to save and sync with my other machine. 
The intent with the empty cases is to fill them in upcoming commits.

abh3: OK, got it.

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -538,7 +544,6 @@ int mapErrNoToHttp(int errNo) {
     case EFAULT:
     case ENXIO:
     case ESPIPE:
-    case ENODEV:
```
amadio, line 541: Why is this being removed?

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2899,6 +2899,11 @@ void XrdHttpReq::reset() {
   final = false;
 
   mScitag = -1;
+
+  httpStatusCode = -1;
+  httpErrorCode = "";
+  httpErrorBody = "";
```
amadio, line 2802: This is not adding a test as the commit says.

src/XProtocol/XProtocol.hh
```diff
@@ -987,6 +987,7 @@ struct ServerResponseBody_ChkPoint { // Only for kXR_ckpQMax
 /******************************************************************************/
 
 enum XErrorCode {
+   kXR_ErrCodeBase =       3000,
```
amadio: This should not be added here.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1770,6 +1748,39 @@ int XrdHttpProtocol::SendSimpleResp(int code, const char *desc, const char *head
...
+  webdavTable[XrdHttpReq::rtPUT][kXR_open - XrdOpBase][kXR_NoSpace - XrdErrBase] = {507, "8.3.1"};
+  webdavTable[XrdHttpReq::rtPUT][kXR_open - XrdOpBase][kXR_overQuota - XrdErrBase] = {507, "8.3.2"};
+  webdavTable[XrdHttpReq::rtPUT][kXR_open - XrdOpBase][kXR_NotAuthorized - XrdErrBase] = {403, "9.3"};
+
+  // PUT + write
+  webdavTable[XrdHttpReq::rtPUT][kXR_write - XrdOpBase][kXR_NoSpace - XrdErrBase] = {507, "8.4.1"};
+  webdavTable[XrdHttpReq::rtPUT][kXR_write - XrdOpBase][kXR_overQuota - XrdErrBase] = {507, "8.4.2"};
+
+  // GET + open
+  webdavTable[XrdHttpReq::rtGET][kXR_open - XrdOpBase][kXR_NotFound - XrdErrBase] = {404, "3.1"};
```
amadio: Any added static table should not be initialized at runtime, it should be declared const and be statically initialized.

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -129,6 +136,16 @@ public:
   static XrdObjectQ<XrdHttpProtocol> ProtStack;
   XrdObject<XrdHttpProtocol> ProtLink;
 
+  struct WebdavMapVal {
+    int httpStatus;
+    const char *httpErrorCode;
+  };
+
+  static WebdavMapVal webdavTable[ReqTypeCount][XrdOpCount][XrdErrCount];
```
amadio: We cannot have statically declared variables in headers, this will create one instance into each source file where this header is included, which is not what we want.

src/XrdHttp/XrdHttpUtils.cc
dynamic-entropy: We discussed this with you that since `ENODEV` is mapped to and from `kXR_FSError`, it is more appropriate to map it to `INTERNAL SERVER ERROR`, hence removed here.

src/XrdHttp/XrdHttpReq.cc
dynamic-entropy: refactored into a separate commit

src/XProtocol/XProtocol.hh
dynamic-entropy: These and the remaining two comments are part of commits that introduced the static table mapping for the `l.m.n` convention. 
I have removed those commits so as to have the base level reporting complete and correct for XRootD 6.

I will introduce them back if need be, with fixes suggested in these comments.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: Resolving here, commented above - this commit has been removed until we reach consensus on the convention.

src/XrdHttp/XrdHttpProtocol.hh
dynamic-entropy: Resolving here, commented above - this commit has been removed until we reach consensus on the convention.

src/XrdHttp/XrdHttpUtils.cc
amadio: Sorry, indeed, I remember it now.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2899,6 +2899,11 @@ void XrdHttpReq::reset() {
...
+
```
amadio, line 2803: Much better! Now I understand why you changed this, whereas before it was not clear at all why this was needed while adding a test. Thanks for changing.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2389,9 +2283,13 @@ int XrdHttpReq::PostProcessHTTPReq(bool final_) {
         break;
       } else {
 
-
         // If we are here it's too late to send a proper error message...
-        if (xrdresp == kXR_error) return -1;
+        // However, we decide to send a response anyway before we close the connection
+        // We are not sure if sending a final response before reading the entire request
```
amadio, line 2288: This sentence seems unfinished. Could you please reword or clarify?

tests/XRootD/http.sh
```diff
@@ -220,62 +220,129 @@ function test_http() {
...
+      # Look for trailer
+      local trailer_line
+      trailer_line=$(grep -i '^X-Transfer-Status:' "$body_file" | sed -E 's/^[^:]+: *//')
+
+      if [[ -n "$trailer_line" ]]; then
+        local trailer_code
+        trailer_code=$(echo "$trailer_line" | cut -d: -f1 | xargs)
+        assert_eq "$expected_trailer_code" "$trailer_code" "$trailer_line"
+
+        # Extract error code (n.m or n.m.l etc.) from trailer message
```
amadio: I suppose this is now out of date relative to the code.

src/XrdHttp/XrdHttpReq.cc
dynamic-entropy: what i wish to convey here is that this is required only for X-Transfer-Status trailer response. for normal http header response of 200, we do not set anything in the body.
Made the comment better, thanks for catching this

dynamic-entropy:
@amadio 
With the CI fix in ABI, this one is ready too. Along with providing a framework for consistent error reporting, it also incorporates a few bug fixes/regressions introduced in the previous error handling commits. E.g. sending `000` instead of `500` as a default response code. 

This would be nice to have in `5.9.1`. 

For XRootD 6, we can consider going along https://www.rfc-editor.org/rfc/rfc9457.html as per your suggestion, after discussing with the other WLCG communities. Though I doubt we gain much from it other than backing from an RFC thus ending the formatting part of the discussion.

src/XrdHttp/XrdHttpReq.cc
amadio: This comment is still not very clear.

dynamic-entropy: Still? (Even after the change?)

How about?
```
// This block is only used when sending an "X-Transfer-Status" trailer response.
// We set the body to "OK" so that the trailer becomes "X-Transfer-Status: 200 OK",
// indicating a successful transfer.
```

dynamic-entropy: Changed it to the above, hope it's very straightforward now :) 

Also rebased on current master

amadio: Yes, this is much better. Now it makes sense, but when we read it a long time from now, we may have forgotten that this is about `X-Transfer-Status`. This looks fine now. Thanks!

PR #2601, opened 2025-09-18
Fcntl pelican Add scaffolding to handle pelican specific query requests.

abh3:
This introduces two additional subcodes to kXR_query: kXR_QFinfo (requires an open file handle) and kXR_QFSinfo (requires a path) and the tunneling of the request http/xroot -> xroot protocol handler -> ofs ->oss/pss and if pss one of two actions may occur: 1) if cache is enabled the request goes to the cache which may forward it to Posix  2) with no Cache the Pss forwards it to  Posix. The posix layer then forwards it to XrdCl which may forward it to the origin or the Pelican plugin (the intended destination).

Generally, if there is no one to handle the query subcode, the client simply gets an unsupported error.

These modifications do not break ABI.

### DIFF
--- a/src/XrdPss/XrdPss.hh
+++ b/src/XrdPss/XrdPss.hh
@@ -94,7 +94,7 @@ virtual int     Close(long long *retsz=0) override;
 virtual int     Open(const char *, int, mode_t, XrdOucEnv &) override;
 
 int     Fchmod(mode_t mode) override {return XrdOssOK;}
-int     Fctl(int cmd, int alen, const char *args, char **resp=0);
+int     Fctl(int cmd, int alen, const char *args, char **resp=0) override;
 int     Fstat(struct stat *) override;
 int     Fsync() override;
 int     Fsync(XrdSfsAio *aiop) override;


abh3:
Well, it won't hurt but I don't think you can include all of A&M's patches in the next release so it's sort of a moot point. My patches enable their patches. You can decide how you want to stage it :-)

amadio:
screenshot

amadio:
@

amadio:
An XCache test recently added by David is failing on this branch (see it in [CDash](https://my.cdash.org/tests/300136032)). It would be good to understand and resolve this failure before merging. Given the ABI changes as noted above, unless there's real urgency, I'd prefer to merge this for XRootD 6.0.

smithdh:
I tried it (running this PR with my new xcache test); but for testing reasons I backported this PR onto 5.8.4 (+ the xrotod test), and it passed, but then I noticed  inconsisetncy in the test result. I think there's aproblem elsewhere (not Csi nor this PR); looks like perhaps errno is not being set in some places for the XrdPosixExtra.cc methods. I'll prepare a PR about this.

abh3 (commented):
Thanks, this looks fine.

src/XrdOss/XrdOss.hh
```diff
@@ -415,6 +416,7 @@ uint16_t        DFType() {return dfType;}
 
 static const int Fctl_ckpObj = 0;
 static const int Fctl_utimes = 1;
+static const int Fctl_QFinfo = 2;
```
amadio: There is a conflict here with a change in `Rubin-TRSE` in 7fc7be6188ceada91360a73cc45517dbb7e66b3f which introduced `Fctl_setFD`:
```
++<<<<<<< HEAD
 +static const int Fctl_setFD  = 2; // alen = sizeof(int), args->int
++=======
+ static const int Fctl_QFinfo = 2;
++>>>>>>> 48f47d90f ([Server] Provide support of kXR_QFinfo and kXR_QFSinfo subcodes of kXR_query.)
```
I'm resolving the conflict by setting `Fctl_QFinfo = 3`.

Issue #2352, opened 2024-10-09
Output Information on Transfer Failures


turetske:
We're trying to generate something similar to an apache access log using reports of transfers from an xrootd server. We'd like for xrootd to report information on failed transfers as well as successful ones. I think it should be similar to a successful report, but maybe include things such as the final server response (404, 500, etc.).

abh3:
We already report such activity but I suspect that no one has written a collector for this information. Also, the monitoring stream likely has not even been enabled. First, checkout the following directive:
xrootd.mongstream which requires you specify tpc (for Third Party Copy) among other needed options.
https://xrootd.slac.stanford.edu/doc/dev57/xrd_config.htm#_monitor

Subsequently, this will enable the TPC g-Stream:
https://xrootd.slac.stanford.edu/doc/dev57/xrd_monitoring.htm#_Toc138968527

Note that this stream repots whether the transfer succeeded or failed. As far as I know, no one is capturing this information even though it is readily available; if configured. You should talk with your mnitoring service providor (e.g. OSG) to make this infrmation available. So, I am closing this ticket since the functionality has been implemented from our side.

abh3:
As per @bbockelm this request was for*all* failures not just TPC events. That's a pretty tall order. This will take a bit of review. I doubt we can do this without changing the existing monitoring streams. I do knw that new monitoring streams via the g-stream usually include that information. Anyway, I am re-opening this ticket.

djw8605:
Hi @abh3,

I think there's some confusion here. Pelican does not use Third Party Copy (TPC), so any errors related to TPC are not relevant to this request.

I believe Emma is looking for the errors (or successes) reported by an XRootD server to the client. Her specific example of logging 404, 403, and other HTTP codes would presumably come from the HTTP server layer. It's been a long time since I've delved into the HTTP XRootD server code, but I assume it would need to log the HTTP status code just before returning the response header to the client.

As far as I know, a file not found error (404) will not show up in the fstream. There might be a user connection event, but certainly, there will not be a file open event. An authentication failure (403) likely will not show up in the stream either.

And... I see you just responded at the same time I was drafting above.

bbockelm:
> That's a pretty tall order. This will take a bit of review. I doubt we can do this without changing the existing monitoring streams

Indeed -- we need to peel this apart into smaller problems and make it approachable.

I think two things are needed:

1.  Simple counters at the protocol level.  Counts the total number of things like SSL alerts, responses by status code, etc, etc.
[...]

PR #2604, opened 2025-09-25
Add gstream monitoring for HTTP

dynamic-entropy:
This PR aims to achieve the first part of the goals discussed _here [#2352](https://github.com/xrootd/xrootd/issues/2352#issuecomment-2407709032)_

> I think two things are needed:
> 
> 1. Simple counters at the protocol level.  Counts the total number of things like SSL alerts, responses by status code, etc, etc.
> 2. A monitoring stream where we could replicate the equivalent of an apache access log.  An entry in the stream for each request.  Now, we already have the f-stream so it makes no sense to reproduce that.  I was envisioning splitting it in half to avoid duplication:
> 
> * A new protocol error stream, covering only errors that don't result in a file record.  (Think: things like "permission denied").
> * An extension to the f-stream that adds error counts, perhaps on the close record.  We already have an example of extending the f-stream in a backward compatible way, I believe, when we added the operations count.
> 
> Something like the above would eventually cover all the pieces but could proceed in smaller steps.

If configured under the [xrootd.mongstream ](https://xrootd.web.cern.ch/doc/dev57/xrd_config.htm#_Toc171719980:~:text=7.8%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%C2%A0%20mongstream) directive, for e.g. 
```
xrootd.mongstream http use flush 2s send json 127.0.0.1:8888
```
Xrootd sends periodic counters of the form under the monitoring `g stream` , refer: https://xrootd.web.cern.ch/doc/dev57/xrd_monitoring.htm 
```
{
  "HTTP_GET_200": {
    "count": 5,
    "errors_network": 0,
    "errors_xrootd": 0,
    "success": 5,
    "duration_seconds": 0.004265
  },
  "HTTP_HEAD_200": {
    "count": 2,
    "errors_network": 0,
    "errors_xrootd": 0,
    "success": 2,
    "duration_seconds": 0.000288
  },
  "HTTP_PUT_201": {
    "count": 3,
    "errors_network": 0,
    "errors_xrootd": 0,
    "success": 3,
    "duration_seconds": 0.442662
  },
  "HTTP_COPY_201": {
    "count": 5,
    "errors_network": 0,
    "errors_xrootd": 0,
    "success": 5,
    "duration_seconds": 1.03564
  }
}
```


src/XrdHttp/XrdHttpMon.cc
```diff
@@ -11,7 +11,10 @@ XrdSysError eDest(0, "HttpMon");
 
 std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_Count>, XrdHttpReq::ReqType::rtCount> XrdHttpMon::statsInfo{};
 
-XrdHttpMon::XrdHttpMon(XrdSysLogger* logP, XrdXrootdGStream* gStream) : gStream(gStream) { eDest.logger(logP); }
+XrdHttpMon::XrdHttpMon(XrdSysLogger* logP, XrdXrootdGStream* gStream) : gStream(gStream) {
+    eDest.logger(logP);
+    flushPeriod = std::chrono::seconds(gStream->GetAutoFlush());
```
dynamic-entropy, line 25: Hello @bbockelm 
Is this a good way to set the flush period? Since this is not under the `http` directive, and I did not want to parse the mongstream here.

This required adding Getters below.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -0,0 +1,219 @@
...
+#include <iostream>
+#include <sstream>
+#include <thread>
+
+#include "XrdSys/XrdSysError.hh"
+#include "XrdXrootd/XrdXrootdGStream.hh"
+
+XrdSysError eDest(0, "HttpMon");
+
+std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_Count>, XrdHttpReq::ReqType::rtCount> XrdHttpMon::statsInfo{};
```
ccaffy: Maybe put this into a typedef?

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -0,0 +1,219 @@
...
+
+void XrdHttpMon::Report() {
+    std::string json = GetMonitoringJson();
+    if (!gStream->Insert(json.c_str(), json.size() + 1)) {
+        eDest.Emsg("HttpMon", "Gstream Buffer Rejected");
+    }
+}
+
+void* XrdHttpMon::Start(void* instance) {
+    XrdHttpMon* mon = static_cast<XrdHttpMon*>(instance);
```
ccaffy: Is this a thread? If yes, should not it be stopped via a `atomic<bool>` ? It looks like this thread can never be joined due to the infinite loop?

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: It's a one-off use for the type. However, if it helps make it more readable, I can call it something like `StatsMatrix`?

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: The thread is not supposed to be joined. This is a detached thread; it will run for the entire lifetime of the process

src/XrdHttp/XrdHttpMon.cc
bbockelm: Yes - actually, a small refactor to make it more readable would be useful.

src/XrdHttp/XrdHttpMon.cc
bbockelm: I think that's reasonable.

However, you can refactor this slightly, I think, so this is set above.  Something like this:

```
XrdHttpMon::XrdHttpMon(XrdSysLogger* logP, XrdXrootdGStream* gStream)
   : gStream(gStream),
     flushPeriod(std::chrono::seconds(gStream->GetAutoFlush())
{
    eDest.logger(logP);
}
```

src/XrdHttp/XrdHttpMon.cc
bbockelm: Yes, it's unfortunate but the XRootD daemons cannot be stopped so no need to add explicit support for stopping.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -0,0 +1,219 @@
...
+        case 502:
+            return sc_502;
+        case 504:
+            return sc_504;
+        case 507:
+            return sc_507;
+        default:
+            return sc_UNKNOWN;
+    }
+}
```
bbockelm: Please add trailing newline, per Unix custom.

src/XrdHttp/XrdHttpMon.hh
```diff
@@ -0,0 +1,76 @@
+#include <array>
```
bbockelm, line 4: C++ language headers should go after package headers; please re-order.

src/XrdHttp/XrdHttpMon.hh
```diff
@@ -0,0 +1,76 @@
...
+        sc_502,
+        sc_504,
+        sc_507,
+        sc_UNKNOWN,
+        sc_Count
+    };
+
+    // Per (operation, status code) statistics
+    struct HttpInfo {
+        RAtomic_uint64_t count{0};
```
bbockelm: Please add comments describing the semantics of each field (right now, only the last two are filled in).

src/XrdHttp/XrdHttpMon.hh
```diff
@@ -0,0 +1,76 @@
...
+        RAtomic_uint64_t error_network{0};
+        RAtomic_uint64_t error_xrootd{0};
+        RAtomic_uint64_t success{0}; // success in responding back with a http response code
+        RAtomic_uint64_t duration{0}; // sum of operation duration in microseconds
```
bbockelm: since we have to do type erasure of the underlying units (`RAtomic` doesn't support the `std::chrono` types), can you explicitly rename this member to include the unit to make it clear?

Perhaps:

```suggestion
        RAtomic_uint64_t duration_us{0}; // sum of operation duration in microseconds
```

(Plus small name changes throughout the rest of the PR)

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -34,6 +34,7 @@
  */
 
 
+#include <chrono>
```
bbockelm: The header ordering is incorrect here.

However, this issue is preexisting so I think it's OK to keep as-is.

src/XrdHttp/XrdHttpReq.hh
```diff
@@ -49,6 +49,7 @@
 #include <vector>
 #include <string>
 #include <map>
+#include <chrono>
```
bbockelm, line 48: Please put headers in alphabetical order.

(I know they aren't already but let's make things slightly better)

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -2854,6 +2830,11 @@ void XrdHttpReq::reset() {
   m_user_agent = "";
   m_origin = "";
 
+  httpStatusCode = -1;
+  initialStatusCode= -1;
+  httpErrorCode = "";
+  httpErrorBody = "";
```
bbockelm, line 2764: The `httpErrorBody` and `httpErrorCode` reset appears to be a small bugfix - good catch!

src/XrdXrootd/XrdXrootdGStream.cc
```diff
@@ -76,6 +76,16 @@ int       XrdXrootdGStream::SetAutoFlush(int afsec)
...
                            }
 
+/******************************************************************************/
+/*                          G e t A u t o F l u s h                           */
+/******************************************************************************/
+  
+int       XrdXrootdGStream::GetAutoFlush()
+                           {
+                            return gStream.GetAutoFlush();
+                            // HELP: if I understand correctly the default it 10 mins (600 sec) - shall we use a smaller one perhaps - 60 seconds as default?
```
bbockelm: Please remove comment here -- much better to open a separate issue if you want the default to be discussed.

dynamic-entropy:
Now, trying the interface of `XrdMonRoll` 

I could achieve something like (though it would need a lot of polishing)
```
{
  "statistics": {
    "tod": 1762186458,
    "ver": "v5.8.2-135-g3b864adc0",
    "src": "rchauhan-pc:7094",
    "tos": 1762186452,
    "pgm": "xrootd",
    "ins": "http",
    "pid": 238405,
    "site": "http",
    "stats_HttpRequestStatistic": {
      "GET": 40,
      "PUT": 36,
      "HEAD": 10
    }
  }
}
```

Although I did not manage to figure out how to turn off XML. This is my test config
```
xrd.report 127.0.0.1:9999 every 3 json all 
```

abh3:
You cannot turn off XML as it is there for backward compatibility. For new collectors you can simply opt in JSON for this report and that is what you will get instead of XML. Anyway, let me know what you think you are missing.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1690,18 +1743,23 @@ int XrdHttpProtocol::StartSimpleResp(int code, const char *desc,
 /******************************************************************************/
 /*                      S t a r t C h u n k e d R e s p                       */
 /******************************************************************************/
-  
+
 int XrdHttpProtocol::StartChunkedResp(int code, const char *desc, const char *header_to_add, long long bodylen, bool keepalive) {
   const std::string crlf = "\r\n";
   std::stringstream ss;
+  CurrentReq.setHttpStatusCode(code);
+  Record();
```
abh3: Have you considered to skip calling Record() if monitoring is not enabled? It seems sort of odd to do all the work ad then simply not report it in the end. This would apply toll calls to Record().

src/XrdXrootd/XrdXrootdConfigMon.cc
```diff
@@ -84,6 +84,8 @@ struct XrdXrootdGSReal::GSParms gsObj[] =
                      XrdXrootdGSReal::fmtBin, XrdXrootdGSReal::hdrNorm},
         {"oss",      0, XROOTD_MON_OSS,   0, -1, XROOTD_MON_GSOSS, 0,
                      XrdXrootdGSReal::fmtBin, XrdXrootdGSReal::hdrNorm},
+        {"http",     0, XROOTD_MON_HTTP,  0, -1, XROOTD_MON_GSHTP, 0,
+                     XrdXrootdGSReal::fmtBin, XrdXrootdGSReal::hdrNorm},
```
abh3, line 88: Please send me text (any format will do) explaining what the additions are that need to be included in the documentation.

abh3 (commented):
I am assuming that @ccaffy signed off on the changes to http relevant processing that appears to have nothing to do with monitoring. I would have prefered that these changes were in a separate pull request. I left one comment that I would like to get resolved about omitting useless work when monitoring is not enabled. Another asking for some text for documentation additions. Once the comment is resolved, I don't see anything that is worrisome.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: I have made all the Record calls conditional based on configuration. 

The other alternative implementation I was thinking of was a function pointer however, this one seems better as it heavly benefits from branch predictions. Please let me know if you think this can be done better.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1002,6 +1003,17 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
 
   pmarkHandle = (XrdNetPMark* ) myEnv->GetPtr("XrdNetPMark*");
 
+  XrdXrootdGStream *gs = nullptr;
+  if ((gs = (XrdXrootdGStream *)myEnv->GetPtr("http.gStream*")) != nullptr) {
+      httpMon = new XrdHttpMon(eDest.logger(), gs);
```
amadio: There is a memory leak here, I don't see this being freed anywhere. I also don't understand why we need a pointer to a class which has only `static` public methods. I think the interfaces can be improved. We shouldn't need to add a data member into `XrdHttpProtocol` to support monitoring. Since this is never intended to be instantiated more than once, turning into a class doesn't make a lot of sense to me.

tests/TPCTests/test.sh
```diff
@@ -82,7 +82,7 @@ cleanup() {
     src=${hosts_abbrev[${src_idx}]}
     dst=${hosts_abbrev[${dst_idx}]}
     rm "${LCLDATADIR}/${src}_empty.dat" || :
-    rm "${LCLDATADIR}/${dst}_empty.ref" || :
+    rm "${LCLDATADIR}/${src}_empty.ref" || :
```
amadio, line 85: This change has no relation to HTTP monitoring. Please remove (and send separately if it's really needed as a bug fix).

src/XrdHttp/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ add_library(XrdHttpUtils SHARED
   XrdHttpChecksum.cc         XrdHttpChecksum.hh
   XrdHttpChecksumHandler.cc  XrdHttpChecksumHandler.hh
   XrdHttpExtHandler.cc       XrdHttpExtHandler.hh
+  XrdHttpMon.cc              XrdHttpMon.hh
```
amadio: This should not be in `XrdHttpUtils`, but likely should be a plugin, like the CORS plugin. The reason is that `XrdHttpUtils` cannot have ABI breakage in minor/patch releases, so it would be very hard to change the monitoring implementation later on, since you are also changing other things like `XrdHttpProtocol` and `XrdHttpReq`. It would probably be better to have monitoring functionality not interfering with data layout of `XrdHttp` classes to avoid ABI breakage in the future.

src/XrdXrootd/XrdXrootdConfigMon.cc
dynamic-entropy: Done!

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -33,14 +33,30 @@ void XrdHttpMon::RecordCount(XrdHttpReq::ReqType op, StatusCodes sc) {
 
     auto& info = statsInfo[op][sc];
     info.count++;
-    // info.duration.fetch_add(latency.count(), std::memory_order_relaxed);
 }
 
-void XrdHttpMon::RecordError(XrdHttpReq::ReqType op, StatusCodes sc) {
+void XrdHttpMon::RecordSuccess(XrdHttpReq::ReqType op, StatusCodes sc, std::chrono::steady_clock::duration duration) {
```
amadio, line 93: When there are temporary changes before merging, they should be folded into the previous commits. I suggest adding first any required changes in XrdHttp not related to monitoring, followed by one commit with the XrdHttpMon implementation, and a final commit making use of the newly introduced interfaces in XrdHttp. See, for example, how XrdClCurl and XrdClS3 plugins are being integrated.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -133,6 +133,8 @@ std::string XrdHttpMon::GetStatusCodeString(StatusCodes sc) {
             return "201";
         case sc_206:
             return "206";
+        case sc_207:
+            return "207";
```
amadio, line 196: This is a fix, it should be folded into previous commits. Also, there's a typo in the commit message.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -131,6 +131,8 @@ std::string XrdHttpMon::GetStatusCodeString(StatusCodes sc) {
             return "200";
         case sc_201:
             return "201";
+        case sc_202:
+            return "202";
```
amadio, line 192: Same here, fixups for previous commits should be used to amend the original commit.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -1,17 +1,22 @@
...
-#include "XrdSys/XrdSysError.hh"
-#include "XrdXrootd/XrdXrootdGStream.hh"
-
 XrdSysError eDest(0, "HttpMon");
 
-std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_Count>, XrdHttpReq::ReqType::rtCount> XrdHttpMon::statsInfo{};
+typedef std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_Count>, XrdHttpReq::ReqType::rtCount>
+    StatsMatrix;
+
+StatsMatrix XrdHttpMon::statsInfo{};
```
amadio, line 14: This change is not related to "make flush period reconfigurable". One logical change per commit, please. Also, this should also be folded into the original commit adding XrdHttpMon, as it changes something introduced there.

abh3 (approved):
This looks OK to me, though the pull request is now out of date; so, that has to be corrected.

amadio (changes requested):
I think adding this to `XrdHttpUtils` will make it hard to change in the future, after 6.0 is released. I can merge this as is, but later on remember that ABI breakage will not be allowed.

At least the memory leak should be fixed before merging, however.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: Actually, I already have a commit for this :) I just didn't want to take the discussion in other direction.
I was also not happy with having to instantiate a class with almost everything static. But since I only recently did that change, I have it on top of #2637, it would be a waste of time to do it here and then rebase the other PR again on top.

tests/TPCTests/test.sh
dynamic-entropy: Sure, yup, it was an oversight before, the `.ref` files are only created at source

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Done

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Done

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Done

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: This commit is a clean extension. The original commit only adds a small framework to show the structure, while this commit adds the additional fields required. I prefer to keep the two separate and not make a big blob like the other PR you mentioned.

src/XrdHttp/CMakeLists.txt
dynamic-entropy: XrdHttpMon is now a shared library https://github.com/xrootd/xrootd/pull/2637/commits/57ea6f6bdd714bcb7d69c5a7d82dd99a86a7f054

The monitState is a pointer within the httpreq object. Its lifetime is tied to the httpreq object as well. Since httpreq object is reused, a cleanup on reset is required.

Doing it here would be very cumbersome with all the refactoring required in each commit.

src/XrdHttp/CMakeLists.txt
```diff
@@ -20,6 +20,11 @@ add_library(XrdHttpUtils SHARED
   XrdHttpHeaderUtils.cc      XrdHttpHeaderUtils.hh
 )
 
+target_sources(XrdHttpUtils PRIVATE
+  XrdHttpMon.hh
+  XrdHttpMon.cc
+)
```
amadio: You can simply list the two extra source files as part of the `add_library(XrdHttpUtils SHARED ...)` command above.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1579,6 +1594,49 @@ int XrdHttpProtocol::BuffgetData(int blen, char **data, bool wait) {
...
 }
 
+// We handle two types of request–response pairs here:
+//  - Chunked response: the HTTP header is sent first, followed by one or more chunks of data
+//  - Simple response: the header and any data (if present) are sent together in a single step
+//
+// A call to SendSimpleResp() indicates that the current request has reached its final state,
+// so we record both the request count and its completion before returning.
+// The only exception is 100-Continue, which is an interim response and must not be recorded as final.
+void XrdHttpProtocol::Record() {
```
amadio: `XrdHttpProtocol` is a public class in XrdHttpUtils, no? This is where my ABI worries come from. Interface changes in it will not be possible later. I'd drop this new method, and instead of just calling `Record()` in several places which hide somewhat what is going on, I'd move this to `XrdHttpMon` and pass in `CurrentReq` to it. In places where you set `CurrentReq.monState`, instead you can just pass in `CurrentReq` + `monState`, and then you don't need to keep track of it in `XrdHttpProtocol` either, and don't need a `switch` to figure out which state you are in, which is inefficient.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1587,24 +1645,26 @@ int XrdHttpProtocol::BuffgetData(int blen, char **data, bool wait) {
...
+  int r{1};
 
   if (body && bodylen) {
     TRACE(REQ, "Sending " << bodylen << " bytes");
     if (ishttps) {
       r = SSL_write(ssl, body, bodylen);
       if (r <= 0) {
         ERR_print_errors(sslbio_err);
-        return -1;
+        CurrentReq.monState = XrdHttpReq::MonitState::ERR_NET;
```
amadio: For example, here you could call `XrdHttpMon::RecordErrNet(CurrentReq, ...)`, and you even have access to what OpenSSL reported to you.

src/XrdHttp/XrdHttpReq.hh
```diff
@@ -362,6 +372,14 @@ public:
...
   std::map<std::string,uint8_t> m_want_repr_digest;
 
+  enum MonitState {
+    NEW,       // Uninitialised state
+    ACTIVE,    // First Call to Process Request
+    ERR_NET,   // Network Error
+    ERR_PROT,  // Filesystem/XRootD error that did not result in a valid HTTP response
+               // We see this only during a chunked response
+    DONE       // Final state 
+  } monState;
```
amadio: This is also added to a public class and won't be possible to easily change later without potentially breaking things. So even if the headers are not distributed, if you change the layout of `XrdHttpReq` which is public and in `XrdHttpUtils`, you won't be able to change it later.

amadio:
I think the exposure of things into public interfaces is still a risk. My suggestion is to add `MonState` into `XrdHttpMon` as an enum, and pass in the request and the monitoring state into a `void XrdHttpMon::Record(XrdHttpReq &, enum MonState, int code)` which does the actual recording without touching the layouts of `XrdHttpProtocol` and `XrdHttpReq`.

src/XrdXrootd/XrdXrootdConfigMon.cc
abh3: I'm sure you've sent me the doc additions but I can't find them. Can you send them again? Better yet, open a ticket with a "documentation" label and insert comments/text there. That was we can track the change.

amadio:
@dynamic-entropy Please rebase this on `master` and please reply to my last comments. As it is, this worries me that we will need ABI breaking changes later on in public classes which are changed in this pull request.

amadio:
Ok, I rebases cleanly at this point, so I just clicked to rebase. Please reply to my open comments before merging, though. Thank you!

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: Removed `Record` from `XrdHttpProtocol`. Thanks! 
Agree that it's cleaner, but since the signature is `void func (void)`, we wouldn't have required a change later. 

The state tracking is required to ensure that we call record only exactly twice, ensuring a transition from `NEW` -> `ACTIVE` and then to `DONE` on the second call.
Thus helping us ensure we are not miscounting.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: We do not wish to report individual errors via the http gstream, only their category

src/XrdHttp/XrdHttpReq.hh
dynamic-entropy: Moved this out of `XrdHttpReq`. Thanks!

src/XrdHttp/XrdHttpProtocol.cc
amadio: It's much better now. Thank you.

amadio (approved):
With the latest changes this is now much better. Thank you!

PR #2605, opened 2025-09-25
Map XrdOss error codes to Linux error codes

### LINKED ISSUE #2526: [XrdHttp] Return error instead of no reply if xrootd server is readonly

dynamic-entropy:
Fix #2526

### DIFF
--- a/src/XrdSys/XrdSysError.cc
+++ b/src/XrdSys/XrdSysError.cc
@@ -95,7 +95,7 @@ const char *XrdSysError::ec2text(int ecode)
 
 int XrdSysError::ec2errno(int ecode)
 {
-    int xcode;
+    int xcode = 0;
     XrdSysError_Table_Errno *etp = etab_errno;
 
     int sign = (ecode < 0 ? -1 : 1);
@@ -104,7 +104,9 @@ int XrdSysError::ec2errno(int ecode)
     while ((etp != 0) && !(xcode = etp->Lookup(absE)))
         etp = etp->next;
 
-    return xcode ? xcode * sign : absE * sign;
+    // if errcode is mapped return
+    // otherwise return the original err code
+    return xcode ? xcode * sign : ecode;
 }
 
 /******************************************************************************/
--- a/tests/XRootD/http.sh
+++ b/tests/XRootD/http.sh
@@ -280,9 +280,9 @@ function test_http() {
   assert davix-put "$bigFilePath" "${HOST}/$bigFilePath"
 
   # Test writing to a readonly file system
-  #should be 403
+  # Writing to a read-only file should return 403 Forbidden
   readOnlyFilePath="/readonly/file";
-  run_and_assert_http_and_error_code 500 "" \
+  run_and_assert_http_and_error_code 403 "" \
     --upload-file "$alphabetFilePath" "${HOST}/$readOnlyFilePath"
 
   # Overwrite a directory with a file - File / Directory conflict
@@ -326,5 +326,8 @@ function test_http() {
   unreadableFilePath="$bigFilePath"
   run_and_assert_http_and_error_code 200 500 \
     "${HOST}/$unreadableFilePath" --with-trailer
+
+  run_and_assert_http_and_error_code 200 "" \
+    --header "Want-Digest: crc32c" -I "${HOST}/$alphabetFilePath"
 
 }


src/XrdSys/XrdSysError.cc
```diff
@@ -87,7 +88,18 @@ const char *XrdSysError::ec2text(int ecode)
     if (!etxt) etxt = XrdSysE2T(xcode);
     return etxt;
 }
-  
+
+int XrdSysError::ecTranslate(int ecode) {
```
dynamic-entropy: Hello @abh3 

So this is a proof of concept of what we can do. I do not want this to be a separate table, but a member `const int* translations`, it was simply easier to tinker this way. But one can add to the `addTable` interface and append a "translation" list, which can be looked up to find the closest / defined linux error code. 
The error texts used remain the same as defined in the static map.

What are your thoughts?

abh3: Well, if you have gone this far you might as well keep it. However, I would suggest naming the method ec2errno() which describes exactly what is being done.

src/XrdSys/XrdSysError.hh
```diff
@@ -114,6 +139,8 @@ int baseFD();
 //
 static const char *ec2text(int ecode);
 
+int ecTranslate(int ecode);
```
abh3: ec2errno() would be better.

src/XrdSys/XrdSysError.hh
```diff
@@ -105,6 +128,8 @@ public:
 //
 static void addTable(XrdSysError_Table *etp) {etp->next = etab; etab = etp;}
 
+static void addTranslation(XrdSysError_TranslationTable *etp) {etp->next = etrans; etrans = etp;}
```
abh3: Since the tables are different classes, use the same addTable() method and let the compiler select which addTable to call based on the signature.

src/XrdSys/XrdSysError.cc
```diff
@@ -110,8 +122,12 @@ int XrdSysError::Emsg(const char *esfx, int ecode, const char *txt1,
                          Set_IOV_Buff(etxt);                         //  8
                          Set_IOV_Item("\n", 1);                      //  9
     Logger->Put(iovpnt, iov);
+    
+    // do not return the actual ecode here but only linux error codes
+    // The add table function should also give a conversion table 
+    // use that to return here 
 
-    return ecode;
+   return ecTranslate(ecode);
```
abh3: No this will likely break things. You should return the ecode as was done here before to make sure the behaviour stays the same. The particular issue you are trying to solve here won't be fully solved by doing this anyway. See the summary note.

src/XrdSys/XrdSysError.cc
```diff
@@ -87,7 +88,18 @@ const char *XrdSysError::ec2text(int ecode)
     if (!etxt) etxt = XrdSysE2T(xcode);
     return etxt;
 }
-  
+
```
abh3, line 91: Would be nice if you added a block header like has been done everywhere else in this fie for methods.

abh3 (changes requested):
OK, if you want to solve the problem using a central call then it has to be applied in the correct place. That place happens to be XrdOfs.cc:2602, here translate the ecode to the corresponding errno and save it as rcode. Then in that method, and only in that method, change the occurrence of "einfo.setErrInfo(ecode" to "einfo.setErrInfo(rcode" (two occurrences) and "einfo.setErrCode(ecode)" to "einfo.setErrCode(rcode)". At this point, the correct error message will be issued and a standard errno will be returned to the caller. Note that the bridge will return a kXR type error number, calling the Ofs() will return a proper errno, but if you call an Oss method then you will need to manually translate the returned code. So, this is why it's hard to come up with a single solution.

src/XrdSys/XrdSysError.cc
dynamic-entropy: Done

PR #2606, opened 2025-09-26
[Pfc] Implement URL CGI parameters for setting block-size and max number of blocks for prefetching on per file basis

osschar:
In the context of WLCG Analysis Facilities it became clear that people want to experiment with different settings for block-size and prefetching. This PR adds the ability to configure ranges of acceptable values for those two parameters and then pass them to the cache via opaque url cgi parameters `pfc.blocksize` and `pfc.prefetch`.

### Configuration
```
pfc.urlcgi [blocksize {ignore | min max}] [prefetch {ignore | min max}]
```
E.g.:
```
pfc.urlcgi blocksize 4k 512k prefetch 0 100
```

### Example usage:
```
xrdcp -f 'root://xcache.site.org//store/user/matevz/xrdmon-far/xmfar-2019-01.root?pfc.blocksize=32k&pfc.prefetch=60' /dev/null
```

### DIFF
--- a/src/XrdPfc/XrdPfcFile.hh
+++ b/src/XrdPfc/XrdPfcFile.hh
@@ -208,7 +208,7 @@ public:
    // Constructor, destructor, Open() and Close() are private.
 
    //! Static constructor that also does Open. Returns null ptr if Open fails.
-   static File* FileOpen(const std::string &path, long long offset, long long fileSize);
+   static File* FileOpen(const std::string &path, long long offset, long long fileSize, XrdOucCacheIO* inputIO);
 
    //! Handle removal of a block from Cache's write queue.
    void BlockRemovedFromWriteQ(Block*);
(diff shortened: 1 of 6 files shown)


src/XrdPfc/XrdPfcFile.hh
```diff
@@ -208,7 +208,7 @@ public:
    // Constructor, destructor, Open() and Close() are private.
 
    //! Static constructor that also does Open. Returns null ptr if Open fails.
-   static File* FileOpen(const std::string &path, long long offset, long long fileSize);
+   static File* FileOpen(const std::string &path, long long offset, long long fileSize, XrdOucCacheIO* inputIO);
```
amadio, line 211: This actually changes ABI of `libXrdPfc-5.so`. However, since it's a plugin and normally used with a matching version of XRootD, I don't see a problem with merging into a normal feature release.

osschar: Thank you! Indeed, getting `libXrdPfc` from another version would be asking for a lot of trouble :)

PR #2608, opened 2025-09-26
[Pfc] Return error code for set cache evict if the file does not exist.

osschar:
_set cache evict_ returned success code when the file does not exist in the cache. This returns the proper error code and message.


abh3 (approved):
Looks fine to me.

PR #2609, opened 2025-10-01
[XrdPosix] Add a missed file object unlock

smithdh:


src/XrdPosix/XrdPosixXrootd.cc
```diff
@@ -1507,18 +1507,20 @@ int XrdPosixXrootd::QueryChksum(const char *path,  time_t &Mtime,
...
-             if (!(fp = XrdPosixObject::File(fd))) return -1;
+       else {if (!(fp = XrdPosixObject::File(fd))) return -1;
              ecmP = fp->getECMsg();
             }
 
 // Return the message information
 //
-   return ecmP->Get(emsg, reset);
+   const int rc = ecmP->Get(emsg, reset);
+   if (fp) fp->UnLock();
```
amadio, line 1522: Where is the file locked such that it needs unlocking here? Should the unlock not be done closer to where it was locked?

smithdh: It gets locked by XrdPosixObject::File(fd).

PR #2615, opened 2025-10-14
[Tests] Fix xcache test failure

ellert:
The test checks that the return code is 54 and that the error message matches "Run: [ERROR] Server responded with an error: [3019]" in case of a corrupted cached file.

However, occasionally the return code is 50 and the error message is "Run: [ERROR] Received corrupted data" in this case instead.

This commit modifies the test so that the test succeeds when this happens.


smithdh:
Hi, thanks @ellert I hadn't noticed the test was flapping. Were you thinking of having this in master soon (e.g. if you're getting the errors in a CI environment and would like to stop it doing that) - I instead for 5.9.1 I'd like to check the root cause. Although this test is supposed to give an error, and both errors you seen look plausible it does mean something is varying and it's probably worth checking.

Issue #2619, opened 2025-10-22
Occasional test failure in XRootD::noauth::test and XrdCl::FileCopyTest.NormalCopyTest


smithdh:
### Operating System

Linux david-dev-al9.cern.ch 5.14.0-503.14.1.el9_5.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Nov 19 21:25:22 EST 2024 x86_64 x86_64 x86_64 GNU/Linux

### XRootD Version

5.9.0

### Description

Occasional failure of XRootD::noauth::test

```
xrdfs root://runnervmwhb2z:3094/ rm /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/4.exists.ref /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/5.exists.ref /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/not_exists_4.ref
rm /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/4.exists.ref : [SUCCESS] 
rm /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/5.exists.ref : [SUCCESS] 
rm /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/not_exists_4.ref : [ERROR] Error response: no such file or directory
error: command "xrdfs root://runnervmwhb2z:3094/ rm /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/4.exists.ref /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/5.exists.ref /home/runner/work/xrootd/xrootd/build/tests/XRootD/noauth/test-CykxtJ/not_exists_4.ref" did not fail
```

i.e. xrdfs <sever> rm f4 f5 not_exist_4

where f4, f5 exist, but file not_exist_4 does not exist, occasionally the xrdfs command indicates success instead of the expected failure.

smithdh:
Will also try to fix occasion failure seen in XrdCl::FileCopyTest.NormalCopyTest, e.g.

```
Running main() from ./googletest/src/gtest_main.cc
Note: Google Test filter = FileCopyTest.NormalCopyTest
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[...]

PR #2620, opened 2025-10-22
Fix case of losing error status (Fix for #2615)

smithdh:


src/XrdCl/XrdClParallelOperation.hh
```diff
@@ -308,12 +308,18 @@ namespace XrdCl
...
-          size_t nb = failed_cnt.fetch_add( 1, std::memory_order_relaxed );
+          if( status.IsOK() )
+          {
+            // ok status: update pending_cnt and return, indicating if all are done
+            return pending_cnt.fetch_sub( 1, std::memory_order_relaxed ) == 1;
+          }
+          // there's a failure: delay decrementing pending_cnt until we've
+          // updated failed_cnt (and posibly recorded the error), otherwise
+          // a succeeding result could return true before we've saved the error
+          const size_t nb = failed_cnt.fetch_add( 1, std::memory_order_relaxed );
```
bbockelm: This code doesn't actually work as advertised, right?  `memory_order_relaxed` only guarantees the `fetch_add` occurs atomically but doesn't guarantee the ordering visibility with respect to other atomics.  From [CPP ref](https://en.cppreference.com/w/cpp/atomic/memory_order.html#Relaxed_ordering)

> Atomic operations tagged memory_order_relaxed are not synchronization operations; they do not impose an order among concurrent memory accesses. They only guarantee atomicity and modification order consistency.

It's perfectly valid for another thread to see the decrement of the pending count before the increment of the failed count.

Without knowing the rest of the rules for this synchronization, I suspect you want `memory_order_acq_rel` instead?  You'll also want the corresponding `memory_order_acquire` for reading as well.

smithdh: Thanks @bbockelm I think you're right; at least memory_order_acq_rel on fetch_sub of pending_cnt (altough from a diagnostic point I think on x86 _64 it will do acq_rel on the atomic op, so problems with this probably won't show up in tests on that).

But I think I'll break this PR into two, as the fix for #2615 seems ok (and was reported externally); with this commit for issue #2619 (opened by me) in a separate PR I can set to draft and possibly take it slower. There are several uses of atomics in XrdClParallelOperation.hh, all order_relaxed and maye I'd better look at them.

bbockelm: Right - x86 has a relatively strong memory model, meaning that switching to acquire/release is possibly zero-cost (assuming the values aren't kept for surprisingly long amounts of time in the registers).

However, without the acquire/release, compiler re-ordering is permitted for the produced assembler.  So, it's possible it works 100% in CI now only to break with a compiler upgrade.

src/XrdCl/XrdClClassicCopyJob.cc
```diff
@@ -824,13 +824,14 @@ namespace
 
           char *buffer = new char[chunkSize];
           ChunkHandler *ch = new ChunkHandler();
-          ch->status = pUsePgRead
+          auto st = pUsePgRead
                      ? reader->PgRead( pCurrentOffset, chunkSize, buffer, ch )
                      : reader->Read( pCurrentOffset, chunkSize, buffer, ch );
           pChunks.push( ch );
```
amadio, line 830: I have a question here. Should this (i.e. `pChunks.push( ch );`) not be done only when `st.IsOK() == true`?

amadio: Actually, reading the code a bit more, I think the above is correct, so marking this as resolved.

PR #2622, opened 2025-10-22
[Cl] Fix some occasional XrdCl::FileCopyTest.NormalCopyTest failures.

smithdh:
Draft for now, to check if memory ordering on atomics is correct.


amadio:
No, it's not needed to split. I'd just like to check the changes throroughly to avoid a similar situation as in #2629.

PR #2613, opened 2025-10-03
Bridge wait/waitrsp synch. (To fix #2602)

smithdh:
Opening as draft at first. For more testing.

Issue #2602, opened 2025-09-18
Bridge re-drive is invoked after protocol is Reset


bbockelm:
### Operating System

Alma Linux 10

### XRootD Version

5.8.4

### Description

We've noticed approximately daily crashes recently with the following stack trace:

```
#0  0x00007fae48357540 in XrdXrootdProtocol::fsError (this=this@entry=0x7fa344065678, rc=3, 
    opC=opC@entry=5 '\005', myError=..., 
    Path=Path@entry=0x7fa19805c400 "/<filename redacted>", Cgi=<optimized out>)
    at /usr/src/debug/xrootd-5.8.4-1.2.osg24.el9.x86_64/xrootd-5.8.4/src/XrdXrootd/XrdXrootdXeq.cc:3780
#1  0x00007fae4835d9b3 in XrdXrootdProtocol::do_Open (this=0x7fa344065678)
    at /usr/src/debug/xrootd-5.8.4-1.2.osg24.el9.x86_64/xrootd-5.8.4/src/XrdXrootd/XrdXrootdXeq.cc:1523
#2  0x00007fae48353848 in XrdXrootdTransit::Redrive (this=0x7fa344065670)
    at /usr/src/debug/xrootd-5.8.4-1.2.osg24.el9.x86_64/xrootd-5.8.4/src/XrdXrootd/XrdXrootdTransit.cc:460
#3  0x00007fae481caf8a in XrdScheduler::Run (this=0x414780 <XrdGlobal::Sched>)
    at /usr/src/debug/xrootd-5.8.4-1.2.osg24.el9.x86_64/xrootd-5.8.4/src/Xrd/XrdScheduler.cc:365
```

The relevant lines of code are:

```
   if (rc >= SFS_STALL)
      {SI->stallCnt++;
=>     TRACEI(STALL, Response.ID() <<"stalling client for " <<rc <<" sec");
       rs = Response.Send(kXR_wait, rc, eMsg);
       if (myError.extData()) myError.Reset();
```

Which seemed to be a funny place for a segfault!  However, unwrapping the layers of macros (and looking at the assembler), the problem is the trace includes the value of `Link->ID`.  However, per GDB, the `Link` member is set to `nullptr`, so `Link->ID` is a null pointer dereference.

This indicates `Reset` was called on the `XrdXrootdProtocol` object before the redrive fired!

I have not yet figured out how this could happen.

amadio:
[...]

PR #2623, opened 2025-10-22
Make sure poller thread waits for all eventfd_writers (Related #2602)

smithdh:
Breaking out part of the PR #2613 for issue #2602.


smithdh:
I rebased this PR with latest master. And removed 1 of the 2 commits: Removed a change that concerned closing a link in XrdLink::Enable if Poller->Enable(linkXQ.PollInfo) fails. I removed it because with pr #2653, if that change was wanted, it likely be better suited in that pr. But as a second reason, while a path with failed enable was present in tests to trigger #2602 on 5.8.4, I believe this is fixed with the comination of this (#2623 & #2613). So it is seems better to have the minimal change.

PR #2627, opened 2025-10-28
Add support for WLCG staging tokens

jafonso:
## Intro

This PR extends the XRootD authorizaton plugin to support WLCG tokens with `storage.stage` and `storage.poll` (see note bellow) capabilities.

It comes as a follow up to the XRootD & FTS Workshop 2025, where we agreed to move forward with discussing and implementing these changes in XRootD.

## Proposes changes

I have added two new operations to `XrdAccAuthorize.hh` and their corresponding priviledges in `XrdAccPrivs.hh`
- `AOP_Stage ` / `XrdAccPriv_Stage `
- `AOP_Poll ` / `XrdAccPriv_Poll `

The changes above modify two header files, and therefore may be non-backard compatible...

Besides these, I have also done other changes that mainly translate the `storage.stage` and `storage.poll` scopes into the new enums.

### To be discussed

At the moment, the discussion about the new scopes `storage.stage` and `storage.poll` is still being finalized by the WLCG DOMA BDW working group, and therefore may be subject to changes:
- https://github.com/WLCG-AuthZ-WG/common-jwt-profile/blob/master/profile.md#221-capability-based-authorization-scope

However, as discussed, we would like to move ahead with an initial version that can already be used with EOSCTA and FTS.
In addition to this, with the upcoming release of XRootD 6.0, it makes sense to merge any non-backward compatible changes soon.

These are the possible changes that may affect this implementation:
- `storage.stage`:
  - It currently allows `AOP_Read/AOP_Readdir` operations, but this may be reverted. However, the changes would be only on `XrdOfs.cc`, not the header file.
- `storage.poll`:
  - The naming of this scope is still under discussion. Other proposed names include `storage.stat` or `tape.poll`.
  - In principle, it will give access to any subset of `stat` metadata that is modified during a stage operation (hence why a new `AOP_Poll` entry may be required), but there is a small chance that it ends up being be the same as `stat`.
  - I wonder if adding it now makes sense, but we may not be able to add it later after 6.0.

### DIFF
--- a/src/XrdAcc/XrdAccPrivs.hh
+++ b/src/XrdAcc/XrdAccPrivs.hh
@@ -36,7 +36,7 @@
   
 // Recognized privileges
 //
-enum XrdAccPrivs {XrdAccPriv_All    = 0x07f,
+enum XrdAccPrivs {XrdAccPriv_All    = 0x1ff,
                   XrdAccPriv_Chmod  = 0x063,  // Insert + Open r/w + Delete
                   XrdAccPriv_Chown  = 0x063,  // Insert + Open r/w + Delete
                   XrdAccPriv_Create = 0x062,  // Insert + Open r/w
@@ -50,6 +50,8 @@ enum XrdAccPrivs {XrdAccPriv_All    = 0x07f,
                   XrdAccPriv_Readdir= 0x020,
                   XrdAccPriv_Write  = 0x040,
                   XrdAccPriv_Update = 0x060,
+                  XrdAccPriv_Stage  = 0x180,  // Stage + Poll
+                  XrdAccPriv_Poll   = 0x100,  // Poll
                   XrdAccPriv_None   = 0x000
                  };
   
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -130,6 +130,12 @@ XrdAccPrivs AddPriv(Access_Operation op, XrdAccPrivs privs)
         case AOP_Update:
             new_privs |= static_cast<int>(XrdAccPriv_Update);
             break;
+        case AOP_Stage:
+            new_privs |= static_cast<int>(XrdAccPriv_Stage);
+            break;
+        case AOP_Poll:
+            new_privs |= static_cast<int>(XrdAccPriv_Poll);
+            break;
     };
     return static_cast<XrdAccPrivs>(new_privs);
 }
@@ -151,6 +157,8 @@ const std::string OpToName(Access_Operation op) {
         case AOP_Rename: return "mv";
         case AOP_Stat: return "stat";
         case AOP_Update: return "update";
+        case AOP_Stage: return "stage";
+        case AOP_Poll: return "poll";
     };
     return "unknown";
 }
@@ -965,6 +973,11 @@ private:
                     xrd_rules.emplace_back(AOP_Chmod, path);
                     xrd_rules.emplace_back(AOP_Stat, path);
                     xrd_rules.emplace_back(AOP_Delete, path);
+                } else if (!strcmp(acl_authz, "storage.stage")) {
+                    xrd_rules.emplace_back(AOP_Stage, path);
+                    xrd_rules.emplace_back(AOP_Poll, path);
+                } else if (!strcmp(acl_authz, "storage.poll")) {
+                    xrd_rules.emplace_back(AOP_Poll, path);
                 } else if (!strcmp(acl_authz, "write")) {
                     paths_write_seen.insert(path);
                 }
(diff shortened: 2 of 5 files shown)


src/XrdOfs/XrdOfs.cc
```diff
@@ -2187,7 +2187,7 @@ int XrdOfs::prepare(      XrdSfsPrep       &pargs,      // In
 //
    if (prepAuth)
       while(tp)
-           {AUTHORIZE(client,0,AOP_Read,"prepare",tp->text,out_error);
+           {AUTHORIZE(client,0,AOP_Stage,"prepare",tp->text,out_error);
```
jafonso: I wonder if this change makes sense...

EOS already uses it's own implementation of `XrdSfsFileSystem::prepare(...)` so this probably has no impact on it.

src/XrdAcc/XrdAccPrivs.hh
```diff
@@ -50,6 +50,8 @@ enum XrdAccPrivs {XrdAccPriv_All    = 0x07f,
                   XrdAccPriv_Readdir= 0x020,
                   XrdAccPriv_Write  = 0x040,
                   XrdAccPriv_Update = 0x060,
+                  XrdAccPriv_Stage  = 0x080,  // Stage
```
abh3: Didn't we decide that "stage' also includes 'poll" as there is no use case without having both of them? If so, this 0x180.

src/XrdOfs/XrdOfs.cc
abh3: No, this is not correct. The prepare here is being vectors to the cmsd in order to find which server has the file. That should only be done if the client has read permission for the file. So, please remove the change. Prepare for staging is handled elsewhere.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -948,13 +956,15 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
                     xrd_rules.emplace_back(AOP_Read, path);
                     xrd_rules.emplace_back(AOP_Readdir, path);
                     xrd_rules.emplace_back(AOP_Stat, path);
+                    xrd_rules.emplace_back(AOP_Poll, path);
```
abh3: I don't think this is correct. The client merely has read privileges since tokens have no special privileges for readdir and stat, these are included. However, there is a storage.poll and so it needs o be present if polling is allowed. Read does not imply poll.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -964,7 +974,16 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
...
                     xrd_rules.emplace_back(AOP_Chmod, path);
                     xrd_rules.emplace_back(AOP_Stat, path);
+                    xrd_rules.emplace_back(AOP_Poll, path);
                     xrd_rules.emplace_back(AOP_Delete, path);
+                } else if (!strcmp(acl_authz, "storage.stage")) {
+                    xrd_rules.emplace_back(AOP_Stage, path);
+                    xrd_rules.emplace_back(AOP_Poll, path);
+                    xrd_rules.emplace_back(AOP_Read, path); // May be removed
+                    xrd_rules.emplace_back(AOP_Readdir, path); // May be removed
+                    xrd_rules.emplace_back(AOP_Stat, path);
```
abh3: Having stage does not imply anything other than stage and (I think) poll. So, one should not add read, readdir, or stat. Note here you do seem to concur that stage implies poll, so my previous comment on privilege bits is correct.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -980,6 +999,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
                 xrd_rules.emplace_back(AOP_Insert, write_path);
                 xrd_rules.emplace_back(AOP_Update, write_path);
                 xrd_rules.emplace_back(AOP_Stat, write_path);
+                xrd_rules.emplace_back(AOP_Poll, write_path);
```
abh3: Again, I don't think poll is implied here.

src/XrdOfs/XrdOfs.cc
jafonso: Ok, thank you for confirming. I will revert this.

src/XrdAcc/XrdAccPrivs.hh
jafonso: Yes, "stage" will always include "poll". I will update this.

src/XrdSciTokens/XrdSciTokensAccess.cc
jafonso: Yes, this was not so clear to me before, but after the discussion with @mpatrascoiu mentioned bellow, I also agree with your comment.

I will revert this.

PR #2632, opened 2025-10-31
Use checked out HEAD reference for PRs

dynamic-entropy:
As can be seen in the below job for a Test PR - the CI loses track of the `HEAD` related to the PR for the ABI job.
https://github.com/xrootd/xrootd/actions/runs/18978259360/job/54203708690

Screenshot From 2025-10-31 17-31-40

This PR reorders the build order for new (PR) and old (latest tagged commit) before ABI reports are compared.

### DIFF
--- a/.github/workflows/ABI.yml
+++ b/.github/workflows/ABI.yml
@@ -47,6 +47,12 @@ jobs:
     - name: Install XRootD build dependencies
       run: mk-build-deps --install --remove -s sudo debian/control <<< yes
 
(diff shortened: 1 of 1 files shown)


amadio:
I thought (wrongly) that `${GITHUB_HEAD_REF}` would contain a hash instead of just the branch name in the fork. The change here builds new first, then old, which at some point I did have while working on it. I'll merge, thanks.

Issue #2625, opened 2025-10-24
Server should return HTTP 202 instead of HTTP 201 when accepting third-party copy transfers


amadio:
The server should not return [HTTP 201 Created](https://httpwg.org/specs/rfc9110.html#status.201) when receiving a TPC, it should return [HTTP 202 Accepted](https://httpwg.org/specs/rfc9110.html#status.202) instead.

Quoting the links above:

> The 201 (Created) status code indicates that the request has been fulfilled and has resulted in one or more new resources being created.

> The 202 (Accepted) status code indicates that the request has been accepted for processing, but the processing has not been completed.

Since by the time the server replies the copy is not yet performed, the latter response seems much more appropriate.

These are the places where it needs to change:

https://github.com/xrootd/xrootd/blob/3fb2a426c32d0e893358b6f2af63c4c9a73e7afa/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L581-L582

https://github.com/xrootd/xrootd/blob/3fb2a426c32d0e893358b6f2af63c4c9a73e7afa/src/XrdHttpTpc/XrdHttpTpcMultistream.cc#L285-L286

See also: https://twiki.cern.ch/twiki/bin/view/LCG/HttpTpcTechnical#Response

mpatrascoiu:
Hello,

On the FTS side, this change falls in the Davix realm, but no impact is expected. Davix would be able to handle both `HTTP 201` and `HTTP 202`. However, to err on the side of caution, when you have a development server running, let me know and I will run some tests.

PR #2637, opened 2025-11-06
HTTP summary monitoring

dynamic-entropy:
This PR builds on top of HTTP GStream monitoring and provides HTTP Summary monitoring using the new XrdMonRoll interface. 

It provides counters for HTTP Request Verbs and HTTP Response Status codes as defined in the comment. https://github.com/xrootd/xrootd/pull/2604#issuecomment-3479530517

```
1. Total requests received by `HTTP Request verb` type
2. Total response count by `HTTP Status Code` type
```

#2604 is a parent PR, and we discussed that it needs to be reviewed and merged before this.

### DIFF
--- a/src/XrdHttp/XrdHttpMon.cc
+++ b/src/XrdHttp/XrdHttpMon.cc
@@ -12,23 +12,81 @@ typedef std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_
     StatsMatrix;
 
 StatsMatrix XrdHttpMon::statsInfo{};
+
 XrdXrootdGStream* XrdHttpMon::gStream = nullptr;
-XrdSysLogger* XrdHttpMon::logger = nullptr;
+XrdMonRoll* XrdHttpMon::mrollP = nullptr;
 std::chrono::seconds XrdHttpMon::flushPeriod{0};
 
-bool XrdHttpMon::Initialize(XrdSysLogger *logP, XrdXrootdGStream *gStreamP) {
-    if (!gStreamP || !logP) {
-        return false;
-    }
-    gStream = gStreamP;
-    logger = logP;
-    flushPeriod = std::chrono::seconds(gStream->GetAutoFlush());
-    eDest.logger(logP);
-    return true;
-}
+bool XrdHttpMon::hasGStream = false;
+bool XrdHttpMon::hasMonRoll = false;
+bool XrdHttpMon::isInitialized = false;
 
-bool XrdHttpMon::IsInitialized() {
-    return gStream != nullptr && logger != nullptr;
+RAtomic_uint64_t XrdHttpMon::verbCounters[XrdHttpReq::ReqType::rtCount] = {0};
+RAtomic_uint64_t XrdHttpMon::statusCounters[XrdHttpMon::StatusCodes::sc_Count] = {0};
+
+// Combined schema for HTTP statistics
+// This creates a JSON structure: {"httpReqStats": {...}, "httpStatusCodeStats": {...}}
+std::vector<XrdMonRoll::Item> XrdHttpMon::statsSchema = {
+// NOTE: Keep this mapping aligned to the XrdHttpReq enum
+    XrdMonRoll::Item("request", XrdMonRoll::Item::Schema::begObject),
+        XrdMonRoll::Item("Unknown",   verbCounters[0]),
+        XrdMonRoll::Item("Malformed", verbCounters[1]),
+        XrdMonRoll::Item("GET",       verbCounters[2]),
+        XrdMonRoll::Item("HEAD",      verbCounters[3]),
+        XrdMonRoll::Item("PUT",       verbCounters[4]),
+        XrdMonRoll::Item("OPTIONS",   verbCounters[5]),
+        XrdMonRoll::Item("PATCH",     verbCounters[6]),
+        XrdMonRoll::Item("DELETE",    verbCounters[7]),
+        XrdMonRoll::Item("PROPFIND",  verbCounters[8]),
+        XrdMonRoll::Item("MKCOL",     verbCounters[9]),
+        XrdMonRoll::Item("MOVE",      verbCounters[10]),
+        XrdMonRoll::Item("POST",      verbCounters[11]),
+        XrdMonRoll::Item("COPY",      verbCounters[12]),
+    XrdMonRoll::Item("request", XrdMonRoll::Item::Schema::endObject),
+
+// NOTE: Keep this mapping strictly aligned with StatusCodes enum XrdHttpMon::StatusCode
+// The order and number of entries MUST match.
+    XrdMonRoll::Item("response", XrdMonRoll::Item::Schema::begObject),
+        XrdMonRoll::Item("100", statusCounters[sc_100]),
+        XrdMonRoll::Item("200", statusCounters[sc_200]),
+        XrdMonRoll::Item("201", statusCounters[sc_201]),
+        XrdMonRoll::Item("202", statusCounters[sc_202]),
+        XrdMonRoll::Item("206", statusCounters[sc_206]),
+        XrdMonRoll::Item("207", statusCounters[sc_207]),
+        XrdMonRoll::Item("302", statusCounters[sc_302]),
+        XrdMonRoll::Item("307", statusCounters[sc_307]),
+        XrdMonRoll::Item("400", statusCounters[sc_400]),
+        XrdMonRoll::Item("401", statusCounters[sc_401]),
+        XrdMonRoll::Item("403", statusCounters[sc_403]),
+        XrdMonRoll::Item("404", statusCounters[sc_404]),
+        XrdMonRoll::Item("405", statusCounters[sc_405]),
+        XrdMonRoll::Item("409", statusCounters[sc_409]),
+        XrdMonRoll::Item("416", statusCounters[sc_416]),
+        XrdMonRoll::Item("423", statusCounters[sc_423]),
+        XrdMonRoll::Item("500", statusCounters[sc_500]),
+        XrdMonRoll::Item("502", statusCounters[sc_502]),
+        XrdMonRoll::Item("504", statusCounters[sc_504]),
+        XrdMonRoll::Item("507", statusCounters[sc_507]),
+        XrdMonRoll::Item("OTHERS", statusCounters[sc_UNKNOWN]),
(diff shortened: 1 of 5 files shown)


abh3:
Not to worry about he uint32 issue. The next version of summary monitoring will accept most data types. We are shooting for that version (likely R6) with this addition. This is just to get the right statistics together.

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -13,9 +13,67 @@ typedef std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_
...
+
+XrdHttpMon::XrdHttpMon(XrdSysLogger *logP, XrdXrootdGStream *gStream, XrdMonRoll *mrollP)
+    : gStream(gStream), mrollP(mrollP) {
+
+    eDest.logger(logP);
+    if (gStream) {
+        flushPeriod = std::chrono::seconds(gStream->GetAutoFlush());
+    }
+    if (mrollP) {
+        mrollP->Register(XrdMonRoll::AddOn, "httpReqStats", verbCountersSchema);
```
abh3: This should be considered a plug-in as the http protocol is really a plug-in not an addon.

src/XrdHttp/XrdHttpMon.hh
```diff
@@ -17,6 +18,7 @@ class XrdHttpMon {
         sc_100,
         sc_200,
         sc_201,
+        sc_202,
```
abh3, line 22: Out of curiosity, why was this omitted from the gStream pull request?

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1003,14 +1003,18 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
...
-      pthread_t tid;
-      int rc;
-      if ((rc = XrdSysThread::Run(&tid, XrdHttpMon::Start, httpMon, 0, "Http Stats thread"))) {
-          eDest.Emsg("httpMon", rc, "create stats thread");
-          return rc;
+  XrdXrootdGStream *gs = (XrdXrootdGStream *)myEnv->GetPtr("http.gStream*");
+  XrdMonRoll *mrollP = (XrdMonRoll *)myEnv->GetPtr("XrdMonRoll*");
+
+  if (gs || mrollP) {
+      httpMon = new XrdHttpMon(eDest.logger(), gs, mrollP);
```
abh3: So, is the idea here that is we are doing summary monitoring (which will always be the case for protocols) we enable full monitoring even though that part may be disabled to simply not duplicate the statistics gathering; right? The question is if the full monitoring is disabled is there any optimization that can be done to avoid the extra work?

abh3 (commented):
Just a couple of questions to get a handle on the approach here. Otherwise, nothing looks worrisome.

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: True, that was an oversight. However, now I am confused as to why is it ever required for the registration? 
Since the result is the same after all. 
```cfg
xrd.report 127.0.0.1:9999 every 3 json plugins
```

I still need the "plugin" in the `xrd.report` config to get the pointer in the http plugin, but other than that, the output itself is not affected.
This is the result when its registered as an `AddOn`.

```json
{"statistics":{"tod":1763391006,"ver":"v5.8.2-135-g3b864adc0","src":"rchauhan-pc:7094","tos":1763391003,"pgm":"xrootd","ins":"http","pid":280113,"site":"http",
"stats_httpReqStats":{"Unknown":0,"Malformed":0,"GET":22,"HEAD":5,"PUT":21,"OPTIONS":1,"PATCH":0,"DELETE":10,"PROPFIND":2,"MKCOL":0,"MOVE":0,"POST":0,"COPY":0}}}

{"statistics":{"tod":1763391006,"ver":"v5.8.2-135-g3b864adc0","src":"rchauhan-pc:7094","tos":1763391003,"pgm":"xrootd","ins":"http","pid":280113,"site":"http",
"stats_httpStatusCodeStats":{"100":0,"200":36,"201":17,"202":0,"206":1,"207":2,"302":0,"307":0,"400":0,"401":0,"403":0,"404":1,"405":0,"409":1,"416":0,"423":0,"500":1,"502":0,"504":0,"507":2,"OTHERS":0}}}
```

src/XrdHttp/XrdHttpMon.hh
dynamic-entropy: It's just because the issue discussing the use of 202 instead of 201 was opened after the http_gstream PR. 
https://github.com/xrootd/xrootd/issues/2625

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Actually, I do not see the output when using `XrdMonRoll::Plugin` during registration. 
It only works for `AddOn`.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: The optimisation I can think of would be more if statements. 
We have theoretically four states now
```
1. Nothing enabled
2. Only summary monitoring is enabled
3. Only gstream monitoring is enabled
4. Both monitoring enabled
```

src/XrdHttp/XrdHttpMon.hh
dynamic-entropy: I have cherry-picked this commit to the gstream PR now

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -13,9 +13,71 @@ typedef std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_
...
+    eDest.logger(logP);
+    if (gStream != nullptr){
+        hasGStream = true;
+        flushPeriod = std::chrono::seconds(gStream->GetAutoFlush());
+    }
+
+    if (mrollP != nullptr) {
+        hasMonRoll = true;
+        mrollP->Register(XrdMonRoll::AddOn, "httpReqStats", verbCountersSchema);
+        mrollP->Register(XrdMonRoll::AddOn, "httpStatusCodeStats", statusCountersSchema);
```
abh3: I would prefer you use the V2 of summary monitoring and instead of using setMember using Item. You an then trivially combine both of these into a nicely structured JSON object.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -1604,29 +1608,39 @@ void XrdHttpProtocol::Record() {
...
     case XrdHttpReq::MonitState::ERR_PROT:
-      httpMon->RecordErrProt(CurrentReq.request, XrdHttpMon::ToStatusCode(code), duration);
+      XrdHttpMon::RecordGStreamErrProt(CurrentReq.request, statusCode, duration);
+      XrdHttpMon::RecordMonRollStatus(statusCode);
       CurrentReq.monState = XrdHttpReq::MonitState::DONE;
       return;
-    
+
     case XrdHttpReq::MonitState::DONE:
       eDest.Emsg("Record", "ERROR: Record called after state was set to DONE");
```
abh3: Does this reset itself somehow or is it now in permanent error state?

src/XrdHttp/XrdHttpMon.cc
```diff
@@ -13,9 +13,71 @@ typedef std::array<std::array<XrdHttpMon::HttpInfo, XrdHttpMon::StatusCodes::sc_
 
 StatsMatrix XrdHttpMon::statsInfo{};
 
-XrdHttpMon::XrdHttpMon(XrdSysLogger *logP, XrdXrootdGStream *gStream)
-    : flushPeriod(std::chrono::seconds(gStream->GetAutoFlush())), gStream(gStream)  {
-  eDest.logger(logP);
+bool XrdHttpMon::hasGStream = false;
+bool XrdHttpMon::hasMonRoll = false;
+
+RAtomic_uint XrdHttpMon::verbCounters[XrdHttpReq::ReqType::rtCount] = {0};
```
abh3: By using V2 of monroll you can use a larger counter, say uint64 or ulong. This addresses @bbockelm concern of wrapping counters.

src/XrdHttp/XrdHttpProtocol.cc
dynamic-entropy: This state resets for each request. 

This should **Never** happen; if we see this in the logs ever, it would mean that I failed to capture the correct entry and exit points.

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Indeed!

src/XrdHttp/XrdHttpMon.cc
dynamic-entropy: Sure! I did not realise that it had merged into master 2 weeks ago. There were a whole lot of commits together without a related issue or PR to follow.

Just added a commit that does that.

dynamic-entropy:
After using the new schema, we attain the following format for the summary packet. 

```json
{
  "statistics": {
    "tod": 1764662596,
    "ver": "v5.9.0-151-g7e3fcbe2a",
    "src": "b7a8de1eeace:7094",
    "tos": 1764662590,
    "pgm": "xrootd",
    "ins": "http",
    "pid": 55573,
    "site": "http",
    "stats_http_plugin": {
      "request": {
        "Unknown": 0,
        "Malformed": 0,
        "GET": 24,
        "HEAD": 11,
        "PUT": 21,
        "OPTIONS": 1,
        "PATCH": 0,
        "DELETE": 10,
        "PROPFIND": 2,
        "MKCOL": 0,
        "MOVE": 0,
        "POST": 0,
        "COPY": 0
      },
      "response": {
        "100": 0,
        "200": 44,
        "201": 17,
        "202": 0,
        "206": 1,
        "207": 2,
        "302": 0,
        "307": 0,
        "400": 0,
        "401": 0,
        "403": 1,
        "404": 1,
        "405": 0,
        "409": 1,
        "416": 0,
        "423": 0,
        "500": 0,
        "502": 0,
        "504": 0,
        "507": 2,
        "OTHERS": 0
      }
    }
  }
}
```

abh3 (approved):
This looks OK. I would suggest, however, to be more discerning on what should be public and what should be private in XrdHttpNib.hh. Right now, everything is public which seems a bit sloppy.

PR #2638, opened 2025-11-08
[XrdMacaroons] Ensure that parent directory creation permission is implied

### LINKED ISSUE #2611: Scitokens v Macaroons: latter too strict on Mkdir operations on parent directories?
### Operating System

N/A

### XRootD Version

5.8.4

### Description

Mihai reports that FTS requests using macaroons against Glasgow's xrootd server fail when their flow is:
macaroon for final file path
mkdirs for parent directories in the path above (to make containing directory)
write file.

They fail at the mkdir stage with permission denied.

(The operations succeed when authorised with scitokens).

Mihai believes this is due to the underlying filesystem implementation (pure rados via xrdceph here) not understanding directories...

...but looking at the xrootd code for Macaroons versus Scitokens, it seems to me that this is caused by differences in the code for each.

The Scitokens flow explicitly allows STAT and MKDIR against parent directories (by comparing the strings and checking if the requested operation is against a substring of the authorised path, terminated with a /)
The Macaroons flow seems to do the same only for STAT (and uses a less complete comparison).

Am I missing something, or should Macaroons also allow MKDIR here?

_No response_

### Additional Information

_No response_

dynamic-entropy:
Fixes #2611 

Extend `AOP_Mkdir` to also imply permission for parent path creation. 

In addition, fix the logic to restrict paths that pass a simple substring match but are not subdirectories. 
For example, do not allow `/data/sub` for a token path of `/data/subdir/myfile` and only allow `/data/subdir`.

### DIFF
--- a/src/XrdMacaroons/XrdMacaroonsAuthz.cc
+++ b/src/XrdMacaroons/XrdMacaroonsAuthz.cc
@@ -505,8 +505,12 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
         return 1;
     }
 
-    int result = strncmp(pred_str.c_str(), m_path.c_str(), pred_str.size());
-    if (!result)
+    // Allow operations under subdirectories and not substrings 
+    // For e.g. pred_str = "/data/sudir/mydir"
+    // Allows m_path = /data/subdir/mydir/newdir
+    // But rejects, m_path = /data/subdir/mydirmycoolname/newdir
+    int is_subdir = is_subdirectory(pred_str, m_path);
+    if (is_subdir)
     {
         m_log.Log(LogMask::Debug, "AuthzCheck", "path request verified for", m_path.c_str());
     }
@@ -514,13 +518,12 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
     // READ_METADATA (i.e AOP_Stat) permission for /foo/bar automatically implies permission
     // to READ_METADATA for /foo.
     // Similarly, MKDIR Pemissions for a parent path is implied.
-
     else if (m_oper == AOP_Stat || m_oper == AOP_Mkdir)
     {
-        result = !is_subdirectory(m_path, pred_str);
+        is_subdir = is_subdirectory(m_path, pred_str);
         const char *opName = (m_oper == AOP_Stat) ? "READ_METADATA" : "MKDIR";
         m_log.Log(LogMask::Debug, "AuthzCheck",
-                  (std::string(opName) + (result ? " Path request NOT allowed for" : " Path request verified for")).c_str(),
+                  (std::string(opName) + (is_subdir? " Path request verified for" : " Path request NOT allowed for")).c_str(),
                   m_path.c_str());
     }
     else
@@ -528,7 +531,7 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
         m_log.Log(LogMask::Debug, "AuthzCheck", "path request NOT allowed", m_path.c_str());
     }
 
-    return result;
+    return !is_subdir;
 }
 
 


src/XrdMacaroons/XrdMacaroonsAuthz.cc
```diff
@@ -509,13 +509,35 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
...
+    */
+
+    else if (m_oper == AOP_Stat || m_oper == AOP_Mkdir)
     {
         result = strncmp(m_path.c_str(), pred_str.c_str(), m_path.size());
-        if (!result) {m_log.Log(LogMask::Debug, "AuthzCheck", "READ_METADATA path request verified for", m_path.c_str());}
-        else {m_log.Log(LogMask::Debug, "AuthzCheck", "READ_METADATA path request NOT allowed", m_path.c_str());}
+        const char *opName = (m_oper == AOP_Stat) ? "READ_METADATA" : "MKDIR";
+
+        if (!result && m_path.size() < pred_str.size() && m_path.back() != '/')
```
dynamic-entropy: @amadio suggested the use of `src/XrdOuc/XrdOucPrivateUtils.hh:static inline bool is_subdirectory(const std::string& dir,...)`

I have modified the code to use that.
However, I realised that the first check is also too permissive. 
It allows permission for non-subdir paths, which I am not sure is intended; adding a commit to use directory path matching for that too.  

We can drop that commit if this is the desired behaviour.
For e.g. 
A macaroon with a path: `/mydir/mysubdir`
Allows creation of `/mydir/mysubdirbutalsonot/path`

src/XrdMacaroons/XrdMacaroonsAuthz.cc
```diff
@@ -504,18 +505,40 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
         return 1;
     }
 
-    int result = strncmp(pred_str.c_str(), m_path.c_str(), pred_str.size());
+    // Allow operations under subdirectories and not substrings 
+    // For e.g. pred_str = "/data/sudir/mydir"
+    // Allows m_path = /data/subdir/mydir/newdir
+    // But rejects, m_path = /data/subdir/mydirmycoolname/newdir
+    int result = !is_subdirectory(pred_str, m_path);
     if (!result)
```
amadio: Why `result = !...` and `!result` in the condition? I guess we can use `int is_subdir = is_subdirectory(...)`, then `if (is_subdir) ...` to be clearer.

src/XrdMacaroons/XrdMacaroonsAuthz.cc
```diff
@@ -504,18 +505,40 @@ AuthzCheck::verify_path(const unsigned char * pred, size_t pred_sz)
...
+    The following logic is intended:
+    Given the the macaroon path, i.e pred_str == /data/subdir/myfile
+    Allowed examples:
+    m_path = "/data/subdir/"  or "/data/subdir"
+    m_path = "/data"          or "/data/"
+
+    Not allowed examples
+    m_path = "/data/sub"      or "/data/sub/"
+    m_path = "/"              // root not allowed
+    */
```
amadio: I think the long comment is not needed, you can keep just something like
```
READ_METADATA permission to a given directory implies the same permissions to its parent directories
```

src/XrdMacaroons/XrdMacaroonsAuthz.cc
dynamic-entropy: Was complying with the existing structure as much as possible. Makes sense to change to `is_subdir`. Done

src/XrdMacaroons/XrdMacaroonsAuthz.cc
dynamic-entropy: Done

PR #2639, opened 2025-11-12
[XrdOuc] Correct is_subdirectory check for dir with trailing /

dynamic-entropy:
The current check fails checks for directories ending with a trailing slash

For e.g.
``` 
subdir = /data/subdir/mydir
dir = /data/subdir/
```

The current code only takes into account paths without a trailing slash: `subdir[dir.size()]` 
This PR adds checks for paths with a trailing slash: `dir[dir.size()-1]` 
The actual code uses string member functions instead of the array indexing shown above.

### DIFF
--- a/src/XrdOuc/XrdOucPrivateUtils.hh
+++ b/src/XrdOuc/XrdOucPrivateUtils.hh
@@ -33,13 +33,13 @@
 static inline bool is_subdirectory(const std::string& dir,
                                    const std::string& subdir)
 {
-    if (subdir.size() < dir.size())
+    if (subdir.size() < dir.size() || dir.empty())
       return false;
 
     if (subdir.compare(0, dir.size(), dir, 0, dir.size()) != 0)
       return false;
 
-    return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir == "/";
+    return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir.back() == '/';
 }
 
 /**


src/XrdOuc/XrdOucPrivateUtils.hh
```diff
@@ -39,7 +39,7 @@ static inline bool is_subdirectory(const std::string& dir,
     if (subdir.compare(0, dir.size(), dir, 0, dir.size()) != 0)
       return false;
 
-    return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir == "/";
+    return dir.size() == subdir.size() || subdir.at(dir.size()) == '/' || dir.back() == '/' || dir == "/";
```
amadio: We don't need to do bounds check on `subdir[dir.size()]` because here `subdir` is guaranteed to be larger than `dir`. Either `dir.size() == subdir.size()` will be `true` and we return true, or `subdir.size() < dir.size()` will be true in line 36 and we return false. We can also then remove the case of `dir == "/"`, since it's already covered by `dir.back() == '/'` after your change, so it's dead code now.

dynamic-entropy: Yup, the use of member function `at()` is for consistency and not to get the added bound check/exception handling.

dynamic-entropy: Done!

src/XrdOuc/XrdOucPrivateUtils.hh
```diff
@@ -39,7 +39,7 @@ static inline bool is_subdirectory(const std::string& dir,
     if (subdir.compare(0, dir.size(), dir, 0, dir.size()) != 0)
       return false;
 
-    return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir == "/";
+    return dir.size() == subdir.size() || subdir[dir.size()] == '/' || dir.back() == '/';
```
amadio, line 42: There is another problem, sigh. `dir.back()` is undefined behavior if `dir == ""`, that's why I had it originally as `dir == "/"` before. We also need to ensure that `dir.size() > 0` somewhere or return false if `dir.empty() == true` right away. So, I think we can do it in line 36 as `if (dir.empty() || subdir.size() < dir.size()) return false;`. Right now `is_subdirectory("", "")` returns true, which is not good.

dynamic-entropy: Hmm, I did not take "empty string" in my test cases. 
I would say, the code calling `is_subdirectory` would already handle the empty "dir" logic. 

> is_subdirectory("", "") returns true, which is not good

Practically, this would not have any meaning in the calling code either

But okay, we can handle this here

dynamic-entropy: Done!

amadio: Better be safe than sorry (i.e. allow undefined behavior).

PR #2650, opened 2025-11-21
Python: expose fSetXattrAdler32 in python

chaen:
When running at sites which have mounted writable file system, we simply use posix semantic to "upload" the files produced by jobs. 
But when we want to access these files from outside (other jobs or FTS), we do it via an xrootd server. 
This triggers a checksum computation which is intensive, and we end up ddossing the server.

The idea is to be able, from python, to write the checksum as extended attributes directly from the workernode.

### DIFF
--- a/python/libs/client/__init__.py
+++ b/python/libs/client/__init__.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, division, print_function
 from .glob_funcs import glob, iglob
 from .filesystem import FileSystem as FileSystem
 from .file import File as File
+from pyxrootd.client import setXAttrAdler32_cpp as setXAttrAdler32
 from .url import URL as URL
 from .copyprocess import CopyProcess as CopyProcess
 from .env import EnvPutString as EnvPutString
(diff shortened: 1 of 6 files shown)


python/libs/client/__init__.py
```diff
@@ -3,6 +3,7 @@
 from .glob_funcs import glob, iglob
 from .filesystem import FileSystem as FileSystem
 from .file import File as File
+from pyxrootd.client import setXAttrAdler32_cpp as setXAttrAdler32
```
amadio, line 6: This should probably be a relative include, like the other ones.

chaen: This cannot be relative, as it comes from another module (`pyxrootd` as opposed to `XRootD`). 
That being said I can move it to a different line for stylistic reasons if you want

amadio: We could expose it here like we do for functions in `env.py`. For example, by adding the new functions into `glob_funcs`, what do you think?

chaen: That's entirely up to your preference ! Let me know how you want it done, or just push it if you prefer :-)

amadio: Let me give this a go then, I will probably merge this next week and do my change on top. Thanks!

amadio: Alright, I'm merging this as is and will apply any changes for the release afterwards. Adding md5 may have to come later in 6.1 as we're pressed on time for this release.

PR #2656, opened 2025-11-28
[XrdHttpTpc] Use HTTP 202 Accepted instead of HTTP 201 Created 

### LINKED ISSUE #2625: Server should return HTTP 202 instead of HTTP 201 when accepting third-party copy transfers

dynamic-entropy:
Fixes #2625 

Note that the description "Created" is replaced with `NULL`, and not "Accepted" since we added a common translation for status codes and string `httpStatusToString` in XrdHttpUtils.

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
@@ -283,7 +283,7 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
     curl_multi_setopt(multi_handle, CURLMOPT_MAX_HOST_CONNECTIONS, streams);
 
     // Start response to client prior to the first call to curl_multi_perform
-    int retval = req.StartChunkedResp(201, "Created", "Content-Type: text/plain");
+    int retval = req.StartChunkedResp(202, NULL, "Content-Type: text/plain");
     if (retval) {
         logTransferEvent(LogMask::Error, rec, "RESPONSE_FAIL",
             "Failed to send the initial response to the TPC client");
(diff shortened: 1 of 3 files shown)


src/XrdHttpTpc/XrdHttpTpcMultistream.cc
```diff
@@ -283,7 +283,7 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
     curl_multi_setopt(multi_handle, CURLMOPT_MAX_HOST_CONNECTIONS, streams);
 
     // Start response to client prior to the first call to curl_multi_perform
-    int retval = req.StartChunkedResp(201, "Created", "Content-Type: text/plain");
+    int retval = req.StartChunkedResp(202, NULL, "Content-Type: text/plain");
```
amadio, line 286: `NULL` is a C construct, in "modern" C++, one is encouraged to use `nullptr` instead.

dynamic-entropy: Yup, but this is to keep it consistent with the rest of the code

amadio: Ok, leave it as is then.

PR #2657, opened 2025-11-28
Remove support for erasure coding (XrdEc)

amadio:
This plugin, as far as we are aware, has no users. It is also buggy and unmaintained. We should consider removing it in the upcoming major release (XRootD 6.0), so I prepared the changes in this pull request in case we decide to remove it. The test coverage within `src/XrdEc` was ok, but the tests all crash when running with sanitizers, and code supporting `XrdEc` elsewhere (i.e. within `XrdCl`, `XrdHttp`, etc) was mostly untested.


src/XrdCl/CMakeLists.txt
```diff
@@ -82,21 +82,6 @@ target_link_libraries(XrdCl
     ${CMAKE_DL_LIBS}
 )
 
-if(BUILD_XRDEC)
-  target_sources(XrdCl
-    PRIVATE
-      ../XrdEc/XrdEcRedundancyProvider.cc
-      ../XrdEc/XrdEcUtilities.cc
-      ../XrdEc/XrdEcStrmWriter.cc
-      ../XrdEc/XrdEcReader.cc
```
amadio, line 91: This is a source of ODR violations in `libXrdCl.so`.

abh3 (approved):
I approve given the way this will be hyandled.

wyang007:
Sorry I missed Guilherme message. Can we defer this decision a bit until SLAC joins HEP-CCE. It is likely this will still be dropped but it would be great if we can take look of what HEP-CCE cares about before we drop it.

But you can exclude it from being compiled and distributed in v6 so as not the delay the release of v6.

abh3:
I don't think that corresponds to the R6 timeline which has a current deadline of 1/18/26. I think the way we are removing it allows us to bring it back with the required fixes. But leaving in code that we know has serious bugs is not a good idea. In the end @amadio gets to decide.

amadio:
Sorry @abh3, didn't see your comment before replying to the list. Let's take a decision together next week, but I tend to agree that since the code is easy to recover, unless a maintainer shows up, it will be hard to take a different decision on this than the removal.

abh3:
I agree. I think we decided that the source cod will be left in github, just won't be included in CMAKE. Also, I believe we decided that the EC addition to POSIX will be removed but we will leave a comment in the EC source directory pointing to the commit hashes for restoration if need be. Of course, we realize that this may produce conflicts that will need to be resolved in the future should restoration be attempted but that part of the overall work.

PR #2658, opened 2025-11-28
Correct path & streamid for aio multi-path response

smithdh:


src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -2430,13 +2430,19 @@ int XrdXrootdProtocol::do_Read()
...
             if (!pathID) pP = this;
                else {if (!(pP = VerifyStream(retc, pathID, false))) return retc;
                      if (pP->linkAioReq >= as_maxperlnk) pP = 0;
                     }
-            if (pP && (aioP = XrdXrootdNormAio::Alloc(pP,Response,IO.File)))
+            if (pP)
+               {XrdXrootdResponse TmpRsp;
+                TmpRsp = Response;
+                TmpRsp.Set(pP->Link);
+                aioP = XrdXrootdNormAio::Alloc(pP,TmpRsp,IO.File);
```
abh3, line 2450: I think it would good to have a comment of why passing TmpResp to Alloc() is OK even though the object will go out of scope upon return (i.e. the contents is copied) would be helpful. The same goes where it gets done elsewhere.

smithdh: thanks Andy. I'll put in the recommendation as soon as i can.

Issue #2512, opened 2025-05-18
[XrdCl] (XRootD 6) Plugin file construction without opening


bbockelm:
There should be an alternate file constructor that internally creates the plugin object _without_ opening.

I have various properties that need to be set on the plugin object that affect its open behavior.  However, if I call `SetProperty` prior to opening the file, the underlying plugin object is not yet created; if I call `SetProperty` after opening the file, it's too late to affect the open.

This is straightforward for a Filesystem object: you create the filesystem, providing the URL to the constructor.  This causes the plugin to be initialized.  Then, you call `SetProperty` as appropriate and then finally the actual actions (make directory, stat, list dir, etc).  For the File object, this is impossible as the plugin is only initialized inside the `XrdCl::Open` call.

(For a workaround, I settled on opening the file in a way that will "safely fail" but load the plugin, then reusing the file object for a real open.  Pretty ugly).

I ran across this when splitting out a curl-only plugin from XrdClPelican.

PR #2661, opened 2025-12-01
[XrdCl] Add constructor to InitialisePlugin without opening the file

dynamic-entropy:
Fixes: #2512 
This PR adds a constructor to `XrdCl::File`, similar to `XrdCl::Filesystem`, that creates the file object on constructor call as opposed to during the Open call, allowing plugins to set flags, i.e `myfile->SetProperty`. 

Then the xrdcl-curl PR https://github.com/xrootd/xrootd/pull/2567 can be patched with

```patch
diff --git a/src/XrdClS3/XrdClS3DownloadHandler.cc b/src/XrdClS3/XrdClS3DownloadHandler.cc
index 8b5b78aa4..1c9884ff4 100644
--- a/src/XrdClS3/XrdClS3DownloadHandler.cc
+++ b/src/XrdClS3/XrdClS3DownloadHandler.cc
@@ -232,15 +232,7 @@ S3DownloadHandler::CloseHandler::HandleResponse(XrdCl::XRootDStatus *status_raw,
 XrdCl::XRootDStatus
 XrdClS3::DownloadUrl(const std::string &url, XrdClCurl::HeaderCallout *header_callout, XrdCl::ResponseHandler *handler, time_t timeout)
 {
-    std::unique_ptr<XrdCl::File> http_file(new XrdCl::File());
-    // Hack - we need to set a few properties on the file object before the open occurs.
-    // However, the "real" (plugin) file object is not created until the open call.
-    // This forces the plugin object to be created, so we can set the properties and Open later.
-    auto status = http_file->Open(url, XrdCl::OpenFlags::Compress, XrdCl::Access::None, nullptr, time_t(0));
-    if (!status.IsOK()) {
-        return status;
-    }
-
+    std::unique_ptr<XrdCl::File> http_file(new XrdCl::File(url));

     if (header_callout) {
         auto callout_loc = reinterpret_cast<long long>(header_callout);
diff --git a/src/XrdClS3/XrdClS3File.cc b/src/XrdClS3/XrdClS3File.cc
index f760ed7f5..ae2c65296 100644
--- a/src/XrdClS3/XrdClS3File.cc
+++ b/src/XrdClS3/XrdClS3File.cc
@@ -146,14 +146,8 @@ File::GetFileHandle(const std::string &s3_url) {
         return std::make_tuple(XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidAddr, 0, "Invalid generated XrdCl URL"), "", nullptr);
     }
     m_url = https_url;
-    std::unique_ptr<XrdCl::File> wrapped_file(new XrdCl::File());
-    // Hack - we need to set a few properties on the file object before the open occurs.
-    // However, the "real" (plugin) file object is not created until the open call.
-    // This forces the plugin object to be created, so we can set the properties and Open later.
-    auto status = wrapped_file->Open(url.GetURL(), XrdCl::OpenFlags::Compress, XrdCl::Access::None, nullptr, time_t(0));
-    if (!status.IsOK()) {
-        return std::make_tuple(status, "", nullptr);
-    }
+
+    std::unique_ptr<XrdCl::File> wrapped_file(new XrdCl::File(url.GetURL()));

     std::stringstream ss;
     ss << std::hex << reinterpret_cast<long long>(&m_header_callout);
```

### DIFF
--- a/src/XrdCl/XrdClFile.cc
+++ b/src/XrdCl/XrdClFile.cc
@@ -74,6 +74,15 @@ namespace XrdCl
   }
 
   //----------------------------------------------------------------------------
+  // Constructor
+  //----------------------------------------------------------------------------
+  File::File(const std::string &url, bool enablePlugIns): pPlugIn(0), pEnablePlugIns(enablePlugIns)
+  {
+    InitPlugin(url);
+    pImpl = new FileImpl(pPlugIn);
+  }
+
+  //----------------------------------------------------------------------------
   // Destructor
   //----------------------------------------------------------------------------
   File::~File()
@@ -93,6 +102,23 @@ namespace XrdCl
     delete pPlugIn;
   }
 
+  void File::InitPlugin(const std::string &url) {
+
+    if (pEnablePlugIns && !pPlugIn) {
+      Log *log = DefaultEnv::GetLog();
+      PlugInFactory *fact = DefaultEnv::GetPlugInManager()->GetFactory(url);
+      if (fact) {
+        pPlugIn = fact->CreateFile(url);
+        if (!pPlugIn) {
+          log->Error(FileMsg,
+                     "Plug-in factory failed to produce a plug-in "
+                     "for %s, continuing without one",
+                     url.c_str());
+        }
+      }
+    }
+  }
+
   //----------------------------------------------------------------------------
   // Open the file pointed to by the given URL - async
   //----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/XrdCl/XrdClFile.hh
```diff
@@ -831,6 +842,14 @@ namespace XrdCl
       template <bool HasHndl>
       friend class ChkptWrtVImpl;
 
+      //------------------------------------------------------------------------
+      //! Helper to Initialise Plugin
+      //!
+      //! @param url     url of the file to intialise the plugin for
+      //! @return        status of the operation
+      //------------------------------------------------------------------------
+      void InitialisePlugin( const std::string &url);
```
amadio: I suppose one may want to call this after creating a blank file, to initialize the plugin before opening the file, so maybe this should be public? Also, to avoid a US vs UK argument on how to spell things, can we call it just `InitPlugin(...)`?

src/XrdCl/XrdClFile.cc
```diff
@@ -73,6 +73,12 @@ namespace XrdCl
     pImpl = new FileImpl( virtRedirect == EnableVirtRedirect, pPlugIn );
   }
 
+  File::File(const std::string &url, bool enablePlugIns): pPlugIn(0), pEnablePlugIns(enablePlugIns) {
+
+    pImpl = new FileImpl(pPlugIn);
+    InitialisePlugin(url);
```
amadio: Shouldn't these two lines be reversed? I mean, you want to get a `pImpl` created after `pPlugIn` has been initialized, rather than with a null pointer, no?

dynamic-entropy: Indeed. I might have messed this up in a rebase, fixed now. Thanks for catching.

src/XrdCl/XrdClFile.hh
dynamic-entropy: Changed to `InitPlugin`. Though we know which spellings are the correct ones ;)

> I suppose one may want to call this after creating a blank file, to initialize the plugin before opening the file, so maybe this should be public?

Not sure what that means. By blank file do you mean an "empty file" or a "file object" without a known URL?
The former would still have a url, so the constructor will initialise it.
For the latter, passing an empty string as a url will initialise the default factory if available.

PR #2676, opened 2025-12-15
[XrdCl] Adds -h option to xrdfs ls for human readable sizes

ccaffy:


src/XrdCl/XrdClFS.cc
```diff
@@ -281,8 +281,57 @@ uint32_t nbDigits( uint64_t nb )
   return uint32_t( log10( double(nb) ) + 1);
 }
 
+std::string generateHumanSize(uint64_t size,uint64_t base) {
```
abh3: It would be better to use the existing library function int XrdOucUtils::fmtBytes() which does this. Is there a reason not to do so?

abh3 (changes requested):
Just a simplification suggestion. I will mark this as request changes but there may be something I am overlooking here.

src/XrdCl/XrdClFS.cc
ccaffy: Hi Andy,

This is because the function `XrdOucUtils::fmtBytes()` could do what I want but is incorrect w.r.t the Linux ls output command.

Here is an example using `fmtBytes()`:
```
$ xrdfs root://localhost:1095// ls -l /tmp/ | grep file
-rw-r--r-- xrootddev xrootddev 1073741845 2025-12-15 15:21:19 /tmp/file.1G
-rw-r--r-- xrootddev xrootddev    1048576 2025-12-12 15:46:59 /tmp/file.1M
-rw-r--r-- xrootddev xrootddev    1088576 2025-12-15 15:33:19 /tmp/file.1M2
$ xrdfs root://localhost:1095// ls -l -h /tmp/ | grep file
-rw-r--r-- xrootddev xrootddev 1.0G 2025-12-15 15:21:19 /tmp/file.1G
-rw-r--r-- xrootddev xrootddev 1.0M 2025-12-12 15:46:59 /tmp/file.1M
-rw-r--r-- xrootddev xrootddev 1.0M 2025-12-15 15:33:19 /tmp/file.1M2
[xrootddev@ccaffy-xrootd-dev-al9 http_server_certificate_tests]$
```

While the output with ls is:

```
$ ls -l /tmp/ | grep file
-rw-r--r--. 1 xrootddev xrootddev 1073741845 Dec 15 16:21 file.1G
-rw-r--r--. 1 xrootddev xrootddev    1048576 Dec 12 16:46 file.1M
-rw-r--r--. 1 xrootddev xrootddev    1088576 Dec 15 16:33 file.1M2
$ ls -lh /tmp/ | grep file
-rw-r--r--. 1 xrootddev xrootddev 1.1G Dec 15 16:21 file.1G
-rw-r--r--. 1 xrootddev xrootddev 1.0M Dec 12 16:46 file.1M
-rw-r--r--. 1 xrootddev xrootddev 1.1M Dec 15 16:33 file.1M2

```

See the difference for the files file.1G and file.1M2?

My function gives the same output as the Linux `ls` and one can choose the base if they want to (if we need that in the future...)

I'm happy to put my function in XrdOucUtils for sure yes!

ccaffy: done

PR #2677, opened 2025-12-16
[XrdSciTokens] Check cleanup timer while holding mutex to prevent cleanup runs from stacking up

jthiltges:
Each call to `Access()` includes a `Check()` to see if the configuration is due to be reloaded.

If the `Reconfig()` call is delayed (we've seen it due to `m_config_lock` contention) while it's due to reload the configuration, many threads will try to call `Reconfig()`. We saw hundreds of waiting threads on a busy server (which was having other related problems related to issuer key retrieval).

To improve this, we add a second check after getting the mutex and immediately update the `m_next_clean` timer.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -521,7 +521,7 @@ public:
         uint64_t now = monotonic_time();
         Check(now);
         {
-            std::lock_guard<std::mutex> guard(m_mutex);
+            std::lock_guard<std::mutex> guard(m_map_mutex);
             const auto iter = m_map.find(authz);
             if (iter != m_map.end() && !iter->second->expired()) {
                 access_rules = iter->second;
@@ -552,7 +552,7 @@ public:
                 m_log.Log(LogMask::Warning, "Access", "Error generating ACLs for authorization", exc.what());
                 return OnMissing(Entity, path, oper, env);
             }
-            std::lock_guard<std::mutex> guard(m_mutex);
+            std::lock_guard<std::mutex> guard(m_map_mutex);
             m_map[authz] = access_rules;
         } else if (m_log.getMsgMask() & LogMask::Debug) {
             m_log.Log(LogMask::Debug, "Access", "Cached token", access_rules->str().c_str());
@@ -1363,14 +1363,22 @@ private:
 
     void Check(uint64_t now)
     {
-        if (now <= m_next_clean) {return;}
-        std::lock_guard<std::mutex> guard(m_mutex);
+        // Bail out if another thread is already checking
+        std::unique_lock<std::mutex> lock(m_check_mutex, std::try_to_lock);
+        if (!lock.owns_lock()) {return;}
 
-        for (auto iter = m_map.begin(); iter != m_map.end(); ) {
-            if (iter->second->expired()) {
-                iter = m_map.erase(iter);
-            } else {
-                ++iter;
+        // Check if cleaning is required
+        if (now <= m_next_clean) {return;}
+
+        // Clean expired m_map entries
+        {
+            std::lock_guard<std::mutex> guard(m_map_mutex);
+            for (auto iter = m_map.begin(); iter != m_map.end(); ) {
+                if (iter->second->expired()) {
+                    iter = m_map.erase(iter);
+                } else {
+                    ++iter;
+                }
             }
         }
         Reconfig();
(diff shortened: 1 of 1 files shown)


amadio:
Looks good. I'm just going on vacations, but I'm tagging this for the next release. Thanks!

bbockelm:
@jthiltges - there's a small (preexisting) race condition here -- `m_next_clean` should be an `std::atomic<uint64_t>`.  Can you fix the declaration and the usage so relaxed reads are used?  XRootD even has a handy `RAtomic` template type that is useful for this particular case that we try to use to make the intent clear.

(if memory serves, this fix is effectively a no-op on x86-64 but potentially meaningful on other platforms and at least technically needed)

amadio:
> m_next_clean should be an `std::atomic<uint64_t>`

Are you sure? There is a mutex being held while accessing it. Maybe you mean to replace the mutex with atomics, and use the mutex only for the reconfig part? In any case, I will wait for this discussion to be resolved before merging this.

bbockelm:
> Are you sure? There is a mutex being held while accessing it.

Only on the second access.  Simplified, with the PR the code is:

```
        if (now <= m_next_clean) {return;}
        std::lock_guard<std::mutex> guard(m_mutex);
        if (now <= m_next_clean) {return;}
        m_next_clean = monotonic_time() + m_expiry_secs;
```

That's undefined behavior technically but fine pragmatically.  By doing a relaxed atomic, you won't change the generated assembler for current compilers but would tidy up the undefined behavior.

Now that you mentioned replacing the mutex, that would be pretty straightforward here but would be 100% fine deferring it to separate work.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -1342,8 +1343,13 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
 
     void Check(uint64_t now)
     {
+        // Check if cleaning is required and bail out if unneeded
         if (now <= m_next_clean) {return;}
+
+        // Cleaning may be needed. Get the mutex, recheck, and reset the timer
         std::lock_guard<std::mutex> guard(m_mutex);
```
abh3: So, here is the problem. If statement 1347 is correct, it means that if the config file has not expired, then it's OK to use the one we have. Line 1360 apparently tries to do two things: 1) allow only one thread to set the time and 2) only one thread to process the config file. However, it doesn't stop thread pile-up, it just reduces it. This is because many threads may get the mutex before the time is actually reset. Each will return once they see that a previous thread reset the time but that will all be serialized and each will wait for the config file to be re-read. Now, getting back to statement 1347. If that statement is correct, then there is no reason to hold on to the mutex while the config file is re-read. Statement 1347 explicitly says it's OK to use the current config file even when the config file is being re-read. Hence, no reason to hold on to the mutex while the file is re-read. Releasing it after setting the time would prevent thread pile-up and only one thread would re-read the config file (assuming re-reading doesn't take longer than the expiration time). This kind of stuff is done in other parts of the code but it relies on creating a shadow config file upon re-reading and replacing it whole-scale in an atomic operation. Alternatively, if that is not possible; threads using the config file get a read lock. The thread altering the config get a write lock. That, of course, still may create a pile-up but does maximize the number of readers,

jthiltges:
Thank you for the thoughtful feedback Brian and Andy. I reworked the PR into what is hopefully a more satisfying state.

The main change is adding a `m_check_mutex` for the Check() function, and immediately returning if it's already held. Then the mutex protecting `m_map` (renamed `m_map_mutex` to better indicate the role) is no longer covering two purposes.

I do like the shadow config with atomic replacement idea, but that will be a bigger change.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -1342,14 +1342,22 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
...
-                ++iter;
+        // Clean expired m_map entries
+        {
+            std::lock_guard<std::mutex> guard(m_map_mutex);
+            for (auto iter = m_map.begin(); iter != m_map.end(); ) {
+                if (iter->second->expired()) {
+                    iter = m_map.erase(iter);
+                } else {
+                    ++iter;
+                }
```
amadio, line 1381: How about using [`std::remove_if`](https://en.cppreference.com/w/cpp/algorithm/remove.html) here (or [`std::erase_if`](https://en.cppreference.com/w/cpp/container/map/erase_if.html)), since we're touching this part of the code in any case?

PR #2678, opened 2025-12-16
[XrdSciTokens] Store m_valid_issuers_array in a shared_ptr to eliminate lock around scitoken_deserialize

jthiltges:
To validate a token, scitokens-cpp downloads current keys from the issuer. If the issuer keys are not currently cached, the xrootd client's request is delayed until the issuer responds or the key retrieval times out. A read lock (`m_config_lock`) is held during this time.
https://github.com/xrootd/xrootd/blob/e78497192f5a91069770a9d4b594cdcea78cf536/src/XrdSciTokens/XrdSciTokensAccess.cc#L793-L795

If a periodic config refresh happens (running every 60 seconds), `Reconfig()` waits for a write lock (also on `m_config_lock`) while also holding `m_mutex` protecting access rules.
https://github.com/xrootd/xrootd/blob/e78497192f5a91069770a9d4b594cdcea78cf536/src/XrdSciTokens/XrdSciTokensAccess.cc#L1318

Other scitokens validations during this time are stalled, waiting for `m_mutex`.
https://github.com/xrootd/xrootd/blob/e78497192f5a91069770a9d4b594cdcea78cf536/src/XrdSciTokens/XrdSciTokensAccess.cc#L525

This change moves `m_valid_issuers_array` to a shared_ptr, allowing the lock around `scitoken_deserialize` to be removed, hopefully reducing contention if there are problems contacting an issuer.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -689,9 +689,8 @@ public:
         SciToken scitoken;
         char *err_msg;
         if (!strncmp(token, "Bearer%20", 9)) token += 9;
-        pthread_rwlock_rdlock(&m_config_lock);
-        auto retval = scitoken_deserialize(token, &scitoken, &m_valid_issuers_array[0], &err_msg);
-        pthread_rwlock_unlock(&m_config_lock);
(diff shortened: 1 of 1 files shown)


jthiltges:
Thanks for catching that, Brian. I was looking for something cleaner than holding a lock to copy `m_valid_issuers_array`, but maybe that's the best path. I'll give that a try unless you recommend another direction.

PR #2680, opened 2025-12-18
[XrdVoms] provide compatibility for timespec on macOS/NetBSD

veprbl:
`st_ctim` is not a thing on some systems.

### DIFF
--- a/src/XrdVoms/XrdVomsMapfile.cc
+++ b/src/XrdVoms/XrdVomsMapfile.cc
@@ -42,6 +42,10 @@
 #include <fcntl.h>
 #include <poll.h>
 
+#if defined(__APPLE__) || defined(__NetBSD__)
+#define st_ctim st_ctimespec
+#endif
+
 bool XrdVomsMapfile::tried_configure = false;
 std::unique_ptr<XrdVomsMapfile> XrdVomsMapfile::mapper;
 


src/XrdVoms/XrdVomsMapfile.cc
```diff
@@ -42,6 +42,10 @@
 #include <fcntl.h>
 #include <poll.h>
 
+#if defined(__APPLE__) || defined(__NetBSD__)
+#define st_ctim st_ctimespec
```
amadio, line 46: Shouldn't this be `st_ctime`? Accorting to the [`stat(2)`](https://man.netbsd.org/stat.2) man page for NetBSD, `st_ctimespec` is specific for NetBSD itself, so this ifdef is not needed for macOS, since there `st_ctime` *is* defined. It might be easier to configure with `-D_NETBSD_SOURCE` for you if that's not already defined, which will define `st_ctime` as an alias to `st_ctimespec`.

amadio (changes requested):
It seems to me that this change is not necessary, if we follow what's in the man pages for stat on macOS and NetBSD. Could you please provide more information on what problem you hit that motivated this change?

src/XrdVoms/XrdVomsMapfile.cc
amadio: Nevermind, I see that this is intended for the `struct timespec st_ctim` field inside `struct stat`, which is only used in this file.

PR #2684, opened 2026-01-10
[XrdClCurl] Improve compatibility with non-XRootD WebDAV

bbockelm:
When testing XrdClCurl against `golang.org/x/net/webdav`, I found some modest incompatibilities worth fixing:

- WebDAV may include a trailing slash for subdirectories which broke our filename parsing.
- The size attribute is not mandatory for a collection; in this case, just set it as zero.
- The other WebDAV server defaults to `Depth: 1` if `Depth` is not specified.  This didn't break our parsing but caused excessively large responses to `stat` (as `stat` became a directory listing even though the client ignored the directory listing response).
- The client did not understand redirects to the existing server (that is, a redirect to `/foo` broke; the client didn't understand that it should implicitly do `https://example.com/foo` in such a case).


amadio:
Rebased on current `master`, to get a fix for a compiler warning introduced recently.

PR #2685, opened 2026-01-12
Remove Davix-based `XrdClHttp` and rename `XrdClCurl` to `XrdClHttp`

amadio:
This makes the new `XrdClCurl` plugin completely replace `XrdClHttp` based on Davix, as mentioned by @abh3 in [GitHub Discussions](https://github.com/xrootd/xrootd/discussions/2668#discussioncomment-15226740).

@abh3 and @bbockelm Please let me know what you think, if both of you are in agreement, I can merge this in for XRootD 6.0.


wyang007:
We are currently using XrdClHtttp in production. While I think XrdClCurl is great, do we have to do this renaming? Will this break existing setup ?

wyang007:
My current use is with xrootd 5.8.x. Does this new package available for xrootd 5.x so that I can try ?

wyang007:
@amadio : I installed xrootd 5.9.1 from EPEL (for el9) and replaced libXrdClHttp-5.so by the one I downloaded. Is this the right way to test? Now I got these in xrootd.log

------ xrootd s3@xrootd-s3gw.aglt2.org:1094 initialization completed.
260114 15:34:03 1302156 XrootdBridge: 71f98131.1:40@fts323.usatlas.bnl.gov login as 71f98131.0
260114 15:34:03 1302133 XrootdBridge: 71f98131.2:37@fts317.usatlas.bnl.gov login as 71f98131.0
260114 15:34:03 1302156 ofs_stat: 71f98131.1:40@fts323.usatlas.bnl.gov Unable to locate /atlaslocalgroupdisk/rucio/user/dntounis/dd/8f/user.dntounis.48076827._000129.CxAOD.root; function not implemented
260114 15:34:03 1302133 ofs_stat: 71f98131.2:37@fts317.usatlas.bnl.gov Unable to locate /atlaslocalgroupdisk/rucio/user/dntounis/3b/ae/user.dntounis.48076818._000010.CxAOD.root; function not implemented
260114 15:34:03 1302132 XrootdBridge: 71f98131.3:30@fts316.usatlas.bnl.gov login as 71f98131.0
260114 15:34:03 1302132 ofs_stat: 71f98131.3:30@fts316.usatlas.bnl.gov Unable to locate /atlaslocalgroupdisk/rucio/user/dntounis/4d/0a/user.dntounis.48076824._000541.CxAOD.root; invalid request code

Is there a document on how to use the new libXrdClHttp.so ? At least I notice that the old one take env var. AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY, etc. but I don't know how provide these keys to the new library.

abh3:
According to curl documentation you need to set envars, as follows:

export AWS_ACCESS_KEY_ID="<your-access-key-id>"
export AWS_SECRET_ACCESS_KEY="<your-secret-access-key>"
export AWS_DEFAULT_REGION="<your-aws-region>" # Optional, but good practice

Typically, you would set these before invoking the plug-in. This can be done in the xrootd config file. For instance:

setenv AWS_ACCESS_KEY_ID < _path_to_file_containing_key_id_
setenv  AWS_SECRET_ACCESS_KEY < _path_to_file_containing_key_

I would have expected hat was what you were doing in the first place, but maybe not.

wyang007:
I have those env var. set in the xrootd config file

amadio:
Maybe you need to try libXrdClS3.so plugin instead? Rather than use just the plugin, I'd install all the RPMs just in case, then use the new plugins based on each protocol.

amadio:
@bbockelm If you could provide advice on migration from XRootD 5 to 6 for the Davix based to curl based plugins in the documentation (i.e. at least in README.md files), it would be nice. If I understand correctly, Pelican was using the current plugin for a while before writing the current plugins based on curl.

wyang007:
@amadio I am not sure what is the different between libXrdClS3.so and libXrdClHttp.so in the xrootd-client-libs-5.9.1.post164+git9542ae13e-1.el9.x86_64.rpm

Are you saying to use  libXrdClS3.so for http and https in a client.plugins.d config file ?

wyang007:
Just set "pss.origin s3://..." in xrootd config ?

abh3:
Almost. You also need to specify that is the plugin to load for "s3" in the XrdCl configuration.:
url = s3://
lib = libXrdClS3.so
enable = true

The config files for the new ClHttp (a.k.a. XrdClCurl) are located in xrootd/src/XrdClS3/config and xrootd/src/XrdClCurl/configs, respectively.

abh3 (commented):
One things that needs to happen is that config options and envars need to be backward compatible. So, if the old XrdClHttp allows x to be specified then XrdClCurl needs to be able to handle that. This would provide backward compatibility. As far as I can tell, that is true but should be verified.

wyang007:
This is from error from the log:

...
Copr.  2019, Stanford University, Pss Version v5.9.1-164-g9542ae13e
=====> all.export /xrootd
=====> all.export /
=====> all.export /atlaslocalgroupdisk
=====> pss.hostarena /atlas
=====> pss.cache pagesize 1m minpages 4096 preread 16 logstats
Config warning: origin port not specified, using port 1094 as default for s3://
=====> pss.origin s3://141.211.169.5
------ Proxy storage system initialization completed.

...

------ xrootd s3@xrootd-s3gw.aglt2.org:1094 initialization completed.
260114 20:51:00 1471257 XrootdBridge: 71f98131.1:34@fts311.usatlas.bnl.gov login as 71f98131.0
260114 20:51:00 1471257 ofs_stat: 71f98131.1:34@fts311.usatlas.bnl.gov Unable to locate /atlaslocalgroupdisk/rucio/user/dntounis/54/40/user.dntounis.48076603._000345.CxAOD.root; operation not supported
Additional context: [ERROR] Operation not supported

wyang007:
This was it.

# cat /etc/xrootd/client.plugins.d/xrdcl-s3-plugin.conf 
url = s3://
lib = libXrdClS3.so
enable = true

I have since put the system back to use the EPEL 5.9.1 rpms and Davix based xddcl-http - it is a production system.

wyang007:
If I am to give this a try, I would not want to use 5.9.1. I think we should try 6. @amadio is there a v6 RPM build for EL9 that we can try ?

amadio:
@wyang007 The build is now running, the results will be available [here](https://github.com/amadio/xrootd/actions/runs/21059608714). Please let me know if you have any problems trying out these RPMs.

PR #2690, opened 2026-01-13
[XrdHttp] Fix a null byte buffer overflow when parsing %-encoded strings

daw1012345:
The unquote function adds 3 to the index without checking whether the new index goes OOB.


daw1012345:
I am slightly unsure what the desired behavior is here. I currently just bail out early, ignoring the invalid %-encoding, but one could argue that the function should instead return an error and halt further processing.

PR #2691, opened 2026-01-13
[XrdHttp] Fix buffer overflow issues when parsing HTTP request lines

daw1012345:
Buffgetline returns the total number of bytes read, while the string stops reading on null bytes. Pass the length of the string to subroutines instead of the return value of Buffgetline.


abh3 (approved):
Hard to believe the previous version ever worked given that rc would rarely be what you wanted it to be.

PR #2697, opened 2026-01-23
[XrdPss] Implement preread for the PSS layer

bbockelm:
This provides a simple implementation of preread functionality for the proxy layer.  After considering several approaches to avoid allocating a buffer to read into, I concluded there's too much external code that may suffer if I simply pass a null buffer.

So, this preread will pull data into memory and then immediately discard it - probably the best option there is without adding new APIs to XrdCl.

### DIFF
--- a/src/XrdPss/XrdPss.cc
+++ b/src/XrdPss/XrdPss.cc
@@ -1201,9 +1201,46 @@ ssize_t XrdPssFile::pgWrite(void     *buffer,
 
 ssize_t XrdPssFile::Read(off_t offset, size_t blen)
 {
+     ssize_t retval;
+
      if (fd < 0) return (ssize_t)-XRDOSS_E8004;
 
-     return 0;  // We haven't implemented this yet!
+     // Implementation notes:
+     // - This implementation requires a buffer to read into even though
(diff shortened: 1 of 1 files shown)


src/XrdPss/XrdPss.cc
```diff
@@ -1201,9 +1201,46 @@ ssize_t XrdPssFile::pgWrite(void     *buffer,
...
 {
+     ssize_t retval;
+
      if (fd < 0) return (ssize_t)-XRDOSS_E8004;
 
-     return 0;  // We haven't implemented this yet!
+     // Implementation notes:
+     // - This implementation requires a buffer to read into even though
+     //   the data is not used.  I considered just providing a `nullptr` for
+     //   a buffer but the XrdCl implementation requires a buffer to read into
```
abh3, line 1211: We are actually implementing an XrdCl API that does not require a buffer to read into. It's not clear why this was not implemented way back when as the protocol supports pre reads without returning any data. Will that change things here?

abh3 (commented):
A heads up that R6 changes a couple of things that will impact this pull request.

amadio:
Closing, as this should no longer be necessary after #2727 and #2743.

PR #2699, opened 2026-01-28
[XrdSciTokens] set `XDG_CACHE_HOME` if not set

### LINKED ISSUE #2667: XrdSciTokens issuer rate limiting produces confusing HTTP status code
### Feature Description

A Pelican user opened this issue against our repo, but I suspect this lives at the interface of XRootD and SciTokens:

> When cilogon.org is used an issuer, it can sometimes respond with a 503 to requests as part of its rate limiting strategy. In this case, currently, the Pelican cache server simply responds to requests with a 4xx code indicating permission denied. While it's true that the user is not authenticated and the owners of the data do not want to give access to the data to unauthenticated users, it is not necessarily the case that the user has provided an invalid token. The user in this case is lead down a path where they try to figure out how they failed to provide a token that would give them access to the data that they want.
>
> I think it would be better, in this case, to respond with something like "5xx could not get metadata for issuer cilogon.org". In this case it would be obvious to everyone involved that the problem was not with the SciToken itself, but with the communication between the cache server and cilogon.org.

I started crawling a bit of the code to figure out what it might take to do this, and I pieced this together:
- The upstream SciTokens library would need a change to be more explicit about what causes a metadata download failure, probably by adding a new exception next to the `CurlException` that lives [here](https://github.com/scitokens/scitokens-cpp/blob/90cabef402b77196a2a4402bdf35fddd5da95b1d/src/scitokens_internal.cpp#L679-L693) or [here](https://github.com/scitokens/scitokens-cpp/blob/90cabef402b77196a2a4402bdf35fddd5da95b1d/src/scitokens_internal.cpp#L729-L731) or by extending the `CurlException`.
- That would need to propagate upwards through `scitoken_deserialize`, which is [the call XrdSciTokens makes](https://github.com/xrootd/xrootd/blob/eff589a46c39fe863256d388bebd92791d1185ae/src/XrdSciTokens/XrdSciTokensAccess.cc#L794) to fetch/check a token's issuer key (not sure if this is the right spot -- that function is also called in [`Validate`](https://github.com/xrootd/xrootd/blob/eff589a46c39fe863256d388bebd92791d1185ae/src/XrdSciTokens/XrdSciTokensAccess.cc#L693).
- (this is where my understanding starts to get pretty fuzzy tracing through the code) The [`XrdAccPrivs Access`](https://github.com/xrootd/xrootd/blob/eff589a46c39fe863256d388bebd92791d1185ae/src/XrdAcc/XrdAccAuthorize.hh#L86-L87) method would need to be extended to return more than just "allow/deny", or to punch the ability to pass a "deny reason" through the call stack.
- The `XrdOfs AUTHORIZE` macro would need to be extended to return values other than `EACCES`, maybe `EBUSY` or `EAGAIN`?

This is almost definitely too many layers to punch through, which leads me to the question -- if I want to make some progress here, how should I be thinking about the problem? 

Since Pelican uses XRootD for HTTP protocol support, what in general can be done to make it easier for plugins to pass information back through all the layers so we can provide more meaningful error messages to clients?

dynamic-entropy:
Fixes #2667 

Set cache file location configuration for Scitokens. 
Use XDG_CACHE_HOME or derive from XRDADMINPATH if not set.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -18,8 +18,7 @@
 #include <fstream>
 #include <unordered_map>
 #include <tuple>
-
-#include "fcntl.h"
+#include <cstdlib>
 
 #include "INIReader.h"
 #include "picojson.h"
@@ -1054,6 +1053,28 @@ private:
                 scitoken_config_set_str("tls.ca_file", params->cafile.c_str(), nullptr);
 #else
                 m_log.Log(LogMask::Warning, "Config", "tls.ca_file is set but the platform's libscitokens.so does not support setting config parameters");
+#endif
+            }
+        }
+
+        // set cache file location
+        if (const char* xdg_cache_home = getenv("XDG_CACHE_HOME")) {
+            m_log.Log(LogMask::Info, "Config", "Scitokens cache file location env var is set to : ", xdg_cache_home);
+        } else {
+            // construct xdg_cache_home to be <adminpath>/.cache
+            const char* adminpath_env = getenv("XRDADMINPATH");
+            if (!adminpath_env || !*adminpath_env) {
+                m_log.Log(LogMask::Warning, "Config", "XRDADMINPATH is not defined; leaving cache location unset");
+            } else {
+                std::string adminpath = adminpath_env;
+                while (adminpath.size() > 1 && adminpath.back() == '/') adminpath.pop_back();
+                std::string xdg_cache_home_str = adminpath + "/.cache";
+                m_log.Log(LogMask::Info, "Config", "Scitokens cache file location env var is not set; using : ", xdg_cache_home_str.c_str());
+
+#ifdef HAVE_SCITOKEN_CONFIG_SET_STR
+                scitoken_config_set_str("keycache.cache_home", xdg_cache_home_str.c_str(), nullptr);
+#else
+                setenv("XDG_CACHE_HOME", xdg_cache_home_str.c_str(), 1);
 #endif
             }
         }


src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -1058,6 +1057,28 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
             }
         }
 
+        // set cache file location
+        if (const char* xdg_cache_home = getenv("XDG_CACHE_HOME")) {
```
amadio, line 1061: There's a typo in the commit message, you call the variable `XRD_CACHE_HOME` instead of `XDG_CACHE_HOME`.

dynamic-entropy:
> * What happens if `adminpath + "/.xdgcache"` doesn't exist? I don't see you creating the directory here, and maybe neither will the scitokens library.

The scitoken library creates it. The sqlite file is then stored in `/.xdgcache/scitoken/sqlite.db`

> I would suggest to use ADMINPATH/.cache as default.

I can do that, just wanted to take a precaution incase there is some other library writing at the same location

> Please note that with systemd, XDG_CACHE_HOME will by default point to /var/cache/xrootd with XRootD 6.0, we may need to change that to /var/cache/xrootd/<instance> if this must be per-instance.

This PR only concerns when `XDG_CACHE_HOME` is not set. If it is already set, it won't modify it.
But as I commented on the issue, I think we do not really need separate cache directories and DBs for each individual instance.

src/XrdSciTokens/XrdSciTokensAccess.cc
dynamic-entropy: Fixed the typo in commit message. Thanks!

PR #2701, opened 2026-01-28
[ClHttp] During shutdown join threads rather than using cv

smithdh:
Associated issue: #2702

### DIFF
--- a/src/XrdClHttp/XrdClHttpOpListdir.cc
+++ b/src/XrdClHttp/XrdClHttpOpListdir.cc
@@ -99,10 +99,10 @@ bool CurlListdirOp::ParseProp(DavEntry &entry, TiXmlElement *prop)
                 return false;
             }
             struct tm tm;
-            if (strptime(lastmod, "%a, %d %b %Y %H:%M:%S %Z", &tm) == nullptr) {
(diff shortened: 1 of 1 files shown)


amadio:
@bbockelm @dynamic-entropy Looks like the thread pool issue was not the only problem we had. Now Alpine is failing due to a call to a pure virtual method. Could you please take a look at it? When I renamed XrdClCurl to XrdClHttp and removed Davix, I also replaced `davix-*` commands with `xrdcp`/`xrdfs`. These are the things which show failures now. Due to a mistake on my part they were disabled before (`XRootD::http` test was disabled). If one of you have time on Monday, this needs to be fixed before I can release. Have a nice weekend!

amadio:
Thank you so much, @smithdh! I'm merging this right away to get the CI back to green. Well, mostly, since we still need a solution for the scitoken failure in the first run. The token used for the test sometimes expires before the test runs, but that can be done separately.

dynamic-entropy:
@bbockelm , yes, this is fixed here, https://github.com/xrootd/xrootd/commit/02bc6ae7715ea7dd7805e27f9420284a6a82e65d 

The issue was that the [strptime implementation in musl](https://git.musl-libc.org/cgit/musl/tree/src/time/strptime.c) does not support the %z format specifier.

PR #2703, opened 2026-01-29
Add statx support

ccaffy:
Mainly ABI-breaking changes

### DIFF
--- a/src/XrdBwm/XrdBwm.cc
+++ b/src/XrdBwm/XrdBwm.cc
@@ -929,6 +929,37 @@ int XrdBwm::stat(const char             *path,        // In
 }
 
 /******************************************************************************/
+/*                                  s t a t x                                 */
+/******************************************************************************/
+#if defined(__linux__)
+int XrdBwm::statx(const char             *path,        // In
+                       struct statx      *buf,         // Out
+                       XrdOucErrInfo    &einfo,       // Out
+                 const XrdSecEntity     *client,      // In
+                 const char             *info)        // In
+
+{
+// Return an error
+//
+   return XrdBwmFS.Emsg("stat", einfo, ENOTSUP, "locate", path);
+}
+
+/******************************************************************************/
+
+int XrdBwm::statx(const char             *path,        // In
+                       mode_t           &mode,        // Out
+                       XrdOucErrInfo    &einfo,       // Out
+                 const XrdSecEntity     *client,      // In
+                 const char             *info)        // In
+
+{
+// Return an error
+//
+   return XrdBwmFS.Emsg("stat", einfo, ENOTSUP, "locate", path);
+}
+#endif
+
+/******************************************************************************/
 /*                              t r u n c a t e                               */
 /******************************************************************************/
 
--- a/src/XrdBwm/XrdBwm.hh
+++ b/src/XrdBwm/XrdBwm.hh
@@ -121,6 +121,9 @@ public:
         int            sync(XrdSfsAio *aiop);
 
         int            stat(struct stat *buf);
+#if defined(__linux__)
+        int            statx(struct statx *buf) { return SFS_OK; }
+#endif
 
         int            truncate(XrdSfsFileOffset   fileOffset);
 
@@ -224,6 +227,19 @@ const   char          *getVersion();
                                   XrdOucErrInfo    &out_error,
                             const XrdSecEntity     *client,
                             const char             *opaque = 0);
+#if defined(__linux__)
+        int            statx(const char             *Name,
+                             struct statx      *buf,
+                             XrdOucErrInfo    &out_error,
+                             const XrdSecEntity     *client,
+                             const char             *opaque = 0);
+
+        int            statx(const char             *Name,
+                             mode_t           &mode,
+                             XrdOucErrInfo    &out_error,
+                             const XrdSecEntity     *client,
+                             const char             *opaque = 0);
+#endif
 
         int            truncate(const char             *Name,
                                       XrdSfsFileOffset fileOffset,
--- a/src/XrdCeph/XrdCephOss.cc
+++ b/src/XrdCeph/XrdCephOss.cc
@@ -519,7 +519,12 @@ int XrdCephOss::Stat(const char* path,
 
 }
 
-
+#if defined(__linux__)
+int XrdCephOss::Statx(const char *path, struct statx *buff, int opts, XrdOucEnv *envP) {
+  XrdCephEroute.Say(__FUNCTION__, " statx not supported, path = ", path);
+  return -ENOTSUP;
+}
+#endif
 
 int XrdCephOss::StatFS(const char *path, char *buff, int &blen, XrdOucEnv *eP) {
 
--- a/src/XrdCeph/XrdCephOss.hh
+++ b/src/XrdCeph/XrdCephOss.hh
@@ -66,6 +66,9 @@ public:
   virtual int     Remdir(const char *, int Opts=0, XrdOucEnv *eP=0);
   virtual int     Rename(const char *, const char *, XrdOucEnv *eP1=0, XrdOucEnv *eP2=0);
   virtual int     Stat(const char *, struct stat *, int opts=0, XrdOucEnv *eP=0);
+#if defined(__linux__)
+  virtual int     Statx(const char *path, struct statx *buff, int opts=0, XrdOucEnv *envP=0);
+#endif
   virtual int     StatFS(const char *path, char *buff, int &blen, XrdOucEnv *eP=0);
   virtual int     StatLS(XrdOucEnv &env, const char *path, char *buff, int &blen);  
   virtual int     StatVS(XrdOssVSInfo *sP, const char *sname=0, int updt=0);
--- a/src/XrdCl/XrdClFile.cc
+++ b/src/XrdCl/XrdClFile.cc
@@ -258,6 +258,33 @@ namespace XrdCl
     return MessageUtils::WaitForResponse( &handler, response );
   }
 
+  //----------------------------------------------------------------------------
+  // Obtain status information for this file - async
+  //----------------------------------------------------------------------------
+  XRootDStatus File::Statx( bool             force,
+                           ResponseHandler *handler,
+                           time_t           timeout )
+  {
+    if( pPlugIn )
+      return pPlugIn->Statx( force, handler, timeout );
+
+    return FileStateHandler::Statx( pImpl->pStateHandler, force, handler, timeout );
+  }
+
+  //----------------------------------------------------------------------------
+  // Obtain status information for this file - sync
+  //----------------------------------------------------------------------------
+  XRootDStatus File::Statx( bool       force,
+                           StatxInfo *&response,
+                           time_t     timeout )
+  {
+    SyncResponseHandler handler;
+    XRootDStatus st = Statx( force, &handler, timeout );
+    if( !st.IsOK() )
+      return st;
+
+    return MessageUtils::WaitForResponse( &handler, response );
+  }
 
   //----------------------------------------------------------------------------
   // Read a data chunk at a given offset - sync
--- a/src/XrdCl/XrdClFile.hh
+++ b/src/XrdCl/XrdClFile.hh
@@ -238,6 +238,36 @@ namespace XrdCl
                          time_t     timeout = 0 )
                          XRD_WARN_UNUSED_RESULT;
 
+      //------------------------------------------------------------------------
+      //! Obtain status information for this file - async
+      //!
+      //! @param force   do not use the cached information, force re-stating
+      //! @param handler handler to be notified when the response arrives,
+      //!                the response parameter will hold a StatInfo object
+      //!                if the procedure is successful
+      //! @param timeout timeout value, if 0 the environment default will
+      //!                be used
+      //! @return        status of the operation
+      //------------------------------------------------------------------------
+      XRootDStatus Statx( bool             force,
+                         ResponseHandler *handler,
+                         time_t           timeout = 0 )
+                         XRD_WARN_UNUSED_RESULT;
+
+      //------------------------------------------------------------------------
+      //! Obtain status information for this file - sync
+      //!
+      //! @param force   do not use the cached information, force re-stating
+      //! @param response the response (to be deleted by the user)
+      //! @param timeout  timeout value, if 0 the environment default will
+      //!                 be used
+      //! @return         status of the operation
+      //------------------------------------------------------------------------
+      XRootDStatus Statx( bool       force,
+                         StatxInfo *&response,
+                         time_t     timeout = 0 )
+                         XRD_WARN_UNUSED_RESULT;
+
 
       //------------------------------------------------------------------------
       //! Read a data chunk at a given offset - async
--- a/src/XrdCl/XrdClFileStateHandler.cc
+++ b/src/XrdCl/XrdClFileStateHandler.cc
@@ -649,6 +649,7 @@ namespace XrdCl
   FileStateHandler::FileStateHandler( FilePlugIn *& plugin ):
     pFileState( Closed ),
     pStatInfo( 0 ),
+    pStatxInfo( 0 ),
     pFileUrl( 0 ),
     pDataServer( 0 ),
     pLoadBalancer( 0 ),
@@ -682,6 +683,7 @@ namespace XrdCl
   FileStateHandler::FileStateHandler( bool useVirtRedirector, FilePlugIn *& plugin ):
     pFileState( Closed ),
     pStatInfo( 0 ),
+    pStatxInfo( 0 ),
     pFileUrl( 0 ),
     pDataServer( 0 ),
     pLoadBalancer( 0 ),
@@ -1044,6 +1046,63 @@ namespace XrdCl
     MessageUtils::CreateRequest( msg, req );
 
     req->requestid = kXR_stat;
+    memcpy( req->fhandle, self->pFileHandle, 4 );
+
+    MessageSendParams params;
+    params.timeout         = timeout;
+    params.followRedirects = false;
+    params.stateful        = true;
+    MessageUtils::ProcessSendParams( params );
+
+    XRootDTransport::SetDescription( msg );
+    StatefulHandler *stHandler = new StatefulHandler( self, handler, msg, params );
+
+    return SendOrQueue( self, *self->pDataServer, msg, stHandler, params );
+  }
+
+  //----------------------------------------------------------------------------
+  // Stat the file
+  //----------------------------------------------------------------------------
+  XRootDStatus FileStateHandler::Statx( std::shared_ptr<FileStateHandler> &self,
+                                       bool                               force,
+                                       ResponseHandler                   *handler,
+                                       time_t                             timeout )
+  {
+    XrdSysMutexHelper scopedLock( self->pMutex );
+
+    if( self->pFileState == Error ) return self->pStatus;
+
+    if( self->pFileState != Opened && self->pFileState != Recovering )
+      return XRootDStatus( stError, errInvalidOp );
+
+    //--------------------------------------------------------------------------
+    // Return the cached info
+    //--------------------------------------------------------------------------
+    if( !force && self->pStatxInfo )
+    {
+      AnyObject *obj = new AnyObject();
+      obj->Set( new StatxInfo( *self->pStatxInfo ) );
+      if (handler)
+        handler->HandleResponseWithHosts( new XRootDStatus(), obj, new HostList() );
+      return XRootDStatus();
+    }
+
+    Log *log = DefaultEnv::GetLog();
+    log->Debug( FileMsg, "[%p@%s] Sending a statx command for handle %#x to %s",
+                (void*)self.get(), self->pFileUrl->GetObfuscatedURL().c_str(),
+                *((uint32_t*)self->pFileHandle), self->pDataServer->GetHostId().c_str() );
+
+    //--------------------------------------------------------------------------
+    // Issue a new stat request
+    // stating a file handle doesn't work (fixed in 3.2.0) so we need to
+    // stat the pat
+    //--------------------------------------------------------------------------
+    Message           *msg;
+    ClientStatRequest *req;
+    std::string        path = self->pFileUrl->GetPath();
+    MessageUtils::CreateRequest( msg, req );
+
+    req->requestid = kXR_statx;
     memcpy( req->fhandle, self->pFileHandle, 4 );
 
     MessageSendParams params;
@@ -2689,6 +2748,18 @@ namespace XrdCl
         response->Get( info );
         delete self->pStatInfo;
         self->pStatInfo = new StatInfo( *info );
+        break;
+      }
+
+      //------------------------------------------------------------------------
+      // Cache the statx response
+      //------------------------------------------------------------------------
+      case kXR_statx:
+      {
+        StatxInfo *info = 0;
+        response->Get( info );
+        delete self->pStatxInfo;
+        self->pStatxInfo = new StatxInfo( *info );
         break;
       }
 
--- a/src/XrdCl/XrdClFileStateHandler.hh
+++ b/src/XrdCl/XrdClFileStateHandler.hh
@@ -201,6 +201,22 @@ namespace XrdCl
                                 time_t                             timeout = 0 );
 
 
+    //------------------------------------------------------------------------
+    //! Obtain status information for this file - async
+    //!
+    //! @param force   do not use the cached information, force re-stating
+    //! @param handler handler to be notified when the response arrives,
+    //!                the response parameter will hold a StatInfo object
+    //!                if the procedure is successful
+    //! @param timeout timeout value, if 0 the environment default will
+    //!                be used
+    //! @return        status of the operation
+    //------------------------------------------------------------------------
+    static XRootDStatus Statx( std::shared_ptr<FileStateHandler> &self,
+                              bool                               force,
+                              ResponseHandler                   *handler,
+                              time_t                             timeout = 0 );
+
       //------------------------------------------------------------------------
       //! Read a data chunk at a given offset - sync
       //!
@@ -951,6 +967,7 @@ namespace XrdCl
       FileStatus              pFileState;
       XRootDStatus            pStatus;
       StatInfo               *pStatInfo;
+      StatxInfo              *pStatxInfo;
       URL                    *pFileUrl;
       URL                    *pDataServer;
       URL                    *pLoadBalancer;
--- a/src/XrdCl/XrdClFileSystem.cc
+++ b/src/XrdCl/XrdClFileSystem.cc
@@ -1585,6 +1585,44 @@ namespace XrdCl
     return MessageUtils::WaitForResponse( &handler, response );
   }
 
+  XRootDStatus FileSystem::Statx( const std::string &path,
+                       ResponseHandler   *handler,
+                       time_t             timeout) {
+
+    if( pPlugIn )
+      return pPlugIn->Statx( path, handler, timeout );
+
+    if( pImpl->fsdata->pUrl->IsLocalFile() )
+      return LocalFS::Instance().Stat( path, handler, timeout );
+
+    std::string fPath = FilterXrdClCgi( path );
+
+    Message           *msg;
+    ClientStatRequest *req;
+    MessageUtils::CreateRequest( msg, req, fPath.length() );
+
+    req->requestid  = kXR_statx;
+    req->options    = 0;
+    req->dlen       = fPath.length();
+    msg->Append( fPath.c_str(), fPath.length(), 24 );
+    MessageSendParams params; params.timeout = timeout;
+    MessageUtils::ProcessSendParams( params );
+    XRootDTransport::SetDescription( msg );
+
+    return FileSystemData::Send( pImpl->fsdata, msg, handler, params );
+  }
+
+  XRootDStatus FileSystem::Statx( const std::string  &path,
+                         StatxInfo          *&response,
+                         time_t              timeout) {
+    SyncResponseHandler handler;
+    Status st = Statx( path, &handler, timeout );
+    if( !st.IsOK() )
+      return st;
+
+    return MessageUtils::WaitForResponse( &handler, response );
+  }
+
   //----------------------------------------------------------------------------
   // Obtain status information for a path - async
   //----------------------------------------------------------------------------
--- a/src/XrdDig/XrdDigFS.cc
+++ b/src/XrdDig/XrdDigFS.cc
@@ -806,6 +806,33 @@ int XrdDigFS::stat(const char              *path,        // In
 }
 
 /******************************************************************************/
+/*                                  s t a t                                   */
+/******************************************************************************/
+#if defined(__linux__)
+int XrdDigFS::statx(const char              *path,        // In
+                         struct statx       *buf,         // Out
+                         XrdOucErrInfo     &error,       // Out
+                   const XrdSecClientName  *client,      // In
+                   const char              *info)        // In
+/*
+  Function: Get info on 'path'.
+
+  Input:    path        - Is the fully qualified name of the file to be tested.
+            buf         - The statx structiure to hold the results
+            error       - Error information object holding the details.
+            client      - Authentication credentials, if any.
+            info        - Opaque information, if any.
+
+  Output:   Returns SFS_OK upon success and SFS_ERROR upon failure.
+*/
+{
+   // We don't support this
+   //
+   error.setErrInfo(ENOTSUP, "statx operation not supported");
+   return SFS_ERROR;
+}
+#endif
+/******************************************************************************/
 /*                              V a l i d a t e                               */
 /******************************************************************************/
   
--- a/src/XrdDig/XrdDigFS.hh
+++ b/src/XrdDig/XrdDigFS.hh
@@ -140,6 +140,9 @@ public:
         int            sync(XrdSfsAio *aiop) {return SFS_OK;}
 
         int            stat(struct stat *buf);
+#if defined(__linux__)
+        int            statx(struct statx *buf) {return SFS_OK;}
+#endif
 
         int            truncate(XrdSfsFileOffset   fileOffset) {return SFS_OK;}
 
@@ -243,7 +246,24 @@ const   char          *getVersion();
                         if (!rc) mode = bfr.st_mode;
                         return rc;
                        }
+#if defined(__linux__)
+      int            statx(const char             *Name,
+                                 struct statx      *buf,
+                                 XrdOucErrInfo    &out_error,
+                           const XrdSecClientName *client = 0,
+                           const char             *opaque = 0);
 
+      int            statx(const char             *Name,
+                                mode_t           &mode,
+                                XrdOucErrInfo    &out_error,
+                          const XrdSecClientName *client = 0,
+                          const char             *opaque = 0)
+                      {struct statx bfr;
+                       int rc = statx(Name, &bfr, out_error, client);
+                       if (!rc) mode = bfr.stx_mode;
+                       return rc;
+                      }
+#endif
         int            truncate(const char             *Name,
                                       XrdSfsFileOffset fileOffset,
                                       XrdOucErrInfo    &out_error,
--- a/src/XrdOfs/XrdOfs.cc
+++ b/src/XrdOfs/XrdOfs.cc
@@ -1711,6 +1711,25 @@ int XrdOfsFile::stat(struct stat     *buf)         // Out
    return SFS_OK;
 }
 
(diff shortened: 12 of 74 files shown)


src/XrdOss/XrdOss.hh
```diff
@@ -162,6 +162,16 @@ virtual void    Flush() {}
...
+//-----------------------------------------------------------------------------
+//! Return statx information for this file.
+//!
+//! @param  buf    - Pointer to the structure where info it to be returned.
+//!
+//! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+//-----------------------------------------------------------------------------
+#if HAVE_STATX
+virtual int     Fstatx(struct statx * buf) {return -EISDIR;}
+#endif
```
amadio, line 174: Mmm, this is probably going to cause problems, since this is a public header. I think we need to avoid having any ifdefs on public headers. The only way to have this working well is to have a config header that is generated, then public headers must include it, otherwise `HAVE_STATX` will not be defined and users will get compilation errors when trying to user our headers.

cmake/XRootDSystemCheck.cmake
```diff
@@ -184,4 +184,4 @@ endif()
 
 if (NOT HAVE_CXX_ATOMICS_WITHOUT_LIB AND NOT HAVE_CXX_ATOMICS_WITH_LIB)
   message(FATAL_ERROR "Compiler must support std::atomic!")
-endif()
+endif()
```
amadio: Can you drop this change, please? Thank you!

tests/XRootD/posix.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env bash
 
+HAVE_STATX=$(
+  printf '%s\n' \
+  '#include <sys/stat.h>' \
+  'void *p = (void*) &statx;' \
+  | c++ -std=c++17 -x c++ -c -o /dev/null - 2>/dev/null \
+  && echo 1 || echo 0
+)
```
amadio: You can use `uname -s`, or reuse `cc -dumpmachine | grep -q linux` directly in the if condition.

tests/XRootD/posix.sh
```diff
@@ -13,30 +21,51 @@ function test_posix() {
 	cc -dumpmachine | grep -q musl && return
 
 	# Make sure all symbols are actually resolved correctly
-	MISSING_SYMBOLS=$(env XRDPOSIX_REPORT=1 xrdposix-cat /dev/null 2>&1)
+	MISSING_SYMBOLS=$(
+    env XRDPOSIX_REPORT=1 xrdposix-cat /dev/null 2>&1
+    env XRDPOSIX_REPORT=1 xrdposix-statx /dev/null 2>&1
```
amadio, line 20: Does this really work? I think you need a `;` between commands here.

ccaffy: Yeah it works the same with or without `;`, I'll add one to better separate the two lines ;)

tests/XRootD/posix.sh
ccaffy: Done

cmake/XRootDSystemCheck.cmake
ccaffy: Done

src/XrdBwm/XrdBwm.cc
```diff
@@ -928,6 +928,37 @@ int XrdBwm::stat(const char             *path,        // In
    return XrdBwmFS.Emsg("stat", einfo, ENOTSUP, "locate", path);
 }
 
+/******************************************************************************/
+/*                                  s t a t x                                 */
+/******************************************************************************/
+#if defined(__linux__)
```
abh3, line 934: See point A in the comments.

src/XrdBwm/XrdBwm.cc
```diff
@@ -928,6 +928,37 @@ int XrdBwm::stat(const char             *path,        // In
...
+int XrdBwm::statx(const char             *path,        // In
```
abh3, line 935: See point B in the comments. This modification is not needed.

src/XrdBwm/XrdBwm.hh
```diff
@@ -121,6 +121,9 @@ public:
         int            sync(XrdSfsAio *aiop);
 
         int            stat(struct stat *buf);
+#if defined(__linux__)
```
abh3, line 124: See point A in the comments.. Also, in point B this file does not need to change.

src/XrdCeph/XrdCephOss.cc
```diff
@@ -519,7 +519,12 @@ int XrdCephOss::Stat(const char* path,
 
 }
 
-
+#if defined(__linux__)
```
abh3, line 522: See point A in the comments. In point B this file does not need to change.

src/XrdCeph/XrdCephOss.hh
```diff
@@ -66,6 +66,9 @@ public:
   virtual int     Remdir(const char *, int Opts=0, XrdOucEnv *eP=0);
   virtual int     Rename(const char *, const char *, XrdOucEnv *eP1=0, XrdOucEnv *eP2=0);
   virtual int     Stat(const char *, struct stat *, int opts=0, XrdOucEnv *eP=0);
+#if defined(__linux__)
```
abh3, line 69: See point A in the comments. As per point B, this file does not need to change.

src/XrdCl/XrdClFile.cc
```diff
@@ -258,6 +258,33 @@ namespace XrdCl
     return MessageUtils::WaitForResponse( &handler, response );
   }
 
+  //----------------------------------------------------------------------------
+  // Obtain status information for this file - async
+  //----------------------------------------------------------------------------
+  XRootDStatus File::Statx( bool             force,
```
abh3, line 264: How does this differ from stat()? Do you get all of the possible fields allowed by statx()? That does not correspond to what statx() is trying to do and makes little sense. This really needs to be rethought. At the vary least a mask of wanted fields should be passed. Otherwise, why bother?

src/XrdCl/XrdClFile.cc
```diff
@@ -258,6 +258,33 @@ namespace XrdCl
...
+    if( pPlugIn )
+      return pPlugIn->Statx( force, handler, timeout );
+
+    return FileStateHandler::Statx( pImpl->pStateHandler, force, handler, timeout );
+  }
+
+  //----------------------------------------------------------------------------
+  // Obtain status information for this file - sync
+  //----------------------------------------------------------------------------
+  XRootDStatus File::Statx( bool       force,
```
abh3, line 277: Same issue here.

src/XrdCl/XrdClFile.hh
```diff
@@ -238,6 +238,36 @@ namespace XrdCl
...
+      //!
+      //! @param force   do not use the cached information, force re-stating
+      //! @param handler handler to be notified when the response arrives,
+      //!                the response parameter will hold a StatInfo object
+      //!                if the procedure is successful
+      //! @param timeout timeout value, if 0 the environment default will
+      //!                be used
+      //! @return        status of the operation
+      //------------------------------------------------------------------------
+      XRootDStatus Statx( bool             force,
```
abh3, line 252: Same issue here.

src/XrdCl/XrdClFile.hh
```diff
@@ -238,6 +238,36 @@ namespace XrdCl
...
+      //------------------------------------------------------------------------
+      //! Obtain status information for this file - sync
+      //!
+      //! @param force   do not use the cached information, force re-stating
+      //! @param response the response (to be deleted by the user)
+      //! @param timeout  timeout value, if 0 the environment default will
+      //!                 be used
+      //! @return         status of the operation
+      //------------------------------------------------------------------------
+      XRootDStatus Statx( bool       force,
```
abh3, line 266: Same issue here.

src/XrdCl/XrdClFileStateHandler.cc
```diff
@@ -1058,6 +1060,63 @@ namespace XrdCl
...
+    // Issue a new stat request
+    // stating a file handle doesn't work (fixed in 3.2.0) so we need to
+    // stat the pat
+    //--------------------------------------------------------------------------
+    Message           *msg;
+    ClientStatRequest *req;
+    std::string        path = self->pFileUrl->GetPath();
+    MessageUtils::CreateRequest( msg, req );
+
+    req->requestid = kXR_statx;
```
abh3, line 1105: This makes no sense. kXR)_statx is already defined with a specific response that has absolutely nothing to do what is being implemented here, This really needs to be completely rethought as you cannot use kXR_statx as it would make the server incompatible with any previous server.
https://xrootd.web.cern.ch/doc/dev6/XRdv520.htm#_Toc173272963

src/XrdCl/XrdClFileStateHandler.cc
```diff
@@ -2692,6 +2751,18 @@ namespace XrdCl
         break;
       }
 
+      //------------------------------------------------------------------------
+      // Cache the statx response
+      //------------------------------------------------------------------------
+      case kXR_statx:
```
abh3, line 2757: This is returning the already defined statx repose which has nothing to do with linux statx(). So, this is a conflict.

src/XrdCl/XrdClFileStateHandler.hh
```diff
@@ -951,6 +967,7 @@ namespace XrdCl
       FileStatus              pFileState;
       XRootDStatus            pStatus;
       StatInfo               *pStatInfo;
+      StatxInfo              *pStatxInfo;
```
abh3, line 970: This will not work. x=kXR_statx already has a definition that does not match this definition. The protocol cannot be changed this way is it totally breaks past servers. This needs to be totally reworked,.

src/XrdCl/XrdClFileSystem.cc
```diff
@@ -1585,6 +1585,44 @@ namespace XrdCl
     return MessageUtils::WaitForResponse( &handler, response );
   }
 
+  XRootDStatus FileSystem::Statx( const std::string &path,
```
abh3, line 1588: This has the same problem and is not compatible with the current definition.

src/XrdDig/XrdDigFS.cc
```diff
@@ -805,6 +805,33 @@ int XrdDigFS::stat(const char              *path,        // In
    return SFS_OK;
 }
 
+/******************************************************************************/
+/*                                  s t a t                                   */
```
abh3, line 809: This does not need to change as per point B.

src/XrdDig/XrdDigFS.hh
```diff
@@ -140,6 +140,9 @@ public:
         int            sync(XrdSfsAio *aiop) {return SFS_OK;}
 
         int            stat(struct stat *buf);
+#if defined(__linux__)
```
abh3, line 143: Point A and point B does not require any changes to this file.

src/XrdOfs/XrdOfs.cc
```diff
@@ -1711,6 +1711,25 @@ int XrdOfsFile::stat(struct stat     *buf)         // Out
    return SFS_OK;
 }
 
+#if defined(__linux__)
```
abh3, line 1714: See point A.

src/XrdOfs/XrdOfs.cc
```diff
@@ -2543,6 +2562,99 @@ int XrdOfs::stat(const char             *path,        // In
    return SFS_OK;
 }
 
+/******************************************************************************/
+/*                                  s t a t                                   */
+/******************************************************************************/
+#if defined(__linux__)
```
abh3, line 2568: See point A.

src/XrdOfs/XrdOfs.cc
```diff
@@ -2543,6 +2562,99 @@ int XrdOfs::stat(const char             *path,        // In
...
+// Now try to find the file or directory
+//
+   if ((retc = XrdOfsOss->Statx(path, buf, 0, &stat_Env)))
+      return XrdOfsFS->Emsg(epname, einfo, retc, "locate", path, "?");
+   return SFS_OK;
+}
+
+/******************************************************************************/
+
+int XrdOfs::statx(const char             *path,        // In
```
abh3, line 2611: Why is this being also defined as statx()? It does exactly the same thing as Stat() with this signature. Please remove it.

src/XrdOfs/XrdOfs.hh
```diff
@@ -341,7 +343,19 @@ const   char          *getVersion();
...
                             const XrdSecEntity     *client,
                             const char             *opaque = 0);
+#if defined(__linux__)
+        int            statx(const char             *Name,
+                                  struct statx      *buf,
+                                  XrdOucErrInfo    &out_error,
+                            const XrdSecEntity     *client,
+                            const char             *opaque = 0);
 
+        int            statx(const char             *Name,
```
abh3, line 353: There is no need for this method as it duplicates the current stat() method.

src/XrdOss/XrdOss.hh
```diff
@@ -162,6 +162,16 @@ virtual void    Flush() {}
...
 virtual int     Fstat(struct stat *buf) {return -EISDIR;}
 
+//-----------------------------------------------------------------------------
+//! Return statx information for this file.
+//!
+//! @param  buf    - Pointer to the structure where info it to be returned.
+//!
+//! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+//-----------------------------------------------------------------------------
+#if defined(__linux__)
```
abh3, line 172: See point A.,

src/XrdOss/XrdOss.hh
```diff
@@ -763,6 +773,24 @@ virtual int       Rename(const char *oPath, const char *nPath,
...
+  //! @param  opts   - Options:
+  //!                  XRDOSS_preop    - this is a stat prior to open.
+  //!                  XRDOSS_resonly  - only look for resident files.
+  //!                  XRDOSS_updtatm  - update file access time.
+  //! @param  envP   - Pointer to environmental information.
+  //!
+  //! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+  //-----------------------------------------------------------------------------
+#if defined(__linux__)
+virtual int       Statx(const char *path, struct statx *buff,
```
abh3, line 790: Please remove as it duplicates the already defined stat() with this signature.

src/XrdOss/XrdOssApi.cc
```diff
@@ -1232,7 +1232,11 @@ int XrdOssFile::Fstat(struct stat *buff)
 {
     return (fstat(fd, buff) ? -errno : XrdOssOK);
 }
-
+#if defined(__linux__)
+int XrdOssFile::Fstatx(struct statx * buff) {
+   return (statx(fd, "", AT_EMPTY_PATH, STATX_ALL, buff) ? -errno : XrdOssOK);
```
abh3, line 1237: This defeats the major purpose of statx() which allowed you to get only the fields of interest as an optimization. It's actually pretty bas for some filesystems because they have to rummage through additional metadata databases to return everything; which is one of the things Linux wanted to avoid. by using a "mask". Perhaps this implementation of statx() should really be rethought to make it more relevant.

src/XrdOss/XrdOssMSS.cc
```diff
@@ -310,7 +310,71 @@ int XrdOssSys::MSS_Stat(const char *path, struct stat *buff)
     delete sfd;
     return 0;
 }
+#if defined(__linux__)
+int XrdOssSys::MSS_Statx(const char *path, struct statx *buff) {
```
abh3, line 314: This implementation has far reaching secondary effects that are much more impactful that one could imagine. Please DO NOT implement it this way. The MSS_Statx() method should, if it exists at all, merely call the MSS_Stat() method as there is no definition on what "statx" should return by external tape systems. The answer here is no, you cannot do this.

src/XrdOss/XrdOssStat.cc
```diff
@@ -126,6 +127,66 @@ int XrdOssSys::Stat(const char *path, struct stat *buff, int opts,
    return XrdOssOK;
 }
 
+#if defined(__linux__)
```
abh3, line 130: See point A.

src/XrdOss/XrdOssStat.cc
```diff
@@ -126,6 +127,66 @@ int XrdOssSys::Stat(const char *path, struct stat *buff, int opts,
...
+int XrdOssSys::Statx(const char *path, struct statx *buff, int opts,
```
abh3, line 131: This is really unfortunate because a lot of code is being duplicated when the only difference is whether stat() or statx() is called. It's a big hit in maintenance and we really need to rethink on how stat() vs statx() is handled in the oss plugin. I don't have a solution but may once all the other issues are addressed.

src/XrdOss/XrdOssWrapper.hh
```diff
@@ -710,6 +710,24 @@ virtual int       Stat(const char *path, struct stat *buff,
...
+//! @param  buff   - Pointer to the structure where info it to be returned.
+//! @param  opts   - Options:
+//!                  XRDOSS_preop    - this is a stat prior to open.
+//!                  XRDOSS_resonly  - only look for resident files.
+//!                  XRDOSS_updtatm  - update file access time.
+//! @param  envP   - Pointer to environmental information.
+//!
+//! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+//-----------------------------------------------------------------------------
+#if defined(__linux__)
```
abh3, line 726: See point A.

src/XrdOssCsi/XrdOssCsi.cc
```diff
@@ -415,6 +415,15 @@ int XrdOssCsi::Stat(const char *path, struct stat *buff, int opts,
    return successor_->Stat(path, buff, opts, EnvP);
 }
 
+#if defined(__linux__)
```
abh3, line 418: This class need not change, it should use the default.

src/XrdOssCsi/XrdOssCsi.hh
```diff
@@ -211,6 +211,10 @@ virtual int       Chmod(const char *path, mode_t mode, XrdOucEnv *envP=0) /* ove
 virtual int       Remdir(const char *path, int Opts=0, XrdOucEnv *eP=0) /* override */;
 virtual int       Stat(const char *path, struct stat *buff, int opts=0,
                     XrdOucEnv  *EnvP=0) /* override */;
+#if defined(__linux__)
```
abh3, line 214: We should use the default here.

src/XrdOssStats/XrdOssStatsFileSystem.cc
```diff
@@ -163,6 +163,15 @@ int       FileSystem::Stat(const char *path, struct stat *buff,
     return wrapPI.Stat(path, buff, opts, env);
 }
 
+#if defined(__linux__)
```
abh3, line 166: This class should use the default implementation. No changes needed here.

src/XrdOssStats/XrdOssStatsFileSystem.hh
```diff
@@ -47,6 +47,10 @@ public:
                          XrdOucEnv  *oEnvP=0, XrdOucEnv *nEnvP=0) override;
     int       Stat(const char *path, struct stat *buff,
                        int opts=0, XrdOucEnv *env=0) override;
+#if defined(__linux__)
```
abh3, line 50: Same as above, this should not be changed,. The default implementation should be used.

src/XrdOuc/XrdOucCache.hh
```diff
@@ -680,6 +680,10 @@ virtual int   Rmdir(const char* dirp) {(void)dirp; return 0;}
 virtual int  Stat(const char *url, struct stat &sbuff)
                  {(void)url; (void)sbuff; return 1;}
 
+#if defined(__linux__)
```
abh3, line 683: Point A. If you must define statx() for the cache (which I see limited use for) it should call stat() not just return 1 with the statx buf undefined. After calling styat() it should set the statx buffer using the stat2stax menthid. Frankly, I wouldn't bother defining this at all unless someone is actually calling statx() which I don't think is the case.

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1688,4 +1688,13 @@ std::string XrdOucUtils::UrlDecode(const std::string &input)
   return out;
 }
 
+#if defined(__linux__)
```
abh3, line 1691: See point A.

src/XrdOuc/XrdOucUtils.hh
```diff
@@ -156,7 +156,9 @@ static uint8_t touint8_t(const std::string_view sv);
 static std::string UrlEncode(const std::string &input);
 
 static std::string UrlDecode(const std::string &input);
-
+#if defined(__linux__)
```
abh3, line 159: See point A.

src/XrdPfc/XrdPfc.cc
```diff
@@ -1244,6 +1244,62 @@ int Cache::Stat(const char *curl, struct stat &sbuff)
    return 0;
 }
 
+#if defined(__linux__)
```
abh3, line 1247: As I pointed out in XrdOucCache which this object uses as a base class,, there is no need for the cache to support statx() beyond what stat() provides. So, this duplication of code is not needed.

src/XrdPfc/XrdPfc.hh
```diff
@@ -193,6 +193,10 @@ public:
    // virtual function of XrdOucCache.
    virtual int  Stat(const char *url, struct stat &sbuff);
 
+#if defined(__linux__)
```
abh3, line 196: See point A and see comments on the "cc".

src/XrdPfc/XrdPfcFile.cc
```diff
@@ -679,6 +679,30 @@ int File::Fstat(struct stat &sbuff)
    return 0;
 }
 
+#if defined(__linux__)
```
abh3, line 682: Same thing here, I don't think we need to implement statx() for the cache.

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,7 +930,29 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
           : Xroot.Stat(myPath, buf));
 }
 }
-  
+
+/******************************************************************************/
+/*                         X r d P o s i x _ S t a t x                        */
+/******************************************************************************/
+#if defined(__linux__)
```
abh3, line 937: See point A.

src/XrdPosix/XrdPosix.hh
```diff
@@ -116,4 +116,7 @@
 
 #define writev(a,b,c)    XrdPosix_Writev(a,b,c)
 
+#if defined(__linux__)
```
abh3, line 119: See point A.

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -37,6 +37,11 @@
 
 #include "XrdNet/XrdNetAddr.hh"
 #include "XrdPosix/XrdPosixAdmin.hh"
+
+#if defined(__linux__)
+#include <sys/sysmacros.h>
```
abh3, line 42: Hmm, why?

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +251,88 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
    delete sInfo;
    return true;
 }
+
+#if defined(__linux__)
+bool XrdPosixAdmin::Statx(mode_t *flags, time_t *mtime)
```
abh3, line 256: This makes no sense. It does exactly what stat() with the same signature does. I don't see why it should duplicate the previous code. Please remove it.

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +251,88 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
...
+  XrdCl::XRootDStatus xStatus;
+  XrdCl::StatxInfo    *sInfo = 0;
+
+  // Make sure admin is ok
+  //
+  if (!isOK()) return false;
+
+  // Issue the stat and verify that all went well
+  //
+  xStatus = Xrd.Statx(Url.GetPathWithParams(), sInfo);
```
abh3, line 296: This is al fine and dandy but it doesn't accomplish what statx() actually does given the client implementation. So, it's irrelevant and quite confusing.

src/XrdPosix/XrdPosixCache.cc
```diff
@@ -89,6 +89,16 @@ int XrdPosixCache::Rename(const char* oldPath, const char* newPath)
 int XrdPosixCache::Stat(const char *path, struct stat &sbuff)
                        {return theCache->Stat(path, sbuff);}
 
+/******************************************************************************/
+/*                                  S t a t                                   */
+/******************************************************************************/
+
+#if defined(__linux__)
+int XrdPosixCache::Statx(const char *path, struct statx &sbuff)
```
abh3, line 97: I don't see the point here at all.

src/XrdPosix/XrdPosixCache.hh
```diff
@@ -98,6 +98,21 @@ int           Rename(const char* oldPath, const char* newPath);
 
 int           Stat(const char *path, struct stat &sbuff);
 
+
+//-----------------------------------------------------------------------------
+//! Rename a file or directory in the cache.
```
abh3, line 103: I think anything to do with statx() should not apply to the Posix cache. It's just not going to work the way it is implemented. The whole concept of statx() in the p[osix realm needs to be rethought.

src/XrdPosix/XrdPosixPreload.cc
```diff
@@ -693,6 +693,16 @@ int        stat64(         const char *path, struct stat64 *buf)
 }
 }
 
+#if defined(__linux__)
+extern "C"
+{
+   int statx(int dirfd, const char *path, int flags, unsigned int mask, struct statx * statxbuf) {
+      static int Init = Xunix.Init(&Init);
+
+      return XrdPosix_Statx(dirfd,path,flags,mask,statxbuf);
```
abh3, line 702: The statx() call either needs to simply use stat() to fill out the statx buffer as I previously described in point C. It can be done here or in the statx() method. but it should not use the client's statx() to do that.

src/XrdPosix/XrdPosixXrootd.cc
```diff
@@ -1221,8 +1221,48 @@ int XrdPosixXrootd::Stat(const char *path, struct stat *buf)
 //
    if (cacheChk) buf->st_atime = 0;
    return 0;
-}        
+}
+
+#if defined(__linux__)
```
abh3, line 1226: Point A and statx() should be implemented as the default -- use a stat() call and initialize the statx buffer from it. This eliminates most of the duplication here.

src/XrdPss/XrdPss.cc
```diff
@@ -594,6 +594,29 @@ int XrdPssSys::Stat(const char *path, struct stat *buff, int Opts, XrdOucEnv *eP
    return (XrdPosixXrootd::Stat(pbuff, buff) ? Info(errno) : XrdOssOK);
 }
 
+/******************************************************************************/
```
abh3, line 597: There is no reason for any changes in the XrdPss objects. They will use the default implementation of statx() supplied by the base class. So none of this is needed.

src/XrdPss/XrdPss.hh
```diff
@@ -188,6 +188,9 @@ int       Remdir(const char *, int Opts=0, XrdOucEnv *eP=0) override;
 int       Rename(const char *, const char *,
                  XrdOucEnv *eP1=0, XrdOucEnv *eP2=0) override;
 int       Stat(const char *, struct stat *, int opts=0, XrdOucEnv *eP=0) override;
+#if defined(__linux__)
```
abh3, line 191: Point A and the cc comment.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -750,6 +750,10 @@ virtual XrdSfsXferSize writev(XrdOucIOVec      *writeV,
 
 virtual int            stat(struct stat *buf) = 0;
 
+#if defined(__linux__)
+virtual int            statx(struct statx *buf) = 0;
```
abh3, line 754: NO, a default implementation as in point B needs to exist in the cc file.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -1301,7 +1305,46 @@ virtual int            stat(const char               *path,
...
+//! @param  buf    - Pointer to the structure where info it to be returned.
+//! @param  eInfo  - The object where error info is to be returned.
+//! @param  client - Client's identify (see common description).
+//! @param  opaque - path's CGI information (see common description).
+//!
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, SFS_STALL, or SFS_STARTED
+//!         When SFS_OK is returned, buf must contain stat information.
+//-----------------------------------------------------------------------------
+
+virtual int            statx(const char               *Name,
```
abh3, line 1322: NO, a default implementation must exist as in point B.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -1301,7 +1305,46 @@ virtual int            stat(const char               *path,
...
+//! @param  eInfo  - The object where error info is to be returned.
+//! @param  client - Client's identify (see common description).
+//! @param  opaque - path's CGI information (see common description).
+//!
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, SFS_STALL, or SFS_STARTED
+//!         When SFS_OK is returned, mode must contain mode information. If
+//!         the mode is -1 then it is taken as an offline file.
+//-----------------------------------------------------------------------------
+
+virtual int            statx(const char               *path,
```
abh3, line 1342: This makes no sense and should be eliminated.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -573,6 +577,31 @@ int XrdSfsNativeFile::stat(struct stat     *buf)         // Out
    return SFS_OK;
 }
 
+/******************************************************************************/
+/*                                  s t a t                                   */
+/******************************************************************************/
+#if defined(__linux__)
+int XrdSfsNativeFile::statx(struct statx    *buf)         // Out
```
abh3, line 584: The default should be used.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -971,6 +1000,32 @@ int XrdSfsNative::stat(const char              *path,        // In
    return SFS_OK;
 }
 
+/******************************************************************************/
+/*                                  s t a t                                   */
```
abh3, line 1004: Wrong header coment.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -971,6 +1000,32 @@ int XrdSfsNative::stat(const char              *path,        // In
...
+/******************************************************************************/
+#if defined(__linux__)
+int XrdSfsNative::statx(const char              *path,        // In
```
abh3, line 1007: The default should be used.

src/XrdSfs/XrdSfsNative.hh
```diff
@@ -227,7 +230,24 @@ const   char          *getVersion();
...
                         return rc;
                        }
+#if defined(__linux__)
+      int            statx(const char             *Name,
+                                  struct statx      *buf,
+                                  XrdOucErrInfo    &out_error,
+                            const XrdSecClientName *client = 0,
+                            const char             *opaque = 0);
 
+      int            statx(const char             *Name,
```
abh3, line 240: As I said, this signature makes no sense. It should be removed.

src/XrdSsi/XrdSsiFile.cc
```diff
@@ -402,6 +402,18 @@ int XrdSsiFile::stat(struct stat     *buf)         // Out
    return SFS_OK;
 }
 
+#if defined(__linux__)
```
abh3, line 405: The default as in point B should be used.

src/XrdSsi/XrdSsiSfs.cc
```diff
@@ -498,6 +498,42 @@ int XrdSsiSfs::stat(const char             *path,        // In
    return SFS_ERROR;
 }
 
+/******************************************************************************/
+/*                                  s t a t x                                 */
```
abh3, line 502: There is no reason not to use the default. Changes to this file are not needed.

src/XrdSsi/XrdSsiSfs.hh
```diff
@@ -123,7 +123,19 @@ const   char          *getVersion();
                                   XrdOucErrInfo    &eInfo,
                             const XrdSecEntity     *client,
                             const char             *opaque = 0);
+#if defined(__linux__)
```
abh3, line 126: No changes to this file are needed.

src/XrdThrottle/XrdThrottle.hh
```diff
@@ -105,6 +105,10 @@ public:
    virtual int
    stat(struct stat *buf) override;
 
+#if defined(__linux__)
```
abh3, line 108: No changes to this file are needed.

src/XrdThrottle/XrdThrottleFile.cc
```diff
@@ -220,6 +220,14 @@ File::stat(struct stat *buf)
    return m_sfs->stat(buf);
 }
 
+#if defined(__linux__)
```
abh3, line 223: No changes to this file are needed.

src/XrdThrottle/XrdThrottleFileSystem.cc
```diff
@@ -194,7 +194,27 @@ FileSystem::stat(const char             *Name,
 {
    return m_sfs_ptr->stat(Name, mode, out_error, client, opaque);
 }
+#if defined(__linux__)
```
abh3, line 197: No changes to this file are needed.

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -59,6 +59,10 @@
 #include "XrdXrootd/XrdXrootdTrace.hh"
 #include "XrdXrootd/XrdXrootdXPath.hh"
 
+#if defined(__linux__)
+#include <sys/sysmacros.h>
```
abh3, line 63: Why?

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -499,6 +503,14 @@ int XrdXrootdProtocol::Process2()
                                  return do_Stat();
                                 }
                              break;
+          case kXR_statx:
+#if defined(__linux__)
```
abh3, line 507: Huh? The defines action of a pre-existing request code is being changed. This is not allowed! The code here cannot exist as written.

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -598,7 +610,9 @@ int XrdXrootdProtocol::Process2()
           case kXR_rmdir:     return do_Rmdir();
           case kXR_set:       return do_Set();
           case kXR_stat:      return do_Stat();
+#if defined(__linux__)
```
abh3, line 613: This makes no sense, this has already been defined and cannot be if/def defined. This is just wrong..

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -850,6 +864,115 @@ int XrdXrootdProtocol::StatGen(struct stat &buf, char *xxBuff, int xxLen,
    return xxBuff - origP;
 }
 
+#if defined(__linux__)
+int XrdXrootdProtocol::StatxGen(struct statx &buf, char *xxBuff, int xxLen,
```
abh3, line 868: This whole duplicate of statgen based on a statx buffer will generate random values because it does not check the stx_mask to see if a value actually exists. So, it's just plain wrong. It's also not clear who would even call this as it only add birth time. There are far better methods to do this and this is not going to work.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -3137,41 +3137,62 @@ int XrdXrootdProtocol::do_Stat()
 /******************************************************************************/
 /*                              d o _ S t a t x                               */
 /******************************************************************************/
-  
+
+#if defined(__linux__)
 int XrdXrootdProtocol::do_Statx()
 {
-   static XrdXrootdCallBack statxCB("xstat", XROOTD_MON_STAT);
+   static XrdXrootdCallBack statCB("statx", XROOTD_MON_STAT);
```
abh3, line 3144: You cannot change this method in the way that it is being changed, It implements a specific protocol response which changes here totally violate. This is not allowed.; The whole statx() implementation needs to be rethought.

Issue #2702, opened 2026-01-28
occasional failre S3ListFixture.Mkdir


smithdh:
### Operating System

Ubuntu 24.04

### XRootD Version

master (4535e133)

### Description

There were occasional failures in [S3ListFixture.Mkdir](https://my.cdash.org/tests/339145826); apparently ubuntu 24.04.03 Clang (debug). I could not reproduce it when I tried running the test with 'ctest' in a docker, but was able to produce some crashes ("1 in a few runs") by running:

```
gdb ./bin/xrdcl-s3-test
run --gtest_filter=\*Mkdir
[...]
Thread 11 "xrdcl-s3-test" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f3026ffd6c0 (LWP 254598)]
[...]
(gdb) where
#0  0x00007f303fe50293 in __gthread_mutex_unlock (__mutex=0x7f303ff228e0 <XrdClHttp::Factory::m_shutdown_lock>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/x86_64-linux-gnu/c++/13/bits/gthr-default.h:779
#1  0x00007f303fe57425 in std::mutex::unlock (this=0x7f303ff228e0 <XrdClHttp::Factory::m_shutdown_lock>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/std_mutex.h:132
#2  0x00007f303fe573f2 in std::unique_lock<std::mutex>::unlock (this=0x7f3026ffc798)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/unique_lock.h:200
#3  0x00007f303fe531f7 in std::unique_lock<std::mutex>::~unique_lock (this=0x7f3026ffc798)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/13/../../../../include/c++/13/bits/unique_lock.h:106
#4  0x00007f303fe4ebab in XrdClHttp::Factory::Monitor (this=0x563385628e20)
    at /home/xrootd/xrootd/src/XrdClHttp/XrdClHttpFactory.cc:269
#5  0x00007f303fe52cb8 in XrdClHttp::Factory::Initialize()::$_0::operator()() const::{lambda()#1}::operator()() const (this=0x563385681528) at /home/xrootd/xrootd/src/XrdClHttp/XrdClHttpFactory.cc:202
[...]

PR #2704, opened 2026-01-29
[CI] Add XRD_PLUGINCONFDIR env var to test confs

dynamic-entropy:
The XrdClCurl/HTTP plugin requires that the lib be specified in the plugin configuration file.
Adding it to various tests that use the client (even though they do not necessarily use HTTP atleast for now).


dynamic-entropy:
This might be related: https://github.com/xrootd/xrootd/issues/2702

PR #2713, opened 2026-02-06
[XrdHttpTPC] Allow missing CRL checking if configured via xrd.tlsca

### LINKED ISSUE #2681: HTTP-TPC fails with missing CRL for CA in the chain
### Operating System

Alma Linux 9

### XRootD Version

5.9.1

### Description

HTTP-TPC pull with XRootD 5.9.1 active party fails with a slightly misleading error message `SSL peer certificate or SSH remote key was not OK`, e.g.
```bash
$ curl -v --capath /etc/grid-security/certificates -L -X COPY -H 'Credential: none' -H "Authorization: Bearer $TDST" -H "TransferHeaderAuthorization: Bearer $TSRC" -H "Source: $SRC" "$DST" 
...
> COPY /atlas/atlasdatadisk/SAM/test.20251226005812.pull HTTP/1.1
> Host: xrd.farm.particle.cz:1094
> User-Agent: curl/7.76.1
> Accept: */*
> Credential: none
> Authorization: Bearer secret1
> TransferHeaderAuthorization: Bearer secret2
> Source: https://dcache.farm.particle.cz:443/atlas/atlasdatadisk/SAM/1M
...
< HTTP/1.1 500 Internal Server Error
< Connection: Keep-Alive
< Server: XrootD/v5.9.1
< Content-Length: 256
< 
...
failure: Internal transfer failure while fetching remote size, local=/atlas/atlasdatadisk/SAM/test.20251226005812, remote=https://dcache.farm.particle.cz:443/atlas/atlasdatadisk/SAM/1M, HTTP library failure=SSL peer certificate or SSH remote key was not OK
```
As a workaround it is possible to set `XRDTPC_CADIR=/etc/grid-security/certificates`, but this basically disable fancy merging of CA and CRLs implemented in XRootD as an another workaround discussed in https://github.com/xrootd/xrootd/pull/1431

HTTP-TPC code should also implement something similar to [crl:{ignore,try,use,require}](https://xrootd.web.cern.ch/doc/dev56/sec_config.htm#_Toc119617433) and at least ignore missing CRLs for e.g. Let's encrypt host certificates ... e.g. this patch allows me to do HTTP-TPC pull on my XRootD testbed from source that use Let's Encrypt host certificates

```patch
diff --git a/src/XrdHttpTpc/XrdHttpTpcTPC.cc b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
index 17d61119e..c22c07981 100644
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
@@ -14,6 +14,7 @@
 #include "XrdHttp/XrdHttpUtils.hh"
 
 #include <curl/curl.h>
+#include <openssl/ssl.h>
 
 #include <dlfcn.h>
 #include <fcntl.h>
@@ -154,6 +155,36 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
   return close(fd);
 }
 
+/*           s s l _ c t x _ c a l l b a c k                                  */
+
+/**
+ * The callback that will be called by libcurl just before the initialization of an SSL connection
+ * after having processed all other SSL related options to give a last chance to an application to
+ * modify the behavior of the SSL initialization.
+ * https://curl.se/libcurl/c/CURLOPT_SSL_CTX_FUNCTION.html
+ */
+int TPCHandler::ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *clientp) {
+  //TPCLogRecord * rec = (TPCLogRecord *)clientp;
+  SSL_CTX* ctx = static_cast<SSL_CTX*>(ssl_ctx);
+  SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, verify_callback);
+  return CURL_SOCKOPT_OK;
+}
+
+int TPCHandler::verify_callback(int preverify_ok, X509_STORE_CTX* ctx) {
+  if (preverify_ok == 1) return 1;
+
+  int err = X509_STORE_CTX_get_error(ctx);
+
+  if (err == X509_V_ERR_UNABLE_TO_GET_CRL) {
+    X509_STORE_CTX_set_error(ctx, X509_V_OK);
+    return 1;
+  }
+
+  return 0;
+}
+
 /*                            p r e p a r e U R L                             */
@@ -215,6 +246,7 @@ TPCHandler::ConfigureCurlCA(CURL *curl)
         std::ifstream in(crl_filename, std::ifstream::ate | std::ifstream::binary);
         if(in.tellg() > 0 && m_ca_file->atLeastOneValidCRLFound()){
             curl_easy_setopt(curl, CURLOPT_CRLFILE, crl_filename.c_str());
+            curl_easy_setopt(curl, CURLOPT_SSL_CTX_FUNCTION, ssl_ctx_callback);
         } else {
             std::ostringstream oss;
             oss << "No valid CRL file has been found in the file " << crl_filename << ". Disabling CRL checking.";
diff --git a/src/XrdHttpTpc/XrdHttpTpcTPC.hh b/src/XrdHttpTpc/XrdHttpTpcTPC.hh
index 0f23f7b38..5f6d18170 100644
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.hh
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.hh
@@ -13,6 +13,7 @@
 #include "XrdHttpTpcPMarkManager.hh"
 
 #include <curl/curl.h>
+#include <openssl/x509_vfy.h>
 
 class XrdOucErrInfo;
 class XrdOucStream;
@@ -62,6 +63,9 @@ private:
 
     static int closesocket_callback(void *clientp, curl_socket_t fd);
 
+    static int ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *clientp);
+    static int verify_callback(int preverify_ok, X509_STORE_CTX* ctx);
+
     struct TPCLogRecord {
 
         TPCLogRecord(XrdHttpExtReq & req, const TpcType tpcType) : bytes_transferred( -1 ), status( -1 ),

```
With this path transfer from test mentioned above succeeds
```bash
$ curl -v --capath /etc/grid-security/certificates -L -X COPY -H 'Credential: none' -H "Authorization: Bearer $TDST" -H "TransferHeaderAuthorization: Bearer $TSRC" -H "Source: $SRC" "$DST.pull" 
*   Trying 2001:718:401:6025:1:0:1:83:1094...
* Connected to xrd.farm.particle.cz (2001:718:401:6025:1:0:1:83) port 1094 (#0)
...
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: CN=xrd.farm.particle.cz
*  start date: Oct 26 02:01:39 2025 GMT
*  expire date: Jan 24 02:01:38 2026 GMT
*  subjectAltName: host "xrd.farm.particle.cz" matched cert's "xrd.farm.particle.cz"
*  issuer: C=US; O=Let's Encrypt; CN=R12
*  SSL certificate verify ok.
* TLSv1.2 (OUT), TLS header, Unknown (23):
> COPY /atlas/atlasdatadisk/SAM/test.20251226005812.pull HTTP/1.1
> Host: xrd.farm.particle.cz:1094
> User-Agent: curl/7.76.1
> Accept: */*
> Credential: none
> Authorization: Bearer secret1
> TransferHeaderAuthorization: Bearer secret2
> Source: https://dcache.farm.particle.cz:443/atlas/atlasdatadisk/SAM/test.20251222102549
> 
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* TLSv1.2 (IN), TLS header, Unknown (23):
* Mark bundle as not supporting multiuse
< HTTP/1.1 201 Created
< Connection: Keep-Alive
< Server: XrootD/v5.9.1
< Content-Type: text/plain
< Transfer-Encoding: chunked
< 
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
Perf Marker
Timestamp: 1766879402
Stripe Index: 0
Stripe Bytes Transferred: 0
Total Stripe Count: 1
RemoteConnections: tcp:[2001:718:401:6025:1:0:1:80]:443
End
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
* TLSv1.2 (IN), TLS header, Unknown (23):
* Connection #0 to host xrd.farm.particle.cz left intact
success: Created
```

1. configure simple XRootD server with hostcerts from Let's Encrypt
2. run HTTP-TPC pull with source & destination from this test server

### Additional Information

_No response_

ccaffy:
One can deactivate missing CRL checking by setting the flag "allowmissingcrl" in the xrd.tlsca directive

Fixes #2681

### DIFF
--- a/src/Xrd/XrdConfig.cc
+++ b/src/Xrd/XrdConfig.cc
@@ -2532,12 +2532,16 @@ int XrdConfig::xtlsca(XrdSysError *eDest, XrdOucStream &Config)
       }
    tlsNoVer = false;
 
-
    do {if (!strcmp(val, "proxies") || !strcmp("noproxies", val))
           {if (*val == 'n') tlsOpts |=  XrdTlsContext::nopxy;
               else          tlsOpts &= ~XrdTlsContext::nopxy;
            continue;
           }
+
+         if (!strcmp(val,"allowmissingcrl")) {
+            tlsOpts |= XrdTlsContext::crlAM;
+            continue;
+         }
 
        if (strlen(val) >= (int)sizeof(kword))
           {eDest->Emsg("Config", "Invalid tlsca parameter -", val);
@@ -2590,8 +2594,7 @@ int XrdConfig::xtlsca(XrdSysError *eDest, XrdOucStream &Config)
                {if (XrdOuca2x::a2i(*eDest,"tlsca verdepth",val,&vd,1,255))
                    return 1;
                 tlsOpts = TLS_SET_VDEPTH(tlsOpts,vd);
-               }
-       else {eDest->Emsg("Config", "invalid tlsca option -",kword); return 1;}
+               } else {eDest->Emsg("Config", "invalid tlsca option -",kword); return 1;}
 
        } while((val = Config.GetWord()));
 
--- a/src/XrdHttpTpc/CMakeLists.txt
+++ b/src/XrdHttpTpc/CMakeLists.txt
@@ -34,6 +34,8 @@ target_link_libraries(${XrdHttpTPC}
     XrdUtils
     XrdHttpUtils
     CURL::libcurl
+    OpenSSL::SSL
+    OpenSSL::Crypto
     ${CMAKE_THREAD_LIBS_INIT}
     ${CMAKE_DL_LIBS}
 )
--- a/src/XrdHttpTpc/XrdHttpTpcConfigure.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcConfigure.cc
@@ -24,7 +24,8 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
 
     // test if XrdEC is used
     usingEC = getenv("XRDCL_EC")? true : false;
-
+    // Test if the CRL checking is enabled
+    allowMissingCRL = (bool) myEnv->GetInt("http.allowmissingcrl");
     std::string authLib;
     std::string authLibParms;
     int cfgFD = open(configfn, O_RDONLY, 0);
--- a/src/XrdTls/XrdTlsContext.hh
+++ b/src/XrdTls/XrdTlsContext.hh
@@ -252,9 +252,13 @@ static const uint64_t rfCRL = 0x0000004000000000; //!< Turn on the CRL refresh t
 static const uint64_t crlON = 0x0000008000000000; //!< Enables crl checking
 static const uint64_t crlFC = 0x000000C000000000; //!< Full crl chain checking
 static const uint64_t crlRF = 0x00000000ffff0000; //!< Mask to isolate crl refresh in min
+static const uint64_t crlAM = 0x0000001000000000; //!< Allow CA validation when CRL is missing (CRL soft-fail)
 static const int      crlRS = 16;                 //!< Bits to shift   vdept
 static const uint64_t artON = 0x0000002000000000; //!< Auto retry Handshake
 static const uint64_t clcOF = 0x0000010000000000; //!< Disable client certificate request
+
+
+static int ctxIndex;
 
        XrdTlsContext(const char *cert=0,  const char *key=0,
                      const char *cadir=0, const char *cafile=0,
--- a/src/XrdHttp/XrdHttpProtocol.cc
+++ b/src/XrdHttp/XrdHttpProtocol.cc
@@ -83,6 +83,7 @@ char *XrdHttpProtocol::sslcert = 0;
 char *XrdHttpProtocol::sslkey = 0;
 char *XrdHttpProtocol::sslcadir = 0;
 int XrdHttpProtocol::crlRefIntervalSec = XrdTlsContext::DEFAULT_CRL_REF_INT_SEC;
+bool XrdHttpProtocol::allowMissingCRL = false;
 char *XrdHttpProtocol::sslcipherfilter = 0;
 char *XrdHttpProtocol::listredir = 0;
 bool XrdHttpProtocol::listdeny = false;
@@ -1136,6 +1137,10 @@ int XrdHttpProtocol::Config(const char *ConfigFN, XrdOucEnv *myEnv) {
        if (!httpsspec && what1) eDest.Say("Config Using ", what1);
        if (!httpsspec && what2) eDest.Say("Config Using ", what2);
        if (!httpsspec && what3) eDest.Say("Config Using ", what3);
+
+       if (cP->opts & XrdTlsContext::crlAM) {
+          allowMissingCRL = true;
+       }
       }
 
 // If a gridmap or secxtractor is present then we must be able to verify certs
@@ -1834,6 +1839,10 @@ bool XrdHttpProtocol::InitTLS() {
    std::string eMsg;
    uint64_t opts = XrdTlsContext::servr | XrdTlsContext::logVF |
                    XrdTlsContext::artON | XrdTlsContext::rfCRL;
+
+  if (allowMissingCRL) {
+    opts |= XrdTlsContext::crlAM;
+  }
 
 // Create a new TLS context
 //
@@ -3171,6 +3180,8 @@ int XrdHttpProtocol::LoadExtHandler(std::vector<extHInfo> &hiVec,
   if (sslcafile) myEnv.Put("http.cafile", sslcafile);
   if (sslcert)  myEnv.Put("http.cert",  sslcert);
   if (sslkey)   myEnv.Put("http.key"  , sslkey);
+  // Add the allowMissingCRL configuration to the environment
+  myEnv.PutInt("http.allowmissingcrl",allowMissingCRL ? 1 : 0);
 
   // Load all of the specified external handlers.
   //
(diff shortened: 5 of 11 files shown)


src/XrdHttpTpc/CMakeLists.txt
```diff
@@ -36,6 +36,8 @@ target_link_libraries(${XrdHttpTPC}
     CURL::libcurl
     ${CMAKE_THREAD_LIBS_INIT}
     ${CMAKE_DL_LIBS}
+    OpenSSL::SSL
+    OpenSSL::Crypto
```
amadio, line 38: These should go above the `${CMAKE...}` ones. Can you change, please?

src/XrdHttpTpc/XrdHttpTpcConfigure.cc
```diff
@@ -24,7 +24,8 @@ bool TPCHandler::Configure(const char *configfn, XrdOucEnv *myEnv)
 
     // test if XrdEC is used
     usingEC = getenv("XRDCL_EC")? true : false;
-
+    // Test if the CRL checking is enabled
+    allowMissingCRL = getenv("XRD_AMCRL") ? true : false;
```
amadio: I wonder if there's no better way of doing this.

src/Xrd/XrdConfig.cc
```diff
@@ -2525,13 +2525,19 @@ int XrdConfig::xtlsca(XrdSysError *eDest, XrdOucStream &Config)
...
 
-
    do {if (!strcmp(val, "proxies") || !strcmp("noproxies", val))
           {if (*val == 'n') tlsOpts |=  XrdTlsContext::nopxy;
               else          tlsOpts &= ~XrdTlsContext::nopxy;
            continue;
           }
 
+         if (!strcmp(val,"allowmissingcrl")) {
+            tlsOpts |= XrdTlsContext::crlAM;
```
amadio, line 2542: I don't see this used anywhere to apply the effects of this option.

src/XrdHttpTpc/CMakeLists.txt
ccaffy: OK :+1:

src/XrdHttpTpc/XrdHttpTpcConfigure.cc
ccaffy: This allowMissingCRL option is set at server start and is valid through all the server runtime. I can pass that via the XrdHttpProtocol and XrdHttpExtReq but not sure this adds value.

src/Xrd/XrdConfig.cc
ccaffy: yeah, it is there if one would need it for the XrootD protocol

src/XrdHttpTpc/XrdHttpTpcConfigure.cc
abh3: Likely not. BTW since the envar was set in the Xrd env object it should be accessible from the HTTP plugin using the env passed to it. So, there was no need to export it into the global environment. The less exported to the global environment the better.

abh3 (commented):
Just a single comment for now.

src/XrdTls/XrdTlsContext.hh
```diff
@@ -252,10 +252,13 @@ static const uint64_t rfCRL = 0x0000004000000000; //!< Turn on the CRL refresh t
 static const uint64_t crlON = 0x0000008000000000; //!< Enables crl checking
 static const uint64_t crlFC = 0x000000C000000000; //!< Full crl chain checking
 static const uint64_t crlRF = 0x00000000ffff0000; //!< Mask to isolate crl refresh in min
+static const uint64_t crlAM = 0x0000001000000000; //!< Allow CA validation when CRL is missing (CRL soft-fail)
 static const int      crlRS = 16;                 //!< Bits to shift   vdept
 static const uint64_t artON = 0x0000002000000000; //!< Auto retry Handshake
 static const uint64_t clcOF = 0x0000010000000000; //!< Disable client certificate request
 
+
+  static int ctxIndex;
```
amadio: This looks odly formatted. Is it intentional?

abh3 (approved):
Looks OK. Ipresume when you say crlcheck last and allowmissingcrl then if there is no crl  then all goes as planned as it's missing for the last entry and hence no crl is applied. Right?

ccaffy:
@abh3  in your case, CRL checking is attempted for the end-entity cert, but if no CRL is found, the connection is allowed anyway.

src/XrdTls/XrdTlsContext.hh
ccaffy: CORRECT! I'll modify and force-push

PR #2720, opened 2026-02-13
[XrdClHttp] Prevent requeue of completed readop

### LINKED ISSUE #2719: Timeout with `xrdfs`/`xrdcp` when performing operations via HTTP protocol
### Operating System

AlmaLinux 10.1 GCC 14.3.1 Debug

### XRootD Version

master@2e999a4b

### Description

The test `XRootD::http::test` fails due to `xrdfs`/`xrdcp` timing out when performing operations via HTTP using the new XrdClHttp plugin.

Problem seen in our regular CI build: https://my.cdash.org/tests/345859839

### Additional Information

_No response_

dynamic-entropy:
Fixes #2719 

Deliver the remaining data in prefetch buffers when `ResponseHandlers` for read operations are called after curl operation is done.

### DIFF
--- a/src/XrdClHttp/XrdClHttpOpRead.cc
+++ b/src/XrdClHttp/XrdClHttpOpRead.cc
@@ -63,12 +63,21 @@ CurlReadOp::Continue(std::shared_ptr<CurlOperation> op, XrdCl::ResponseHandler *
         }
     }
 
-    try {
-        m_continue_queue->Produce(op);
-    } catch (...) {
-        Fail(XrdCl::errInternal, ENOMEM, "Failed to continue the curl operation");
-        return false;
+    // This handles the case where the transfer finished but its last WriteCallback
+    // produced more data than the client buffer could hold, the excess being stored
+    // in the prefetch buffer
+    // we just need to deliver the response without re-queuing to the continue queue
+    if (op->IsDone()) {
+        DeliverResponse();
+    } else {
+        try {
+            m_continue_queue->Produce(op);
+        } catch (...) {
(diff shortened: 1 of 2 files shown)


amadio:
LGTM.

@bbockelm Could you please review? Thank you!

src/XrdClHttp/XrdClHttpOpRead.cc
```diff
@@ -63,12 +63,21 @@ CurlReadOp::Continue(std::shared_ptr<CurlOperation> op, XrdCl::ResponseHandler *
         }
     }
 
-    try {
-        m_continue_queue->Produce(op);
-    } catch (...) {
-        Fail(XrdCl::errInternal, ENOMEM, "Failed to continue the curl operation");
-        return false;
+    // This handles the case where the transfer finished but its last WriteCallback
```
dynamic-entropy, line 66: Brian has reviewed this commit before it was a PR. He requested an addition of a comment explaining this check for `Done`, which has been made.

Maybe @smithdh wants to express his thoughts, if this is a fitting solution, or are there cases we have left out?

smithdh: I'd say let's go with this to fix the issue. I had some worries about hypothetical cases in this area, but so far without a speicifc sequence. If there's something more we can open that as another issue (maybe with demonstrator, or listing specific ordering of actions where one can show a bad outcome).

amadio: Alright, if Brian already reviewed the commit, we can proceed then. Thanks.

PR #2721, opened 2026-02-20
Davs scheme support in XrdCl

dynamic-entropy:
Add support for davs scheme in XrdCl. 

The aim is to simply rewrite the URI with the http(s) counterpart. Curl itself does not provide a mechanism to register arbitrary (davs) schemes even for known protocols. As per this answer by Daniel (curl maintainer), they do not intend to either. https://github.com/curl/curl/discussions/11700 

Davix takes the same approach:
https://github.com/cern-fts/davix/blob/3054c828c4b63ff194880845c4a93a908b984a91/src/utils/davixuri.cpp#L238-L245

https://github.com/cern-fts/davix/blob/3054c828c4b63ff194880845c4a93a908b984a91/src/utils/davixuri.cpp#L133 

Haven't tested extensively, only in the XRootD tests in the second commit - but seemingly should be enough, since the stages after this translation are the same as http.

### DIFF
--- a/tests/XRootD/http.sh
+++ b/tests/XRootD/http.sh
@@ -172,13 +172,14 @@ function test_http() {
   expectedDigest="Digest: crc32c=$alphabetcrc32c"
   receivedDigest=$(grep "Digest" "$outputFilePath")
   assert_eq "$expectedDigest" "$receivedDigest" "HEAD request test failed (crc32c)"
+  
   # Test with md5 checksum to test the base64 encoding of the md5 byte-representation
-  cat /home/ccaffy/CLionProjects/xrootd/tests/XRootD/http.sh
   curl -v -I -H 'Want-Digest: md5' "${HTTP_HOST}/$alphabetFilePath" | tr -d '\r' > "$outputFilePath"
   cat "$outputFilePath"
   expectedDigest="Digest: md5=$alphabetmd5sumb64"
   receivedDigest=$(grep "Digest" "$outputFilePath")
   assert_eq "$expectedDigest" "$receivedDigest" "HEAD request test failed (md5)"
+
   curl -v -I -H 'Want-Digest: NotSupported, adler32, crc32c' "${HTTP_HOST}/$alphabetFilePath" | tr -d '\r' > "$outputFilePath"
   cat "$outputFilePath"
   expectedDigest="Digest: adler32=$alphabetadler32"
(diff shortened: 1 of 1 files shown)


tests/XRootD/http.sh
```diff
@@ -20,20 +20,22 @@ function teardown_http() {
 }
 
 function test_http() {
+
+	export HTTP_HOST="http://localhost:${XRD_PORT}"
+	export DAV_HOST="dav://localhost:${XRD_PORT}"
```
amadio: The script `test.sh` exports a host already, let's try to reuse that in favor of adding more hard-coded `localhost`. I'd like to keep things as centralized as possible so when a change is needed it is only needed in one place, not many scattered places throughout the test suite.

https://github.com/xrootd/xrootd/blob/56612fe85e34f51cabdd5f4902c34ce3c5222a63/tests/XRootD/test.sh#L156-L157

I think it's better to make all tests run with both `http://` and `dav://` instead of having some with one of them, and some with the other.

For later, since this test is not the only one that would benefit from this, we can split the tests into a separate function and run it in a for loop on `$HOST` using `${HOSTNAME}`, `localhost`, `127.0.0.1`, and `[::1]` to ensure that all of these work, and prepend each of the forms with a protocol `root(s)://`, `http(s)://`, and `dav(s)://` and run some basic set of tests covering `xrdcp`/`xrdfs` functionality.

dynamic-entropy: > The script test.sh exports a host already, let's try to reuse that in favor of adding more hard-coded localhost. I'd like to keep things as centralized as possible so when a change is needed it is only needed in one place, not many scattered places throughout the test suite.

Will do!

dynamic-entropy: > I think it's better to make all tests run with both http:// and dav:// instead of having some with one of them, and some with the other.

It won't be of much use; there is essentially only a string replacement here.

dynamic-entropy:
> We still need to understand, for example, subtle things like when to use HEAD vs PROPFIND, as there are some minor differences between https:// and davs:// there, but that can be a separate thing.

No, these are orthogonal concepts. We can use both the `HEAD` and `PROPFIND` verbs for both the schemes (`https://` and `davs://`). 

When to use `HEAD` vs `PROPFIND` is based on the operation we wish to perform
1. The kind of information we wish to retrieve, directory listing can be a crude example
2. Weather `PROPFIND` is supported, for which we use the `OPTIONS` request first, and a missing propfind would mean using head instead, for e.g.

tests/XRootD/http.sh
dynamic-entropy: > Will do!

Done!

PR #2724, opened 2026-02-26
[XrdHttp] Refactor URL parsing function and add unit tests

ccaffy:


src/XrdHttp/XrdHttpUtils.hh
```diff
@@ -122,10 +124,18 @@ enum : int {
...
-// GetHost from URL
-// Parse an URL and extract the host name and port
-// Return 0 if OK
-int parseURL(char *url, char *host, int &port, char **path);
+// Parsed result of parseURL.
+struct ParsedURL {
+  std::string host; // hostname or bare IPv6 address (without brackets)
+  int         port = 0; // 0 when no port is present in the URL
+  std::string path; // path component, always starts with '/'
+};
```
amadio, line 132: This is such that we always copy data and allocate space for `std::string`, which negates the benefits of using a `std::string_view` as input to `parseURL`. If you want to use `std::string_view`, then you should return string views for host and path based on the original, without copying (as long as there are no lifetime issues introduced, of course). The alignment of the struct with two strings with an int in between is also suboptimal. See output of `pahole` below:
```
struct ParsedURL {
	string                     host;                 /*     0    32 */
	int                        port;                 /*    32     4 */

	/* XXX 4 bytes hole, try to pack */

	string                     path;                 /*    40    32 */
	void ~ParsedURL(struct ParsedURL *, int);

	void ParsedURL(struct ParsedURL *);


	/* size: 72, cachelines: 2, members: 3 */
	/* sum members: 68, holes: 1, sum holes: 4 */
	/* last cacheline: 8 bytes */
};
```

PR #2725, opened 2026-02-27
New statx implementation

ccaffy:
DO NOT MERGE.
The PR is not ready yet, but you can review as I progress.


src/XrdSys/XrdSysStatx.hh
```diff
@@ -0,0 +1,142 @@
...
+  StatT2StatxT(st.st_atim, stx.stx_atime);
+  StatT2StatxT(st.st_mtim, stx.stx_mtime);
+  StatT2StatxT(st.st_ctim, stx.stx_ctime);
+  stx.stx_dev_major  = major(st.st_dev);
+  stx.stx_dev_minor  = minor(st.st_dev);
+  stx.stx_rdev_major = major(st.st_rdev);
+  stx.stx_rdev_minor = minor(st.st_rdev);
+#else
+  stx.stx_mask = STATX_BASIC_STATS;
+  stx.statx    = st;
```
abh3, line 131: Just to be a clear, a simple comment stating they are identical definitions.

abh3 (commented):
Looks good so far! I'll be looking forward to the rest of the code.

src/XrdSys/XrdSysStatx.hh
```diff
@@ -0,0 +1,142 @@
...
+#endif
+}
+
+inline void XrdSysStatxHelpers::Statx2Stat(const XrdSysStatx & stx, struct stat & st) {
+#ifdef HAVE_STATX
+  memset(&st, 0, sizeof(st));
+  st.st_blksize = stx.stx_blksize;
```
abh3: Technically, one should test each bit in stmask to see if the value in statx is actually present before converting. Alternatively, one could do a short-circuit test to see if basic stats are all present and then forge ahead; otherwise do an individual test, While that is technically correct, this code will work fine for all the current Linux supported filesystems. However, that may not be the case in the future. On the other hand, I think we will need to cross that bridge when it actually comes. A comment regarding this would likely be a good addition.

src/XrdSys/XrdSysStatx.hh
ccaffy: Yup :+1:

src/XrdSys/XrdSysStatx.hh
ccaffy: Done, indeed better prepare for the future. I also added a comment

src/XrdSfs/XrdSfsInterface.cc
```diff
@@ -44,9 +44,9 @@
 /*                              a u t o s t a t                               */
 /******************************************************************************/
 
-int XrdSfsDirectory::autoStat(struct stat *buf)
+int XrdSfsDirectory::autoStat(XrdSysStatx *buf, int mask)
```
abh3: Please don't do this. It will create havoc with plugin writers as they will have to actual;ly change their plugins to use R6. Simply add a new autostat() with the new signature with a default implementation that is based on the old autostat(). That way you don't have to change DigFS at all,. There really is no reason to modify DigFS as tit doesn't need to use the new stax() interface.

I'm sorry I didn't make that clear. We want to add new APIs to statx() enablement but keep all the old API's for backward compatibility. This allows plugin writers to simply recompile. They won't need to change actual source code if they don't want to.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -300,17 +302,22 @@ virtual int         close() = 0;
...
+//!
 //! @return If supported, SFS_OK should be returned. If not supported, then
 //!         SFS_ERROR should be returned with error.code set to ENOTSUP.
 //!
 //! @note: When autoStat() is in effect, directory entries that have been
 //!        deleted from the target directory are quietly skipped.
 //-----------------------------------------------------------------------------
 
-virtual int         autoStat(struct stat *buf);
+virtual int         autoStat(XrdSysStatx *buf, int mask = STATX_ALL);
```
abh3: Again, this should be kept and a new autstat() defined whose default implementation uses the old autostat().

src/XrdSsi/XrdSsiDir.cc
```diff
@@ -138,11 +138,11 @@ int XrdSsiDir::close()
 /*                              a u t o S t a t                               */
 /******************************************************************************/
 
-int XrdSsiDir::autoStat(struct stat *buf)
+int XrdSsiDir::autoStat(XrdSysStatx *buf, int mask)
```
abh3: This plugin does not need to change if autostat() is not changed. I think you should see that changing an existing API causes a cascade of changes that can be avoided. We want to avoid creating this cascade of unnecessary changes.

abh3 (changes requested):
So, the big problem here is that autostat() was changed and it did not need to change. The same idea of adding statx-specific API's to enable use of statx() should be extended to autostat() as well. This avoid a cascading set of changes and does not force plugin writers to make source code changes just to use R6. The new autostat() signature has a default implementation of using the old autostat API and converting the stat to a statx, This is the same philosophy we have for the new statx() API. Our plugins will, of course, override the default implementation to actually use the new statx(). So, no need to change DigFS nor Ssi.

abh3:
OK, now I see why autostat() was changed as opposed to adding a new autostat(). It's not that simple because it's "automatic". So, I propose the following solution:
1) We still add autostat(statx, uint32_t mask)
2) The default implementation returns ENOSYS (function not implemented)
3) I have added a new flag to kXR_dirllist = kXR_dstatx
4) When that flag is on dirlist first calsl the new autostat() if that is not implemented it reverts to using the old autostat()
5) When the flag is not set, it should use the old autostat()

Obviously, the logic in dirlist is a bit trickier to manage but it should be doable.

I think this will address he problem here and also negate the need for changing DigFS and Ssi.

ccaffy:
Thanks @abh3, very nice feedback. I wrote down on my notebook when we met at CERN that I had to modify this autoStat() function, while probably you meant creating a new one and calling the old one from it. Will do the proper changes.

src/XrdSfs/XrdSfsInterface.cc
```diff
@@ -50,7 +50,14 @@ int XrdSfsDirectory::autoStat(struct stat *buf)
    error.setErrInfo(ENOTSUP, "Not supported.");
    return SFS_ERROR;
 }
-  
+
+int XrdSfsDirectory::autoStat(XrdSysStatx *buf, int mask)
```
abh3: Let's follow the Linux statx() implementation and make the mask unsigned int everywhere.

abh3 (changes requested):
Minor change to follow the Linux API.

src/XrdSfs/XrdSfsInterface.cc
ccaffy: Absolutely!

src/XrdOfs/XrdOfs.cc
```diff
@@ -444,6 +444,15 @@ int XrdOfsDirectory::autoStat(struct stat *buf)
     if ((retc = dp->StatRet(buf))) return retc;
     return SFS_OK;
 }
+
+int XrdOfsDirectory::autoStat(XrdSysStatx * buf, unsigned int mask) {
+   struct stat tmpBuf;
+   int retc = autoStat(&tmpBuf);
```
abh3: This is not how autostat() works. It itself only registers the buffer, it does not fill in the buffer, That is done automatically when a directory entry is returned. So, this code is really not doing what it should be doing. It should do exactly what the old autostat() does and call the directory autostat() method for statx and return the result of the call. Simple as that. See XrdOfs.cc::444

src/XrdOss/XrdOss.hh
```diff
@@ -763,6 +764,30 @@ virtual int       Rename(const char *oPath, const char *nPath,
...
+  //! @param  mask   - The statx mask indicating which fields are requested.
+  //! @param  opts   - Options:
+  //!                  XRDOSS_preop    - this is a stat prior to open.
+  //!                  XRDOSS_resonly  - only look for resident files.
+  //!                  XRDOSS_updtatm  - update file access time.
+  //! @param  envP   - Pointer to environmental information.
+  //!
+  //! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+  //-----------------------------------------------------------------------------
+virtual int       Stat(const char *path, XrdSysStatx *buff, unsigned int mask = STATX_ALL, int opts=0, XrdOucEnv *envP=0) {
```
abh3: There can be no default for mask. Shouldwe not have STATX then XrdSysStatx is define as struct stat and the signature is identical to the method already defined under the old regime. The compiler will not allow this. The signatures must be distinguishable and they can only be so under all circumstances when mask is a required argument.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -312,6 +313,22 @@ virtual const char *FName() = 0;
...
+//! @param  mask   - The statx mask indicating which fields are requested.
+//!
+//! @return If supported, SFS_OK should be returned. If not supported, then
+//!         SFS_ERROR should be returned with error.code set to ENOTSUP.
+//!
+//! @note: When autoStat() is in effect, directory entries that have been
+//!        deleted from the target directory are quietly skipped.
+//-----------------------------------------------------------------------------
+
+virtual int         autoStat(XrdSysStatx *buf, unsigned int mask = STATX_ALL);
```
abh3: There can be no default for mask. Shouldwe not have STATX then XrdSysStatx is define as struct stat and the signature is identical to the method already defined under the old regime. The compiler will not allow this. The signatures must be distinguishable and it can only be so under all circumstances when mask is a required argument.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -750,6 +767,24 @@ virtual XrdSfsXferSize writev(XrdOucIOVec      *writeV,
...
+//-----------------------------------------------------------------------------
+//! Return state information on the file.
+//!
+//! @param  buf    - Pointer to the structure where info it to be returned.
+//! @param mask    - The statx mask indicating which fields are required
+//!
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, or SFS_STALL. When SFS_OK
+//!         is returned, buf must hold stat information.
+//-----------------------------------------------------------------------------
+virtual int            stat(XrdSysStatx *buf, unsigned int mask = STATX_ALL) {
```
abh3: There can be no default for mask. Shouldwe not have STATX then XrdSysStatx is define as struct stat and the signature is identical to the method already defined under the old regime. The compiler will not allow this. The signatures must be distinguishable and they can only be so under all circumstances when mask is a required argument.

src/XrdOss/XrdOss.hh
```diff
@@ -763,6 +764,30 @@ virtual int       Rename(const char *oPath, const char *nPath,
...
+  (void) mask;
+  struct stat statbuf;
+  int retc = Stat(path,&statbuf,opts,envP);
```
abh3: Please use the HAVE_STATX variable for optimization here. If defined then use your logic. If not defined you can simply code:
return Stat(path, buff, opts, envP)
because you know that XrdSysStatx is defined as struct stat and you can use that directly as a short circuit.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -750,6 +767,24 @@ virtual XrdSfsXferSize writev(XrdOucIOVec      *writeV,
...
+  struct stat statbuf;
+  int retc = stat(&statbuf);
```
abh3: Same here, use the HAVE_STATX macro to short circuit the call because if HAVE_STYATX is not defined you know that XrdSysStatx is defined as struct stat and the call can proceed in one line with no extra variables or calls involved.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -1282,6 +1317,34 @@ virtual int            stat(const char               *Name,
...
+//! @param  client - Client's identify (see common description).
+//! @param  opaque - path's CGI information (see common description).
+//!
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, SFS_STALL, or SFS_STARTED
+//!         When SFS_OK is returned, buf must contain stat information.
+//-----------------------------------------------------------------------------
+virtual int             stat(const char               *Name,
+                                   XrdSysStatx        *buf,
+                                   XrdOucErrInfo      &eInfo,
+                             unsigned int             mask = STATX_ALL,
```
abh3: There can be no default for mask as the signatures must be distinguishable even when XrdSysStatx is defined as struct stat.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -93,6 +93,30 @@ static int Statfd(int fd, struct stat *buf) {return  fstat(fd, buf);}
 
 static int Statfn(const char *fn, struct stat *buf) {return stat(fn, buf);}
 
+static int Statxfn(int dirFd,const char *path,int flags, unsigned int mask, XrdSysStatx * statxbuf) {
+#ifdef HAVE_STATX
+   return statx(dirFd,path,flags,mask,statxbuf);
+#endif
+   struct stat statbuf;
```
abh3: I think you meant to have an #else here otherwise it makes no sense as the codee would not be reachable.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -93,6 +93,30 @@ static int Statfd(int fd, struct stat *buf) {return  fstat(fd, buf);}
...
+   if (!retc) {
+      XrdSysStatxHelpers::Stat2Statx(statbuf,*statxbuf);
+   }
+   return retc;
+}
+
+static int Statxfd(int fd, unsigned int mask, XrdSysStatx * statxbuf) {
+#ifdef HAVE_STATX
+   return statx(fd,"",AT_EMPTY_PATH,mask,statxbuf);
+#endif
```
abh3: Same here there should be an #else.

src/XrdSsi/XrdSsiDir.cc
```diff
@@ -157,6 +157,15 @@ int XrdSsiDir::autoStat(struct stat *buf)
    return XrdSsiUtils::Emsg(epname, EBADF, epname, "???", error);
 }
 
+int XrdSsiDir::autoStat(XrdSysStatx *buf, unsigned int mask) {
```
abh3: I would not bother with providing statx support for the Ssi, it's not used for file access in most cases and when it is it's for limited purposes. Is there something I am missing or can you simply remove the changes to an Ssi component?

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -843,6 +846,24 @@ int XrdXrootdProtocol::StatGen(struct stat &buf, char *xxBuff, int xxLen,
        if (!(n = XrdOucUtils::GidName(buf.st_gid,xxBuff,xxLen,keepT))) return m;
       }
    xxBuff += n+1;
+   xxLen -= (n+1);
+
+#ifdef HAVE_STATX
```
abh3: So, we have a problem here, Notice that dirlist has \three extended options:  kXR_dstat. kXR_dstatx and  kXR_dcksm which affect the output. When kXR_dstat  you get the normal stat line just what statgen() provides. With kXR_dstatx, you get what dstat provides along with a CGI string for bdate. However, when dcksm is specified you get the dstat info *followed* by checksum info and then the CGI bdate string should dstatx be specified.  The checksum information in the form of "[algorithm:value]" such as [adler32:xxxxxx]. The code here prevents that order. In other words, statgen() should be left alone to generate the standard response, to dstat allowing the code to possibly add the checksum information and only then ass statx() information as cgi . This varies by who called statgen(). For regular stat calls no checksum info is returned but the bdata may be returned if asked for.

So, I suggest leave statgen as it is. Create a new statgenx() that generates the CGI bdate should it be called. Of course, I can see it may need a mask to selectively add other key-value pairs. For now we can keep it simple.

src/XrdXrootd/XrdXrootdProtocol.hh
```diff
@@ -341,7 +342,7 @@ static bool  ConfigFS(const char *path, XrdOucEnv &xEnv, const char *cfn);
 static bool  ConfigGStream(XrdOucEnv &myEnv, XrdOucEnv *urEnv);
 static bool  ConfigRedirPI(const char*, XrdOucEnv&, const char*, const char*);
 static int   Squash(char *);
-       int   StatGen(struct stat &buf, char *xxBuff, int xxLen, bool xa=false);
+       int   StatGen(XrdSysStatx &statxbuf, char *xxBuff, int xxLen, bool xa=false);
```
abh3: As mentioned previously, there is no reason to give statgen a statx struct as it will never be able to use it as such.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -809,7 +810,20 @@ int XrdXrootdProtocol::do_DirStat(XrdSfsDirectory *dp, char *pbuff,
 
 // We always return stat information, see if we can use autostat
 //
-   manStat = (dp->autoStat(&Stat) != SFS_OK);
+   if (Request.dirlist.options[0] & kXR_dstatx) {
```
abh3: As I stated in the comment for statgen() this part will need to be somewhat rewritten to accommodate the additional statgenx().

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -1719,7 +1736,10 @@ int XrdXrootdProtocol::do_Open()
 // If client wants a stat in open, return the stat information
 //
    if (retStat)
-      {retStat = StatGen(statbuf, ebuff, sizeof(ebuff));
+      {
+       XrdSysStatxHelpers::Stat2Statx(statbuf,statxBuf);
```
abh3: Again, this part will need to be revised as per keeping the old statgen() and adding a statgens().

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -3095,8 +3116,12 @@ int XrdXrootdProtocol::do_Stat()
                               "stat does not refer to an open file");
        rc = fp->XrdSfsp->stat(&buf);
        TRACEP(FS, "fh=" <<fh.handle <<" stat rc=" <<rc);
-       if (SFS_OK == rc) return Response.Send(xxBuff,
-                                StatGen(buf,xxBuff,sizeof(xxBuff)));
+       if (SFS_OK == rc) {
+          //TODO statx call above?
```
abh3: Noted.

abh3 (changes requested):
We are converging. Yes, some changes will be needed. There are also parts that I think were not fully addressed because, I admit, the existing code is not clear on what the format of a stat response should look like. That's just historical baggage which as some point should be corrected but nor now.

src/XrdOfs/XrdOfs.cc
ccaffy: Mmmh but then the problem is that it actually sets a pointer and not a copy of the stat buffer. So here that would mean that if I create a stat buffer from that function, it will be destroyed when out of scope and the pointer will be invalid.

src/XrdOss/XrdOss.hh
ccaffy: Yes, very good point. Done

src/XrdSfs/XrdSfsInterface.hh
ccaffy: Yes, very good point. Done

src/XrdSfs/XrdSfsInterface.hh
ccaffy: Yes, done.

src/XrdSsi/XrdSsiDir.cc
ccaffy: OK yes, happy to remove that change to the Ssi component

src/XrdOfs/XrdOfs.cc
abh3: Correct, that's why you don't do it and return ENOSYS as he caller would automatically revert to using regular autostat(). The sequence is:
1) If needed: call autostat_new()
2) if returned ENOSYS call autostat_old()
3) ...
Yes, the caller has to remember which stat buffer is actually being used but that is easy. The dirlist function has to also know which stat call to make depending on which buffer it is using. A bit harder but not that much.

abh3: I was afraid of that. It likely works in gcc but not in clang. Some compilers are clever enough to know what type is actually being used and some insist that the type you defined is unique even when it is based on another well defined type. Looks like Clang is of the "advanced" variety.
Solution #1: change the typedef to a using statiment. So you would one of:
using XrdSysStatx = struct statx;
or 
using XrdSysStatx = struct stat;

If Clang still complains (which it really should not) then the brute force solution will need to be adopted:
#2: Simply make XrdSysStatx a define symbol. However, in order to do that, you will need to use a unique name. Since XrdSysStatx is already in use for other things, let's call the "stat" buffer XrdSysStatBuf or choose a different name but make it reasonably descriptive.

src/XrdOss/XrdOss.hh
```diff
@@ -763,6 +764,34 @@ virtual int       Rename(const char *oPath, const char *nPath,
...
+  if (!retc) {
+    XrdSysStatxHelpers::Stat2Statx(statbuf,*buff);
+  }
+  return retc;
+#else
+  return Stat(path,buff,opts,envP);
+#endif
```
abh3: See my comment earlier on how to fix this.

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -744,7 +744,7 @@ void XrdXrootdProtocol::Recycle(XrdLink *lp, int csec, const char *reason)
 /*                               S t a t G e n                                */
 /******************************************************************************/
   
-int XrdXrootdProtocol::StatGen(struct stat &buf, char *xxBuff, int xxLen,
+int XrdXrootdProtocol::StatGen(XrdSysStatx &statxBuf, char *xxBuff, int xxLen,
```
abh3: I thought that in order to keep the correct stat response order we would keep the old stat (though we could change it to use a statx buffer as input) and add a new genstatx() that is responsible for adding the new cgi encoded info. I don't see that reflected here. In fact, I don't much about statgen. o, what's the story with that?

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -809,7 +810,20 @@ int XrdXrootdProtocol::do_DirStat(XrdSfsDirectory *dp, char *pbuff,
 
 // We always return stat information, see if we can use autostat
 //
-   manStat = (dp->autoStat(&Stat) != SFS_OK);
+   if (Request.dirlist.options[0] & kXR_dstatx) {
+      // A statx call will be attempted
+      int retc = dp->autoStat(&Statx,STATX_ALL);
+      if (retc == SFS_ERROR && dp->error.getErrInfo() == ENOTSUP) {
```
abh3: This assumes that autostat() can only fail with ENOSUP (though I used ENPOSYS earlier - no matter). That's a dangerous assumption. Any error should cause the code to try using the old autostat. So, kill the && clause.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -809,7 +810,20 @@ int XrdXrootdProtocol::do_DirStat(XrdSfsDirectory *dp, char *pbuff,
...
+         //Statx call is not supported, we will do a "stat" autoStat call
+         manStat = (dp->autoStat(&Stat) != SFS_OK);
+      } else {
```
abh3: I think you will need to remember which type of buffer you are using to avoid undefined behaviour later on.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -826,6 +840,7 @@ int XrdXrootdProtocol::do_DirStat(XrdSfsDirectory *dp, char *pbuff,
 // client to issue individual stat requests in that case.
 //
    memset(&Stat, 0, sizeof(Stat));
+   memset(&Statx,0,sizeof(Statx));
```
abh3: Please move this into the clause relevant to which buffer is being used. You don't need to clear both buffers as only one will be active.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -843,15 +858,16 @@ int XrdXrootdProtocol::do_DirStat(XrdSfsDirectory *dp, char *pbuff,
...
                     if (rc == SFS_ERROR && myError.getErrInfo() == ENOENT)
                        {dname = 0; continue;}
                     if (rc != SFS_OK)
                        return fsError(rc, XROOTD_MON_STAT, myError,
                                           argp->buff, opaque);
                    }
                 strcpy(buff, dname); buff += dlen; *buff = '\n'; buff++; cnt++;
-                dlen = StatGen(Stat, buff, sizeof(XB.epad));
+               //TODO: statx call above?
+                dlen = StatGen(Statx, buff, sizeof(XB.epad));
```
abh3: So, this seems to be wrong. You may be using the stat buffer or the statx bufferm, depending on what actually worked. But here the assumption is that you are using the statx buffer. That''s only true if you did a manual statx in the preceding if clause. Otherwise, you are using stat or statx . This seems awfully like what Claude would do....

Further, I don't see how this is keepingthe old response structure in tact:
<old stat info> [<checksum info>] [<new stat cgi string>]
Am I missing something?

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -1719,7 +1736,10 @@ int XrdXrootdProtocol::do_Open()
 // If client wants a stat in open, return the stat information
 //
    if (retStat)
-      {retStat = StatGen(statbuf, ebuff, sizeof(ebuff));
+      {
+       XrdSysStatxHelpers::Stat2Statx(statbuf,statxBuf);
+      //TODO: statx call above?
+       retStat = StatGen(statxBuf, ebuff, sizeof(ebuff));
```
abh3: This is problematic and I forgot about this issue. We can't just return a new stat response unless if the client isn't expecting one., So, I will need to add an option to open() to enable the return of additional information. Onlythen you can add the extra statx information. I will do so. The flag will be set in the optt variable and will be called kXR_retstatx.

abh3 (changes requested):
Some additional changes plus a new option for kXR_open that I should have remembered - sorry about hat. Anyway, I am still confused on use of stagen() and how correct response order is maintained.

src/XrdSfs/XrdSfsInterface.hh
ccaffy: Done

src/XrdOfs/XrdOfs.cc
ccaffy: I'm trying this out, let's see with solution 1.

src/XrdXrootd/XrdXrootdProtocol.cc
ccaffy: Not sure I understood then. How would the call to StatxGen() look like in do_DirStat(), do_Open() and do_Stat()?

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: One tries to do a statx call, but if there's an error and the error is ENOTSUP, then we fallback to the regular autoStat. No?

abh3: Yes, you do the fallback. My point is that it shouldn't matter what the error code is. If it failed for whatever reason you try the fallback. Notice that the existing code uses the manual method (one stat call at a time) should the fallback fail.

src/XrdXrootd/XrdXrootdProtocol.cc
ccaffy: Should be better now, feedback's welcome :)

src/XrdXrootd/XrdXrootdProtocol.cc
ccaffy: OK I modified the code accordingly, your feedback is welcomed :)

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: Yes, good point, should be good now

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: Yes, done!

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: No no, I pushed it a bit too fast to come back to this later. Now should look much better :crossed_fingers:

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: OK I modified the do_Open() to take this into consideration. I had to modify the constructor of XrdXrootdFile to take a XrdSysStatx object in parameter. Feedback is welcomed too :)

src/XrdOss/XrdOss.hh
```diff
@@ -763,6 +764,34 @@ virtual int       Rename(const char *oPath, const char *nPath,
...
+  //! @param  opts   - Options:
+  //!                  XRDOSS_preop    - this is a stat prior to open.
+  //!                  XRDOSS_resonly  - only look for resident files.
+  //!                  XRDOSS_updtatm  - update file access time.
+  //! @param  envP   - Pointer to environmental information.
+  //!
+  //! @return 0 upon success or -errno or -osserr (see XrdOssError.hh).
+  //-----------------------------------------------------------------------------
+virtual int       Stat(const char *path, XrdSysStatx *buff, unsigned int mask, int opts=0, XrdOucEnv *envP=0) {
+#ifdef HAVE_STATX
```
abh3: Please do not inline if-def code in header files. This should be moved to the cc file. In fact, simply remember this is a rule. If you see a violation, please call it our so we can correct it.

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -750,6 +767,29 @@ virtual XrdSfsXferSize writev(XrdOucIOVec      *writeV,
...
+//!
+//! @param  buf    - Pointer to the structure where info it to be returned.
+//! @param mask    - The statx mask indicating which fields are required
+//!
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, or SFS_STALL. When SFS_OK
+//!         is returned, buf must hold stat information.
+//-----------------------------------------------------------------------------
+virtual int            stat(XrdSysStatx *buf, unsigned int mask) {
+  (void)mask;
+#ifdef HAVE_STATX
```
abh3: I am confused. Has this been reverted? I thought I already pointed out that we never include if-def code in header files. I also recall thi was moved the the cc file. So, why is this appearing again?

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -1282,6 +1322,34 @@ virtual int            stat(const char               *Name,
...
+//! @return One of SFS_OK, SFS_ERROR, SFS_REDIRECT, SFS_STALL, or SFS_STARTED
+//!         When SFS_OK is returned, buf must contain stat information.
+//-----------------------------------------------------------------------------
+virtual int             stat(const char               *Name,
+                                   XrdSysStatx        *buf,
+                                   XrdOucErrInfo      &eInfo,
+                             unsigned int             mask,
+                             const XrdSecEntity       *client = 0,
+                             const char               *opaque = 0) {
+  (void)mask;
```
abh3: Move to CC file.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -93,6 +93,22 @@ static int Statfd(int fd, struct stat *buf) {return  fstat(fd, buf);}
 
 static int Statfn(const char *fn, struct stat *buf) {return stat(fn, buf);}
 
+static int Statxfn(int dirFd,const char *path,int flags, unsigned int mask, XrdSysStatx * statxbuf) {
```
abh3: OK, I am skeptical here. I know that stax() is an all-inluis=ve beas but I don't think you can force that on the existing xrootd code. There is no way, at the moment, you can supply an FD to a directory that makes sense. We simply don't support "at" directory operations at this level. So, this SPI is confusing and misleading. If the staxfn is to mimic statfn() then delete the fd and flags. Internally supply the correct flags (the fd will be ignored because the flags say use the path).

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -573,6 +589,21 @@ int XrdSfsNativeFile::stat(struct stat     *buf)         // Out
    return SFS_OK;
 }
 
+int XrdSfsNativeFile::stat(XrdSysStatx * buf,unsigned int mask) {
+#ifdef HAVE_STATX
+   // Execute the function
+   //
+   static const char *epname = "stat";
+   if (XrdSfsUFS::Statxfd(oh, mask, buf))
```
abh3: I didn't see this added to XrdSfsUFS in the same file. I presume it must be there as this would not have compiled.

src/XrdSfs/XrdSfsNative.cc
```diff
@@ -971,6 +1002,41 @@ int XrdSfsNative::stat(const char              *path,        // In
...
+            info        - Opaque information, if any.
+
+  Output:   Returns SFS_OK upon success and SFS_ERROR upon failure.
+*/
+{
+   static const char *epname = "statx";
+
+   // Execute the function
+   //
+   if (XrdSfsUFS::Statxfn(0,path,AT_FDCWD,mask, buf))
```
abh3: I don't think this is correct, looks like Claude must have made a mistake (do you think you can stop using Claude, please?). Looks at the man poage,. This is incorrect and will not work.

src/XrdXrootd/XrdXrootdFile.cc
```diff
@@ -116,8 +116,8 @@ XrdXrootdFile::XrdXrootdFile(const char *id, const char *path, XrdSfsFile *fp,
 //
    if (sP || !isMMapped)
       {if (!sP) sP = &buf;
-       fp->stat(sP);
-       if (!isMMapped) Stats.fSize = static_cast<long long>(sP->st_size);
+       fp->stat(sP,STATX_ALL);
```
abh3: Ummm, no this is why we have a selection in stat in the protocol. If you want btme you must ask for it. Please do not always ask for STATX_ALL when people do not want the extra information. This has the potential of killing performance in distributed filesystems.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -1723,19 +1769,26 @@ int XrdXrootdProtocol::do_Open()
 
 // If client wants a stat in open, return the stat information
 //
-   if (retStat)
-      {retStat = StatGen(statbuf, ebuff, sizeof(ebuff));
+   if (retStat) {
```
abh3: Recakk, we added a flags to open that indicates whether you want to get stat or stax information. I don't see that reflected here. The particular flag to test is optt & kXR_retstatx. The retstat flag will always be set if kXR_retstat or kXR_retstatx is set.

abh3 (changes requested):
Slowly converging.

src/XrdOss/XrdOss.hh
ccaffy: Yes, done :+1:

src/XrdSfs/XrdSfsInterface.hh
ccaffy: Yes, done :+1:

src/XrdSfs/XrdSfsNative.cc
ccaffy: Yes it is there :
```cpp
static int Statxfd(int fd, unsigned int mask, XrdSysStatx * statxbuf) {
#ifdef HAVE_STATX
   return statx(fd,"",AT_EMPTY_PATH,mask,statxbuf);
#else
   return Statfd(fd,&statxbuf->statx);
#endif
}
```

src/XrdSfs/XrdSfsNative.cc
ccaffy: It is not claude-generated, I wrongly read the doc ! Will replace by 0.

src/XrdXrootd/XrdXrootdFile.cc
ccaffy: In the open() we did not add the flag to ask what people want right? I can see the optt & kXR_retstatx, but there's no flag to specify what user wants?

The wants integer was added for the kXR_stat request but not for open.
```
enum XOpenRequestOption {
   kXR_compress = 0x0001, //     1   // also locate (return unique hosts)
   kXR_delete   = 0x0002, //     2
   kXR_force    = 0x0004, //     4
   kXR_new      = 0x0008, //     8
   kXR_open_read= 0x0010, //    16
   kXR_open_updt= 0x0020, //    32
   kXR_async    = 0x0040, //    64
   kXR_refresh  = 0x0080, //   128   // also locate
   kXR_mkpath   = 0x0100, //   256
   kXR_prefname = 0x0100, //   256   // only locate
   kXR_open_apnd= 0x0200, //   512
   kXR_retstat  = 0x0400, //  1024
   kXR_4dirlist = 0x0400, //  1024   // for locate intending a dirlist
   kXR_replica  = 0x0800, //  2048
   kXR_posc     = 0x1000, //  4096
   kXR_nowait   = 0x2000, //  8192   // also locate
   kXR_seqio    = 0x4000, // 16384
   kXR_open_wrto= 0x8000  // 32768
};

enum XOpenRequestOption2 {// Set in optiont
   kXR_dup      = 0x0001, //     1
   kXR_samefs   = 0x0002, //     2
   kXR_retstatx = 0x0004, //     4   // open: return extended stat -> retstat
   kXR_directio = 0x0008  //     8
};

struct ClientOpenRequest {
   kXR_char  streamid[2];
   kXR_unt16 requestid;
   kXR_unt16 mode;
   kXR_unt16 options;
   kXR_unt16 optiont;
   kXR_char  reserved[6];
   kXR_char  fhtemplt[4];
   kXR_int32  dlen;
};
```

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: Yes it is reflected few lines below that one
```cpp
if (retStat) {
       struct stat buf;
       XrdSysStatxHelpers::Statx2Stat(statbuf,buf);
       retStat = StatGen(buf, ebuff, sizeof(ebuff));
       if (optt & kXR_retstatx) {
          int sxLen = StatxGen(statbuf, ebuff+(retStat-1),
                                sizeof(ebuff)-(retStat-1));
          if (sxLen > 0) retStat += (sxLen - 1);
       }
       IOResp[1].iov_base = (char *)&myResp; IOResp[1].iov_len = sizeof(myResp);
       IOResp[2].iov_base =         ebuff;   IOResp[2].iov_len = retStat;
       resplen = sizeof(myResp) + retStat;
   }
```

src/XrdSfs/XrdSfsNative.cc
ccaffy: OK that makes complete sense. I modified accordingly ! Indeed I just mapped the statx(...) definition to XrootD like it was done for Statfn. But I definitely agreed and just performed the changes.

src/XrdSfs/XrdSfsNative.hh
```diff
@@ -228,6 +230,13 @@ const   char          *getVersion();
                         return rc;
                        }
 
+      int            stat(const char              *path,
+                           XrdSysStatx       *buf,
+                           XrdOucErrInfo     &error,
+                     unsigned int            mask,
+                     const XrdSecClientName  *client,
+                     const char              *info);
```
amadio: The formatting looks odd here. It's probably mixing tabs and spaces.

src/XrdXrootd/XrdXrootdFile.cc
amadio: From the man page, `STATX_ALL` only adds `BTIME` to `STATX_BASIC_STATS`:
```
        STATX_TYPE             Want stx_mode & S_IFMT
        STATX_MODE             Want stx_mode & ~S_IFMT
        STATX_NLINK            Want stx_nlink
        STATX_UID              Want stx_uid
        STATX_GID              Want stx_gid
        STATX_ATIME            Want stx_atime
        STATX_MTIME            Want stx_mtime
        STATX_CTIME            Want stx_ctime
        STATX_INO              Want stx_ino
        STATX_SIZE             Want stx_size
        STATX_BLOCKS           Want stx_blocks
        STATX_BASIC_STATS      [All of the above]
        STATX_BTIME            Want stx_btime
        STATX_ALL              The same as STATX_BASIC_STATS | STATX_BTIME.
```
So, we can either leave this as it was (i.e. using regular stat instead of statx), use `STATX_BASIC_STATS` here, although why would we change it in that case, or use `STATX_ALL`, which adds birth time. I'd suggest to leave as is or add a separate constructor that takes also the mask to use with `statx`.

src/XrdXrootd/XrdXrootdProtocol.cc
```diff
@@ -850,6 +850,23 @@ int XrdXrootdProtocol::StatGen(struct stat &buf, char *xxBuff, int xxLen,
...
 }
 
+// Returns the byte count including the \0 terminator (same convention as
+// StatGen), or 0 if nothing was written.
+//
+int XrdXrootdProtocol::StatxGen(XrdSysStatx &buf, char *xxBuff, int xxLen) {
+  int n = 0;
+#ifdef HAVE_STATX
+  if (buf.stx_mask & STATX_BTIME) {
+    n = snprintf(xxBuff, xxLen, " \nbtime_s=%lld&btime_n=%lld",
```
amadio: Is the leading space really intentional? Are we sure that `xxBuff` is never `nullptr`?

src/XrdXrootd/XrdXrootdProtocol.hh
```diff
@@ -47,6 +47,7 @@
 #include "XrdXrootd/XrdXrootdReqID.hh"
 #include "XrdXrootd/XrdXrootdResponse.hh"
 #include "XProtocol/XProtocol.hh"
+#include "XrdSys/XrdSysStatx.hh"
```
amadio: You can just forward declare `XrdSysStatx`, there's no need to include the header, I think.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -1723,19 +1769,26 @@ int XrdXrootdProtocol::do_Open()
...
+       struct stat buf;
+       XrdSysStatxHelpers::Statx2Stat(statbuf,buf);
```
amadio: I'm confused, shouldn't this come after `StatGen(...)` below?

amadio (commented):
Looks like we're getting there!

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: No no, Statgen() just prepare the string returned to the client after the stat

src/XrdXrootd/XrdXrootdProtocol.hh
ccaffy: I don't think I can ;) One is doing `using` and the other one is a struct ;)

src/XrdSfs/XrdSfsNative.hh
ccaffy: Yes :/

src/XrdXrootd/XrdXrootdProtocol.cc
ccaffy: Well spotted, it was relevant in the code when it was in StatGen(), but not relevant in StatxGen as StatGen() automatically adds the ' ' at the end...
Will modify!

src/XrdXrootd/XrdXrootdProtocol.hh
amadio: Yes, agreed, keep it as is.

src/XrdCl/XrdClFileSystem.hh
```diff
@@ -534,11 +534,13 @@ namespace XrdCl
...
       //! @param timeout timeout value, if 0 the environment default will
       //!                be used
+      //! @param wants   statx stx_mask e.g: STATX_BASIC_STATS | STATX_BTIME.
       //! @return        status of the operation
       //------------------------------------------------------------------------
       XRootDStatus Stat( const std::string &path,
                          ResponseHandler   *handler,
-                         time_t             timeout = 0 )
+                         time_t             timeout = 0,
+                         uint32_t           wants = 0)
```
amadio: Handler and timeout must always be the last options. It's like this in a lot of interfaces already, we don't want to have different things depending on the operation.

ccaffy: I saw this with Andy when he was at CERN and he told me to put it after timeout ;)

src/XrdXrootd/XrdXrootdProtocol.cc
ccaffy: No actually, this space is needed! The client-side parses the server response by splitting with the space character. I've added this back.

src/XrdCl/XrdClFile.cc
```diff
@@ -235,23 +235,26 @@ namespace XrdCl
   //----------------------------------------------------------------------------
   XRootDStatus File::Stat( bool             force,
                            ResponseHandler *handler,
-                           time_t           timeout )
+                           time_t           timeout ,
+                           uint32_t         wants)
   {
     if( pPlugIn )
+      // should the plugin have the wants mask as well?
```
abh3: So, here we have a conundrum. We can't just give the mask to the plugin because existing plugins won't match that signature. Furthermore, we can't really require that plugin writers retool all there plugins to provide a methods that accepts a mask. The only solution here is as follows:
1) in XrdClPlugInInterface.hh define a new method that accepts a mask before response handler.
2) The new method should, as a default implementation, simply return the results of a call the existing stat() without the mask.

There are other considerations here explained in the "hh" comment.

src/XrdCl/XrdClFile.hh
```diff
@@ -217,11 +218,13 @@ namespace XrdCl
...
       //! @param timeout timeout value, if 0 the environment default will
       //!                be used
+      //! @param wants   statx stx_mask e.g: STATX_BASIC_STATS | STATX_BTIME.
       //! @return        status of the operation
       //------------------------------------------------------------------------
       XRootDStatus Stat( bool             force,
                          ResponseHandler *handler,
-                         time_t           timeout = 0 )
+                         time_t           timeout = 0,
+                         uint32_t wants = 0)
```
abh3: So, while I said this is an acceptable strategy (adding mak at the end), as @amadio points out it is a lousy one. After additional thought, I do agree with him, even of that means additional methods to make mask a first class parameter. See comments later where @amadio explains why putting mask at the end is not good). So, unfortunately, new methods where mask is mandatory because it appears before ResponseHandler and/or timeout.

src/XrdCl/XrdClFileSystem.cc
```diff
@@ -1545,9 +1545,11 @@ namespace XrdCl
   //----------------------------------------------------------------------------
   XRootDStatus FileSystem::Stat( const std::string &path,
                                  ResponseHandler   *handler,
-                                 time_t             timeout )
+                                 time_t             timeout,
+                                 uint32_t           wants)
   {
     if( pPlugIn )
+      // Should the plugin have the wants mask as well?
```
abh3: This suffers the same problem as the ClFile stat(). The solution will be pretty much identical.

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,7 +931,42 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
...
+       if (Xunix.Stat(path, &sbuf) != 0)
+          return -1;
+       XrdSysStatxHelpers::Stat2Statx(sbuf, *stx);
+       return 0;
+      }
+#endif
+
+// Return the results of our version of statx()
+//
+   return Xroot.Statx(myPath, mask, stx);
```
abh3: You could, of course, simply default to calling Xroot.Stat() and converting the result to a statx structure. For Posix that is totally acceptable. It would also shield you from any statx aberrations father down the line. Something to consider.

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +248,63 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
...
+   XrdCl::XRootDStatus xStatus;
+   XrdCl::StatInfo    *sInfo = 0;
+
+// Make sure admin is ok
+//
+   if (!isOK()) return false;
+
+// Issue the stat with wants mask and verify that all went well
+//
+   xStatus = Xrd.Stat(Url.GetPathWithParams(), sInfo, 0, mask);
```
abh3: Remember that the mask will move here so this will need to change.

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +248,63 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
...
+   if (!xStatus.IsOK()) {
+      XrdPosixMap::Result(xStatus, ecMsg);
+      delete sInfo;
+      return false;
+   }
+
+// First fill a stat structure, then convert to statx
+//
+   struct stat buf;
+   XrdPosixConfig::initStat(&buf);
```
abh3: I don't understand, you will be returning a statx version, why not convert here directly to the stax buffer. I don't see the benefit of doing a stat initialization followed by a stat to stax conversion., I don't think you need that extra step here.

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +248,63 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
    delete sInfo;
    return true;
 }
+
+/******************************************************************************/
+/*                                S t a t x                                   */
+/******************************************************************************/
+
+bool XrdPosixAdmin::Statx(unsigned int mask, XrdSysStatx &stx)
```
abh3: So, this all works because Posix is only defined for Linux and its derivatives. Posix does not work on MacOS. Keep that in mind as you can simplify the steps here, as noted below.

src/XrdPosix/XrdPosixCache.cc
```diff
@@ -89,6 +90,17 @@ int XrdPosixCache::Rename(const char* oldPath, const char* newPath)
 int XrdPosixCache::Stat(const char *path, struct stat &sbuff)
                        {return theCache->Stat(path, sbuff);}
 
+/******************************************************************************/
+
+int XrdPosixCache::Stat(const char *url, XrdSysStatx &stx, unsigned int mask)
```
abh3: I think you can dispense with this. The PosixCache can be considered to not support statx and it should just use the default implementation well in advance of this call. There is no reason for this code to support statx.

src/XrdPosix/XrdPosixFile.cc
```diff
@@ -744,14 +744,14 @@ void XrdPosixFile::ReadV(XrdOucCacheIOCB &iocb, const XrdOucIOVec *readV, int n)
...
-bool XrdPosixFile::Stat(XrdCl::XRootDStatus &Status, bool force)
+bool XrdPosixFile::Stat(XrdCl::XRootDStatus &Status, bool force, uint32_t wants)
 {
    XrdCl::StatInfo *sInfo = 0;
 
 // Get the stat information from the open file
 //
    Ref();
-   Status = clFile.Stat(force, sInfo);
+   Status = clFile.Stat(force, sInfo,0,wants);
```
abh3: This will like need to change as per @amadio.

src/XrdPosix/XrdPosixLinkage.cc
```diff
@@ -173,8 +173,10 @@ XrdPosixLinkage Xunix;
                          {       Xunix.Load_Error("rewinddir"); abort();}
       Retv_Rmdir       Xrd_U_Rmdir(Args_Rmdir)
                          {return (Retv_Rmdir)Xunix.Load_Error("rmdir");}
-      Retv_Seekdir     Xrd_U_Seekdir(Args_Seekdir)
+      Retv_Seekdir     Xrd_U_Seekdir(Args_Seekdir) //@Andy, should I fallback to normal stat? I guess no but prefer to ask
```
abh3: I am confused. The comment is for seekdir which has nothing to do with stat. Are you taking about the subsequent statx? If so, what you have for statx if sine, the implementation, of course, could be greatly simplified. But that sort of begs the question once you have the whole statx infrastructure implemented.

src/XrdSsi/XrdSsiDir.hh
```diff
@@ -50,7 +50,6 @@ inline  void        copyError(XrdOucErrInfo &einfo) {einfo = error;}
 const   char       *FName();
 
         int         autoStat(struct stat *buf);
-
```
abh3, line 53: I suppose, but why eliminate the space? It just clutters the pull request.

src/XrdXrootd/XrdXrootdFile.cc
abh3: Sort of, my point was instead of saying STATX_ALL simply add btime to the basic stat. I think that is what @amadio is saying. Yes, open does not have a "want" field. The retstatx is meant to return the most basic extra information. On the other hand, I am tempted to remove the retstatx and just say we never return statx information on an open call. If you want the extra stuff, follow it up with an additional statx call -- I doubt anyone will. So, that is a suggested alternative here.

src/XrdXrootd/XrdXrootdProtocol.cc
abh3: Good tach. I think that is why t was specified in the original spec.

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: Yes, done above

src/XrdCl/XrdClFile.cc
```diff
@@ -126,7 +126,8 @@ namespace XrdCl
                            OpenFlags::Flags   flags,
                            Access::Mode       mode,
                            ResponseHandler   *handler,
-                           time_t             timeout )
+                           time_t             timeout,
+                           uint32_t           wants)
```
amadio: I really think this should come before `*handler` and not have a default. Every other interface has `*handler` and `timeout` as the last parameters, this should not be an exception. It should always be completely unabiguous in calling code which `Stat()` or `Open()` is being invoked, and with the default for `wants` that's not the case, which may lead to bugs like the one fixed in 8e7fd73f297756bdb569659b84aadaf0b8377a02.

src/XrdCl/XrdClFile.cc
```diff
@@ -235,23 +236,26 @@ namespace XrdCl
...
   }
 
   //----------------------------------------------------------------------------
   // Obtain status information for this file - sync
   //----------------------------------------------------------------------------
   XRootDStatus File::Stat( bool       force,
                            StatInfo *&response,
-                           time_t     timeout )
+                           time_t     timeout ,
+                           uint32_t   wants)
```
amadio: To make this closer to `statx` itself, which has the signature
```
       int statx(int dirfd, const char *_Nullable restrict path,
                 int flags, unsigned int mask, struct statx *restrict statxbuf);
```
I'd make `wants` an `unsigned int` instead of `uint32_t`, and consider renaming it to `mask`, just like in the original syscall signature.

src/XrdCl/XrdClXRootDResponses.cc
```diff
@@ -202,6 +205,44 @@ namespace XrdCl
...
+          {
+            // We only take the birthtime in seconds to match what is done with access time and the rest
+            // if we need the nanoseconds for birth time, one can extract btime_n.
+            pBirthTime = ::strtoll( extraAttrs["btime_s"].c_str(), &result, 0 );
+            if( *result != 0 )
+            {
+              pBirthTime = 0;
+              return false;
+            }
+            pHasBirthTime = true;
```
amadio: This looks less efficient than it could be. If we're only going to set `pBirthTime`, can we try not to go via a `std::map` to do it?

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,7 +931,55 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
...
+}
+
+int XrdPosix_Statx(int dirfd, const char *path, int flags,
+                   unsigned int mask, XrdSysStatx *stx)
+{
+   char *myPath, buff[2048];
+
+// Make sure a path was passed
+//
+   if (!path) {errno = EFAULT; return -1;}
```
amadio: For when it's called on a local path using `path == nullptr` and valid `dirfd`, this looks like it will error out unnecessarily.

amadio: From "Invoking statx" section of the man page:
```
       By file descriptor
              If  path is an empty string (or NULL since Linux 6.11) and the AT_EMPTY_PATH flag is specified in flags (see below), then the target file is the one referred to by the file descriptor
              dirfd.
```

src/XrdPosix/XrdPosixAdmin.cc
```diff
@@ -246,3 +254,69 @@ bool XrdPosixAdmin::Stat(struct stat &Stat)
...
+      delete sInfo;
+      return false;
+   }
+
+// First fill a stat structure, then convert to statx
+//
+   struct stat buf;
+   XrdPosixConfig::initStat(&buf);
+   buf.st_size   = static_cast<size_t>(sInfo->GetSize());
+   buf.st_blocks = buf.st_size/512 + buf.st_size%512;
```
amadio: Is this really correct? It seems wrong to hard-code 512 here.

src/XrdCl/XrdClFile.cc
ccaffy: @abh3, I followed your instruction on this one, any idea?

src/XrdCl/XrdClXRootDResponses.cc
ccaffy: This was for the future, it was in case we wanted to put more things later on, the code would have already be there. I can definitely change it then.

src/XrdPosix/XrdPosix.cc
ccaffy: yes good catch, I thought that there would be no way one could call Linux stat without a path. I can remove that check :+1:

src/XrdPosix/XrdPosixAdmin.cc
ccaffy: I took it from https://github.com/xrootd/xrootd/blob/master/src/XrdPosix/XrdPosixFile.cc#L435 (XrdPosixFile::Fstat(...))

src/XrdCl/XrdClFile.cc
ccaffy: I think the name came from the fact that it will set a kXR_wants flag. @abh3 gave me this name while discussing, happy to change it however.

In my notes from my meeting with Andy, I wrote to use uint32_t for the wants flag. Again, happy to change it to unsigned int, but before doing so, there may have been a rationale I am probably not aware of?

src/XrdCl/XrdClXRootDResponses.cc
ccaffy: If we only use this for the birthtime and is going to be used only for this, the map is probably over kill. What do you think @abh3 ?

src/XrdPosix/XrdPosixAdmin.cc
abh3: Correct. The stat() call returns blocks in 512 byte units. See https://man7.org/linux/man-pages/man3/stat.3type.html

src/XrdCl/XrdClXRootDResponses.cc
abh3: Indeed, it is overkill. However, the way we designed it, we can now returns a lot more "other" information which would make sense to use a map. So, I have no problems with overkill to bite the bullet now rather than later.

amadio:
> It's on my list. Hopefully it is squashed to be one commit for the full implementation is viewable.

Yes, it's already squashed.

amadio:
LGTM. I tested locally:
```sh
gentoo build $ env LD_LIBRARY_PATH=$PWD/lib LD_PRELOAD=libXrdPosixPreload.so ls root://gentoo.cern.ch//
cluster  dd.dat  file.png  fio    gentoo    hugo  hugo.txt    mkdir_test1  new.txt  newdir  python  server  test  test.txt  tmp  tokens  toremove.txt  tpc1  tpc2  xcache
gentoo build $ env LD_LIBRARY_PATH=$PWD/lib LD_PRELOAD=libXrdPosixPreload.so ls -F root://gentoo.cern.ch//
cluster/  dd.dat  file.png  fio/  gentoo/  hugo/  hugo.txt  mkdir_test1/  new.txt  newdir/  python/  server/  test/  test.txt  tmp/  tokens/  toremove.txt  tpc1/  tpc2/  xcache/
gentoo build $ env LD_LIBRARY_PATH=$PWD/lib LD_PRELOAD=libXrdPosixPreload.so ls -F root://gentoo.cern.ch//file.png
root://gentoo.cern.ch//file.png
gentoo build $ env LD_LIBRARY_PATH=$PWD/lib LD_PRELOAD=libXrdPosixPreload.so ls -l root://gentoo.cern.ch//file.png
-rw-rw-r--? 1 amadio zh 126582 Jan 23  2024 root://gentoo.cern.ch//file.png
gentoo build $ ls -l /srv/xrootd/file.png 
-rw-rw-r-- 1 xrootd xrootd 126582 Jan 23  2024 /srv/xrootd/file.png
gentoo build $ env LD_LIBRARY_PATH=$PWD/lib LD_PRELOAD=libXrdPosixPreload.so stat root://gentoo.cern.ch//file.png
  File: root://gentoo.cern.ch//file.png
  Size: 126582        Blocks: 365        IO Block: 65536  regular file
Device: 0,38    Inode: 249917601424080944  Links: 1
Access: (0664/-rw-rw-r--)  Uid: (75748/  amadio)   Gid: ( 1399/      zh)
Access: 2026-02-24 12:05:21.000000000 +0100
Modify: 2024-01-23 16:08:09.000000000 +0100
Change: 2024-04-04 10:00:24.000000000 +0200
 Birth: -
gentoo build $ stat /srv/xrootd/fi
file.png  fio/      
gentoo build $ stat /srv/xrootd/file.png 
  File: /srv/xrootd/file.png
  Size: 126582        Blocks: 248        IO Block: 4096   regular file
Device: 0,48    Inode: 58188476    Links: 1
Access: (0664/-rw-rw-r--)  Uid: (  469/  xrootd)   Gid: (  469/  xrootd)
Access: 2026-02-24 12:05:21.159395042 +0100
Modify: 2024-01-23 16:08:09.575129109 +0100
Change: 2024-04-04 10:00:24.068854096 +0200
 Birth: 2024-01-23 16:08:09.525128974 +0100
```
For me, this is good for the release.

@abh3 As soon as you approve this, I will be able to release 6.0. I have everything else (release notes, etc) in a local branch, and will merge v5.9.2 into master before releasing to get the release notes as well.

Thanks, @ccaffy!

amadio (approved):

src/XrdSfs/XrdSfsInterface.cc
amadio: I'm marking things as resolved to allow merging later.

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,6 +931,21 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
...
 }
+
+extern "C"
+{
+
+int XrdPosix_Statx(int dirfd, const char *path, int flags,
+                   unsigned int mask, XrdSysStatx *stx)
+{
+  struct stat buf;
+  int ret = XrdPosix_Stat(path,&buf);
```
abh3: So, while this works, it likely will "crash" if the call is made using a different set of parameters for dirfd and path. The question is how much do we want to support statx for Linux. If just the standard command line tools, likely this will work modulo some protection code. Otherwise, we will need to simulate all the variations allowed. Please look at the man page https://man7.org/linux/man-pages/man2/statx.2.html which explains the myriad of combinations that dirfd, path, and flags may be specified with each meaning something else. If we only want to support absolute path, then all of the other possibilities should return an error not an undefined result. Now, of course, we start the debate what error code should be returned for unsupported combinations. I'd say the "not implemented" one as we may well want to implement them at some point (i.e. ENOSYS).

src/XrdPosix/XrdPosixExtern.hh
```diff
@@ -190,6 +191,10 @@ extern ssize_t    XrdPosix_Write(int fildes, const void *buf, size_t nbyte);
 
 extern ssize_t    XrdPosix_Writev(int fildes, const struct iovec *iov, int iovcnt);
 
+// statx support
```
abh3: The header file lists the externs in alphabetic order. Let's keep it that way and move this to be after XrdPosix_Statvfs().

src/XrdPosix/XrdPosixLinkage.cc
```diff
@@ -175,6 +175,21 @@ XrdPosixLinkage Xunix;
                          {return (Retv_Rmdir)Xunix.Load_Error("rmdir");}
       Retv_Seekdir     Xrd_U_Seekdir(Args_Seekdir)
                          {       Xunix.Load_Error("seekdir"); abort();}
+#ifdef HAVE_STATX
+      Retv_Statx       Xrd_U_Statx(Args_Statx)
```
abh3, line 191: Save thing jere. The entries are in alphantic order. So, let's leep it that way and place this after 
Retv_Statvfs64   Xrd_U_Statvfs64()

src/XrdPosix/XrdPosixLinkage.cc
```diff
@@ -262,6 +277,11 @@ int XrdPosixLinkage::Resolve()
   LOOKUP_UNIX(Rewinddir)
   LOOKUP_UNIX(Rmdir)
   LOOKUP_UNIX(Seekdir)
+#ifdef HAVE_STATX
+  LOOKUP_UNIX(Statx)
```
abh3, line 281: Again, let's place this after  "LOOKUP_UNIX(Statvfs64)".

src/XrdPosix/XrdPosixLinkage.hh
```diff
@@ -506,6 +511,7 @@ class XrdPosixLinkage
       Retv_Stat64      (*Stat64)(Args_Stat64);
       Retv_Statfs      (*Statfs)(Args_Statfs);
       Retv_Statfs64    (*Statfs64)(Args_Statfs64);
+      Retv_Statx       (*Statx)(Args_Statx);
```
abh3: Place this after Retv_Statvfs64 to keep the list in alphabetic order.

src/XrdPosix/XrdPosixPreload.cc
```diff
@@ -792,3 +792,20 @@ ssize_t writev(int fildes, const struct iovec *iov, int iovcnt)
    return XrdPosix_Writev(fildes, iov, iovcnt);
 }
 }
+
+/******************************************************************************/
```
abh3: Pllease move this to be in alphabetic order (i.e. after statvfs).

abh3 (changes requested):
Only one major issue here and that is ignoring that statx() variations, The other comments are more in line with formatting and following the current name ordering scheme used in this package.

src/XrdPosix/XrdPosix.cc
amadio: Indeed, I agree that some checks here should be performed to avoid problems in corner cases. We can use something like [this](https://github.com/torvalds/linux/blob/master/samples/vfs/test-statx.c) test program from Linux to check that things work by linking it with XrdPosixPreload and calling it on local and remote paths to cover the different usage cases.

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,6 +931,21 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
...
+   if (!ret) {
+      XrdSysStatxHelpers::Stat2Statx(buf,*stx);
+   }
+   return ret;
```
amadio: I have implemented this locally as the following:
```suggestion
  if (path) {
    char buff[2048];
    if (char *myPath = XrootPath.URL(path, buff, sizeof(buff))) {
      struct stat st{};

      if (int ret = XrdPosix_Stat(myPath, &st))
        return ret;

      XrdSysStatxHelpers::Stat2Statx(st, *stx);
      return 0;
    }
  }
#ifdef SYS_statx
  return syscall(SYS_statx, dirfd, path, flags, mask, stx);
#else
  errno = ENOSYS;
  return -1;
#endif
```

This works to have full `statx()` functionality for local paths, and also works for `root://` URLs using `XrdPosix_Stat()`. I think we could actually use `syscall()` with this same style of wrapper functions to significantly simplify how XrdPosix is implemented, without the need to use `LOOKUP_UNIX`, etc. What do you think, @abh3?

src/XrdPosix/XrdPosixLinkage.hh
amadio: Done.

src/XrdPosix/XrdPosix.cc
amadio: I added a fallback to the `statx` syscall and added also a test to make sure it works.

amadio:
@abh3 Since @ccaffy is on holiday today, I've addressed your comments and force-pushed to his repository to update this pull request. I've moved things into alphabetical order, and added a fallback to the `statx` syscall to `XrdPosix_Statx`, which I hope could become the basis for simplifying the implementation of the other functions as well, removing some of these indirections. Please let me know what you think of this. Cheers,

I'm going to resolve the comments which are for things we will work on in XRootD 7.0 to allow me to merge, otherwise GitHub's interface prevents merging with:

screenshot

src/XrdPosix/XrdPosixAdmin.cc
amadio: Posix *does* actually work on macOS, but due to pedantic security restrictions they have, you have to link to `XrdPosixPreload` instead of using something like `LD_PRELAOD`. Only for `statx`, since it's Linux-only, it will not work on macOS.

src/XrdPosix/XrdPosix.cc
```diff
@@ -930,6 +932,32 @@ int XrdPosix_Stat(const char *path, struct stat *buf)
           : Xroot.Stat(myPath, buf));
 }
 }
+
+extern "C"
+{
+int XrdPosix_Statx(int dirfd, const char *path, int flags,
+                   unsigned int mask, XrdSysStatx *stx)
+{
+  if (path && *path) {
```
amadio, line 941: @abh3 The change is here, now this should be ready to merge.

abh3: OK.

abh3 (approved):

PR #2727, opened 2026-03-03
[Client] Add ClFile API to preread a list of offset/lengths from the …

abh3:
This version of pre read only handles reads from a single file, though has the capability to handle multi-file pre reads.  That said, since it has not been test nor are there any test programs, the call to the implementation is commented out until all that is done. However, the API is usable and can be added to the next release that allows for ABI breakage (this definitely breaks ABI).


amadio (approved):
This looks good. I'm merging it. Thank you, @abh3.

PR #2728, opened 2026-03-03
[XrdFrm] Remove local XrdXrootdTrace clashing with global symbol

amadio:
Fixes warnings like the one below (ODR violations reported by the address sanitizer):
```
[ 64%] Linking CXX executable ../../bin/frm_admin
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: warning:
  size of symbol `XrdXrootdTrace' changed from 848 in CMakeFiles/frm_purged.dir/XrdFrmPurgMain.cc.o
  (symbol from plugin) to 8 in /tmp/cc6BZkMF.ltrans0.ltrans.o
/usr/lib/gcc/x86_64-pc-linux-gnu/14/../../../../x86_64-pc-linux-gnu/bin/ld: warning:
  NOTE: size discrepancies can cause real problems.  Investigation is advised.
```
This is because `XrdXrootdTrace` is declared as a global symbol in `XrdXrootd/XrdXrootdProtocol.cc` with type `XrdSysTrace` (size 848), and in XrdFrm binaries it's declared as `XrdOucTrace*` (size 8).

### DIFF
--- a/src/XrdFrm/XrdFrmAdminMain.cc
+++ b/src/XrdFrm/XrdFrmAdminMain.cc
@@ -93,7 +93,6 @@ using namespace XrdFrm;
 
 // The following is needed to resolve symbols for objects included from xrootd
 //
-       XrdOucTrace       *XrdXrootdTrace;
        XrdSysError        XrdLog(0, "");
        XrdOucTrace        XrdTrace(&Say);
 
@@ -143,10 +142,6 @@ int main(int argc, char *argv[])
    Say.logger(&Logger);
    XrdLog.logger(&Logger);
    if (!Config.Configure(argc, argv, 0)) exit(4);
-
-// Fill out the dummy symbol to avoid crashes
-//
-   XrdXrootdTrace = new XrdOucTrace(&Say);
 
 // We either have a command line or need to enter interactive mode
 //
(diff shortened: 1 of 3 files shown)


abh3 (changes requested):
I'm surprised that this error appeared all of a sudden. What changed? 

The proposed fix is not compatible with the stated purpose of the existence of these symbols in the XrdFrm component. Specifically, notice that in XrdFrmXfrMain.cc:102 specifically states that the need to be defined in case certain object files are included from the Xrd/XrdXrootd components that exist in the XrdServer library that is used as a link target. Mind you, things have substantially changed over the years and it is likely that no symbols are included because the Frm assumed that  XrdXrootdTrace and XrdTrace were defined as XrdOucTrace but actually were changed some time ago to be XrdSysTrace. Also note that XrdXrootdTrace is never defined as a pointer except in XrdFrm, further indicating that this problem likely no longer exists. 

So, given that the original intent was to supply missing symbols that are no longer inadvertently dragged into building various XrdFrm components, this fix would do the trick but is not fully complete. All occurences of the definition of XrdTrace should also be emmoved as it's definition has the same problem (i.e. XrdTrace isdefined as XrdOucTrace by XrdFrm when it is now defined as XrdSysTrace)..

That leaves one symbol in the same category, XrdLog(). This symbol is not used anywhere in Frc/Frm so it is clearly meant to provide existence in case some other component was included that depends on it. That is more difficult to determine whether this is the case or not. So, I would leave that in as it does not conflict with anything; though it is somewhat ugly.

All of the above is to provide a historical context. The change I am requesting is to also remove XrdTrace wherever it is defined.  One day we need to determine if the same holds true for XrdLog. I suspect it likely is.

PR #2733, opened 2026-03-05
[ClHttp] Call curl_global_init/cleanup

smithdh:


src/XrdClHttp/XrdClHttpWorker.hh
```diff
@@ -153,9 +153,10 @@ private:
     static std::mutex m_worker_stats_mutex;
 
     // shutdown trigger
-    static struct shutdown_s {
-      ~shutdown_s() { ShutdownAll(); }
-    } m_shutdowns;
+    static struct initcontrol {
+      initcontrol();
+      ~initcontrol();
+    } m_initcontrol;
```
dynamic-entropy, line 159: My approach was to call `curl_global_init`  in XrdCl::HttpFactory::Initialize and cleanup later in Shutdown. 
```
        if (curl_global_init(CURL_GLOBAL_DEFAULT) != 0) {
            m_log->Error(kLogXrdClHttp, "curl_global_init failed");
            return;
        }
```

But having the static guard is better. One suggestion would be to log a failure in case the global init fails as in the snippet above.

Having said that, we only have the issue in curl versions < 7.84  
Since starting curl `7.84`, `curl_global_init` is thread safe and we canrely on it being internally called by `curl_multi_init`. 
i.e ubuntu24, alma9

PR #2735, opened 2026-03-05
[Cl] Rework DefaultEnv finalization with atomics

smithdh:


abh3 (commented):
My concern here is that all of these atomic variables will use strict full barrier fetching. Some of these values are fetched often in the code like postmaster, monitor and the like. Having all the threads serialize memory access just to prevent access prior to initialization is a very heavy price to pay. At the very least they could be make RAtomic (see XrdSysRAtomic) which only require relaxed  memory constraints which have a very low penalty.

abh3:
Is there a way to change RAtomics constructor to avoid this problem?

amadio:
Hi @smithdh, is this something you still plan to work on? If ABI breaking changes are required, we will need to defer this to XRootD 7.0 (which I'd estimate will happen sometime in 2029).

smithdh:
let's close this for now.

PR #2737, opened 2026-03-09
[XrdClHttp] Enable libcurl header traces in Debug mode

dynamic-entropy:
Allow a user to enable curl header dumps with `XRD_LOGLEVEL=Debug`

Currently, one can only achieve this by recompiling; this PR allows a user to set it with xrd env vars. 
It is especially useful when debugging issues reported by users.

If we should make these header traces available at `Debug` or `Dump` can be discussed.

### DIFF
--- a/src/XrdClHttp/XrdClHttpUtil.cc
+++ b/src/XrdClHttp/XrdClHttpUtil.cc
@@ -30,6 +30,7 @@
 #include <XrdCl/XrdClURL.hh>
 #include <XrdCl/XrdClXRootDResponses.hh>
 #include <XrdOuc/XrdOucCRC.hh>
+#include <XrdOuc/XrdOucPrivateUtils.hh>
 #include <XrdSys/XrdSysPageSize.hh>
 #include <XrdVersion.hh>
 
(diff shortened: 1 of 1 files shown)


amadio:
I like this, thanks! Can you please add to the PR description an example of what this looks like when enabled? If you want, you could add it to the commit message as well.

abh3:
I also hope you suppress printing full authorization headers as well.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -569,20 +570,36 @@ HandlerQueue::HandlerQueue(unsigned max_pending_ops) :
...
-        printf("Header > %s\n", std::string(data, size).c_str());
+        direction = ">";
         break;
-    default:
-        printf("Info: %s", std::string(data, size).c_str());
+    case CURLINFO_HEADER_IN:
+        direction = "<";
         break;
+    default:
+        return 0;
```
amadio, line 598: I guess you don't want to return here, do you? Maybe you want to just set `direction = ""`?

dynamic-entropy: No, we want to, because curl invokes this callback with other data as well, such as ssl info and the actual data itself, too. But we only want to provide a header trace. 


(If you wish we can extend to also provide binary ssl data, but that is usually not consulted. The error message already has the hint.
A list of other `curl_infotype` is https://curl.se/libcurl/c/CURLOPT_DEBUGFUNCTION.html

amadio: Ok, I asked because previously this didn't return in the other case, so just confirming. This is good to go then.

PR #2739, opened 2026-03-10
Fix xrdcl error propagation

### LINKED ISSUE #2738: XrdCl send generic error on failure
### Operating System

All

### XRootD Version

-

### Description

A client can receive an error via two paths in XrdCl write requests
1. On issuing a new write
2. ResponseHandlers being invoked

This race between which thread responds first determines which error message the client will see.
However, the first path returns a generic error message instead of one reported by curl.

Example call for the same file:

```zsh
❯ bin/xrdcp /tmp/myfile http://localhost:7094/out_of_space_quota.txt
[8MB/47.68MB][ 16%][========>                                         ][8MB/s]  
Run: [ERROR] Server responded with an error: [3021] Insufficient Storage (write operation at offset 0) (destination)

❯ bin/xrdcp /tmp/myfile http://localhost:7094/out_of_space_quota.txt
[8MB/47.68MB][ 16%][========>                                         ][8MB/s]  
Run: [ERROR] Invalid operation:  (destination)
```

We already have the error information available in `QueueWrite` while responding to subsequent write requests.

 

1. Start an XRootD server for e.g. 
```
ctest -R XRootD::http::setup
```
2. Copy a dummy file with the name `out_of_space_quota.txt` with xrdclhttp client: 
```
bin/xrdcp /tmp/myfile http://localhost:7094/out_of_space_quota.txt
```
3. Do it a couple more times and observe the error reported

---
Note: you might want to use the ` put_handler_callback` branch, because of an existing bug with the file not closing on put, and thus no error is reported.

### Additional Information

_No response_

dynamic-entropy:
Fixes #2738 

Make error reporting similar on both error paths. 

A new write request may arrive and be responded to in `QueueWrite` after the state has been set to `Done` after a failure. This error message need not be generic "Cannot continue PUT after a failure". The response message has already been set by curl. 

This error message now matches the case when the first response is made by the PutResponseHandler instead.

### DIFF
--- a/src/XrdClHttp/XrdClHttpFile.cc
+++ b/src/XrdClHttp/XrdClHttpFile.cc
@@ -1361,7 +1361,11 @@ File::PutResponseHandler::QueueWrite(std::variant<std::pair<const void *, size_t
         auto sc = m_op->GetStatusCode();
         if (HTTPStatusIsError(sc)){
             auto httpErr = HTTPStatusConvert(sc);
-            return XrdCl::XRootDStatus(XrdCl::stError, httpErr.first, httpErr.second, m_op->GetStatusMessage());
+            auto err_msg = m_op->GetCurlErrorMessage();
+            if (err_msg.empty()) {
+                err_msg = m_op->GetStatusMessage();
+            }
+            return XrdCl::XRootDStatus(XrdCl::stError, httpErr.first, httpErr.second, err_msg);
         }
         return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidOp, 0, "Cannot continue writing to open file after error");
     }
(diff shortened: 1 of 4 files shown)


src/XrdClHttp/XrdClHttpOps.cc
```diff
@@ -531,7 +531,9 @@ CurlOperation::Setup(CURL *curl, CurlWorker &worker)
     m_last_header_reset = m_last_reset = m_start_op = m_header_start = m_header_lastop = std::chrono::steady_clock::now();
 
     m_curl.reset(curl);
+    m_curl_error_buffer[0] = '\0';
     curl_easy_setopt(m_curl.get(), CURLOPT_URL, m_url.c_str());
+    curl_easy_setopt(m_curl.get(), CURLOPT_ERRORBUFFER, m_curl_error_buffer);
```
dynamic-entropy, line 536: I noticed that the error message received by the client was not informative enough and compared it to one from the curl CLI. Hence, the use of ERRORBUFFER to get the runtime error from curl.

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -1381,7 +1390,7 @@ File::PutResponseHandler::QueueWrite(std::variant<std::pair<const void *, size_t
     } else {
         m_pending_writes.emplace_back(std::move(buffer), handler);
     }
-    return XrdCl::Status{};
+    return XrdCl::XRootDStatus{};
```
amadio, line 1393: You could simply `return {};` here, but it's just cosmetic.

PR #2741, opened 2026-03-11
Use correct format specifier for user connections in monitoring report

jhiemstrawisc:
Caught this from a user-reported error in Pelican: `msg="Pelican failed to handle monitoring packet received from XRootD: Provided ID, -7865, is not a valid uint32"`

It looks like `Did` [is in fact a unsigned](https://github.com/xrootd/xrootd/blob/e1ec1c4bc71c8f2d21483634f683a4070627f802/src/XrdXrootd/XrdXrootdMonitor.hh#L179), but the `snprintf` statements had the wrong format specifier.


abh3 (approved):
Thanks!

PR #2743, opened 2026-03-14
[Preread] Add missing server-side preread API's (breaks ABI)

abh3:
This is a redo fix of the previous preread branch. It is meant to add ABI-breaking API's with actual implementation to follow.

### DIFF
--- a/src/XrdSfs/XrdSfsInterface.hh
+++ b/src/XrdSfs/XrdSfsInterface.hh
@@ -40,7 +40,6 @@
 #include "XrdOuc/XrdOucErrInfo.hh"
 #include "XrdOuc/XrdOucIOVec.hh"
 #include "XrdOuc/XrdOucRange.hh"
-//#include "XrdOuc/XrdOucSFVec.hh"
 
 #include "XrdSfs/XrdSfsGPFile.hh"
 


src/XrdOss/XrdOss.cc
```diff
@@ -229,10 +229,26 @@ int XrdOssDF::pgWrite(XrdSfsAio *aioparm, uint64_t opts)
...
+/******************************************************************************/
+
+ssize_t XrdOssDF::Read(XrdOucRangeList& rlist)
+{
+   ssize_t rc, retval = 0;
+
+   for (auto it = rlist.begin(); it != rlist.end(); it++)
+       {rc = Read(it->offset, it->size);
+        if (rc < 0) return rc;
+        retval |= rc;
```
amadio, line 243: Doesn't `+=` make more sense here? I see below that you say a positive value means the request was accepted, but given that on negative value we shortcut things, we might as well without much extra cost return the number of bytes which were read using `+=` instead of `|=`.

src/XrdOuc/XrdOucCache.hh
```diff
@@ -307,6 +308,17 @@ static const int SingleUse = 0x0001; //!< Mark pages for single use
...
 
+//------------------------------------------------------------------------------
+//! Perform an asynchronous vector preread (may be ignored).
+//!
+//! @param  rlist a vector of byte ranges to preread.
+//------------------------------------------------------------------------------
+
+virtual void Preread(XrdOucRangeList& rlist)
+                    {for (auto it = rlist.begin(); it != rlist.end(); it++)
+                         Preread(it->offset, it->size);
```
amadio, line 319: You can use C++17 to make the code much nicer here (if instead of defining a `XrdOucRangeList` you take in `std::vector<std::pair<off_t, size_t>> const&`, which would be mostly self-explanatory anyway given the type names):
```cpp
for (const auto& [offset, size] : rlist)
    Preread(offset, size);
```

src/XrdOuc/XrdOucRange.hh
```diff
@@ -0,0 +1,59 @@
...
+//! The struct defined here is a generic data structure that is used whenever
+//! we need to pass a vector of file offsets, lengths, and possible other
+//! information. It is used by the sfs, ofs, and oss components for pre-reads.
+//-----------------------------------------------------------------------------
+
+struct XrdOucRange
+{
+       long long offset;    // Offset into the file.
+       int       size;      // Size of I/O to perform.
+       int       info;      // Available for arbitrary use
```
amadio, line 46: I'd suggest using `off_t` for offsets and `size_t` for sizes. Alternatively, I'd use fixed size types, like `int64_t` and `int32_t`, but matching signatures of POSIX I/O functions would be better, I think. It's also inconsistent to use `long long` here and `off_t` elsewhere.

src/XrdPosix/XrdPosixExtra.cc
```diff
@@ -265,3 +265,33 @@ ssize_t XrdPosixExtra::pgWrite(int   fildes, void*    buffer,
    fp->XCio->pgWrite(*cbp, (char *)buffer, offs, (int)iosz, csvec);
    return 0;
 }
+
+/******************************************************************************/
+/*                               P r e R e a d                                */
+/******************************************************************************/
+
+int XrdPosixExtra::PreRead(int fildes, off_t offs, int size)
```
amadio, line 273: `size_t size` is safer and more future proof, I believe.

src/XrdSfs/XrdSfsInterface.cc
```diff
@@ -185,10 +185,21 @@ XrdSfsXferSize XrdSfsFile::pgWrite(XrdSfsAio *aioparm, uint64_t opts)
...
 
+/******************************************************************************/
+/*                                  r e a d                                   */
+/******************************************************************************/
+
+XrdSfsXferSize XrdSfsFile::read(XrdOucRangeList& rlist)
+{
+   for (auto it = rlist.begin(); it != rlist.end(); it++)
+       if (read(it->offset, it->size) == SFS_ERROR) return SFS_ERROR;
+   return 1;
```
amadio, line 196: Should `1` be `SFS_OK`?

src/XrdSfs/XrdSfsInterface.hh
```diff
@@ -39,7 +39,8 @@
 
 #include "XrdOuc/XrdOucErrInfo.hh"
 #include "XrdOuc/XrdOucIOVec.hh"
-#include "XrdOuc/XrdOucSFVec.hh"
+#include "XrdOuc/XrdOucRange.hh"
+//#include "XrdOuc/XrdOucSFVec.hh"
```
amadio: Here it would be better not to add commented out code.

src/XrdOss/XrdOss.cc
abh3: I chose "or" instead of plus equals to avoid overflows. I just need a positive value. Of course, that won't happen if one of the items is negative. I suppose I should check for that. On the other hand, Matevz and I have yet to decode how to indicate a pre-read to the end of the file. fadvise uses zero but some other API's (which I can't recall now) use a negative value to mean "end of file". So, we don't know how we are going to handle that yet.

src/XrdOuc/XrdOucCache.hh
abh3: Hmmm, let me mull it over. It's beauty in he eye of the beholder sort of thing here.

src/XrdOuc/XrdOucRange.hh
abh3: Well, that's not what the protocol allows. Lengths are defined as integers so I'm just following the size constraint. You simply can't read/write/pre-read more than 2GB-1 at a time. We'd also wind up with a narrowing problem when passing through a proxy server. Additionally, this nicely fits into two doublewords. Making size bigger requires three doublewords. Not that we are hurting for space but it's just awkward.

src/XrdPosix/XrdPosixExtra.cc
abh3: Again, that may be true but that is not what the protocol allows. Normally we use XrdSfsXferSize which is defined as an int. However, by convention we don't include XrdSfsInterface, where that is defined, for XrdPosix APIs.

src/XrdSfs/XrdSfsInterface.cc
abh3: No, SFS_OK is defied as zero and we need to return a positive value to indicate the request was accepted.

src/XrdOuc/XrdOucRange.hh
amadio: Ok, I agree with you, let's close this one as resolved then.

src/XrdOss/XrdOss.cc
amadio: Ok, knowing the reason, it makes sense.

src/XrdOuc/XrdOucCache.hh
amadio: Since this is the only thing left, I guess let's just keep it as is for the sake of getting the release out sooner.

amadio (approved):

PR #2757, opened 2026-04-01
Xrd cl security fixes

apeters1971:
Proposed security fixes.


src/XrdCl/XrdClFileStateHandler.cc
```diff
@@ -1894,18 +1894,26 @@ namespace XrdCl
 
     ChunkList *list   = new ChunkList();
 
-    uint32_t size = 0;
+    uint64_t totalSize = 0;
```
amadio, line 1897: This is a real fix, which has been reported also by Coverity. There a bunch of places where we use a 32bit integer for representing time or sizes where we need to replace them with a 64bit equivalent. The ideal fix would be a commit changing all the places using 32bit integers for time with 64bit integers, without unrelated changes together.

amadio: The right fix here is to use `size_t`:
```
       struct iovec {
           void   *iov_base;  /* Starting address */
           size_t  iov_len;   /* Size of the memory pointed to by iov_base. */
       };
```

PR #2759, opened 2026-04-01
[XrdSeckrb5] Fix top thread-safety and crash issues

apeters1971:
Fix a null-pointer crash in <uid> token-export expansion, protect static initialization with a mutex, and make static parameter storage updates synchronized and instance-safe.


src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -84,6 +84,10 @@ extern "C" {
 
 typedef  krb5_error_code krb_rc;
 
+namespace {
+XrdSysMutex initMutex;
+}
```
amadio, line 89:
```suggestion
static XrdSysMutex initMutex;
```

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -106,7 +110,12 @@ static  int                Init(XrdOucErrInfo *einfo, char *KP=0, char *kfn=0);
 
 static  void               setOpts(int opts) {options = opts;}
 static  void               setClientOpts(int opts) {client_options = opts;}
-static  void               setParms(char *param) {Parms = param;}
+static  void               setParms(char *param)
+                                    {initMutex.Lock();
+                                     if (Parms) free(Parms);
+                                     Parms = param;
+                                     initMutex.UnLock();
+                                    }
```
amadio, line 117: Please use a scoped lock with `XrdSysMutexHelper` (i.e. RAII-style rather than explicit lock/unlock).

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -800,7 +808,7 @@ int XrdSecProtocolkrb5::exp_krbTkn(XrdSecCredentials *cred, XrdOucErrInfo *erp)
         if (ln != 5) {
            // Adjust the space
            int lm = strlen(ccfile) - (int)(puid + 5 - &ccfile[0]); 
-           memmove(puid+ln, pusr+5, lm);
+           memmove(puid+ln, puid+5, lm);
```
amadio, line 810: This should be a separate commit, as it has no relation to initialization. It should also come with an explanation in the commit message why it's wrong and why the new code is right.

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -908,11 +916,18 @@ char  *XrdSecProtocolkrb5Init(const char     mode,
    XrdOucTokenizer inParms(parmbuff);
    int options = XrdSecNOIPCHK;
    static bool serverinitialized = false;
+   bool alreadyInitialized = false;
```
amadio, line 918: Why is a second bool needed? Can we not use just the pre-existing `serverinitialized` instead?

src/XrdSeckrb5/XrdSecProtocolkrb5.cc
```diff
@@ -908,11 +916,18 @@ char  *XrdSecProtocolkrb5Init(const char     mode,
...
 // For client-side one-time initialization, we only need to set debug flag and
 // initialize the kerberos context and cache location.
 //
-   if ((mode == 'c') || (serverinitialized))
+   if (mode != 'c')
+      {initMutex.Lock();
+       alreadyInitialized = serverinitialized;
+       if (!serverinitialized) serverinitialized = true;
+       initMutex.UnLock();
+      }
```
amadio, line 927: Same as before, please use a scoped lock.

amadio (changes requested):
Please break the changes down into one logical change per commit with a proper explanation why the change is needed. That is, one commit for the initialization, and another for the fix in `memmove`. It seems to me that we'd have observed crashes if there was really a problem with the initialization of this plugin, so I'm skeptical as to what exactly this is fixing.

PR #2760, opened 2026-04-01
[XrdSecztn] Harden token parsing and validation logic

apeters1971:
Fix unsafe protocol-id parsing, correct expiry handling and -expiry option parsing, remove token-file TOCTOU by validating metadata on opened FDs, and make global runtime ztn settings thread-safe with atomics.

### DIFF
--- a/src/XrdSecztn/XrdSecProtocolztn.cc
+++ b/src/XrdSecztn/XrdSecProtocolztn.cc
@@ -462,23 +462,35 @@ XrdSecCredentials *XrdSecProtocolztn::readToken(XrdOucErrInfo *erp,
 //
    isbad = true;
 
-// Get the size of the file
+// Open the token file and inspect metadata from the open file descriptor.
+// This avoids TOCTOU races between stat/open/read.
 //
-   if (stat(path, &Stat))
+   int open_flags = O_RDONLY;
+#ifdef O_NOFOLLOW
+   open_flags |= O_NOFOLLOW;
+#endif
+   if ((tokFD = open(path, open_flags)) < 0)
       {if (errno != ENOENT) return readFail(erp, path, errno);
        isbad = false;
        return 0;
       }
+   if (fstat(tokFD, &Stat))
+      {int rc = errno;
+       close(tokFD);
+       return readFail(erp, path, rc);
+      }
 
-// Make sure token is not too big
+// Make sure token is not too big and inaccessible by group/other.
 //
-   if (Stat.st_size > maxTSize) return readFail(erp, path, EMSGSIZE);
+   if (Stat.st_size > maxTSize)
+      {close(tokFD);
+       return readFail(erp, path, EMSGSIZE);
+      }
+   if (Stat.st_mode & (S_IRWXG | S_IRWXO))
+      {close(tokFD);
+       return readFail(erp, path, EPERM);
+      }
    buff = (char *)alloca(Stat.st_size+1);
-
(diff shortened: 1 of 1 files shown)


src/XrdSecztn/XrdSecProtocolztn.cc
```diff
@@ -614,7 +614,7 @@ int XrdSecProtocolztn::Authenticate(XrdSecCredentials *cred,
 
 // Check if this is our protocol
 //
-   if (strcmp(tResp->hdr.id, "ztn"))
+   if (memcmp(tResp->hdr.id, "ztn", 3) || tResp->hdr.id[3] != '\0')
```
amadio, line 617: Since `strcmp` will compare `tResp->hdr.id` with the constant string `"ztn"`, it will never read more than the 4 first bytes of `tResp->hdr.id`, right?

src/XrdSecztn/XrdSecProtocolztn.cc
```diff
@@ -658,20 +658,21 @@ int XrdSecProtocolztn::Authenticate(XrdSecCredentials *cred,
...
-          {if (eTime < 0 && expiry > 0)
+   int expiryMode = expiry.load();
+   if (tokenlib.load() && sthP->Validate(tResp->tkn, msgRC, (expiryMode ? &eTime : 0), &Entity))
+      {if (expiryMode)
+          {if (eTime < 0 && expiryMode > 0)
               {Fatal(erp, "'ztn' token expiry missing", EINVAL, false);
                return -1;
               }
-           if ((monotonic_time() - eTime) <= 0)
+           if ((eTime >= 0) && (time(nullptr) >= eTime))
```
amadio, line 668: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_real_time/9/html/optimizing_rhel_9_for_real_time_for_low_latency_operation/assembly_application-timestamping_optimizing-rhel9-for-real-time-for-low-latency-operation

src/XrdSecztn/XrdSecProtocolztn.cc
apeters1971: I think this was recommended because hdr.id is actually not a string and not guaranteed to be null terminated.

src/XrdSecztn/XrdSecProtocolztn.cc
apeters1971: You think this matters for an authentication handshake with one call to time() in a 20ms exchange ?

amadio: No, just telling you why it was using `clock_gettime`. @abh3 might want to keep this at it was before.

src/XrdSecztn/XrdSecProtocolztn.cc
```diff
@@ -461,24 +448,36 @@ XrdSecCredentials *XrdSecProtocolztn::readToken(XrdOucErrInfo *erp,
...
    buff = (char *)alloca(Stat.st_size+1);
 
-// Open the token file
-//
-   if ((tokFD = open(path, O_RDONLY)) < 0)
-      return readFail(erp, path, errno);
-
 // Read in the token
 //
    if ((rdLen = read(tokFD, buff, Stat.st_size)) != Stat.st_size)
```
amadio, line 483: New warning is on this line: https://my.cdash.org/builds/3453305/build
Should have been shown by the warning parser, but seems that failed.

PR #2761, opened 2026-04-01
[XrdClHttp] Harden parsing and redirect host handling

apeters1971:
Prevent remote-triggered crashes by validating XML roots and numeric conversions, harden Content-Range parsing against invalid ranges, and correctly parse IPv6/invalid authorities before connection-callout remapping.


amadio (changes requested):
Please make one commit per logical change. It's very difficult to review a lot of unrelated changes together.

bbockelm:
@apeters1971 - looks like quite good catches!  Could you take the opportunity to add regression tests to the parsing code as well?  If there's one bug hiding, there's likely others...

PR #2763, opened 2026-04-01
[XrdSecsss] Fix security and correctness issues

apeters1971:
- Replace alloca(cred->size) with a heap-allocated RAII buffer that zeroes itself on scope exit, eliminating key material on the stack and the risk of stack exhaustion from attacker-controlled size
- Check malloc(idTLen) return value to avoid null dereference DoS
- Guard strlen(hostIP) in RR_Data() with a null check to fix crash when hostIP is legitimately null
- Use random suffix and O_EXCL in Rewrite() temp file open to prevent a predictable-name symlink-race attack
- Fix == vs = typo in Refresh() that caused spurious error log on every successful keytable refresh
- Cap per-packet attribute count at 64 to bound entity memory growth


amadio (approved):
This looks good, I will likely merge for the next release.

PR #2769, opened 2026-04-21
Some minor fixes for issues found when building 6.0.0

ellert:
This PR collects some fixes for issues found when building xrootd 6.0.0.

### DIFF
--- a/src/XrdClHttp/XrdClHttpUtil.cc
+++ b/src/XrdClHttp/XrdClHttpUtil.cc
@@ -86,16 +86,18 @@ struct WaitingForBroker {
 namespace {
 
 pid_t getthreadid() {
-#ifdef __APPLE__
+#if   defined(__APPLE__)
     uint64_t pth_threadid;
     pthread_threadid_np(pthread_self(), &pth_threadid);
     return pth_threadid;
-#else
+#elif defined(__linux__)
     // NOTE: glibc 2.30 finally provides a gettid() wrapper; however,
     // we currently support RHEL 8, which is based on glibc 2.28.  Until
     // we drop that platform, it's easier to do the syscall directly on Linux
     // instead of additional ifdef calls.
     return syscall(SYS_gettid);
+#else
+   return getpid();
 #endif
 }
 
--- a/src/XrdSys/XrdSysStatx.hh
+++ b/src/XrdSys/XrdSysStatx.hh
@@ -35,7 +35,7 @@
 #include <cstring>
 #include <fcntl.h>
 
-#ifdef __linux__
+#if defined(__linux__) || defined(__GNU__)
 #include <sys/sysmacros.h>
 using XrdSysStatx = struct statx;
 #define HAVE_STATX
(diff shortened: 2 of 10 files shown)


tests/krb5/kerberos.sh
```diff
@@ -36,16 +36,16 @@ function setup() {
 	popd >/dev/null || exit 1
 
 	# Create the KDC database
-	kdb5_util create -s -r XROOTD.ORG -P xrootd
+	/usr/sbin/kdb5_util create -s -r XROOTD.ORG -P xrootd
```
amadio: Instead of hard-coding `/usr/bin/` as a prefix here, I'd rather prefer that we add `/usr/sbin/` to `$PATH` if it's not there (or even unconditionally). Otherwise some system may install these tools into `/usr/bin` and these commands will fail too.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -86,16 +86,18 @@ struct WaitingForBroker {
...
-#else
+#elif defined(__linux__)
     // NOTE: glibc 2.30 finally provides a gettid() wrapper; however,
     // we currently support RHEL 8, which is based on glibc 2.28.  Until
     // we drop that platform, it's easier to do the syscall directly on Linux
     // instead of additional ifdef calls.
     return syscall(SYS_gettid);
+#else
+   return getpid();
 #endif
```
amadio, line 101: Please change the commit message to explain why this change is needed, i.e., for which platform the previous version did not work (I suspect GNU/Hurd). A better option may be using the implementation from `XrdSysPthread` here, or the implementation below:
```cpp
#if defined(SYS_gettid)
    return syscall(SYS_gettid);
#elif defined(__APPLE__)
    uint64_t pth_threadid;
    pthread_threadid_np(pthread_self(), &pth_threadid);
    return pth_threadid;
#else
   return getpid();
#endif
```

src/XrdSys/XrdSysStatx.hh
```diff
@@ -35,7 +35,7 @@
 #include <cstring>
 #include <fcntl.h>
 
-#ifdef __linux__
+#if defined(__linux__) || defined(__GNU__)
```
amadio, line 38: Here, an `#if defined(SYS_statx)` might be better if it works on GNU/Hurd as well. I think it's cleaner and may cover other systems offering `statx` without needing a change.

Please improve the commit message, I think something like `[XrdSys] Adjust check for statx to also work on GNU/Hurd`, which is much clearer. I'd encourage adding a tag whenever it makes sense to better follow our [conventions](https://github.com/xrootd/xrootd/blob/master/docs/CONTRIBUTING.md#describe-your-changes-well) in commit messages.

amadio (changes requested):
Please try to follow our commit message conventions, adding a tag like `[XrdSys]` or `[XrdClHttp]` where it makes sense. It's all good, just needs some minor changes before merging. Thanks!

amadio:
Alright, this looks ok now. Thanks for the corrections! I will soon create patch releases 5.9.3 and 6.0.1, so if you prefer to wait for those, they should be out sometime next week.

PR #2771, opened 2026-04-22
[CI] Build and publish Python wheels on macOS and Linux arm64

### LINKED ISSUE #2778: MacOS wheels
### Feature Description

XRootD recently started building wheels and uploading them on pypi.
But the wheels are `manylinux` wheels only at the moment.
Could we get macos wheels too? Ideally for arm64 and x86 macos too but arm64 would be a priority.

One can argue that arm64 linux wheels would be nice to have too but that's a separate issue.

lobis:
## Summary
- add native macOS wheel jobs for arm64 and x86_64 to the Python workflow
- fix PyPI-build rpaths on macOS so delocate can resolve and bundle XRootD dylib dependencies
- make genversion.sh's sanitize check work with macOS's Bash 3 and trigger the Python workflow when wheel inputs change

## Validation
- built an arm64 macOS wheel locally from sdist with `_PYTHON_HOST_PLATFORM=macosx-26.0-arm64`
- repaired the wheel with `delocate-wheel`
- installed the repaired wheel and verified `import XRootD` and `import pyxrootd` succeeded

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -9,6 +9,7 @@ on:
       - setup.py
       - pyproject.toml
       - 'bindings/**'
+      - .github/workflows/python.yml
     tags-ignore:
   pull_request:
     paths:
@@ -16,16 +17,52 @@ on:
       - pyproject.toml
       - 'bindings/**'
   workflow_dispatch:
+  release:
+    types:
+      - published
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
+  sdist:
+    name: Source distribution
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+        fetch-tags: true
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.14"
+
+    - name: Install Python build dependencies
+      run: python -m pip install build
+
+    - name: Build source distribution tarball
+      run: |
+        git config --global --add safe.directory "$GITHUB_WORKSPACE"
+        ./genversion.sh >| VERSION
+        python -m build --sdist
+
+    - name: Upload source distribution
+      uses: actions/upload-artifact@v7
+      with:
+        name: python-sdist
+        path: dist/*.tar.gz
+        retention-days: 14
+
   manylinux:
     name: Python ${{ matrix.version }} (Linux, ${{ matrix.platform.arch }})
     runs-on: ${{ matrix.platform.runs_on }}
     container: ${{ matrix.platform.container }}
+    needs: [ sdist ]
 
     strategy:
       fail-fast: false
@@ -47,20 +84,13 @@ jobs:
     - name: Install dependencies
       run: dnf install -y git krb5-devel libuuid-devel openssl-devel
 
-    - name: Clone repository
-      uses: actions/checkout@v6
+    - name: Download source distribution
+      uses: actions/download-artifact@v7
       with:
-        fetch-depth: 0
-        fetch-tags: true
-
-    - name: Build source distribution tarball
-      run: |
-        git config --global --add safe.directory "$GITHUB_WORKSPACE"
-        ./genversion.sh >| VERSION
-        ${PYTHON} -m build --sdist
+        name: python-sdist
 
     - name: Build binary wheel
-      run: ${PYTHON} -m pip wheel --use-pep517 --verbose dist/*.tar.gz
+      run: ${PYTHON} -m pip wheel --use-pep517 --verbose *.tar.gz
 
     - name: Fixup binary wheel
       run: auditwheel repair *.whl
@@ -87,6 +117,7 @@ jobs:
   macos:
     name: Python ${{ matrix.version }} (macOS, ${{ matrix.platform.arch }})
     runs-on: ${{ matrix.platform.runs_on }}
+    needs: [ sdist ]
 
     strategy:
       fail-fast: false
@@ -110,31 +141,24 @@ jobs:
       _PYTHON_HOST_PLATFORM: ${{ matrix.platform.host_platform }}
 
     steps:
-    - name: Clone repository
-      uses: actions/checkout@v6
-      with:
-        fetch-depth: 0
-        fetch-tags: true
-
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
         python-version: ${{ matrix.version }}
 
     - name: Install dependencies with Homebrew
-      run: brew install nlohmann-json openssl@3
+      run: brew install nlohmann-json
 
     - name: Install Python build dependencies
       run: python -m pip install build delocate
 
-    - name: Build source distribution tarball
-      run: |
-        git config --global --add safe.directory "$GITHUB_WORKSPACE"
-        ./genversion.sh >| VERSION
-        python -m build --sdist
+    - name: Download source distribution
+      uses: actions/download-artifact@v7
+      with:
+        name: python-sdist
 
     - name: Build binary wheel
-      run: python -m pip wheel --use-pep517 --verbose dist/*.tar.gz
+      run: python -m pip wheel --use-pep517 --verbose *.tar.gz
 
     - name: Fixup binary wheel
       run: delocate-wheel -w wheelhouse -v *.whl
@@ -157,3 +181,25 @@ jobs:
         name: python-${{ matrix.version }}-macos-${{ matrix.platform.arch }}
         path: wheelhouse
         retention-days: 14
+
+  pypi:
+    if: ${{ github.event_name == 'release' && github.repository_owner == 'xrootd' }}
+    name: Publish to PyPI
+    needs: [ macos, manylinux ]
+    runs-on: ubuntu-latest
+    environment: pypi
+    permissions:
+      attestations: write
+      contents: read
+      id-token: write
+
+    steps:
+    - name: Download Artifacts
+      uses: actions/download-artifact@v7
+      with:
+        path: dist
+        pattern: python-*
+        merge-multiple: true
+
+    - name: Publish to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -23,15 +23,24 @@ concurrency:
 
 jobs:
   manylinux:
-    name: Python
-    runs-on: ubuntu-latest
-    container: quay.io/pypa/manylinux_2_28_x86_64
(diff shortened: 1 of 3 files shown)


.github/workflows/python.yml
```diff
@@ -6,12 +6,24 @@ permissions:
 on:
   push:
     paths:
+      - .github/workflows/python.yml
+      - CMakeLists.txt
+      - VERSION
```
amadio: This file is not meant to be changed

src/CMakeLists.txt
```diff
@@ -24,6 +30,10 @@ set_target_properties(XrdUtils
   VERSION ${XRootD_LIBVERSION}
 )
 
+if( PYPI_BUILD )
+  set_property(TARGET XrdUtils PROPERTY BUILD_RPATH "${XROOTD_PYPI_RPATH}")
+endif()
```
amadio: Please change `PROPERTIES` in `set_target_properties` above instead.

genversion.sh
```diff
@@ -66,7 +66,7 @@ while [[ $# -gt 0 ]]; do
 	esac
 done
 
-if [[ -v SANITIZE ]]; then
+if [[ -n "${SANITIZE+x}" ]]; then
```
amadio: What is this fixing? This is mixing up things, should be a separate commit.

.github/workflows/python.yml
amadio: Maybe let's remove the path restrictions and let this run on every push, we've missed problems before due to these restrictions. The list of paths is getting quite long too.

genversion.sh
lobis: MacOS ships with an old bash version. This is from my M4 macbook pro:
```
❯ bash --version
GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25)
Copyright (C) 2007 Free Software Foundation, Inc.
```

Apparently this syntax is too new for this version of bash, so we need to use the more backwards compatible syntax.

amadio:
Please apply fixes in the original commits rather than as separate commits, using `git rebase -i` and marking the commits as `fixup` in the right place. Otherwise, we can squash on merge. I'd like to avoid intermediate broken states in the git history if possible.

genversion.sh
amadio: Ok, then this should be explained in the commit message. Maybe there's some less ugly syntax for this too than the strange `+x` inside the variable curly braces, which I've never seen before.

.github/workflows/python.yml
```diff
@@ -73,3 +65,85 @@ jobs:
...
+            runs_on: macos-15
+            host_platform: macosx-15.0-arm64
+            openssl_prefix: /opt/homebrew/opt/openssl@3
+          - arch: x86_64
+            runs_on: macos-15-intel
+            host_platform: macosx-15.0-x86_64
+            openssl_prefix: /usr/local/opt/openssl@3
+
+    env:
+      CMAKE_ARGS: -DPython_FIND_UNVERSIONED_NAMES=FIRST -DCMAKE_PREFIX_PATH=${{ matrix.platform.openssl_prefix }}
```
amadio: For `CMAKE_PREFIX_PATH`, I think you can just set it in the environment itself, see https://cmake.org/cmake/help/latest/envvar/CMAKE_PREFIX_PATH.html#envvar:CMAKE_PREFIX_PATH

.github/workflows/python.yml
```diff
@@ -73,3 +65,85 @@ jobs:
...
+        missing=()
+        for formula in bash libzip nlohmann-json openssl@3; do
+          if ! brew list --versions "${formula}" >/dev/null 2>&1; then
+            missing+=("${formula}")
+          fi
+        done
+
+        if [ ${#missing[@]} -gt 0 ]; then
+          brew install "${missing[@]}"
+        fi
```
amadio: I think it's fine to just `brew install bash libzip nlohmann-json openssl@3` here. `libzip` is only needed by the server (XrdOssArc plugin), so it can be dropped. As for `bash`, it might be better to fix `genversion.sh` to work with plain `/bin/sh` instead so we can drop `bash` as a dependency on `macOS`.

python/setup.py
```diff
@@ -48,6 +48,7 @@ def get_cmake_args():
 
     cmake = which("cmake3") or which("cmake")
 
+    cmdline_args += [ '-DCMAKE_BUILD_TYPE=Release' ]
```
amadio: This is set on line 39, please change it there instead. We should only use `Release` if it's not set by the user to something else as well, otherwise when we build from CMake with `-DCMAKE_BUILD_TYPE=Debug`, the python bindings will not be in debug mode.

setup.py
```diff
@@ -70,6 +70,7 @@ def build_extensions(self):
 
             cmake_args = [
                 '-DPython_EXECUTABLE={}'.format(sys.executable),
+                '-DCMAKE_BUILD_TYPE=Release',
```
amadio: I actually think that the right way to do this is to set a build type in the Python wheel CI workflow, not here, by exporting `CMAKE_BUILD_TYPE` in the environment and ensuring we use CMake 3.22 or above in the CI:
https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_TYPE.html#envvar:CMAKE_BUILD_TYPE

amadio: Either that or by using `CMAKE_ARGS`, which is looked at in `get_cmake_args()` above.

.github/workflows/python.yml
lobis: removed paths

genversion.sh
lobis: refactored the script to use `sh` instead

.github/workflows/python.yml
```diff
@@ -73,3 +65,76 @@ jobs:
...
+            runs_on: macos-15
+            host_platform: macosx-15.0-arm64
+            openssl_prefix: /opt/homebrew/opt/openssl@3
+          - arch: x86_64
+            runs_on: macos-15-intel
+            host_platform: macosx-15.0-x86_64
+            openssl_prefix: /usr/local/opt/openssl@3
+
+    env:
+      CMAKE_ARGS: -DPython_FIND_UNVERSIONED_NAMES=FIRST -DCMAKE_BUILD_TYPE=Release
```
amadio, line 137: Does this really work? I'd expect something like
```
CMAKE_ARGS: "-DPython_FIND_UNVERSIONED_NAMES=FIRST;-DCMAKE_BUILD_TYPE=Release"
```
instead, or CMake will get confused, but I might be wrong.

genversion.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
amadio, line 1: Please split changes to `genversion.sh` as a separate commit, to have one logical change per commit. Something like `Convert genversion.sh from bash to plain sh` or such.

.github/workflows/python.yml
lobis: I think this should work, I'll resolve this conversation once I verify it in the CI.

genversion.sh
lobis: Refactored to use more standard syntax but with a small change in behaviour if the variable is set to empty string which does not really matter given how the script is used it makes it more readeable this way.

genversion.sh
```diff
@@ -66,8 +66,8 @@ while [[ $# -gt 0 ]]; do
 	esac
 done
 
-if [[ -v SANITIZE ]]; then
-	VERSION=$(sed -e 's/v//; s/-rc/~rc/; s/-g/+git/; s/-/.post/; s/-/./' <<< "${VERSION}")
+if [ -n "${SANITIZE}" ]; then
```
amadio, line 69: Good, this is better than that strange syntax with `+x`. :+1:

.github/workflows/python.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Build source distribution tarball
       run: |
         git config --global --add safe.directory "$GITHUB_WORKSPACE"
-        ./genversion.sh --sanitize >| VERSION
+        ./genversion.sh >| VERSION
```
amadio: Why are you making this change? The `--sanitize` was made to satisfy [PEP440](https://peps.python.org/pep-0440/) and we also try to follow [RPM versioning guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_snapshots):
```sh
$ genversion.sh
v6.0.0-4+git0fe8a5ddb
$ genversion.sh --sanitize
6.0.0.post4+git0fe8a5ddb
```

src/CMakeLists.txt
```diff
@@ -3,8 +3,17 @@
...
+  else()
+    set( XROOTD_PYPI_RPATH "$ORIGIN" )
+  endif()
+
+  set( CMAKE_INSTALL_RPATH "${XROOTD_PYPI_RPATH}" )
   set( CMAKE_INSTALL_RPATH_USE_LINK_PATH true )
+  set( XROOTD_UTILS_BUILD_RPATH "${XROOTD_PYPI_RPATH}" )
+else()
+  set( XROOTD_UTILS_BUILD_RPATH "$ORIGIN" )
 endif()
```
amadio, line 12: I actually think that this particular change may not be necessary since [MACOSX_RPATH](https://cmake.org/cmake/help/latest/prop_tgt/MACOSX_RPATH.html) is enabled by default by policy [CMP0042](https://cmake.org/cmake/help/latest/policy/CMP0042.html#policy:CMP0042). Either that, or we should be setting [INSTALL_NAME_DIR](https://cmake.org/cmake/help/latest/prop_tgt/INSTALL_NAME_DIR.html) instead of RPATH in any case.

.github/workflows/python.yml
lobis: The reasoning is that it was removed because its ~rc form is fine for RPM/DEB but invalid for Python packaging. The Python build already sanitizes VERSION in setup.py / python/setup.py using PEP 440 rules, so it needs the raw XRootD version string, not the generic package-manager one.

src/CMakeLists.txt
lobis: Addressed, thanks!

.github/workflows/python.yml
amadio: Indeed, I added PEP440 processing to setup.py as well, and there's actually a [problem](https://github.com/xrootd/xrootd/security/quality/rules/py%2Ffile-not-closed) with it which needs fixing as well.

lobis: image

amadio: Yeah, it's a mess between RPMs, Python, and git describe to keep things sane. The Python setup.py sanitization is here:

https://github.com/xrootd/xrootd/blob/ddb061816b5044d4a13b773979b3c9a76ab8963e/setup.py#L26-L39

It has two problems, 5.9 is outdated, and the file is not closed properly 😞

PR #2772, opened 2026-04-22
[XrdSciToken] Security Hardening

apeters1971:
These vulnerabilities were identified and fixed by Opus 4.6. Let me know if I should stash all commits and add this information to the commit message.

### DIFF
--- a/src/XrdSciTokens/XrdSciTokensAccess.cc
+++ b/src/XrdSciTokens/XrdSciTokensAccess.cc
@@ -9,6 +9,7 @@
 #include "XrdTls/XrdTlsContext.hh"
 #include "XrdVersion.hh"
 
+#include <cctype>
 #include <map>
 #include <memory>
 #include <mutex>
@@ -174,6 +175,19 @@ std::string AccessRuleStr(const AccessRulesRaw &rules) {
        first = false;
     }
     return ss.str();
+}
+
+// Returns true iff every character in the string is a valid POSIX username
+// character: [A-Za-z0-9._@-], with a non-empty length and no leading '-'.
+// This prevents attacker-controlled JWT claim values from being forwarded as
+// OS usernames containing path separators, shell metacharacters, or null bytes.
+bool IsSafeUsername(const std::string &name) {
+    if (name.empty() || name[0] == '-') return false;
+    for (unsigned char c : name) {
+        if (!isalnum(c) && c != '_' && c != '.' && c != '@' && c != '-')
+            return false;
+    }
+    return true;
 }
 
 bool MakeCanonical(const std::string &path, std::string &result)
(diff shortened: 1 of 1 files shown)


amadio:
We do not yet have documentation on how to handle AI-based contributions, but the Linux kernel did come up with something reasonable that we can try to follow as well:
- https://docs.kernel.org/process/coding-assistants.html#attribution

Could you please add the appropriate `Assisted-by: ...` tags as described above to these commits? Thank you!

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -257,10 +272,17 @@ struct MapRule
...
 
-        if (!m_path_prefix.empty() &&
-            strncmp(req_path.c_str(), m_path_prefix.c_str(), m_path_prefix.size()))
-        {
-            return "";
+        if (!m_path_prefix.empty()) {
+            if (req_path.compare(0, m_path_prefix.size(), m_path_prefix)) {
+                return "";
+            }
+            // Path prefixes must match on full path components, not partial names.
```
amadio: I'm not sure about this one, actually. We have to be mindful of Ceph object stores too, so that `/atlas:file.root` can be read with a token containing a scope of `storage.read:/atlas`.

amadio (commented):
Looks good now, can you please adjust the commit messages? We just need to resolve the doubt I have about Ceph (i.e. non-hierarchical systems), then we can merge this.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -19,6 +19,8 @@
 #include <unordered_map>
 #include <tuple>
 #include <cstdlib>
+#include <cstring>
+#include <ctime>
```
amadio: Is `<ctime>` really needed? Also, the commit message now is out of sync with the changes (no more prefix validation).

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -827,10 +827,17 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
             return false;
         }
         if (expiry > 0) {
-            expiry = std::max(static_cast<int64_t>(monotonic_time() - expiry),
-                static_cast<int64_t>(60));
+            const auto now_wall = static_cast<long long>(std::time(nullptr));
+            const auto remaining = expiry - now_wall;
```
amadio, line 828: Actually, `<ctime>` might be needed for this change, or we could keep the old `monotonic_time()` as before.

src/XrdSciTokens/XrdSciTokensAccess.cc
amadio: Ok, the conclusion is that we do fix this one, as in SciTokens the paths are considered hierarchical. See e.g. commits 50c540afecb414b8cfea03badc9d376b8494c5d1, and 548fdc2a153514eca099ecd49bac96fdbf74632d.

src/XrdSciTokens/XrdSciTokensAccess.cc
amadio: You are still including `<ctime>` on the wrong commit.

amadio:
I've force-pushed to the branch here to be based on v5.9.x so that we can get the fixes for both 5.9.3 and 6.0.1 later when I merge 5.9.3 into it. Once the CI runs, I will merge this. Thank you!

PR #2775, opened 2026-04-27
Improve XrdPosix support: add support for links, add openat and fstatat, and fix multiple minor bugs

schwicke:
This patch introduces improved support for links in XrdPosix, specifically for XrdPosixPreload. Current override functions for  stat, statx, open etc do check if the current file is to be handled by xrootd or not. If the file is a local link pointing to e.g. EOS, these functions should resolve the link location and check the target rather than the local link itself (exception: lstat).
In addition, this patch introduces overrides for openat and fstatat (albeit they are likely not used much), and implements a few minor bug fixes.

### DIFF
--- a/src/XrdPosix/XrdPosix.cc
+++ b/src/XrdPosix/XrdPosix.cc
@@ -48,11 +48,11 @@
 /******************************************************************************/
 /*                        G l o b a l   O b j e c t s                         */
 /******************************************************************************/
-  
+
        XrdPosixXrootd    Xroot;
 
        XrdPosixXrootPath XrootPath;
-  
+
 extern XrdPosixLinkage   Xunix;
 
 /******************************************************************************/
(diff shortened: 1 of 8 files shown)


amadio:
The ABI check build points to a [removed symbol](https://github.com/xrootd/xrootd/actions/runs/24990757572/job/73175480086?pr=2775#step:11:366). The diff is not clear to me (I don't see where the symbol is removed), but before merging, this needs to be understood/resolved.

schwicke:
Indeed, this needs careful checking. The patch contains a few bug fixes which may have an impact here. Checking on 
``
__fxstat
__lxstat
__lxstat64
XrdPosix_FstatV
__xstat
__xstat64
``
These should not be in the library for this version of Linux I think. Instead, it now has these which were missing before, and therefore not properly overridden I think:
``
fstat
lstat
lstat64
stat
stat64
``
The test failure might be triggered by XrdPosix_FstatV, to be checked though.

amadio:
It would be good to add to the existing tests something that checks what you are fixing here, so that it doesn't break in the future.

schwicke:
I have re-added XrdPosix_FstatV generally albeit it should not be needed on recent Linux versions. This fixes the ABI issue. 
For the tests, I will take a look.

schwicke:
Added tests, based on existing test executables (statx, cat).

amadio (approved):
I'll squash due to some intermediate changes being undone in subsequent commits, but the end result looks fine. Thanks!

PR #2779, opened 2026-05-04
[XrdSys] Include C style headers here

schwicke:
Fixing an issue when trying to link with XrdPosix from C.


amadio:
The errors are CDash failing to receive the upload, and some warnings on Fedora which are unrelated. Thank you, I'm merging it.

PR #2780, opened 2026-05-06
[XrdPosix] Remove duplicate definitions and ensure that XrdPosix.hh works in C context

schwicke:
This patch fixes some issues spotted on Fedora44. 
Specifically, there are issues trying to use XrdPosix.hh from a C context.


amadio:
@schwicke I changed the last commit to retain the header mostly as it was before, just using an `#ifdef` to skip C++ stuff. This is for better compatibility. Please let me know if this works for you.

amadio (approved):
This is ready for merge as far as I'm concerned. It will go into 6.0.2. Thanks!

PR #2781, opened 2026-05-06
[Python] Fix macOS wheel plugin loading

lobis:
Resolve bundled PyPI wheel plugins relative to the library directory on macOS so the client can load authentication modules without relying on DYLD_LIBRARY_PATH.

This also fixes the XrdSecgsiAUTHZVO preload name to match the packaged plugin filename.

Assisted-by: Codex:gpt-5.5

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -5,6 +5,7 @@
 if( PYPI_BUILD )
   if( APPLE )
     set(CMAKE_INSTALL_NAME_DIR "@rpath")
+    set(CMAKE_INSTALL_RPATH "@loader_path")
   else()
     set(CMAKE_INSTALL_RPATH "$ORIGIN")
   endif()


amadio:
This is very hacky stuff, cannot merge it as is. I'd rather investigate why the wheels' `install_name` is not working, because it passes tests that require it to be correct already, in this part of the CI workflow (I know because these same tests failed when I had it wrong):

https://github.com/xrootd/xrootd/blob/3b90ae9d05b1f5d26f0b0bf4eefeb00fe516f321/.github/workflows/python.yml#L169-L176

amadio (changes requested):
This is where the changes should happen, and it should likely be a simple one liner change with the correct RPATH/install_name:

https://github.com/xrootd/xrootd/blob/3b90ae9d05b1f5d26f0b0bf4eefeb00fe516f321/src/CMakeLists.txt#L5-L12

amadio:
Yes, the new changes are exactly what I would expect as a fix, thanks a lot!

amadio (approved):

amadio:
The commit message refers to stuff in the broken implementation, please fix, though.

PR #2782, opened 2026-05-07
Add the resource type tag for non-collections in xml response

### LINKED ISSUE #2783: Add resourcetype in xml response for non collections
### Feature Description

Currently, the XRootD PROPFIND response doesn't include 'resourcetype' in the response for non-collections.

Some webdav clients expect that tag to be there for all responses, not just if it's a collection. As such, I've adjusted the return so it adds the tag in the response for non collections.

This is an issue that showed up in PelicanFS using the aoiwebdav client from python. This is a relatively simple addition, just add the  '\<resourcetype/>' tag for non-collections

turetske:
This is a PR to address #2783


amadio:
Is this needed in 5.9.x as well, or just in 6.x releases? If it's needed in 5.9.x, please rebase the change against the `v5.9.x` branch and target that branch for merging. Thank you!

PR #2788, opened 2026-05-11
[XrdXrootd, XrdHttpTPC] Redirect plugin is called in XrdHttpTPC

### LINKED ISSUE #2767: [XrdHttpTPC] Redirector plugin is not called upon COPY request
### Operating System

Alma 9

### XRootD Version

6.0.0

### Description

The redirect plugin does not seem to work with XrdHttpTPC. 

```
260320 23:00:27 057 cms_Decode: redir redirects ? to server.cern.ch:1094 /store/hello-2
260320 23:00:27 065 anon.0:37@uaf-2 http_Protocol: Sending resp: 307 header len:1381
```

It works properly with HTTP GET.
I'll get in touch with the original poster to get more information on how this plugin should be configured.

More about this plugin: https://indico.cern.ch/event/1542060/contributions/6710219/

_No response_

### Additional Information

_No response_

ccaffy:
Fixes #2767

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
@@ -8,7 +8,9 @@
 #include "XrdSys/XrdSysFD.hh"
 #include "XrdVersion.hh"
 
+#include "XrdXrootd/XrdXrootdRedirHelper.hh"
 #include "XrdXrootd/XrdXrootdTpcMon.hh"
+#include "XrdOuc/XrdOucPrivateUtils.hh"
 #include "XrdOuc/XrdOucTUtils.hh"
 #include "XrdHttpTpc/XrdHttpTpcUtils.hh"
 #include "XrdHttp/XrdHttpUtils.hh"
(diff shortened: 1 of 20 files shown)


src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -4111,107 +4112,48 @@ int XrdXrootdProtocol::fsRedirNoEnt(const char *eMsg, char *Cgi, int popt)
...
+                                             *(Link->AddrInfo()),
+                                             outTarget, errMsg);
        newPort = static_cast<int>(TPort);
       } else {
 
-// This is a url which requires additional handling. If the url is not
-// valid we skip calling the plugin as the situation is not salvageable
-// and the client will complain. We also require that the host and optional
-// port be atleast two characters long.
+// URL form. Parse scheme://host[:port][/tail] before handing the pieces to
```
abh3: So, this is where I am confused. There are two cases: 1) the redirect is of the form of naked host and numeric port, and 2) the redirect is of the form of a full URL where the numeric port is less than 0. It is obvious that case 1 only applies to xroot protocol and case 2 is common to both http/tpc and xroot protocol. So, one would think the abstract would be the common case. So, the splitting of the url should have been inside of RedirectHelper, should it not be so? I mean the http/tpc would simply call RedirectHelper with the full URL and xroot would do the same if it was handed a full URL. In each case it would get back a result as you have here so that the final action can be taken. Now, since we cache host names for redirects to avoid hitting DNS every time, that part should also be in redirect helper. That would also mean the redirecr helper would have a method to handle case 1 as well since the big chore here is handling the caching.

So, the XRootdXeq part should be simply callingthe redirect helper (one of two methods) that would do all the work and hand back the result that XrootdXeq would process. The same would occur in http/tpc except that it really has only one method to call -- the one for handling a full URL. That would really simplify all of this.

Am I missing something here?

abh3 (commented):
I have a big question, see my comment.

src/XrdXrootd/XrdXrootdXeq.cc
ccaffy: The splitting of the URL could indeed go into the RedirectHelper yes that would simplify the calling sites. The caching of names is in LookupTarget(...) which is called by the two RedirectHelper::Apply() and RedirectHelper::ApplyURL(). I'll push a new version with the URL splitting in the RedirectHelper.

ccaffy: Done, the code has been simplified :)

src/XrdHttpTpc/XrdHttpTpcTPC.cc
abh3, line 1045: I know you did not write this part but it is somewhat concerning. This is in reference to the original file XrdHttpTpcTpc.cc lines 1040 to 1045. My inline comment landed in the wrong place. 

First, "brackets" which are being eliminated are only applicable to IPv6 addresses and here it could be one or the other and no differentiation is made. Secondly, there is an option for this case in format(). Instead of using XrdNetAddrInfo::noPort you can use XrdNetAddrInfo::noPortRaw where you don't get the port and if it is an IPv6 address you don't get the brackets either. Making this odd looking suspicious code superfluous. Could you check what this is really trying to accomplish and why it is asking for all IP addresses (which is expensive) and using only the first one rather than just asking for the canonical address?

abh3 (approved):
OK, this looks good. I assume you checked if SonarCloud had anything reasonable to say and acted appropriately. I must admit the inline comment boxes make a review more challenging. I left one comment for a follow-up on some code you did not write/change but looks just plain wrong to me. Please look at that and maybe you can explain what it is really doing.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
amadio: Hi Andy, I noticed with Coverity that there was something wrong at that point in the code, and created commit https://github.com/xrootd/xrootd/commit/145b6e5db32dbcbc4312f1ac9c41e5a760e77e4a to fix it exactly by using `noPortRaw` instead of removing the brackets by hand. This is something that slipped through review when `tpc.fixed_route` got implemented. I put the commit on the `v5.9.x` branch, but that was merged into master and released both in 5.9.4 and 6.0.2 (you can see that by clicking on the commit and checking which tags contain it). I guess this work just needs to be rebased on the latest master now.

ccaffy: Hi @abh3, I did all the SonarCloud-recommended fixes, this tool is AMAZING !

Now, about your inline comment:
```
Could you check what this is really trying to accomplish and why it is asking for all IP addresses (which is expensive) and using only the first one rather than just asking for the canonical address?
```

Actually, we do not need to do that I guess? What if we use the local socket address of the incoming HTTP connection ? Then we would not need to do any address resolution! What `tpc.fixed_route` really needs: “which local IP did the client connect to?” not “what does the Host header resolve to?”

The code would look like:

```cpp
if (m_fixed_route) {
      char ip[64];
      char ipType = 0;

      XrdNetAddrInfo *addrInfo = req.GetSecEntity().addrInfo;
      int sockFD = addrInfo ? addrInfo->SockFD() : -1;

      if (sockFD < 0 || XrdNetUtils::GetSokInfo(-sockFD, ip, sizeof(ip), ipType) < 0) {
          std::stringstream ss;
          ss << "Failed to determine local address of incoming request";
          rec.status = 500;
          logTransferEvent(LogMask::Error, rec, "PULL_FAIL", ss.str());
          return req.SendSimpleResp(rec.status, NULL, NULL,
                                    generateClientErr(ss, rec).c_str(), 0);
      }

      logTransferEvent(LogMask::Info, rec, "LOCAL IP", ip);
      curl_easy_setopt(curl, CURLOPT_INTERFACE, ip);
  }
```

Please tell me if you're happy with my proposal and then I can create another issue and another PR to fix this :)

amadio: @abh3 Please let us know your comments here. I'd like to merge this for 6.0.3. Thanks.

abh3: Congratulations! Indeed, you have found the simplest way of handling fixed_route. Now, given that the original mechanism really didn't do the right thing anyway and ignored errors in the process, we ae left with the decision of what todo when an error occurs. Perhaps this should not be treated as an error and fixed_route should be treated as I would like but it isn't absolutely necessary type of option. In that case, we issue the message to the log and skip setting CURLOPT_INTERFACE. What do you all think?

Also please change the message from "Failed to determine local address of incoming request" to Failed to determine local address of incoming fixed route request" so we know the error involves the fixed_route option.

amadio: Yes, the code above further simplifies what we already have, which is shown below after my fix for the leaks I saw in Coverity:

https://github.com/xrootd/xrootd/blob/be3016068dcfce5f58d284498364e0dcdae97948/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L1027-L1048

@ccaffy You can add the proposed modification above as a separate commit to this pull request if you want. After thinking a bit more about it, I will release 6.0.3, then I will merge this to master and release it with 6.1.0. Thank you!

abh3: OK, this is a totally confusing response. So what was the actual resolution? I thought we would go with SockInfo only. This seems to say otherwise.

amadio: Ok, talking to @ccaffy, this will be in a separate pull request, so I will just merge this once I start getting ready to release 6.1.

ccaffy: SockInfo is the right way @abh3 ;) I'll put this in a different commit and PR! Thanks for the feeback! Fine for having tpc.fixed_route as a best-effort option.

PR #2789, opened 2026-05-11
[Python] Add native fsspec protocol support

lobis:
## Summary
- add a native fsspec filesystem for root, xroot, roots, and xroots URLs
- expose sync and async filesystem operations backed by XRootD.client
- document the integration and cover protocol aliases, URL parsing, async paths, errors, range reads, and vector reads

## Validation
- python3 -m py_compile python/libs/fsspec.py python/tests/test_fsspec.py
- PYTHONPATH=/tmp/xrootd-pytest python3 -m pytest python/tests/test_fsspec.py -q
- python3 setup.py --name --version
- python3 python/setup.py --name --version
- git diff --check

### DIFF
--- a/python/docs/source/index.rst
+++ b/python/docs/source/index.rst
@@ -29,10 +29,10 @@ API Reference
    :maxdepth: 1
 
    modules/client/filesystem
+   modules/client/fsspec
    modules/client/file
    modules/client/copyprocess
    modules/client/responses
    modules/client/flags
    modules/client/url
    modules/client/utils
-
--- a/python/docs/source/modules/client/fsspec.rst
+++ b/python/docs/source/modules/client/fsspec.rst
@@ -1,0 +1,30 @@
+``XRootD.fsspec``: fsspec integration
+=====================================
+
+The Python bindings can register the ``root``, ``xroot``, ``roots``, and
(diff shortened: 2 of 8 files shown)


amadio:
Looks pretty simple, but mostly looks good. If we iterate a bit this can be a nice addition to XRootD 6.1! I will write a proper review tomorrow.

lobis:
A good way to spot any issues is to run the full `uproot` test suite with our fsspec implementation.

I have this running here: https://github.com/lobis/uproot5/pull/1

So far it has highlighted some potential issues that I will address. I will let you know once its fully green. I will squash into a single commit before requesting your review again.

abh3:
OK, I do understand that we have a catch-22 here. However, it doesn't seem right to duplicate xroot specific fsspec code into our repo when that exact same code exists in another repo. This is just asking for inconsistency when both repo diverge which I guarantee will happen. So, let's rethink this. For instance, we could make the xroot fsspec port in the fsspec repo a submodule of our repo or, likely better, the other way around (assuming the fsspec repo owners agree). So, let's rethink how to eliminate duplication.

lobis:
After some offline discussions I will close this PR for the moment as this is not currently on the roadmap.

PR #2798, opened 2026-05-21
[Server] Encrypt sessionid to prevent spoofing

abh3:
This patch encrypts the session id used in the kXR_bind and kXR_endsess requests. The information itself does not need to be secret so we can use blowfish ecb for simplicity. However, by using a large enough key (i.e. 256 effective  bits) makes it very difficult to spoof a session id without knowing the encryption key and cracking the key would take more time than its lifetime. While the standard solution would be to sign the session id; backward compatibility makes it impossible to take this route. So, a size preserving encryption mechanism is used instead.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -86,6 +86,7 @@ if( NOT XRDCL_ONLY )
   target_link_libraries(XrdServer
     PRIVATE
     XrdUtils
+    XrdCryptoLite
     ${CMAKE_DL_LIBS}
     ${CMAKE_THREAD_LIBS_INIT}
     ${ATOMIC_LIBRARY}
--- a/src/XrdCrypto/CMakeLists.txt
+++ b/src/XrdCrypto/CMakeLists.txt
@@ -18,7 +18,8 @@ target_sources(XrdUtils PRIVATE
 )
 
 add_library(XrdCryptoLite SHARED
-  XrdCryptoLite.cc  XrdCryptoLite.hh
+  XrdCryptoLite.cc       XrdCryptoLite.hh
+  XrdCryptoLite_bfecb.cc XrdCryptoLite_bfecb.hh
   XrdCryptoLite_bf32.cc
 )
 
--- a/src/XrdCrypto/CMakeLists.txt
+++ b/src/XrdCrypto/CMakeLists.txt
@@ -19,7 +19,7 @@ target_sources(XrdUtils PRIVATE
 
 add_library(XrdCryptoLite SHARED
   XrdCryptoLite.cc       XrdCryptoLite.hh
-  XrdCryptoLite_bfecb.cc XrdCryptoLite_bfecb.hh
+  XrdCryptoLite_BFecb.cc XrdCryptoLite_BFecb.hh
   XrdCryptoLite_bf32.cc
 )
 
--- a/src/XrdCrypto/XrdCryptoLite_BFecb.cc
+++ b/src/XrdCrypto/XrdCryptoLite_BFecb.cc
@@ -1,0 +1,65 @@
+/******************************************************************************/
+/*                                                                            */
+/*                X r d C r y p t o L i t e _ B F e c b . c c                 */
+/*                                                                            */
+/* (c) 2026 by the Board of Trustees of the Leland Stanford, Jr., University  */
+/*   Produced by Andrew Hanushevsky for Stanford University under contract    */
+/*              DE-AC02-76-SFO0515 with the Department of Energy              */
+/*                                                                            */
+/* This file is part of the XRootD software suite.                            */
+/*                                                                            */
+/* XRootD is free software: you can redistribute it and/or modify it under    */
+/* the terms of the GNU Lesser General Public License as published by the     */
+/* Free Software Foundation, either version 3 of the License, or (at your     */
+/* option) any later version.                                                 */
+/*                                                                            */
+/* XRootD is distributed in the hope that it will be useful, but WITHOUT      */
+/* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or      */
+/* FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public       */
+/* License for more details.                                                  */
+/*                                                                            */
+/* You should have received a copy of the GNU Lesser General Public License   */
+/* along with XRootD in a file called COPYING.LESSER (LGPL license) and file  */
+/* COPYING (GPL license).  If not, see <http://www.gnu.org/licenses/>.        */
+/*                                                                            */
+/* The copyright holder's institutional names and contributor's names may not */
+/* be used to endorse or promote products derived from this software without  */
+/* specific prior written permission of the institution or contributor.       */
+/******************************************************************************/
+
+#include <openssl/blowfish.h>
+
+#include "XrdCrypto/XrdCryptoLite_BFecb.hh"
+
+/******************************************************************************/
+/*                           C o n s t r u c t o r                            */
+/******************************************************************************/
+
+XrdCryptoLite_BFecb::XrdCryptoLite_BFecb(const unsigned char* key,
+                                               unsigned int   keylen)
+{
+// Allocate and set the key. This is an expensive operation so we do it once.
+//
+   bfKey = new BF_KEY;
+   BF_set_key(bfKey, keylen, key);
+}
+
+/******************************************************************************/
+/*                            D e s t r u c t o r                             */
+/******************************************************************************/
+
+XrdCryptoLite_BFecb::~XrdCryptoLite_BFecb() {delete bfKey;}
+
+/******************************************************************************/
+/*                                 A p p l y                                  */
+/******************************************************************************/
+
+void XrdCryptoLite_BFecb::Apply(const unsigned char* in8, unsigned char* out8,
+                                bool encrypt)
+{
+   int action = (encrypt ? BF_ENCRYPT : BF_DECRYPT);
+
+// Perform the action
+//
+   BF_ecb_encrypt(in8, out8, bfKey, action);
+}
--- a/src/XrdCrypto/XrdCryptoLite_BFecb.cc
+++ b/src/XrdCrypto/XrdCryptoLite_BFecb.cc
@@ -27,28 +27,83 @@
 /* specific prior written permission of the institution or contributor.       */
 /******************************************************************************/
 
-#include <openssl/blowfish.h>
+#include <openssl/evp.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/provider.h>
+#endif
 
 #include "XrdCrypto/XrdCryptoLite_BFecb.hh"
+#include "XrdOuc/XrdOucUtils.hh"
 
 /******************************************************************************/
 /*                           C o n s t r u c t o r                            */
 /******************************************************************************/
 
-XrdCryptoLite_BFecb::XrdCryptoLite_BFecb(const unsigned char* key,
+XrdCryptoLite_BFecb::XrdCryptoLite_BFecb(bool&                aOK,
+                                         const unsigned char* key,
                                                unsigned int   keylen)
+                    : decCTX(0), encCTX(0)
 {
-// Allocate and set the key. This is an expensive operation so we do it once.
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+// With openssl v3 the blowfish cipher is only available via the "legacy"
+// provider. Legacy is typically not enabled by default (but can be via
+// openssl.cnf) so it is loaded here. Explicitly loading a provider will
+// disable the automatic loading of the "default" one. The default might
+// not have already been loaded, or standard algorithms might be available
+// via another configured provider, such as FIPS. So an attempt is made to
+// fetch a common default algorithm, possibly automaticlly loading the
+// default provider. Afterwards the legacy provider is loaded.
 //
-   bfKey = new BF_KEY;
-   BF_set_key(bfKey, keylen, key);
+   static struct loadProviders {
+      loadProviders() {
+         EVP_MD *mdp = EVP_MD_fetch(NULL, "SHA2-256", NULL);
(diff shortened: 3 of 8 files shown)


src/XrdCrypto/CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ target_sources(XrdUtils PRIVATE
 )
 
 add_library(XrdCryptoLite SHARED
-  XrdCryptoLite.cc  XrdCryptoLite.hh
+  XrdCryptoLite.cc       XrdCryptoLite.hh
+  XrdCryptoLite_bfecb.cc XrdCryptoLite_bfecb.hh
```
amadio: The filenames are lower case here, but the actual files have uppercase BF in their name, this is why it's failing everywhere except on macOS, where the filesystem is case-insensitive.

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1186,6 +1187,22 @@ char *XrdOucUtils::parseHome(XrdSysError &eDest, XrdOucStream &Config, int &mode
...
 }
+/******************************************************************************/
+/*                                R a n d o m                                 */
+/******************************************************************************/
+
+void XrdOucUtils::Random(unsigned char* buff, unsigned int inblen)
+{
+   std::random_device rd;
+   std::mt19937 gen(rd());
+   std::uniform_int_distribution<unsigned short> dist(0, 255); // Byte distrib
```
amadio: You are initializing a whole random number generator every time this is called, which is not very good for performance. It's better to create the generator once and reuse later.

src/XrdXrootd/XrdXrootdConfig.cc
```diff
@@ -536,6 +540,14 @@ int XrdXrootdProtocol::Configure(char *parms, XrdProtocol_Config *pi)
           }
       }
 
+// Create the blowfish objects to be used to mask/unmask the session ID
+//
+   {struct {unsigned char key1[16]; unsigned char key2[164];} bf;
+    XrdOucUtils::Random((unsigned char*)&bf, sizeof(bf));
+    XrdXrootd::bfEcb1 = new XrdCryptoLite_BFecb(bf.key1, sizeof(bf.key1));
+    XrdXrootd::bfEcb2 = new XrdCryptoLite_BFecb(bf.key2, sizeof(bf.key2));
```
amadio: Is there anything preventing these from being just file-level static variables? These objects seem to be leaked at the end of the process, and I'd like to fix all such leaks so that we can actually enable a leak checker for the other parts of the code where leaks could lead to running out of memory. However, it will not work unless the whole codebase is clean from memory leaks.

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -907,9 +929,11 @@ int XrdXrootdProtocol::do_Endsess()
 // Extract out the FD and Instance from the session ID
 //
    sp = (XrdXrootdSessID *)Request.endsess.sessid;
+   memcpy((void *)&sessID.Sid,  &sp->Pid,  sizeof(sessID.Sid));
```
amadio: Looks like it should be `&sp->Sid`, no? Is that a typo?

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -100,6 +109,18 @@ struct XrdXrootdSessID
...
+                     XrdXrootd::bfEcb1->Apply((unsigned char*)&Sid, buff);
+                     XrdXrootd::bfEcb2->Apply((unsigned char*)&FD,  buff+8);
+                     memcpy((void*)&Sid, (const void*)buff, sizeof(int)*4);
+                    }
+
+        void UnMask() {unsigned char buff[sizeof(int)*4];
+                       XrdXrootd::bfEcb1->Apply((unsigned char*)&Sid,buff,  false);
+                       XrdXrootd::bfEcb2->Apply((unsigned char*)&FD, buff+8,false);
+                       memcpy((void*)&Sid, (const void*)buff, sizeof(int)*4);
+                      }
```
amadio, line 121: All this raw pointer casting feels rather yucky. Also, `Apply()` is doing both encryption and decryption, so it's hard to look at the above and know which one it is without going back to the actual implementation to check. I'd replace `Apply()` with two separate methods: `Encrypt()` and `Decrypt()` to make it clearer. In fact, I think using a class may be a bit overkill for this. A couple of functions should be sufficient, no?

src/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ if( NOT XRDCL_ONLY )
   target_link_libraries(XrdServer
     PRIVATE
     XrdUtils
+    XrdCryptoLite
```
amadio, line 89: If instead of adding this stuff into `XrdCryptoLite`, if you added it to `XrdUtils`, which already links against `libssl.so` and `libcrypto.so`, then you'd avoid adding the extra link dependency to `XrdServer`.

amadio (changes requested):
I did not check the compatibility implications of this, but I think that the implementation could be improved to not waste resources building a random number generator in each call and to avoid leaking memory. I'd also replace the class used for encryption/decryption with a couple of free-standing functions in `XrdUtils` to avoid adding a new library dependency to `XrdServer`.

src/XrdCrypto/CMakeLists.txt
abh3: Well, that would explain that! Thanks.

src/CMakeLists.txt
abh3: I though of that as well and that was my first take. Then I reconsidered because all of our crypto stuff really should belong in XrdCrypto package but things with local implementations) are put into XrdLryptoLite.so so as not to drag in more than is needed. One could assign the compilation to XrdOucUtils build but I find that very confusing. So, in the end having the dependency to CryptoLite actually says you have a dependency on security stuff, so it seem reasonable.

src/XrdXrootd/XrdXrootdXeq.cc
abh3: Six of one half a dozen of the other. I thought of doing it that way but decided to follow the blowfish model that does it this way. I can change it.

src/XrdXrootd/XrdXrootdXeq.cc
abh3: Good call.

src/XrdXrootd/XrdXrootdConfig.cc
abh3: Sure, I can move it into an anonymous naespace.

src/XrdOuc/XrdOucUtils.cc
abh3: Didn't think much of it given the expected low level use. On the other hand, since they are not thread-safe making it static would require serialization, which I could do. The alternative recommendation is to make it thread local which it totally overkill in this scenario. So, given the low level of use, serialization is the way  to go.

src/CMakeLists.txt
abh3: After looking at it, the blowfish thing needs to be self-contained and carry state. That pretty mucj makes it an object/class for a clean implementation. Sure, we could putr an intermediate library infront of it but that would move the dependency later. I don't see it being bad to have a dependency on server. One could argue the only reason that we wind up having such dependencies is becauuse too much is being put in server -- like the whole xroot protocol (though I'm not against that).

src/XrdXrootd/XrdXrootdXeq.cc
```diff
@@ -93,31 +93,38 @@
...
 /******************************************************************************/
 /*                      L o c a l   S t r u c t u r e s                       */
 /******************************************************************************/
 
+namespace
+{
+XrdCryptoLite_BFecb* get_BFecb()
+                        {unsigned char bfKey[16];
+                         XrdOucUtils::Random(bfKey, sizeof(bfKey));
+                         return new XrdCryptoLite_BFecb(bfKey, sizeof(bfKey));
```
amadio: This is still a raw pointer being leaked at the end of the process.

src/XrdCrypto/XrdCryptoLite_BFecb.cc
```diff
@@ -0,0 +1,75 @@
...
+/*                           C o n s t r u c t o r                            */
+/******************************************************************************/
+
+XrdCryptoLite_BFecb::XrdCryptoLite_BFecb(const unsigned char* key,
+                                               unsigned int   keylen)
+{
+// Allocate and set the key. This is an expensive operation so we do it once.
+//
+   bfKey = new BF_KEY;
+   BF_set_key(bfKey, keylen, key);
```
amadio: This is a [deprecated](https://docs.openssl.org/3.6/man3/BF_encrypt/) interface which is causing compilation warnings, therefore the CI failures. The recommended replacement is [EVP_EncryptInit](https://docs.openssl.org/3.6/man3/EVP_EncryptInit/) and other higher-level interfaces. Please update the implementation to use the new interfaces. Thank you.

src/XrdXrootd/XrdXrootdXeq.cc
abh3: I'm sorry this is not a leak. We do this in many other places where we want an object that will be used elsewhere. Since when is returning "new" considered a leak?

src/XrdCrypto/XrdCryptoLite_BFecb.cc
abh3: EVP_Encrypt is one of the most overkill interfaces on earth for blowfish in the way it is intended to be used here,. Gees! Yeah yeah yeah, deprecated not that it is not used behind EVP. I might as well just do xor masking as bad as it is to get something that is simple.. Also, please change the default in cmake to treat warnings as error so this kind of stuff is caught early in the development phase. If we are turning it on then is should be turned on up front, period!

src/XrdXrootd/XrdXrootdXeq.cc
abh3: To make it more palitable, the pointer has been moved into the class. This pointer must exist until the process terminates Hence, this is totally valid.

src/XrdCrypto/XrdCryptoLite_BFecb.cc
abh3: OK, switched to using EVP the intyerface that no one likes. It's so badly documented that you need several iterations to make it work. It works now with copious comments on why it does.

amadio: @abh3 Sorry for the trouble, but unfortunately, sooner or later this would have been needed anyway, as the older interfaces could be removed. Thank you for updating the implementation. The reason I don't use `-Werror` is that you can see the build through completion and address all the warnings at once rather than just the first one. You can, of course add `-DCMAKE_CXX_FLAGS=-Werror` in your local build to catch this earlier. However, if the warning happens only on another OS, you will not catch it in any case. If the CI fails, you can go to CDash and check what happened. There's always a link at the end of the log on GitHub directly to it, and if there was a warning, there's a message in the log indicating that as well. In addition, the warnings appear in the "Files changed" tab on GitHub too, so between all those, I think it should be enough, no?

abh3:
Seems I can't respond directly on github as the conversation has gotten way to complicated. Yes, @amadio it was trouble, but you are correct that we might as well address now rather than later. My peeve was with the EVP interface that has been replaced by more modern and more efficient interfaces. Unfortunately, none are currently distributed in standard Linux releases. It took way more time to make EVP work than expected (though it seems everyone has issues with it). Anyway, it's done. I hope you don't mind at this point the stain on the dress of an additional dependency for the server.

amadio:
BTW, you can ignore the build "failures". They're just failures to send results to CDash. I will think of a way of addressing this problem. When you make a pull request from a branch in the main repo, the CI runs twice, and CDash fails to receive all the builds in that case. I can see that everything is green otherwise. We could disable sending to CDash if this becomes too much of a problem.

abh3:
Well, maybe, though sending to cdash does give you a much faster entry into the view of what the issue is (well now with copilot help).

Andy

abh3:
I think all of the requested changes have been done. The only problem left is to figure out why libz is not being linked in every release other that almaxx. What so special here? The dependency is listed but seems not to matter for anything but almaxx.

amadio:
@abh3 Given that changes in this area have caused problems for EOS in the past, before the final merge, I'd like to first test this locally within EOS. If all goes well, then we can merge.

xrootd-dev:
Not a problem. This should interacts with the client and the client isn't supposed to look at it (it's defined as opaque). If EOS violated that, we could make this optional, and EOS can turn it off for their cluster.

Andy

amadio:
@abh3 Looks like this works just fine with EOS. I will wait to merge after our meeting tomorrow, though, as I'd like to ask a few final questions about the implementation.

abh3:
You can always add you questions now so that I can formulate a reasonable answer ahead of time :-)

amadio:
@abh3 Ok, I wonder what the performance impact of this would be, and whether using uuid as session id could be just as good, with a lower performance overhead. Probably the masking/unmasking of the session id happens infrequently enough that we don't really have to care, and using uuid we'd be likely changing behavior that EOS relies on, so the easiest option is to simply include this in 6.1 as the fix for the two reports you mentioned. If you don't mind, however, I'd prefer to squash this and remove the "fixes two advisories" note from the commit message.

abh3:
The performance impact is minimal. We encrypt the session ID once upon login. We need to decrypt it for each bind request (not very often used). We also need to decrypt it when endsess is called, which is called only on client reconnects, again fairly uncommon. As for using uuid, not really doable without maintaining a map of uuid to FD/Instance. So, likely about the same amount of overhead overall. However, I am not adverse to changing this to be map based since the actual implementation cannot be anything but always backward compatible by definition. Different servers can have different implementations since servers only accept session ID's that they generated.

abh3:
What was the solution to the zlib conundrum?

amadio:
@abh3 I will work on that after I finish merging other pending things. Hopefully by the end of the day today it will also be merged in. If the only problem is CMake/packaging, I can easily fix it before tomorrow.

PR #2800, opened 2026-05-21
[XrdClHttp] Add a knob to disable prefetch

bbockelm:
The prefetch code is designed for the Pelican use case (typical use trends to whole-files); we are just now starting to test more heavily the sub-file requests and its performance impact.

It was probably a mistake to upstream XrdClHttp leaving this on-by-default; it was definitely a mistake to not make it configurable.

This is a RFC for the latter.  Whether we consider the former correct behavior, needing a bugfix (6.0.x), or changing it is a new feature (6.1.x), I'll defer to the collective wisdom.

### DIFF
--- a/src/XrdClHttp/XrdClHttpFactory.cc
+++ b/src/XrdClHttp/XrdClHttpFactory.cc
@@ -154,6 +154,17 @@ Factory::Initialize()
         }
         XrdClHttp::CurlOperation::SetStallTimeout(stall_timeout);
 
+        // Whether prefetching is enabled by default for newly-opened files.
+        // Per-file overrides are still possible via the XrdClHttpPrefetch property.
+        env->PutInt("HttpPrefetch", 1);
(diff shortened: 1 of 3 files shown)


bbockelm:
@amadio - what should I do with the Sonar Qube issues?

amadio:
@bbockelm For now we are just testing SonarQube, since [Coverity](https://scan.coverity.com) is down since over two weeks. If you think the comments from SonarQube are good, please address them, but it's not necessary. However, this pull request is only hiding the problem, I think it's better to fix prefetching support instead of disabling (we can still disable by default if that's not going to hurt performance too much).

amadio:
I believe this is no longer necessary, due to the recent fixes introduced by @dynamic-entropy . Can you confirm, @dynamic-entropy?

amadio:
Closing due to inactivity. Please feel free to reopen if needed.

PR #2801, opened 2026-05-21
Add the resource type tag for non-collections in xml response

turetske:
This is a PR to address https://github.com/xrootd/xrootd/issues/2783 in the 5.9.x branch.

Actually choosing the proper branch this time.


amadio:
Ok, I will pick this up. However, I did ask in the previous pull request if you wanted this on both branches and did not receive a reply.

PR #2802, opened 2026-05-23
[XrdCrypto/XrdTls/XrdVoms] Adapt to OpenSSL 4.0

ellert:


amadio:
Hi @ellert. Thank you very much for this. As the patch is not trivial, and the coverage of VOMS in the CI is not very good, this might take a bit more time than usual to review/merge. Is this change required only on Fedora Rawhide for now? I suppose so, as we did build the latest versions for Fedora 44 recently.

src/XrdCrypto/XrdCryptosslgsiAux.cc
```diff
@@ -823,16 +847,32 @@ int XrdCryptosslX509CreateProxyReq(XrdCryptoX509 *xcpi,
...
    if ((i2d_PROXY_CERT_INFO_EXTENSION(pci.get(), &pp)) <= 0) {
       PRINT("problem converting data for extension");
       return -kErrPX_Error;
    }
+   ASN1_OCTET_STRING *os = ASN1_OCTET_STRING_new();
+   if (!os) {
+      PRINT("could not allocate data field for extension");
+      return -kErrPX_NoResources;
+   }
+   if (ASN1_OCTET_STRING_set(os, data, len) == 0) {
```
abh3, line 873: This looks like a memory leak where "os" and "data" varaiables can remain allocated even if the request fails. Is this intentional?

src/XrdCrypto/XrdCryptosslgsiAux.cc
```diff
@@ -1140,16 +1184,32 @@ int XrdCryptosslX509SignProxyReq(XrdCryptoX509 *xcpi, XrdCryptoRSA *kcpi,
...
+      PRINT("could not allocate data field for extension");
+      return -kErrPX_NoResources;
```
abh3, line 1219: Same problem here, os is still allocated upon an error return.

abh3 (commented):
Just a couple of observations about possible memory leaks. I am suprised how little depends on the actual version of OpenSSL being < 4. So, that brings to question what is the minimum version you need to have for all the changes to actually compile?

src/XrdCrypto/XrdCryptosslgsiAux.cc
ellert: Fixed.

src/XrdCrypto/XrdCryptosslgsiAux.cc
ellert: Fixed.

ellert:
> what is the minimum version you need to have for all the changes to actually compile?

The code compiles with the changes on RHEL 8 which has openssl 1.1.1k. This is consistent with the current minimal required version:
https://github.com/xrootd/xrootd/blob/a8fb957c302b89330d6b43bc23be4ca8d6a583e8/cmake/XRootDFindLibs.cmake#L31
As far as I am aware the changes should be fine back to 1.1.0. It would require some tweeks for openssl 1.0 since ASN1_STRING_get0_data was new in openssl 1.1.0.

abh3:
Thanks! That means there really are no practical compatibility problems with this patch.

amadio:
> Thanks! That means there really are no practical compatibility problems with this patch.

Yes, CentOS 7 was the last OS needing OpenSSL 1.0.x, but after its EOL I've removed support for OpenSSL 1.0.x in XRootD. Now OpenSSL 1.1.1, which is the version on Alma 8 as observed by @ellert, is the earliest version we need to support.

PR #2806, opened 2026-05-29
Streaming real-time checksums

abh3:
This pull request implements real-time streaming checksums. This feature is turned off but may be enabled via configuration options. All checksums are eligible to be streamed. However, only those that can algorithmically combine two checksums from adjacent blocks can be efficiently computed when data arrives out of order. This includes adler32m crc32, and crc32c. However, the current implementation only includes adler32. Other checksums use a somewhat less efficient read-back method to handle out of order data. A real-time (a.k.a. streaming) checksum is treated just like an offline checksum. Once computed a query for the checksum does not trigger another offline computation for the checksum but simply uses the real-time one. Since checksums are implemented at the Sfs layer (i.e. implemented by an Ofs plug-in) they apply to all possible storage types that are implemented by Oss layer plug-ins. Directives dealing with real-time checksums are described in the "Open File System & Open Storage System Configuration Reference". Note that this pull request is based on an older copy of master and may generate misleading ABI incompatibility warnings. Nothing in this pull request breaks ABI.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -32,6 +32,7 @@ set_target_properties(XrdUtils
 target_link_libraries(XrdUtils
   PRIVATE
     OpenSSL::SSL
+    ZLIB::ZLIB
     ${CMAKE_THREAD_LIBS_INIT}
     ${CMAKE_DL_LIBS}
     ${SOCKET_LIBRARY}
--- a/src/XrdCks/CMakeLists.txt
+++ b/src/XrdCks/CMakeLists.txt
@@ -5,7 +5,6 @@ target_sources(XrdUtils
     XrdCksCalccrc32C.cc  XrdCksCalccrc32C.hh
     XrdCksCalcmd5.cc     XrdCksCalcmd5.hh
     XrdCksConfig.cc      XrdCksConfig.hh
-    XrdCksFile.cc        XrdCksFile.hh
     XrdCksLoader.cc      XrdCksLoader.hh
     XrdCksManager.cc     XrdCksManager.hh
     XrdCksManOss.cc      XrdCksManOss.hh
--- a/src/XrdCks/CMakeLists.txt
+++ b/src/XrdCks/CMakeLists.txt
@@ -5,6 +5,7 @@ target_sources(XrdUtils
     XrdCksCalccrc32C.cc  XrdCksCalccrc32C.hh
     XrdCksCalcmd5.cc     XrdCksCalcmd5.hh
     XrdCksConfig.cc      XrdCksConfig.hh
+    XrdCksFile.cc        XrdCksFile.hh
     XrdCksLoader.cc      XrdCksLoader.hh
     XrdCksManager.cc     XrdCksManager.hh
     XrdCksManOss.cc      XrdCksManOss.hh
--- a/src/XrdCks/XrdCksCalcadler32.hh
+++ b/src/XrdCks/XrdCksCalcadler32.hh
@@ -41,7 +41,7 @@
                    * Copyright (C) 1995-1998 Mark Adler
    Below are the zlib license terms for this implementation.
 */
-  
+
 /* zlib.h -- interface of the 'zlib' general purpose compression library
   version 1.1.4, March 11th, 2002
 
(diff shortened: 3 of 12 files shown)


amadio:
@abh3 Please rebase on master, your branch is many commits behind.

abh3:
The rebase has been completed.

amadio:
@abh3 Due to the usage of `zlib` in `XrdCks`, I had to link `XrdUtils` with `ZLIB::ZLIB`. I suppose we should squash this to prevent broken intermediate state (which breaks git bisect). What do you think?

amadio:
@abh3 The ABI build failed, with reason, there are changes to the size of XrdOfs class, and some members in other places that changed. There's also a warning about a `char` changed to `bool`, but that can be safely ignored. The ABI breakage only affects XrdServer (i.e. EOS), not XrdUtils (which is shared with the client), so we can probably merge despite the ABI issues, as long as no more ABI-breaking changes are necessary in this area later on. EOS has no release with XRootD 6.x yet, so as long as the first one requires XRootD 6.1 or later, we should be ok.

One problem, however, is that this adds a lot of code paths not covered by any tests. Maybe consider using AI to add some tests for the functionality included here. I'm hesitant to merge this in 6.1 in its current state. I asked Copilot to help with the review and it flagged the following points:

- Code coverage: Non-combinable checksum re-read path needs testing with out-of-order I/O
- Memory usage: Segment map can grow with heavily out-of-order writes — consider bounds checking
- Error handling: Some edge cases in RTC_NCXX() (e.g., buffer allocation failures) could propagate silently

It also suggests the following potential improvements:

- Add bounds checking on segment map: Consider a maximum size limit or LRU eviction for the `std::map<off_t, inSeg>` to prevent unbounded memory growth with pathological out-of-order I/O patterns.
- Enhanced error propagation: The `Dirty` flag silently disables checksum processing on errors. Consider logging more details about which write operation triggered the failure.
- Configuration validation: Add a check in `ConfigCksRT()` to verify that the selected cipher actually supports the `Combinable()` method before relying on it.
- Test coverage for CGI parameter: The `chkcgi` path in `SetupCksRT()` could benefit from explicit test scenarios for parameter validation and cipher availability.

So, looking at the above, I wonder what will happen when someone uploads a 1TB file with multi-stream `xrdcp`? This does happen in EOS, we have the database backups which often reach 1.6TB in size. Without bounds checking or some other measure, it seems this could cause problems in production if enabled.

My suggestion would be to add this into 6.2 and use the time between now and then to add more tests and run this within EOS in a test instance as well. What do you think?

abh3:
I will look as there should not have been any ABI changes and certainly no change to the OFS class. As for the AI suggestion, some of them are totally ridiculous as they are already there, it just seems to ignore them. I will give you info today.

abh3:
OK, I don't see why this matters as XrdOfs.hh is a private header. It is the implementation for a public header. It is a standalone plugin that is instantiated in a private class. So, size changes do not affect anyone except the objects that reference it in the release in which it is contained.  It cannot leak into any other release. Why is the ABI checker let loose on this?

The change from char to bool was., again, made for consistency. A) the size was not changed, and b) the char was simulating a bool so it should have been a bool in the first place. The logic did not change whatsoever.

So, please explain where the concern comes from.

amadio:
The ABI check covers both private and public headers now, I will consider adjusting it to only cover things in public headers, but we may let a change that breaks EOS slip up in this case, so I'd rather selectively enforce that it passes depending on what it reports as problem. As I mentioned above, the changes introduced here are ok, what concerns me more is the lack of testing for changes that are not simple. I will merge this into 6.1, since it's disabled by default, but I hope that the second half of the work comes with tests for the new functionality. Cheers,

amadio:
I have squashed the commits to prevent broken intermediate state.

abh3:
Just for the record, we had an agreement with the EOS team when we allowed internal header to be exposed to EOS and then calling them private headers. Thea agreement was that they were well aware than we could change them any time and it was up to them to recompile any programs that used private headers when going to a new release. Any change to this agreement will result is a serious degradation of forward progress on our part. Sure, tell EOS that private headers were changed so they get a heads up but this should not be a roadblock to the release process.

esindril:
Hi Andy,

The established agreement remains firmly in place: EOS functionality is never disrupted by changes made by the XRootD team, whether to private or public headers. All EOS releases maintain a strict dependency on a specific version of XRootD; when necessary, we create a new release compiled with the latest or required XRootD version. Consequently, there is no need for concern regarding EOS compatibility for these types of changes.

Thanks,
Elvin

PR #2809, opened 2026-06-02
[Server] add SSO/OIDC/OAUTH0 token auth plugin XrdSecoidc (prot: oidc)

### LINKED ISSUE #2573: Implement support for native OAuth2 authentication (e.g. CERN SSO tokens)
The use case for this is to use [id tokens](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) [from CERN's SSO](https://auth.docs.cern.ch/user-documentation/oidc/config/) (Keycloak) with [SWAN](https://swan.cern.ch). One could imagine also configuring access based on other online accounts such as from Google, GitHub, etc.

apeters1971:
Add a new XRootD security protocol plugin, XrdSecoidc, that authenticates bearer-style OIDC/OAuth2 JWT access tokens over TLS. The plugin performs JWT validation entirely in-process (no SciTokens dependency): it verifies RS256 signatures against issuer JWKS keys, checks issuer, audience and time-based claims, and maps a configurable token claim to the local XrdSecEntity identity.

Plugin (src/XrdSecoidc/XrdSecProtocoloidc.cc):
- TLS-only: both the client and server entry points reject non-TLS connections.
- RS256-only signature verification; tokens with alg=none, HMAC variants or a missing alg are rejected to prevent algorithm-confusion attacks.
- Per-issuer policy blocks (issuer, audience list, OIDC discovery URL, JWKS URL, forced identity claim). OIDC discovery and JWKS endpoints must use https://, redirects are pinned to https, and response bodies are size-capped.
- JWKS retrieved via libcurl with peer/host verification, refreshed on a configurable interval and optionally persisted to a per-issuer on-disk cache that survives restarts (with falls-back-to-cache behaviour on fetch failure).
- Identity mapping via an ordered claim list (preferred_username, upn, username, name, sub by default) or a per-issuer forced claim; the special forced claim "email" maps addresses to local usernames via an [email-map] section.
- Validated tokens are cached keyed by the SHA-256 of the token, with a bounded size and TTL and soonest-to-expire eviction to keep the auth hot path cheap.
- INI configuration (/etc/xrootd/oidc.cfg or -config-file) with [global], [issuer "..."] and [email-map] sections. Config files must be regular, owner-only, non-symlink (O_NOFOLLOW) files; the issuer and email-map sections are reloaded at runtime on inode/mtime changes (a stat-only fast path avoids re-reading an unchanged file on every authentication).
- Selectable expiry policy (required|optional|ignore); in ignore mode a present-but-expired exp claim is not enforced.
- Optional debug logging of the decoded JWT header/payload or selected claims.

Client token discovery searches XRD_SSO_TOKEN[_FILE], BEARER_TOKEN[_FILE], XDG_RUNTIME_DIR/bt_u<uid> and /tmp/bt_u<uid>; token files must be regular, owner-only and are opened with O_NOFOLLOW. Client debug logging is enabled via the XrdSecDEBUG environment variable.

Helper utilities (utils/xrdsso, utils/xrdsso_show.py) create and inspect tokens for CERN SSO (OAuth2/OIDC) and Google (device/PKCE) flows.

Build and packaging: register the plugin in src/CMakeLists.txt and src/XrdVersionPlugin.hh, and add the Debian/RPM install entries and docs note.

Tests (tests/XrdSec): add a dedicated XrdSec unit-test bucket with a GoogleTest suite (34 cases) covering the JWT validation paths (valid/expired/not-yet-valid, audience string and array match/mismatch, unknown issuer and multi-issuer policies), algorithm-confusion guards (none/HS256/missing alg), kid selection, the three expiry modes, forced and email-mapped identities, and the supporting helpers (Strip, base64url, JWKS parsing, the token cache, the INI parser and runtime reload, and the server Authenticate credential framing).

Assisted-by: Cursor:Opus-4.8 CursorAI


python/docs/source/install.rst
```diff
@@ -166,6 +166,7 @@ In this case, the structure is a bit different than before::
   |   |-- libXrdSecgsiAUTHZVO-5.so
   |   |-- libXrdSecgsiGMAPDN-5.so
   |   |-- libXrdSeckrb5-5.so
+  |   |-- libXrdSecoidc-5.so
```
amadio: The Python build is running because of this line, could you please remove it if you push again? It's not needed to add it here. Thanks.

src/Xrdsecoidc/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+if( XRDCL_ONLY )
+  return()
+endif()
```
amadio: This is obviously wrong, the client also needs the security plugins to be able to authenticate, of course.

src/Xrdsecoidc/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+set(XrdSecoidc XrdSecoidc-${PLUGIN_VERSION})
+
+find_package(CURL REQUIRED)
+find_package(OpenSSL REQUIRED)
+
+add_library(${XrdSecoidc} MODULE ../XrdSecoidc/XrdSecProtocoloidc.cc)
```
amadio: Not sure why the file is added in this way, Claude must be drunk.
Please apply `s@../XrdSecoidc/XrdSecProtocoloidc.cc@XrdSecProtocoloidc.cc@`.

src/XrdSecoidc/README.md
```diff
@@ -0,0 +1,303 @@
...
+## Client token pickup order
+
+The client side `getCredentials()` searches for a token in this order:
+
+1. `XRD_SSO_TOKEN` (raw token value)
+2. `XRD_SSO_TOKEN_FILE` (path to token file)
+3. `BEARER_TOKEN` (raw token value, compatibility fallback)
+4. `BEARER_TOKEN_FILE` (path to token file, compatibility fallback)
+5. `XDG_RUNTIME_DIR/bt_u<uid>`
+6. `/tmp/bt_u<uid>`
```
amadio: I think to follow the path of least surprise, we should do just the same thing as the `XrdSecztn` plugin (and [`oidc-agent`](https://github.com/indigo-dc/oidc-agent)) does in terms of credential discovery, which is to follow what's described in [WLCG Bearer Token Discovery](https://zenodo.org/records/3937438).

tldr; Let's not read `XRD_SSO_TOKEN` and `XRD_SSO_TOKEN_FILE`, but just `BEARER_TOKEN` and `BEARER_TOKEN_FILE` in the initial implementation.

src/XrdSecoidc/README.md
```diff
@@ -0,0 +1,303 @@
...
+audience = public-client
+```
+
+## Standard Google configuration
+
+Example `oidc.cfg` using Google issuer + email mapping:
+
+```ini
+[issuer "https://accounts.google.com"]
+audience = 780271439668-1aukl5va8p6rbf5i81sgpg5ppr8s2p63.apps.googleusercontent.com
```
amadio: This will probably be flagged by secret detection tools, maybe consider changing the example.

src/XrdSecoidc/README.md
```diff
@@ -0,0 +1,303 @@
...
+  environment variable to a truthy value (e.g. `1`, `on`, `true`, `yes`); it
+  logs which token source/file the client selected.
+
+## Local helper scripts
+
+For quick manual testing, helper scripts are available in `utils/`:
+
+- `utils/xrdsso create [tokenfile] [CERN|CERNOIDC|GOOGLE]`: creates token
+  (`CERN` uses OAuth2 access token flow, `CERNOIDC` requests OIDC scopes and
+  requires/stores `id_token`, `GOOGLE` defaults to device flow).
```
amadio: We could follow the command line interface of `oidc-agent` where it makes sense and instead of leaving this in `utils/` we could install this so that users can easily obtain tokens from SSO and use them with XRootD/EOS. This is not needed, though.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+bool hasSuffix(const std::string &s, const char *sfx)
+{
+   size_t n = strlen(sfx);
+   return s.size() >= n && s.compare(s.size()-n, n, sfx) == 0;
+}
+
+bool hasPrefix(const char *s, const char *pfx)
+{
+   return s && pfx && strncmp(s, pfx, strlen(pfx)) == 0;
+}
```
amadio: If possible, let's use `std::string` and C++20's `.starts_with()` and `.ends_with()`. If things fail on Alma 8, we can consider moving to a devtoolset compiler that supports this. I joked about it in the EOS workshop, but this is actually a good reason to use C++20 in the code.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+      {uint32_t v = uint32_t((unsigned char)in[i]) << 16;
+       out.push_back(tbl[(v >> 18) & 0x3F]);
+       if (i + 1 < in.size())
+          {v |= uint32_t((unsigned char)in[i + 1]) << 8;
+           out.push_back(tbl[(v >> 12) & 0x3F]);
+           out.push_back(tbl[(v >> 6) & 0x3F]);
+          } else out.push_back(tbl[(v >> 12) & 0x3F]);
+      }
+   return out;
+}
```
amadio: I'd prefer to avoid code duplication if possible. We already have functions for base64 encoding/decoding in `XrdOucUtils`. At the very least, if these functions are really necessary in addition to what we already have, they should be moved to `XrdOucUtils` so they can be shared with other parts of the code.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+bool getStringClaim(const std::string &json, const char *claim, std::string &val)
+{
+   nlohmann::json obj;
+   if (!parseJsonObject(json, obj)) return false;
+   return getStringClaim(obj, claim, val);
+}
+
+// Strip whitespace and optional Bearer prefix from a length-bounded buffer.
+// maxLen must be the number of usable bytes at bTok; strlen is never called.
+const char *Strip(const char *bTok, int &sz, int maxLen = -1)
```
amadio: We already have a `trim` function somewhere that performs a similar task to this. You can use that here and just strip an optional `Bearer*` prefix here in addition to that. I'd also try to refactor things in terms of `std::string` since this is new code, and avoid raw `const char*` whenever possible. This is Claude following the general coding style, but for new code, idiomatic C++ should be preferred.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+
+   while (sTok < endPtr && isspace(static_cast<unsigned char>(*sTok))) sTok++;
+   if (sTok >= endPtr) return 0;
+
+   while (endPtr > sTok && isspace(static_cast<unsigned char>(*(endPtr-1)))) endPtr--;
+   sz = static_cast<int>(endPtr - sTok);
+   return (sz > 0 ? sTok : 0);
+}
+
+std::atomic<int> expiry{1}; // 1=require, 0=ignore, -1=optional
```
amadio: I think you need to declare these plugin globals as `static` to stop SonarQube from complaining about global variables which are not const. I also think that rather than have several atomic variables, it might be cheaper to use read/write locks to make things thread-safe against configuration updates.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+
+XrdSysLogger OIDCLogger;
+XrdSysError  OIDCLog(0, "secoidc_");
+
+struct CachedTokenEntry {
+   std::string identity;
+   std::string identityMethod; // e.g. "default", "claim:preferred_username", "email-map:foo@bar.com"
+   std::string headerJSON;
+   std::string payloadJSON;
+   uint64_t    expiresAt = 0; // epoch seconds
```
amadio: Use `time_t` here, please.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+      {
+       if (!p) continue;
+       std::lock_guard<std::mutex> lock(p->keysMtx);
+       freeKeys(p->jwksKeys);
+      }
+   IssuerPolicies.clear();
+   IssuerPolicyByIssuer.clear();
+}
+
+std::string joinURL(const std::string &base, const char *sfx)
```
amadio: Let's not mix `std::string` and `const char*` if possible.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+   return s.substr(b, e - b);
+}
+
+std::string toLowerCopy(const std::string &s)
+{
+   std::string out(s);
+   for (size_t i = 0; i < out.size(); ++i)
+      out[i] = static_cast<char>(tolower(static_cast<unsigned char>(out[i])));
+   return out;
+}
```
amadio: We can use more idiomatic C++ here:
```cpp
std::string toLowerCopy(std::string s)
{
    std::transform(s.begin(), s.end(), s.begin(),
        [](unsigned char c){ return std::tolower(c); });
    return s;
}
```

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+
+std::string normalizeEmailKey(const std::string &email)
+{
+   return toLowerCopy(trimCopy(email));
```
amadio: If this is copying twice (likely), we can probably optimize it. For example, we likely won't need the original (non-normalized) version of the email string, we can fix it in place.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+
+bool parseBool(const std::string &s, bool &out)
+{
+   std::string v = toLowerCopy(trimCopy(s));
+   if (v.empty() || v == "1" || v == "true" || v == "yes" || v == "on")
+      {out = true; return true;}
+   if (v == "0" || v == "false" || v == "no" || v == "off")
+      {out = false; return true;}
+   return false;
+}
```
amadio: This looks like a good candidate to go into a more generic place to be reused.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+
+void appendCliOpt(std::string &opts, const std::string &key, const std::string *val = 0)
+{
+   if (!opts.empty()) opts.push_back(' ');
+   opts += key;
+   if (val)
+      {opts.push_back(' ');
+       opts += *val;
+      }
+}
```
amadio: This function looks suspicious to me. Taking a pointer to `std::string` as argument is not great.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+#ifdef CURLOPT_REDIR_PROTOCOLS_STR
+   curl_easy_setopt(c, CURLOPT_REDIR_PROTOCOLS_STR, "https");
+#else
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#elif defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
amadio: You don't need all these warning suppressions. You can switch between the old and new API based on the curl version, like in XrdHttpTpc here:

https://github.com/xrootd/xrootd/blob/ace1d2a5aa3d0e3081a558337a069bfbc57bf05b/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L929-L936

but even switching as you did, the old version obviously will not have the API deprecated, so this part suppressing warnings can be removed.

amadio (commented):
Mostly good. I'm sending my first round of comments so you can review them while I review the rest of the code. Some of this stuff we can work on after merging, in subsequent releases, like moving code around to be shared with other modules.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+   char *bp = (char *)malloc(bsz);
+   if (!bp) return 0;
+   strcpy(bp, "oidc");
+   memcpy(bp+5, tok, tsz);
+   bp[5+tsz] = '\0';
+   return new XrdSecCredentials(bp, bsz);
+}
+
+XrdSecCredentials *XrdSecProtocoloidc::readToken(XrdOucErrInfo *erp,
+                                                 const char *path, bool &isbad)
```
amadio, line 155: This is a good candidate function to be shared later with `XrdSecztn`, same for other similar functions for token discovery. Since the discovery mechanism is the same, we need a way to differentiate `ztn` and `oidc` tokens, so that the right protocol is picked by the client when reading `BEARER_TOKEN`, `BEARER_TOKEN_FILE`, etc. Maybe an optional prefix, like `ztn:<TOKEN>` and `openid:<TOKEN>` for the format, rather than just the raw token. Then the ztn plugin can reject oidc tokens and vice-versa.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+   free(buff);
+   if (!ret) return Fatal(erp, "Token value malformed.", EINVAL);
+   isbad = true;
+   return ret;
+}
+
+XrdSecCredentials *XrdSecProtocoloidc::findToken(XrdOucErrInfo *erp, bool &isbad)
+{
+   static const char *loc[] = {
+     "XRD_SSO_TOKEN", "XRD_SSO_TOKEN_FILE",
```
amadio: Let's remove `XRD_SSO_TOKEN` and `XRD_SSO_TOKEN_FILE` for now, and just stick with the standard WLCG token discovery as I mentioned before. At least for the initial implementation.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,2366 @@
...
+       OIDCConfigPath = selectedCfgPath;
+       OIDCConfigIno = cfgIno;
+       OIDCConfigMTime = cfgMtime;
+       OIDCConfigStatValid = true;
+       OIDCConfigWatch = true;
+      }
+
+   char buff[256];
+   snprintf(buff, sizeof(buff), "TLS:%" PRIu64 ":%d:", opts, MaxTokSize.load());
+   return strdup(buff);
```
amadio: I wonder if this leaks memory or not.

tests/XrdCl/CMakeLists.txt
```diff
@@ -4,10 +4,11 @@ add_executable(xrdcl-unit-tests
   XrdClPoller.cc
   XrdClSocket.cc
   XrdClUtilsTest.cc
-  )
+)
+find_package(CURL REQUIRED)
 
 target_link_libraries(xrdcl-unit-tests
-  XrdTestUtils GTest::gtest GTest::gtest_main)
+  XrdTestUtils GTest::GTest GTest::gtest_main OpenSSL::Crypto CURL::libcurl)
```
amadio: Please keep it as `::gtest`. See c41edb1c11a27a9ba963727b86f22f191818279b.

tests/XrdCl/CMakeLists.txt
```diff
@@ -4,10 +4,11 @@ add_executable(xrdcl-unit-tests
   XrdClPoller.cc
   XrdClSocket.cc
   XrdClUtilsTest.cc
-  )
+)
+find_package(CURL REQUIRED)
```
amadio: Maybe move this to the top of the file.

utils/xrdsso
```diff
@@ -0,0 +1,484 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  cat <<'EOF'
+Usage:
+  xrdsso create [tokenfile] [CERN|CERNOIDC|GOOGLE] [--provider CERN|CERNOIDC|GOOGLE] [--flow device|pkce] [--client-id ID] [--client-secret SECRET] [--scope SCOPE] [--auth-ep URL] [--device-ep URL] [--token-ep URL]
+  xrdsso show   [tokenfile]
```
amadio: How about calling this `xrdtoken`? We can expand it later to get also macaroons and scitokens. It would become a very useful command line tool for clients.

utils/xrdsso
```diff
@@ -0,0 +1,484 @@
...
+      result["code"] = code
+    self.send_response(200); self.send_header("Content-Type", "text/plain; charset=utf-8"); self.end_headers()
+    self.wfile.write(b"Authentication completed. You can close this tab.\n")
+  def log_message(self, *args): pass
+srv = HTTPServer(("127.0.0.1", port), H)
+srv.handle_request()
+print(result["code"])
+if result["error"]:
+  print(result["error"], file=sys.stderr)
+PY
```
amadio: This is an odd way to write a shell script, maybe it's better to write it in Python, or even in C++, to be able to plug in macaroon support later.

utils/xrdsso
```diff
@@ -0,0 +1,484 @@
...
+  if (($# < 1)); then
+    usage
+    exit 1
+  fi
+
+  local cmd="$1"
+  shift
+
+  case "${cmd}" in
+    CERN|CERNOIDC|GOOGLE)
```
amadio: What about GitHub? I hope there's a "standard" way of doing this (getting a token), so that it works across providers.

src/Xrdsecoidc/CMakeLists.txt
amadio: I guess this explains it:
```
gentoo xrootd $ ls src/Xrdsecoidc
CMakeLists.txt
gentoo xrootd $ ls src/XrdSecoidc
README.md  XrdSecProtocoloidc.cc
```
Bad Claude.

src/XrdHttp/XrdHttpProtocol.hh
```diff
@@ -170,6 +170,10 @@ private:
   /// @return 0 if successful, otherwise error
   int HandleAuthentication(XrdLink* lp);
 
+  /// Validate a Bearer OIDC token and populate SecEntity.
+  /// @return 0 if successful or not applicable, otherwise error
+  int HandleOidcAuthentication();
```
amadio, line 175: I suggest to call this `HandleTokenAuthentication()`, and use this also for the other token types, passing the token as input and getting out an `XrdSecEntity` object.

src/XrdOuc/CMakeLists.txt
```diff
@@ -81,3 +82,16 @@ add_library(${XrdN2No2p} MODULE XrdOucN2No2p.cc)
 target_link_libraries(${XrdN2No2p} PRIVATE XrdUtils)
 
 install(TARGETS ${XrdN2No2p} LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
+#-------------------------------------------------------------------------------
+# Shared OIDC JWT validation library
+#-------------------------------------------------------------------------------
+set(XrdOucOIDC XrdOucOIDC-${PLUGIN_VERSION})
```
amadio, line 89: I think this should have `JWT` and/or `OAuth(2)` in the name rather than `XrdOuc`. I will suggest something once I read the code in it.

src/XrdOuc/XrdOucOIDC.cc
```diff
@@ -0,0 +1,2236 @@
...
+bool hasSuffix(const std::string &s, const char *sfx)
+{
+   size_t n = strlen(sfx);
+   return s.size() >= n && s.compare(s.size()-n, n, sfx) == 0;
+}
+
+bool hasPrefix(const char *s, const char *pfx)
+{
+   return s && pfx && strncmp(s, pfx, strlen(pfx)) == 0;
+}
```
amadio, line 111: I'm working to enable devtoolset-14 for Alma 8, which will allow us to start using `.starts_with()` and `.ends_with()` from C++20 in the code. Just FYI, you can rebase later and then drop this.

src/XrdOuc/XrdOucOIDC.cc
```diff
@@ -0,0 +1,2236 @@
...
+std::atomic JwksRefresh{300};
+bool customIdentityClaims = false;
+std::atomic<bool> DebugToken{false};
+std::atomic<bool> DebugTokenClaims{false};
+std::atomic<int> TokenCacheMax{10000};
+std::atomic TokenCacheNoExpTTL{60};
+std::string JwksCacheFile;
+int JwksCacheTTL = 0;
+std::vector<std::string> IdentityClaims = {
+   "preferred_username", "upn", "username", "name", "sub"
```
amadio, line 436: Please drop `preferred_username` and `name` from this list. These fields are not expected to be unique, so they must not be used to map ids. See https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability for more information. The order should also probably be reversed, with `"sub"` first, so `"sub", "username", "upn"`.

src/XrdOuc/XrdOucOIDC.cc
```diff
@@ -0,0 +1,2236 @@
...
+
+std::atomic<int> expiry{1}; // 1=require, 0=ignore, -1=optional
+std::atomic<int> MaxTokSize{8192};
+std::atomic<int> ClockSkew{60};
+std::atomic JwksRefresh{300};
+bool customIdentityClaims = false;
+std::atomic<bool> DebugToken{false};
+std::atomic<bool> DebugTokenClaims{false};
+std::atomic<int> TokenCacheMax{10000};
+std::atomic TokenCacheNoExpTTL{60};
```
amadio, line 432: I wonder if these all really need to be atomic or not...

src/XrdOuc/XrdOucOIDC.cc
```diff
@@ -0,0 +1,2236 @@
...
+
+std::mutex TokenCacheMtx;
+std::unordered_map<std::string, CachedTokenEntry> TokenCache;
+std::atomic<uint64_t> TokenCacheHits(0);
+std::atomic<uint64_t> TokenCacheMisses(0);
+std::mutex JwksDiskCacheMtx;
+
+struct JwksDiskEntry {
+   std::string jwksUrl;
+   time_t fetchedAt = 0;
```
amadio, line 526: Uh, the AI messed up here, I'd put both strings before the `time_t` field to pack it better.

amadio (commented):
Some more comments for you, but the plugin is getting better, which is very good.

src/XrdHttp/CMakeLists.txt
```diff
@@ -28,11 +28,14 @@ set_target_properties(XrdHttpUtils
...
 )
 
+set(XrdOucOIDC XrdOucOIDC-${PLUGIN_VERSION})
+
 target_link_libraries(XrdHttpUtils
   PRIVATE
     XrdServer
     XrdUtils
     XrdCrypto
+    ${XrdOucOIDC}
```
amadio: Either it's a plugin or it's part of `XrdOuc`, i.e of the `XrdUtils` library. Linking in plugins is wrong.

src/XrdHttp/XrdHttpProtocol.cc
```diff
@@ -806,6 +810,11 @@ int XrdHttpProtocol::Process(XrdLink *lp) // We ignore the argument here
 
 
 
+  // Bearer OIDC authentication over HTTPS (via sec.protocol oidc / CIA).
+  if (ishttps && ssldone && oidcHttpMode && HandleOidcAuthentication()) {
```
amadio, line 814: Note that WLCG token profiles are OAuth2, but not OpenID Connect. The distinction between OAuth2 and OIDC is important, the code should reflect that too. The initial version submitted in this pull request only handled OIDC, then was later expanded to handle also WLCG tokens, which are *not* compatible with OIDC specs. This should be ironed out before merging in the code.

src/XrdHttp/XrdHttpProtocol.hh
amadio: This is not addressed. Given that you are targetting to support both OIDC and non-OIDC token types (e.g. WLCG tokens), this should be renamed. It would actually be even better if this could be folded into the existing `HandleAuthentication(XrdLink* lp)` above, but that would be a nice to have if possible.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -49,6 +61,81 @@ const char *TraceID = "Security";
...
+   if ((endPtr - sTok) >= 9 && hasPrefix(sTok, endPtr, "Bearer%20")) sTok += 9;
+   else if ((endPtr - sTok) >= 7 && hasPrefix(sTok, endPtr, "Bearer ")) sTok += 7;
+
+   while (sTok < endPtr && isspace(static_cast<unsigned char>(*sTok))) ++sTok;
+   if (sTok >= endPtr) return nullptr;
+
+   while (endPtr > sTok && isspace(static_cast<unsigned char>(*(endPtr - 1)))) --endPtr;
+   sz = static_cast<int>(endPtr - sTok);
+   return (sz > 0 ? sTok : nullptr);
+}
```
amadio, line 105: This stuff looks a lot like C+, I'd ask the AI for a refactor into idiomatic C++17, and that applies for all new code, actually.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
       }
 
+// Load the security framework when HTTP bearer OIDC is enabled.
+//
+   if (oidcHttpMode && !CIA)
+      {if (!oidcConfigFN)
+          {eDest.Say("Error: http.oidc requires a configuration file path");
+           return false;
+          }
+       if (!(CIA = XrdSecLoadSecService(&eDest, oidcConfigFN)))
```
amadio, line 178: Was this `CIA` really just always `nullptr` before in XrdHttp? It seems to me that `XrdSecLoadSecService` should only ever be called once, and this would add a second call, with unpredictable effects. See below:

https://github.com/xrootd/xrootd/blob/074871a9752c1ee314222572403fd5cad685e69f/src/XrdSec/XrdSecInterface.hh#L652-L653

@abh3, your advice is welcome here, this is a part of the code is still somewhat unclear to me.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+int
+XrdHttpProtocol::HandleOidcAuthentication()
+{
+#undef  TRACELINK
+#define TRACELINK Link
+
+  if (!oidcHttpMode || !CIA) return 0;
+
+  // Client-certificate identity is fixed for the TLS connection.
+  if (SecEntity.name && strncmp(SecEntity.prot, "oidc", 4) != 0) return 0;
```
amadio, line 202: The comment above is not very clear. I think it means you want to skip token authentication if there is already another form of authentication (that is, `SecEntity` is already populated, with certificate currently being the only other option to a token), so I'd clarify the comment a bit for this, especially if we add Kerberos authentication to HTTP.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+
+  std::string bearer;
+  if (!extractBearerToken(CurrentReq, bearer))
+     {if (oidcHttpMode == 2 && oidcBearerTokKey.empty())
```
amadio, line 206: I wonder why `oidcBearerTokKey.empty()` is needed here? If `extractBearerToken()` already returned false, it's because there's no valid token available, no? Also, `extractBearerToken()` doesn't change `oidcBearerTokKey`, is it expected to be already set at this point by something else?

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+         {TRACEI(REQ, " OIDC bearer token required but not provided.");
```
amadio, line 207: This may create a lot of log output with not much value, I'd say.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+          SendSimpleResp(401, nullptr, nullptr, "Authentication required", 0, false);
+          return 1;
+         }
+      return 0;
+     }
+
+  int tlen = 0;
+  const char *tok = stripBearerToken(bearer.c_str(), tlen);
+  if (!tok || tlen <= 0)
```
amadio, line 216: This is C+, please refactor `stripBearerToken`, as much as I dislike the usage of `std::string_view`, this is a good use case for that.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+     {TRACEI(REQ, " OIDC bearer token malformed.");
+      SendSimpleResp(401, nullptr, nullptr, "Authentication failed", 0, false);
```
amadio, line 218: Do we need a body here? Also, the response should likely contain a [`WWW-Authenticate: Bearer`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/WWW-Authenticate). Similarly to how I've done by adding an `Accept:` header in XrdMacaroons here:

https://github.com/xrootd/xrootd/blob/4ec63a24e099f108f1b7af8cd0dd4e41ace6a8c5/src/XrdMacaroons/XrdMacaroonsHandler.cc#L301-L302

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+      return 1;
+     }
+
+  const std::string tokKey = bearerTokenKey(tok, tlen);
+  if (tokKey.empty())
+     {TRACEI(REQ, " OIDC bearer token fingerprint failed.");
+      SendSimpleResp(500, nullptr, nullptr, "Authentication failed", 0, false);
```
amadio, line 225: Is this really a server-side error? Most likely not.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+      return 1;
+     }
+
+  if (!oidcBearerTokKey.empty() && tokKey == oidcBearerTokKey) return 0;
+
+  const int bsz = 5 + tlen + 1;
+  std::vector<char> credBuf(static_cast<size_t>(bsz));
+  strcpy(credBuf.data(), "oidc");
+  memcpy(credBuf.data() + 5, tok, static_cast<size_t>(tlen));
+  credBuf[static_cast<size_t>(5 + tlen)] = '\0';
```
amadio, line 235: This is hurting my eyes 🥲 It can probably be done in a simpler way with with a normal `std::string`. Another side effect of `tok` being the result of C+ stuff above.

I should add that this is the kind of thing that makes reviewing AI code heavy on maintainers, and it's also the kind of thing that I would hope you could review and fix prior to submitting the code...

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+  cred.buffer = credBuf.data();
+  cred.size = bsz;
+
+  XrdOucErrInfo eMsg;
+  XrdSecProtocol *authProt = CIA->getProtocol(Link->Host(), *(Link->AddrInfo()),
+                                            &cred, eMsg);
+  if (!authProt)
+     {int ec = 0;
+      const char *et = eMsg.getErrText(ec);
+      TRACEI(REQ, " OIDC protocol unavailable: " << (et && *et ? et : "unknown"));
```
amadio, line 247: This should very likely come before the `OIDC bearer token required but not provided` which is towards the beginning of the function. There's no point providing a token when OIDC is disabled...

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+      Bridge = nullptr;
+      DoingLogin = false;
+      DoneSetInfo = false;
+      if (!oidcBearerTokKey.empty())
+         TRACEI(REQ, " OIDC bearer token changed; re-authenticating.");
+     }
+
+  if (SecEntity.name) free(SecEntity.name);
+  SecEntity.name = authProt->Entity.name ? strdup(authProt->Entity.name) : nullptr;
+  strncpy(SecEntity.prot, authProt->Entity.prot, sizeof(SecEntity.prot));
```
amadio, line 281: Please use `strlcpy` in new code, instead of `strncpy`.

src/XrdOuc/CMakeLists.txt
amadio: This is still unresolved. The name should reflect also which type of plugin this is (e.g. `XrdAccSciTokens`).

src/XrdOuc/XrdOucOIDC.hh
```diff
@@ -0,0 +1,58 @@
...
+//! Initialize OIDC JWT validation using the same options as sec.protocol oidc.
+//! @return true on success; false with emsg set on failure.
+//------------------------------------------------------------------------------
+bool Init(XrdSysLogger *logger, const char *parms, std::string &emsg);
+
+//------------------------------------------------------------------------------
+//! sec.protocol-compatible initializer. Returns a malloc'd protocol parameter
+//! string on success, or nullptr on failure (details in erp).
+//------------------------------------------------------------------------------
+char *InitSecProtocol(const char *parms, XrdOucErrInfo *erp);
```
amadio, line 35: Why is a `SecProtocol` thing in `XrdOuc`? It seems misplaced here.

src/XrdSecoidc/configs/wlcg-iam.example.cfg
```diff
@@ -0,0 +1,25 @@
...
+#
+# Replace <iam-client-id> with the OAuth client ID registered at your IAM
+# instance. Production VO deployments use VO-specific issuer URLs; the CNAF
+# test instance is shown below.
+#
+# Use together with XrdAccToken (ofs.authlib libXrdAccToken.so) so storage.*
+# scopes and wlcg.ver from the access token are enforced.
+
+[global]
+expiry = required
```
amadio, line 11: Should we offer anything else than forcing tokens to not be expired? I'd rather not provide an option that degrades security (we already have the `*.onmissing allow` which I don't like in scitokens and macaroons).

src/XrdOuc/XrdOucOIDC.cc
amadio: This should be addressed.

src/XrdOuc/XrdOucOIDC.cc
```diff
@@ -0,0 +1,2419 @@
...
+}
+
+std::string OIDCConfigPath = "/etc/xrootd/oidc.cfg";
+bool OIDCConfigWatch = false;
+bool OIDCConfigStatValid = false;
+ino_t OIDCConfigIno = 0;
+time_t OIDCConfigMTime = 0;
+
+XrdSysLogger OIDCLogger;
+XrdSysError  OIDCLog(0, "secoidc_");
```
amadio, line 508: `Secoidc` does not belong in `XrdOuc`. I know I asked to put things which are common facilities into XrdOuc, but the code moved into this new file includes a lot of things that should have stayed in XrdSecoidc. Token caches, configuration parsing for XrdSecoidc, etc, should not be in XrdOuc, as it's not common to other parts of the codebase. Only things which are truly common utilities should be moved here.

src/XrdOuc/XrdOucUtils.cc
```diff
@@ -1566,6 +1567,33 @@ void XrdOucUtils::trim(std::string_view & sv) {
...
+
+bool XrdOucUtils::parseBool(const std::string &str, bool &out)
+{
+   std::string v = toLowerCopy(trimCopy(str));
+   if (v.empty() || v == "1" || v == "true" || v == "yes" || v == "on")
+      {out = true; return true;}
+   if (v == "0" || v == "false" || v == "no" || v == "off")
+      {out = false; return true;}
+   return false;
+}
```
amadio, line 1595: This is a good example of common utility function that does belong in `XrdOuc`.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,366 @@
...
+            erp->setErrInfo(rc, eVec, 2);
+           }
+   return 0;
+}
+
+bool hasSuffix(const std::string &s, const char *sfx)
+{
+   size_t n = strlen(sfx);
+   return s.size() >= n && s.compare(s.size()-n, n, sfx) == 0;
+}
```
amadio, line 60: This is another example that is a good candidate to move to XrdOuc, but without using C+ as done here. Please take also a `const std::string &sfx` or `std::string_view` for the second argument.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,366 @@
...
+
+void clientDebugLog(const std::string &msg)
+{
+   const char *dbg = getenv("XrdSecDEBUG");
+   if (!dbg || !*dbg) return;
+   std::string v(dbg);
+   for (char &c : v) c = static_cast<char>(tolower(static_cast<unsigned char>(c)));
+   if (v == "0" || v == "off" || v == "false" || v == "no") return;
```
amadio, line 68: Why is this not using the `parseBool` helper which you added? It should.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,366 @@
...
+   buff[st.st_size] = 0;
+   int tlen = 0;
+   const char *tok = XrdOucOIDC::StripToken(buff.data(), tlen);
+   XrdSecCredentials *ret = (tok ? retToken(tok, tlen) : nullptr);
+   if (!ret) return Fatal(erp, "Token value malformed.", EINVAL);
+   isbad = true;
+   return ret;
+}
+
+XrdSecCredentials *XrdSecProtocoloidc::findToken(XrdOucErrInfo *erp, bool &isbad)
```
amadio, line 221: Same as with `readToken` above, this duplicates things already in ztn. Only a single version of token discovery should exist, otherwise we have multiple places to where to fix bugs when they appear.

src/XrdSecoidc/XrdSecProtocoloidc.cc
```diff
@@ -0,0 +1,366 @@
...
+      }
+
+   if (mode == 'c')
+      {bool aOK = false;
+       auto prot = std::make_unique<XrdSecProtocoloidc>(parms, erp, aOK);
+       if (aOK) return prot.release();
+       return nullptr;
+      }
+
+   return std::make_unique<XrdSecProtocoloidc>(hostname, endPoint).release();
```
amadio, line 364: There is no point in wrapping in a `std::unique_ptr` here, since it's released right away.

amadio (changes requested):
This is still quite some way from being ready to merge, I think. The reorganization by Cursor put most code into XrdOuc that doesn't really belong there. This is probably the more important thing to fix in the next iteration. Please review each of my comments and reply. There are several comments from last round which were not considered in this iteration.

src/XrdHttp/XrdHttpSecurity.cc
```diff
@@ -81,11 +168,125 @@ bool XrdHttpProtocol::InitSecurity() {
...
+      SendSimpleResp(401, nullptr, nullptr, "Authentication failed", 0, false);
+      return 1;
+     }
+
+  if (!oidcBearerTokKey.empty() || Bridge)
+     {if (Bridge && !Bridge->Disc())
+         {TRACEI(REQ, " OIDC token changed but bridge is busy.");
+          authProt->Delete();
+          SendSimpleResp(503, nullptr, nullptr, "Authentication busy", 0, false);
+          return 1;
```
amadio, line 270: This is a strange thing to happen, and I wonder in which situation it could happen. Maybe add a retry instead.

src/XrdOuc/XrdOucOIDC.cc
amadio: Actually, they don't promise ABI stability with devtoolset, so I've reconsidered this. We will wait a bit more before using devtoolset, so we're stuck on what GCC 8.5 supports for now (which doesn't include `.starts_with()` unfortunately).

src/XrdHttp/XrdHttpSecurity.cc
amadio: Note that returning a `WWW-Authenticate` in the response is mandatory, per [RFC6750](https://www.rfc-editor.org/info/rfc6750/#section-3), so this must be addressed before merge.

apeters1971:
I close this one, in particular because this is all refactored and renamed into a new oauth2 security protocol working with https: and roots: in the same way. Will open for that a new merge request.

amadio:
Ok, but please do not forget the review comments when doing the refactoring. Thanks.

PR #2810, opened 2026-06-03
[XrdClHttp] Cancel prefetch requests when the file is closed

bbockelm:
The prefetch requests should not be allowed to linger when the file is closed.  While they would eventually expire and timeout, they still count as an "active" request, meaning a large number of GET requests that read out partial data from an object would result in a large number of requests in the curl worker threads.  We observed use cases where this ultimately would be enough to hit the maximum number of pending and active requests, causing HTTP GETs to fail.

This was most visible with the HTTP fsspec implementation in Python which implements "stat" by issuing a GET (not a HEAD!), looking for the response status and then closing the TCP socket.  Simply use fsspec to `stat` a few hundred files and then your XCache instance would be unresponsive for a few tens of seconds.

Includes unit test and human before/after verification.  This is a backport of the equivalent work in xrdcl-pelican for 5.x; note this has been relicensed to LGPL to match the xrootd license (in fact, both repos now have the same license to allow easier back/forward porting).  Marking it as draft for now until we get this released on the Pelican side (or, at least, run in a test cache for a few days); however, it's available for comment!

### DIFF
--- a/src/XrdClHttp/XrdClHttpFile.cc
+++ b/src/XrdClHttp/XrdClHttpFile.cc
@@ -27,6 +27,7 @@
 using namespace XrdClHttp;
 
 std::atomic<uint64_t> File::m_prefetch_count = 0;
+std::atomic<uint64_t> File::m_prefetch_cancelled_count = 0;
 std::atomic<uint64_t> File::m_prefetch_expired_count = 0;
 std::atomic<uint64_t> File::m_prefetch_failed_count = 0;
 std::atomic<uint64_t> File::m_prefetch_reads_hit = 0;
@@ -190,6 +191,36 @@ private:
     XrdCl::ResponseHandler *m_handler;
 };
 
+// Response handler installed by File::CancelPrefetch() when a paused prefetch op needs
+// to be woken up just to be torn down.  Holds a strong ref to the PrefetchDefaultHandler
+// (whose lifetime is independent of File) so the cancellation is delivered safely even
+// if File has been destroyed before the worker drains the continue queue.
+//
+// Holds the strong ref via shared_ptr so PrefetchDefaultHandler outlives any in-flight
+// cancellation delivery; otherwise File::~File could destroy the handler before this
+// callback fires.  The 1-byte buffer is unused once Write() short-circuits on
+// IsCancelled(), but a non-null buffer is still required by CurlReadOp::Continue().
+class CancelResponseHandler : public XrdCl::ResponseHandler {
+public:
+    explicit CancelResponseHandler(std::shared_ptr<XrdCl::ResponseHandler> dh)
+        : m_dh(std::move(dh)) {}
+
+    char *GetBuffer() { return m_buf; }
+
+    virtual void HandleResponse(XrdCl::XRootDStatus *status, XrdCl::AnyObject *response) override {
+        std::unique_ptr<CancelResponseHandler> self(this);
+        delete response;
+        if (m_dh) {
+            m_dh->HandleResponse(status, nullptr);
+        } else {
+            delete status;
+        }
+    }
+private:
+    std::shared_ptr<XrdCl::ResponseHandler> m_dh;
+    char m_buf[1];
+};
+
 } // anonymous namespace
 
 // Note: these values are typically overwritten by `CurlFactory::CurlFactory`;
@@ -199,7 +230,57 @@ struct timespec XrdClHttp::File::m_default_header_timeout = {9, 5};
 struct timespec XrdClHttp::File::m_fed_timeout = {5, 0};
 
 
+void
+File::CancelPrefetch() {
+    // If Open() never succeeded, the prefetch handler is null and there is nothing to do.
+    if (!m_default_prefetch_handler) return;
+    // We must observe and modify m_prefetch_op under m_prefetch_mutex, the same lock
+    // that ReadPrefetch() / PrefetchResponseHandler take, to avoid racing with a
+    // concurrent Read that may be appending to the prefetch handler chain.
+    std::shared_ptr<XrdClHttp::CurlReadOp> op;
+    bool was_paused = false;
+    {
+        std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
+        if (m_prefetch_op && !m_prefetch_op->IsDone()) {
+            op = m_prefetch_op;
+            // Set the cancel flag while the lock is held: this happens-before any
+            // subsequent worker-thread read of IsCancelled() in CurlReadOp::Write(),
+            // which is the primary deterministic cancellation channel.
+            op->Cancel();
+            // Snapshot IsPaused() under the lock.  Even with m_is_paused atomic, taking
+            // the snapshot here avoids the worker concurrently transitioning the op
+            // through a pause+deliver cycle that would leave it parked indefinitely.
+            was_paused = op->IsPaused();
+        }
+        // Once we've decided to cancel, disable further prefetch decisions on this file
+        // so a racing Read() can't observe the cancel-in-progress op and try to extend it.
+        m_default_prefetch_handler->m_prefetch_enabled.store(false, std::memory_order_relaxed);
+        m_prefetch_op.reset();
+    }
+    if (!op) return;
+    if (was_paused) {
+        // The op is parked in libcurl waiting for the next Read.  Wake the worker via the
+        // continue queue so it can drain the slot promptly.  CurlReadOp::Continue calls
+        // m_continue_queue->Produce, which may briefly block — done outside the lock.
+        // CancelResponseHandler holds a shared_ptr to m_default_prefetch_handler, so the
+        // delivery is safe even if `this` File is destroyed before the worker drains it.
+        auto *ch = new CancelResponseHandler(m_default_prefetch_handler);
+        op->Continue(op, ch, ch->GetBuffer(), 1);
+    }
+    // If not paused, the worker thread will reach CurlReadOp::Write(), observe IsCancelled(),
+    // return 0 from the write callback, and tear the slot down via CURLE_WRITE_ERROR.
+}
+
 File::~File() noexcept {
+    CancelPrefetch();
+    // Detach the prefetch default handler from this File so any in-flight
+    // PrefetchResponseHandler that wakes up after destruction completes sees a
+    // null File pointer (under the lock) and skips touching File-owned memory.
+    // Must happen after CancelPrefetch (which still needs the live File) and
+    // before any other File-owned member is destroyed.
+    if (m_default_prefetch_handler) {
+        m_default_prefetch_handler->DetachFile();
+    }
     auto handler = m_put_handler.load(std::memory_order_acquire);
     if (handler) {
         // We must wait for all ongoing writes to complete; the XrdCl::File
@@ -289,6 +370,7 @@ File::GetMonitoringJson()
 {
     return "{\"prefetch\": {"
         "\"count\": " + std::to_string(m_prefetch_count) + ","
+        "\"cancelled\": " + std::to_string(m_prefetch_cancelled_count) + ","
         "\"expired\": " + std::to_string(m_prefetch_expired_count) + ","
         "\"failed\": " + std::to_string(m_prefetch_failed_count) + ","
         "\"reads_hit\": " + std::to_string(m_prefetch_reads_hit) + ","
@@ -415,6 +497,10 @@ File::Close(XrdCl::ResponseHandler *handler,
         return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidOp);
     }
     m_is_opened = false;
+
+    // Tear down any outstanding prefetch op so its worker slot is freed immediately
+    // instead of waiting up to the transfer-stall timeout (60s).
+    CancelPrefetch();
 
     std::unique_ptr<XrdCl::XRootDStatus> status(new XrdCl::XRootDStatus{});
     if (m_put_op && !m_put_op->HasFailed()) {
@@ -626,8 +712,18 @@ File::Read(uint64_t                offset,
         }
         return status;
     } else if (m_full_download.load(std::memory_order_relaxed)) {
-        std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
-        if (m_prefetch_op && m_prefetch_op->IsDone() && (static_cast<off_t>(offset) == m_prefetch_offset.load(std::memory_order_acquire))) {
+        bool eof = false;
+        {
+            std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
+            eof = m_prefetch_op && m_prefetch_op->IsDone() &&
+                  (static_cast<off_t>(offset) == m_prefetch_offset.load(std::memory_order_acquire));
+        }
+        // Release the lock before calling the user's handler.  The handler may close
+        // the file synchronously (e.g., XrdClS3DownloadHandler reacts to a 0-byte
+        // chunk by calling File::Close), and File::Close → CancelPrefetch re-acquires
+        // the same non-recursive mutex.  Holding the lock across the callback would
+        // deadlock the curl worker thread.
+        if (eof) {
             if (handler) {
                 auto ci = new XrdCl::ChunkInfo(offset, 0, buffer);
                 auto obj = new XrdCl::AnyObject();
@@ -995,6 +1091,16 @@ File::GetProperty(const std::string &name,
         return true;
     }
 
+    // Pseudo-property: returns the plugin-global monitoring JSON.  Useful from
+    // tests, which are linked against a separate copy of libXrdClHttp and cannot
+    // call File::GetMonitoringJson() directly — the plugin and the test SO each
+    // own their own copies of the static atomics under RTLD_LOCAL.  Routing
+    // through GetProperty crosses the plugin boundary correctly.
+    if (name == "XrdClHttpMonitoringJson") {
+        value = GetMonitoringJson();
+        return true;
+    }
+
     std::shared_lock lock(m_properties_mutex);
     if (name == "LastURL") {
         value = m_last_url;
@@ -1135,7 +1241,8 @@ File::PrefetchResponseHandler::PrefetchResponseHandler(
     File &parent, off_t offset, size_t size, std::atomic<off_t> *prefetch_offset, char *buffer,
     XrdCl::ResponseHandler *handler, std::unique_lock<std::mutex> *lock, time_t timeout
 )
-    : m_parent(parent),
+    : m_default_handler(parent.m_default_prefetch_handler),
+    m_parent(parent),
     m_handler(handler),
     m_buffer(buffer),
     m_size(size),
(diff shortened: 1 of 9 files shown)


bbockelm:
@amadio - the SonarCloud issues are mostly reasonable (ok, I might find a few nitpick-y but why not get things extra clean?).  However, it wants to turn all atomic memory operations to sequentially consistent which is definitely **not** needed here and, I'd argue, bad advice in a lot of cases.

Is there a way to turn that particular check off?

amadio:
Don't worry about SonarQube. I disabled it because indeed it's too noisy. I will run it only on my fork for now and relay any important info. Feel free to fix previous reports, though, if relevant. Next push should be clean as SonarQube will not run.

amadio:
@bbockelm I have a few questions about this patch.
- What is the current status of it within Pelican? Has it been released/tested there?
- Have AI coding assistants been used in the creation of the patch? If yes, please add an appropriate `Assisted-by` tag as described in https://docs.kernel.org/process/coding-assistants.html

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -1135,7 +1241,8 @@ File::PrefetchResponseHandler::PrefetchResponseHandler(
     File &parent, off_t offset, size_t size, std::atomic<off_t> *prefetch_offset, char *buffer,
     XrdCl::ResponseHandler *handler, std::unique_lock<std::mutex> *lock, time_t timeout
 )
-    : m_parent(parent),
+    : m_default_handler(parent.m_default_prefetch_handler),
```
amadio, line 1244: So, `m_default_handler` is a copy of the parent's prefetch handler? Is the mutex really the same as the parent file's mutex?

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -1173,51 +1280,76 @@ File::PrefetchResponseHandler::HandleResponse(XrdCl::XRootDStatus *status, XrdCl
...
-                if (missing_bytes) {
-                    mismatched_size = true;
-                    m_prefetch_offset->fetch_sub(missing_bytes, std::memory_order_relaxed);
+    bool parent_alive = false;
+    PrefetchResponseHandler *next = nullptr;
+    std::shared_ptr<XrdClHttp::CurlReadOp> next_op; // op shared_ptr copied out for the Continue() call below
+
+    {
+        std::unique_lock lock(m_default_handler->m_prefetch_mutex);
+        File *parent = m_default_handler->GetFileLocked();
```
amadio, line 1296: Is this supposed to lock `m_parent`? Is `parent` the same as `m_parent`? How is our mutex going to prevent `m_parent` from being destroyed while we hold this lock if it's a copy? This is very confusing at least.

src/XrdClHttp/XrdClHttpFile.hh
```diff
@@ -304,9 +314,21 @@ private:
...
+        // mutex and File*-tracking state alive even after the parent File has been
+        // destroyed, so a worker-thread callback firing after ~File can safely
+        // synchronise and skip accesses to dead File memory.
+        std::shared_ptr<PrefetchDefaultHandler> m_default_handler;
+
+        // The open file we are associated with.  Raw reference because it is set at
+        // construction time on the live File, but no member function may dereference
+        // this except via m_default_handler->GetFileLocked() under the lock — that
+        // is the only thread-safe way to know File is still alive.
         File &m_parent;
```
amadio, line 332: If `m_parent` can be destroyed while this class instance is active, this reference will become stale and cause segmentation faults due to use after free. How does the current code prevent that?

src/XrdClHttp/XrdClHttpOps.hh
```diff
@@ -53,6 +53,13 @@ class CurlWorker;
 class File;
 class ResponseInfo;
 
+// Sentinel errNum attached to the XRootDStatus delivered when a prefetch op is
+// cancelled by File::Close() / File::~File().  Routed through PrefetchDefaultHandler
+// so it can be classified into m_prefetch_cancelled_count without string-matching
+// the error message.  Picked to be distinct from any plausible value of `errno`,
+// `XErrorCode` (kXR_*), or libcurl's CURLcode.
+constexpr uint32_t kPrefetchCancelledOnClose = 0x4F434C53; // "OCLS"
```
amadio, line 61: This looks hacky, I'm not a big fan of magic numbers like this. Why not use `ECANCELED` errno value?

amadio (commented):
I'm not sure the code introduced really fixes the problem, please see my comments. I will also ask for copilot to review, maybe it catches more than me. I also think that the SonarQube comments about using sequentially consistent memory ordering could be considered (while keeping relaxed ordering in the print outs). In most cases, the difference in performance by using `memory_order_seq_cst` (or simply omitting that argument) is probably negligible.

src/XrdClHttp/XrdClHttpOps.hh
bbockelm: Grepp'ing around, it looks like `ECANCELED` is used for the `kXR_Cancelled` response in the xroot protocol.  I fairly sure that it is otherwise unused.  My only hesitation is making it clear to future folks that it not be used for other error messages.  I just feel it's less likely to accidentally reuse random / magic strings for a second use case while standard integers may be introduced elsewhere.

src/XrdClHttp/XrdClHttpFile.cc
bbockelm: Yes.  `m_default_prefetch_handler` is a shared pointer unique to the open file but may outlive the open file.  Callbacks may be invoked by the XrdClHttp worker thread at any time, including after `delete` is invoked on the File.  So, this small object tracks whether the original File is still live (state synchronized by the mutex `m_default_handler->m_prefetch_mutex`)

src/XrdClHttp/XrdClHttpFile.cc
bbockelm: The lock is protects the reference to the parent (which is the same as `m_parent`) in the default handler.  It is also acquired in the parent's destructor where the reference is nil'd.

For as long as the lock is held, the guarantee is one of the two following things happen:
1. `m_default_handler->GetFileLocked()` returns a null pointer, OR
2. The parent file object is not destroyed.

src/XrdClHttp/XrdClHttpFile.hh
bbockelm: - All direct references to `m_parent` are done in places where the parent can't have been destroyed yet.
- In places where `m_parent` may be stale (particularly, callbacks), we instead get a reference (`parent`) via `m_default_handler` because the `m_default_handler` mutex can prevent the parent reference from becoming stale.

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -199,7 +230,57 @@ struct timespec XrdClHttp::File::m_default_header_timeout = {9, 5};
...
+        // continue queue so it can drain the slot promptly.  CurlReadOp::Continue calls
+        // m_continue_queue->Produce, which may briefly block — done outside the lock.
+        // CancelResponseHandler holds a shared_ptr to m_default_prefetch_handler, so the
+        // delivery is safe even if `this` File is destroyed before the worker drains it.
+        auto *ch = new CancelResponseHandler(m_default_prefetch_handler);
+        op->Continue(op, ch, ch->GetBuffer(), 1);
+    }
+    // If not paused, the worker thread will reach CurlReadOp::Write(), observe IsCancelled(),
+    // return 0 from the write callback, and tear the slot down via CURLE_WRITE_ERROR.
+}
```
amadio, line 272: Looking at the `CancelPrefetch()` path when the operation is paused: if `op->Continue()` fails (returns false), the error is silently swallowed — there's no check of the return value. This means the cancellation request may never reach the worker thread, and the curl operation slot could remain occupied indefinitely instead of being freed.

```cpp
// Current code (lines 267-269)
auto *ch = new CancelResponseHandler(m_default_prefetch_handler);
op->Continue(op, ch, ch->GetBuffer(), 1);
// No check if Continue() succeeded or failed
```

**Suggested fix:**

```cpp
auto *ch = new CancelResponseHandler(m_default_prefetch_handler);
if (!op->Continue(op, ch, ch->GetBuffer(), 1)) {
    m_logger->Warning(kLogXrdClHttp, 
        "Failed to continue paused prefetch operation for cancellation; slot may not be freed promptly");
    delete ch;
}
```

This ensures that if the operation's continue queue is unavailable or the continuation fails, we at least log the condition and avoid leaking the `CancelResponseHandler`.

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -1173,51 +1280,76 @@ File::PrefetchResponseHandler::HandleResponse(XrdCl::XRootDStatus *status, XrdCl
...
-        }
-        if (!status || !status->IsOK()) {
-            m_parent.m_prefetch_op.reset();
-            m_parent.m_default_prefetch_handler->m_prefetch_enabled = false;
+        std::unique_lock lock(m_default_handler->m_prefetch_mutex);
+        File *parent = m_default_handler->GetFileLocked();
+        if (parent) {
+            if (parent->m_last_prefetch_handler == this) {
+                parent->m_last_prefetch_handler = nullptr;
+            }
```
amadio, line 1348: The synchronization of `m_last_prefetch_handler` is incomplete. While the reset at lines 1346–1347 happens inside the lock, concurrent writes to this pointer in the `PrefetchResponseHandler` constructor (lines 1254–1255) are **not** guarded by the lock:

```cpp
// Constructor (lines 1253-1255) - NO lock held
if (parent.m_last_prefetch_handler) {
    parent.m_last_prefetch_handler->m_next = this;
    parent.m_last_prefetch_handler = this;  // ← Write without lock!
}

// HandleResponse (lines 1343-1347) - Lock held
{
    std::unique_lock lock(m_default_handler->m_prefetch_mutex);
    File *parent = m_default_handler->GetFileLocked();
    if (parent) {
        if (parent->m_last_prefetch_handler == this) {
            parent->m_last_prefetch_handler = nullptr;  // ← Write with lock
        }
```

This creates a race window: one thread may be linking a new handler to the chain (lines 1254–1255) while another thread is clearing the tail pointer (line 1347). The linked-list invariant could be corrupted, leading to dangling pointers or missed handlers in the chain.

**Suggested fix:** Always guard writes to `m_last_prefetch_handler` with the same lock:

```cpp
// In PrefetchResponseHandler constructor
std::unique_lock lock(parent.m_default_prefetch_handler->m_prefetch_mutex);
if (parent.m_last_prefetch_handler) {
    parent.m_last_prefetch_handler->m_next = this;
}
parent.m_last_prefetch_handler = this;
// lock released here
```

The constructor already acquires the lock for `Continue()`, so this adds minimal overhead.

src/XrdClHttp/XrdClHttpFile.hh
amadio: I really wanted to be sure here, so asked GitHub Copilot to check. It came up with the comment below:

---

The `m_parent` raw reference is a maintenance hazard. The code correctly documents (via comments) that it must only be dereferenced via `m_default_handler->GetFileLocked()` under the lock, but there's **no compile-time enforcement** of this contract. A future maintainer could easily add code that directly references `m_parent` without realizing the thread-safety implications.

```cpp
// Current design (line 328-332)
// The open file we are associated with.  Raw reference because it is set at
// construction time on the live File, but no member function may dereference
// this except via m_default_handler->GetFileLocked() under the lock — that
// is the only thread-safe way to know File is still alive.
File &m_parent;
```

**Suggested mitigation:** Replace the raw reference with a private accessor method that enforces the locking contract at the call site:

```cpp
private:
    File &m_parent;
    
    // Caller MUST be holding m_default_handler->m_prefetch_mutex
    File *GetParentUnsafe() const {
        // Note: deliberately non-const lock to force caller to hold it explicitly
        return &m_parent;
    }
```

---

Back to my own thoughts, as I don't really like the suggestion above. Better yet is to eliminate the need for `m_parent` entirely by using an approach with `std::shared_ptr` + `std::weak_ptr` to simplify this whole thing, which is described in the main review comment.

src/XrdClHttp/XrdClHttpFile.cc
amadio, line 1274: There's a potential memory leak in the `PrefetchResponseHandler` constructor when `Continue()` fails. If the `Continue()` call returns false, the handler is removed from the linked list and an exception is thrown, but the handler object itself is never cleaned up:

```cpp
// Current code (lines 1260-1272)
if (lock && m_parent.m_prefetch_op) {
    lock->unlock();
    if (!parent.m_prefetch_op->Continue(parent.m_prefetch_op, this, buffer, size)) {
        lock->lock();
        if (parent.m_last_prefetch_handler == this)
            parent.m_last_prefetch_handler = nullptr;
        throw std::runtime_error("Failed to continue prefetch operation");
        // ← Handler 'this' is never deleted; caller loses reference
    }
}
```

The `PrefetchResponseHandler` is allocated as `new PrefetchResponseHandler(...)` by the caller (line 796), and if this constructor throws an exception, the caller is expected to handle cleanup. However, since the handler has already been added to the chain and then removed, no cleanup occurs—the allocated object is orphaned.

amadio (commented):
So, I think that the current implementation has some landmines which could be avoided by using a different approach, using `std::shared_ptr` and `std::weak_ptr` (it would be similar to how I used these in `TBufferMerger` back in the day in ROOT). To give you a more concrete idea of what I mean, I prompted GitHub Copilot with:

```
Please suggest a potential simplification by using a combination of std::shared_ptr and std::weak_ptr, using the expired() method to check if the file has been destroyed, rather than the current mechanism. The response handler will then just return immediately if the file has been already closed.
```

GitHub Copilot then responded with the suggestion below, based on my input. Can you maybe try to rework it that way? If a prefetch callback sees that the file is gone (via `f.expired()` call on the `std::weak_ptr`, then it can just exit immediately, which is what we want. It's not *cancelled*, just becomes a no-op automatically.

BTW, if you are using Claude or other type of AI model in the development of these changes, please add the appropriate `Assisted-by:` tag to the commits.

---

Excellent suggestion! Let me create a detailed analysis with code examples showing how to simplify this design using `std::shared_ptr` and `std::weak_ptr`.

## Proposed Simplification: `shared_ptr<File>` + `weak_ptr<File>`

The current design maintains a nullable raw pointer (`File *m_file`) in `PrefetchDefaultHandler` and relies on explicit `DetachFile()` calls to null it during destruction. **A cleaner approach would use reference counting via `shared_ptr` and `weak_ptr`**.

### Current Design Problems

```cpp
// Current approach in PrefetchDefaultHandler (File.hh)
class PrefetchDefaultHandler {
    File *m_file{nullptr};  // ❌ Raw pointer - requires manual DetachFile()
    
    // Called from ~File()
    void DetachFile() {
        std::unique_lock lock(m_prefetch_mutex);
        m_file = nullptr;
    }
    
    // Called from HandleResponse()
    File *GetFileLocked() const { return m_file; }
};
```

**Issues:**
1. Requires explicit `DetachFile()` call in the destructor
2. No automatic cleanup if the call is forgotten
3. Race conditions are possible if `DetachFile()` and callback racing
4. Feels manual and error-prone

---

### Proposed Simplification

```cpp
// Simplified approach using weak_ptr
class PrefetchDefaultHandler {
private:
    std::weak_ptr<FileImpl> m_file;  // ✓ No manual detach needed
    
public:
    // Try to get a live File; returns nullptr if already destroyed
    std::shared_ptr<FileImpl> GetFileLocked() const {
        // expired() is thread-safe for this check
        if (m_file.expired()) {
            return nullptr;
        }
        // lock() atomically tries to promote weak_ptr → shared_ptr
        return m_file.lock();
    }
};
```

**Key advantages:**
1. **No manual cleanup**: `weak_ptr::expired()` checks if the original `shared_ptr` is gone
2. **Automatic**: The File is destroyed when the last `shared_ptr` is dropped
3. **Atomic**: `weak_ptr::lock()` is thread-safe (atomic under the hood)
4. **Idiomatic C++**: This is the standard pattern for break-cycle references

---

### Implementation Sketch

#### File.hh (Header)

```cpp
// Make File ref-counted so callbacks can hold weak_ptr to it
class File final : public XrdCl::FilePlugIn, 
                   public std::enable_shared_from_this<File> {
public:
    // Constructor and public API unchanged...
    
private:
    friend class PrefetchDefaultHandler;
    
    // PrefetchDefaultHandler is created with a weak_ptr to this File.
    // The File holds the strong ref via m_default_prefetch_handler.
    std::shared_ptr<PrefetchDefaultHandler> m_default_prefetch_handler;
    
    class PrefetchDefaultHandler : public XrdCl::ResponseHandler {
    public:
        // Constructor captures a weak_ptr; File holds strong ref
        PrefetchDefaultHandler(std::weak_ptr<File> file_ref)
            : m_file(std::move(file_ref)) {}
        
        // No DetachFile() needed! weak_ptr handles it automatically.
        
        File *GetFileLocked() const {
            // Check if File still exists (hold the lock!)
            auto file = m_file.lock();
            return file ? file.get() : nullptr;
        }
        
        // Try to lock and get a shared_ptr for safer holding
        std::shared_ptr<File> TryGetFile() const {
            return m_file.lock();
        }
        
    private:
        std::weak_ptr<File> m_file;  // Doesn't keep File alive
    };
};
```

#### File.cc (Implementation)

```cpp
// In File::Open() or the constructor
void File::InitializePrefetchHandler() {
    // Pass weak_ptr; File itself is created as shared_ptr by XrdCl infrastructure
    m_default_prefetch_handler.reset(
        new PrefetchDefaultHandler(weak_from_this())
    );
}

// In ~File() — NO DetachFile() call needed!
File::~File() noexcept {
    CancelPrefetch();
    // m_default_prefetch_handler is destroyed normally
    // Any in-flight callbacks that call m_file.lock() will get nullptr
    // because the last shared_ptr to File is being destroyed now.
}

// In PrefetchResponseHandler::HandleResponse()
void File::PrefetchResponseHandler::HandleResponse(
    XrdCl::XRootDStatus *status, XrdCl::AnyObject *response) {
    
    std::unique_ptr<PrefetchResponseHandler> owner(this);
    
    bool parent_alive = false;
    PrefetchResponseHandler *next = nullptr;
    std::shared_ptr<XrdClHttp::CurlReadOp> next_op;
    
    {
        std::unique_lock lock(m_default_handler->m_prefetch_mutex);
        
        // ✓ Cleanly promoted weak_ptr to shared_ptr
        auto parent = m_default_handler->TryGetFile();
        parent_alive = (parent != nullptr);
        
        if (status) {
            if (status->IsOK() && response) {
                XrdCl::ChunkInfo *ci = nullptr;
                response->Get(ci);
                if (ci) {
                    auto missing_bytes = m_size - ci->GetLength();
                    if (missing_bytes && parent_alive) {
                        // Safe: parent is held by shared_ptr for this scope
                        m_prefetch_offset->fetch_sub(missing_bytes, 
                                                     std::memory_order_relaxed);
                    }
                    m_prefetch_bytes_used.fetch_add(ci->GetLength(), 
                                                    std::memory_order_relaxed);
                }
            } else if (!status->IsOK()) {
                m_prefetch_failed_count.fetch_add(1, std::memory_order_relaxed);
            }
        }
        
        next = m_next;
        if (parent_alive) {
            next_op = parent->m_prefetch_op;
        }
        // ✓ Lock released here; parent may be destroyed if this was the last ref
    }
    
    // Continue processing...
    if (next) {
        if (parent_alive && next_op && status && status->IsOK() && !mismatched_size) {
            next_op->Continue(next_op, next, next->m_buffer, next->m_size);
        } else {
            m_default_handler->DisablePrefetch();
            next->ResubmitOperation(status);
        }
    }
    
    if (m_handler) m_handler->HandleResponse(status, response);
    else delete response;
}

// In PrefetchResponseHandler::ResubmitOperation()
void File::PrefetchResponseHandler::ResubmitOperation(
    XrdCl::XRootDStatus *fallback_status) {
    
    m_default_handler->m_logger->Debug(kLogXrdClHttp,
        "Resubmitting waiting prefetch operations as new reads due to prefetch failure");
    
    PrefetchResponseHandler *next = this;
    while (next) {
        auto cur = next;
        XrdCl::XRootDStatus *delivered_status = nullptr;
        
        {
            std::unique_lock lock(next->m_default_handler->m_prefetch_mutex);
            
            // ✓ Try to lock; if File is gone, parent will be nullptr
            auto parent = next->m_default_handler->TryGetFile();
            
            if (parent) {
                lock.unlock();
                auto st = parent->Read(next->m_offset, next->m_size, 
                                      next->m_buffer, next->m_handler, 
                                      next->m_timeout);
                if (!st.IsOK()) {
                    delivered_status = new XrdCl::XRootDStatus(st);
                }
            } else if (fallback_status) {
                // Parent gone; deliver the cancellation reason
                delivered_status = new XrdCl::XRootDStatus(*fallback_status);
            } else {
                // No parent, no fallback → generate cancellation error
                delivered_status = new XrdCl::XRootDStatus(
                    XrdCl::stError, XrdCl::errOperationExpired,
                    XrdClHttp::kPrefetchCancelledOnClose,
                    "File closed before prefetched read completed");
            }
        }
        
        if (delivered_status && next->m_handler) {
            next->m_handler->HandleResponse(delivered_status, nullptr);
        } else if (delivered_status) {
            delete delivered_status;
        }
        
        {
            std::unique_lock lock(next->m_default_handler->m_prefetch_mutex);
            next = next->m_next;
        }
        delete cur;
    }
}
```

---

### Comparison: Current vs. Proposed

```cpp
// ============ CURRENT DESIGN ============
// File.hh
class PrefetchDefaultHandler {
    File *m_file{nullptr};  // ❌ Raw pointer, manual lifecycle
    
    void DetachFile() {
        std::unique_lock lock(m_prefetch_mutex);
        m_file = nullptr;  // ❌ Manual null-out
    }
    
    File *GetFileLocked() const { 
        return m_file;  // Can return dangling pointer if DetachFile() missed
    }
};

// File.cc in ~File()
File::~File() noexcept {
    CancelPrefetch();
    if (m_default_prefetch_handler) {
        m_default_prefetch_handler->DetachFile();  // ❌ Manual cleanup
    }
    // ...
}

// ============ PROPOSED DESIGN ============
// File.hh
class PrefetchDefaultHandler {
    std::weak_ptr<File> m_file;  // ✓ Auto-managed lifecycle
    
    // ✓ No DetachFile() needed!
    
    std::shared_ptr<File> TryGetFile() const {
        return m_file.lock();  // ✓ Safe, automatic null if destroyed
    }
};

// File.cc in ~File()
File::~File() noexcept {
    CancelPrefetch();
    // ✓ m_default_prefetch_handler destroyed normally
    // ✓ No explicit DetachFile() call
    // ✓ Any in-flight callbacks automatically see nullptr
    // ...
}
```

---

### Key Improvements

| Aspect | Current | Proposed |
|--------|---------|----------|
| **Lifecycle Management** | Manual `DetachFile()` | Automatic via `weak_ptr` |
| **Race Conditions** | Possible if `DetachFile()` missed | Impossible; `weak_ptr::lock()` is atomic |
| **Code Clarity** | Requires understanding raw pointer tricks | Idiomatic C++; familiar pattern |
| **Lines of Code** | ~30 for synchronization | ~15 for synchronization |
| **Dangling Pointer Risk** | High if maintainer forgets `DetachFile()` | None; reference counting handles it |
| **Testing Burden** | Must verify `DetachFile()` is called | Fewer edge cases to test |

---

### Why This Works

1. **`File` holds `shared_ptr<PrefetchDefaultHandler>`**: Keeps handler alive until File is destroyed.
2. **`PrefetchDefaultHandler` holds `weak_ptr<File>`**: Doesn't keep File alive; allows File to be destroyed.
3. **Worker thread callback (in-flight)**: Calls `m_file.lock()` → gets `shared_ptr` or `nullptr`.
4. **No deadlock**: `weak_ptr::lock()` doesn't require any locks; it's atomic.
5. **Automatic cleanup**: When File is destroyed, the last `shared_ptr` is dropped, and all `weak_ptr::lock()` calls return `nullptr`.

---

### Caveats & Constraints

1. **XrdCl::FilePlugIn lifecycle**: The PR must verify that `File` objects are created/stored as `shared_ptr` by the XrdCl infrastructure. If File is always stack-allocated, this won't work.
   
   Check: Look for `File::Open()` and object creation patterns in the calling code.

2. **`enable_shared_from_this<File>`**: File must inherit from `enable_shared_from_this<File>` so it can call `weak_from_this()` in its constructor.

3. **Header includes**: Need `<memory>` for `std::shared_ptr`, `std::weak_ptr`, `std::enable_shared_from_this`.

---

### Recommendation

**Adopt the `weak_ptr` approach if:**
- The XrdCl infrastructure already manages Files as `shared_ptr` (very likely for plugin lifecycle)
- The team is comfortable with reference-counted lifecycle models
- You want to eliminate the manual `DetachFile()` footgun

**The refactoring is straightforward:**
1. Replace `File *m_file` with `std::weak_ptr<File> m_file`
2. Replace `GetFileLocked()` with `TryGetFile()` that returns `shared_ptr` or `nullptr`
3. Remove all `DetachFile()` calls
4. Remove the `DetachFile()` method entirely

This **cuts synchronization complexity by ~40%** while making the code more maintainable and less error-prone.

src/XrdClHttp/XrdClHttpOps.hh
amadio: I believe that with the refactor using `std::shared_ptr` + `std::weak_ptr` this will no longer be necessary, but if it is, please use `ECANCELED`, as that is the right error to use in cases like this.

amadio:
@bbockelm Do you still plan to work on this? I'd like to avoid having stale PRs in draft lingering for a long time in the repository if possible.

amadio:
Closing due to inactivity. Please feel free to reopen if needed.

PR #2816, opened 2026-06-04
[Python] Fix standalone build against an installed XRootD

chrisburr:
PyXRootDAdler32.cc includes XrdCks/XrdCksXAttr.hh and XrdOuc/XrdOucXAttr.hh (which in turn includes XrdSys/XrdSysFAttr.hh), none of which were installed by 'make install'. This broke the fallback in python/src/CMakeLists.txt that allows building the Python bindings against a pre-installed XRootD, since the bindings could no longer be compiled without the full source tree.

Install the three headers into the private include tree, which the Python bindings already use for XrdCl private headers, and install XrdCksData.hh (included by XrdCksXAttr.hh) unconditionally since the XrdCks code is built into libXrdUtils even for client-only builds. Also link XrdUtils explicitly, as the bindings depend on XrdSysFAttr from libXrdUtils, which until now was only resolved transitively through libXrdCl.

### DIFF
--- a/src/XrdHeaders.cmake
+++ b/src/XrdHeaders.cmake
@@ -33,6 +33,7 @@ set( XROOTD_PUBLIC_HEADERS
   XrdNet/XrdNetPMark.hh
   XrdNet/XrdNetSockAddr.hh
   XrdNet/XrdNetSocket.hh
+  XrdCks/XrdCksData.hh
   XrdOuc/XrdOucBuffer.hh
   XrdOuc/XrdOucCRC.hh
   XrdOuc/XrdOucCache.hh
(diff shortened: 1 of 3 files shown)


amadio:
@chrisburr The CI error (ABI check build) is due to linking against XrdUtils directly. Could you please take a look and fix? Thanks.

amadio:
I will look at adding a test to ensure this doesn't happen again (breaking standalone Python bindings build).

amadio:
@chrisburr Could you please add a `Fixes: <commithash>` tag to this and force-push? Other than that, this is ready to be merged. Thank you!

src/XrdHeaders.cmake
```diff
@@ -33,6 +33,7 @@ set( XROOTD_PUBLIC_HEADERS
   XrdNet/XrdNetPMark.hh
   XrdNet/XrdNetSockAddr.hh
   XrdNet/XrdNetSocket.hh
+  XrdCks/XrdCksData.hh
```
amadio: @chrisburr Is it really needed to make this header public? I believe it can be kept private given that it's only included by other private headers and .cc files.

amadio: Ok, unfortunately the header indeed needs to be public.

PR #2818, opened 2026-06-08
[XrdOssMirage] Add XrdOssMirage, a server-side plugin that simulates the existence of files

diogomattioli:
XrdOssMirage is a server-side plugin that simulates the existence of files on an XRootD server. It requires no storage: any uploaded file content is discarded, while only the file size is retained.

When reading a file, the plugin supports three different modes: returning arbitrary garbage data, generating a repeated character pattern, or generating a repeated string pattern.

It is also possible to upload an empty or very small file and then truncate it to simulate a huge file on the server side, avoiding the need for lengthy data transfers.

No data is persisted: the entire filesystem exists only in the volatile memory.

The plugin is targeted for testing and development purposes.

This PR contains the plugin itself, documentation, unit tests and end-to-end tests.


src/XrdOssMirage/XrdOssMirage.cc
```diff
@@ -0,0 +1,164 @@
...
+
+    if (!has_entry(path) || is_entry_being_written(path))
+        return {};
+
+    return entries[path];    
+}
+
+bool XrdOssMirage::has_entry(const char *path)
+{
+    return entries.contains(path);
```
amadio: I guess this C++20 method is not available on Alma 8. Either I need to upgrade the compiler, or you could try using a C++17 equivalent here.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
```diff
@@ -8,85 +8,115 @@ class XrdOssMirageFileFixture : public XrdOssMirageFixture
 {
 };
 
-TEST_F(XrdOssMirageFileFixture, Stat)
+TEST_F(XrdOssMirageFileFixture, StatingReturnsOk)
```
amadio: I'd keep verbs in infinitive form, instead of using gerund.
So, `StatReturnsOK` in this case. It still clearly states what is being tested.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
```diff
@@ -8,85 +8,115 @@ class XrdOssMirageFileFixture : public XrdOssMirageFixture
...
-TEST_F(XrdOssMirageFileFixture, Truncate)
+TEST_F(XrdOssMirageFileFixture, TruncatingReturnsOk)
 {
     file.Open("/dummy", O_WRONLY, {}, env);
 
     ASSERT_EQ(XrdOssOK, file.Ftruncate(1000));
-    ASSERT_EQ(XrdOssOK, file.Close());
+}
+
+TEST_F(XrdOssMirageFileFixture, TruncatingChangesItsSize)
```
amadio: `TruncateChangesFileSize` here, but I'd add two tests: `TruncateReducesFileSize` and `TruncateIncreasesFileSize` to make sure that it works both ways.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
```diff
@@ -8,85 +8,115 @@ class XrdOssMirageFileFixture : public XrdOssMirageFixture
...
+{
+    file.Open("/dummy", O_WRONLY, {}, env);
+    file.Ftruncate(1000);
+    file.Close();
+
     ASSERT_EQ(1000, oss.get_entry_read("/dummy").value().size);
 }
 
-TEST_F(XrdOssMirageFileFixture, OpenWriteMode)
+TEST_F(XrdOssMirageFileFixture, OpeningInWriteModeReturnsOk)
```
amadio: An alternative to `RetunsOK` is `Succeeds`, so `OpenForWriteSucceeds`.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
```diff
@@ -8,85 +8,115 @@ class XrdOssMirageFileFixture : public XrdOssMirageFixture
...
 {
     ASSERT_EQ(XrdOssOK, file.Open("/dummy", O_WRONLY, {}, env));
 }
 
-TEST_F(XrdOssMirageFileFixture, OpenWriteModeInexistentFile)
+TEST_F(XrdOssMirageFileFixture, OpeningInWriteModeFileThatDoesNotExistReturnsNOENTError)
```
amadio: Is this correct? Shouldn't the file just be created on open in this case? I guess you want to ensure that you need `O_CREAT` too for that?

For naming, I'd suggest `OpenNonExistentFileForWriteFailsWithENOENT` to follow the best practices. But I also think that the names are becoming quite verbose. I think we should consider balancing granularity and verbosity by allowing tests with more than one assertion, as long as they test closely related things. In this case, it could be like shown below:

```cpp
TEST_F(XrdOssMirageFileFixture, OpenForWrite)
{
  {
    EXPECT_EQ(-ENOENT, file.Open("/inexistent1", O_WRONLY, {}, env)) <<
      "Open non-existent file for write without O_CREAT did not fail with ENOENT";
  }
  {
    EXPECT_EQ(0, file.Open("/inexistent2", O_WRONLY | O_CREAT, {}, env)) <<
      "Open non-existent file for write with O_CREAT did not succeed";

    /* do not leak file descriptor */
    EXPECT_EQ(0, file.Close()) << "Failed to close file."; 
  }
  ...
}
```
I think the above still reasonably satisfies the requirements, while keeping things as clear and easy to debug as recommended. In other works, the "test", is the assertion itself, all of which should be independent of each other so that within, say, the `OpenForWrite` test case above, they can fail independently without affecting other assertions and the output is always clear.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
```diff
@@ -8,85 +8,115 @@ class XrdOssMirageFileFixture : public XrdOssMirageFixture
...
-TEST_F(XrdOssMirageFileFixture, ReadWithOffset)
+TEST_F(XrdOssMirageFileFixture, ReadingWithOffsetFromFileWhereRemainingSizeIsBiggerThanBufferReturnsBufferSizeAsReadBytes)
 {
     file.Open("/dummy", O_RDONLY, {}, env);
 
     ASSERT_EQ(1000, file.Read(nullptr, 1000, 1000));
 }
 
-TEST_F(XrdOssMirageFileFixture, ReadWithOffsetRemainingSmallerThanSize)
+TEST_F(XrdOssMirageFileFixture, ReadingWithOffsetFromFileWhereRemainingSizeIsSmallerThanBufferReturnsRemainingSizeAsReadBytes)
```
amadio: This is also a very verbose name. I think it's better to use the feature to print out a detailed failure message after the `ASSERT_EQ` and keep the shorter names for tests.

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
```diff
@@ -73,43 +94,73 @@ TEST_F(XrdOssMirageXAttrFixture, Get)
...
-    ASSERT_EQ(1, xattr.Get("U.read.return_position", &value, 1, "/dummy", 0));
+    xattr.Get("U.read.return_position", &value, 1, "/dummy", 0);
     ASSERT_EQ('3', value);
 
-    ASSERT_EQ(1, xattr.Get("U.write.return_code", &value, 1, "/dummy", 0));
+    xattr.Get("U.write.return_code", &value, 1, "/dummy", 0);
     ASSERT_EQ('4', value);
 
-    ASSERT_EQ(1, xattr.Get("U.write.return_position", &value, 1, "/dummy", 0));
+    xattr.Get("U.write.return_position", &value, 1, "/dummy", 0);
```
amadio: I think it's fine to keep the check above, but use `EXPECT_EQ` instead of `ASSERT_EQ`, so the test continues it execution on expectation failure.

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
```diff
@@ -73,43 +94,73 @@ TEST_F(XrdOssMirageXAttrFixture, Get)
...
 }
 
+TEST_F(XrdOssMirageXAttrFixture, DeletingPropertiesReturnsOk)
+{
+    ASSERT_EQ(0, xattr.Del("U.open.return_code", "/dummy", 0));
+    ASSERT_EQ(0, xattr.Del("U.read.return_code", "/dummy", 0));
+    ASSERT_EQ(0, xattr.Del("U.read.return_position", "/dummy", 0));
+    ASSERT_EQ(0, xattr.Del("U.write.return_code", "/dummy", 0));
+    ASSERT_EQ(0, xattr.Del("U.write.return_position", "/dummy", 0));
+    ASSERT_EQ(0, xattr.Del("U.pattern", "/dummy", 0));
```
amadio: Replace `ASSERT_EQ` with `EXPECT_EQ` here to keep tests independent. `ASSERT_EQ` shortcuts the test case execution.

amadio (commented):
I think we need to fine a tradeoff that satisfies the best practices while keeping things as easy as possible to debug when something goes wrong. This can be done by choosing well between `EXPECT_EQ` and `ASSERT_EQ` in tests, and ensuring that each of them is completely independent of each other, splitting into a separate test case whenever it becomes sufficiently long (i.e. more than just a couple of lines per expect/assert). What do you think?

src/XrdOssMirage/XrdOssMirage.cc
diogomattioli: After discussion on upgrading the compiler to fully support c++20, we decided to remove the unsupported c++20 elements for the moment

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
diogomattioli: Fixed!

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
diogomattioli: fixed!

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
diogomattioli: fixed!

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
diogomattioli: I'm replacing `ReturnsOk` with `ShouldSucceed`, because it indicates the intention of the test.

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
diogomattioli: done!

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
diogomattioli: done!

tests/XrdOssMirageTests/XrdOssMirageFileTests.cc
diogomattioli: I'd prefer to write it into separate tests. The granularity is more evident, there's no need to open the test to check exactly what's being tested or not.

In the example above, you're just aggregating multiple tests into one. Apart what I mentioned above, you're just repeating the test name in the description (for instance `"Open non-existent file for write without O_CREAT fails with ENOENT"` could just be the name of the test. You also run at the risk of putting unrelated things together (should writing methods being tested there?). This also makes very tempting to do the initialization of those tests in a common place thus not following the isolation of the tests. Last but not least, if one of those assertions become obsolete in the future you cannot simply delete the test, you've to change an existing test and be careful to not break it.

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
```diff
@@ -0,0 +1,207 @@
...
+    ASSERT_EQ(-EINVAL, xattr.Set("U.pattern", "1", 1, "/dummy", 0, 0));
+}
+
+TEST_F(XrdOssMirageXAttrFixture, SetPropertiesWithValuesOutOfRangeFailsWithINVAL)
+{
+    ASSERT_EQ(-EINVAL, xattr.Set("U.open.return_code", "18446744073709551615", 20, "/dummy", 0, 0));
+    ASSERT_EQ(-EINVAL, xattr.Set("U.read.return_code", "18446744073709551615", 20, "/dummy", 0, 0));
+    ASSERT_EQ(-EINVAL, xattr.Set("U.read.return_position", "18446744073709551615", 20, "/dummy", 0, 0));
+    ASSERT_EQ(-EINVAL, xattr.Set("U.write.return_code", "18446744073709551615", 20, "/dummy", 0, 0));
+    ASSERT_EQ(-EINVAL, xattr.Set("U.write.return_position", "18446744073709551615", 20, "/dummy", 0, 0));
```
amadio: Should these also be `EXPECT_EQ`, maybe?

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
```diff
@@ -0,0 +1,207 @@
...
+        entry->write.return_code = 4;
+        entry->write.return_position = 5;
+        entry->pattern = "6";
+    }
+
+    char value = 0;
+    xattr.Get("U.open.return_code", &value, 1, "/dummy", 0);
+    EXPECT_EQ('1', value);
+
+    xattr.Get("U.read.return_code", &value, 1, "/dummy", 0);
```
amadio, line 121: Tests after this point may depend on outcome of previous tests (by reusing `value`).

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
```diff
@@ -0,0 +1,207 @@
...
+}
+
+TEST_F(XrdOssMirageXAttrFixture, DeletePropertiesShouldSucceed)
+{
+    EXPECT_EQ(0, xattr.Del("U.open.return_code", "/dummy", 0));
+    EXPECT_EQ(0, xattr.Del("U.read.return_code", "/dummy", 0));
+    EXPECT_EQ(0, xattr.Del("U.read.return_position", "/dummy", 0));
+    EXPECT_EQ(0, xattr.Del("U.write.return_code", "/dummy", 0));
+    EXPECT_EQ(0, xattr.Del("U.write.return_position", "/dummy", 0));
+    EXPECT_EQ(0, xattr.Del("U.pattern", "/dummy", 0));
```
amadio, line 165: Do these maybe depend on tests named `SetProperties*`?

tests/XRootD/mirage.cfg
```diff
@@ -0,0 +1,15 @@
+set name = mirage
+set port = 6543
+
+set src = $SOURCE_DIR
+set pwd = $PWD
+
+xrootd.trace all
+pfc.trace info
+
+xrootd.seclib libXrdSec.so
```
amadio: This one is not needed unless you want to set something up with `sec.protocol`.

tests/XRootD/mirage.cfg
```diff
@@ -0,0 +1,15 @@
+set name = mirage
+set port = 6543
+
+set src = $SOURCE_DIR
+set pwd = $PWD
+
+xrootd.trace all
+pfc.trace info
```
amadio: Not sure about this one, probably not needed. You may add here `ofs.trace all` and not continue to `common.cfg` if you want.

tests/XRootD/mirage.sh
```diff
@@ -0,0 +1,55 @@
...
+	require_commands diff
+}
+
+function test_mirage() {
+	echo
+	echo "client: XRootD $(xrdcp --version 2>&1)"
+	echo "server: XRootD $(xrdfs "${HOST}" query config version 2>&1)"
+	echo
+
+	ORIGIN_HOST="root://localhost:6543"
```
amadio: `${HOST}` is exported by `test.sh`, it should be the same as the above.

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
diogomattioli: somehow i missed those. fixed!

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
diogomattioli: good catch!

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
diogomattioli: Not really. Delete should work independently of the value stored. The expected 0 is just the same as Ok but `XAttr` AFAIK does not have a specific return code mapped to enum, so zero is used.

tests/XRootD/mirage.cfg
diogomattioli: I can remove the `pfc.trace`. In reality I don't understand why it's needed or not. I just would like to continue with the `common.cfg` to maintain the same configuration as the other tests. What you think?

tests/XRootD/mirage.cfg
diogomattioli: done

tests/XRootD/mirage.sh
diogomattioli: done

tests/XrdOssMirageTests/XrdOssMirageXAttrTests.cc
amadio: As long as the attribute doesn't have to pre-exist in order to be deleted, no problem. But then it's also not deleting something, because it was not there before anyway, so consider adding the xattrs, then deleting them.

tests/XRootD/mirage.cfg
amadio: Yes, no problem, `pfc` stands for `Proxy File Cache`, so you don't need it. The docs are here, in case you want to have a look: https://xrootd.web.cern.ch/doc/dev6/pss_config.htm#_Toc209818908

src/XrdOssMirage/XrdOssMirage.cc
```diff
@@ -0,0 +1,164 @@
...
+            if (XrdOssMirageXAttr * const xattr = dynamic_cast<XrdOssMirageXAttr*>(XrdSysFAttr::Xat); xattr != nullptr)
+                xattr->setOss(*this);
+        });
+
+    return XrdOssOK;
+}
+
+uint64_t XrdOssMirage::Features()
+{
+    return XRDOSS_HASNAIO | XRDOSS_HASFICL;
```
abh3: This may cause a problem here. Since, unimplemented features but not so declared causes unnecessary errors to be returned to the client. In this case, since file cloning is not supported, please remove XRDOSS_HASFICL from the feature set.

src/XrdOssMirage/XrdOssMirage.cc
```diff
@@ -0,0 +1,164 @@
...
+}
+
+int XrdOssMirage::Stat(const char *path, struct stat *buff, int opts, XrdOucEnv *envP)
+{
+    const std::lock_guard lock(mutex);
+
+    if (!has_entry(path))
+        return -ENOENT;
+
+    buff->st_size = entries[path]->size;
```
abh3, line 105: I suggest that the buff be cleared to zeroes first to avoid random stuff being returned.

src/XrdOssMirage/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+When reading a file, the plugin supports three different modes: returning arbitrary garbage data, generating a repeated character pattern, or generating a repeated string pattern.
+
+It is also possible to upload an empty or very small file and then truncate it to simulate a huge file on the server side, avoiding the need for lengthy data transfers.
+
+No data is persisted: the entire filesystem exists only in the volatile memory.
+
+## Configuration
+
+The plugin is a runtime-only plugin that requires minimal configuration.
```
abh3: It looks like it requires no configuration given that all you need to do is load it. If this is true, please say so otherwise I am left wondering what minimal amount am I missing here.

src/XrdOssMirage/README.md
```diff
@@ -0,0 +1,164 @@
...
+Id:     25769803871
+Size:   1073741824
+MTime:  1970-01-01 00:00:00
+Flags:  4 (Other)
+```
+
+## Usage of extended options
+
+The extended options are implemented through the extended attributes plugin. This enables additional functionality useful for testing and development without requiring manual file setup or custom implementations.
+
```
abh3, line 65: Quite clever I would say.

src/XrdOssMirage/XrdOssMirageFile.hh
```diff
@@ -0,0 +1,42 @@
...
+    XrdOssMirage &oss;
+    std::variant<XrdOssMirageEntry, XrdOssMirageEntryPtr> entry_storage;
+    XrdOssMirageEntry *entry;
+
+public:
+    XrdOssMirageFile(XrdOssMirage &oss);
+    virtual ~XrdOssMirageFile() = default;
+
+    virtual int     StatRet(struct stat *buff) override;
+    virtual int     Clone(XrdOssDF& srcFile) override;
```
abh3: For this and all subsequent methods that are not supported, please simply rely on the default implementation of the base class (most of which do exactly what the override does). This would greatly simply this class. I notice that was done in the class derived from XrdOss.

abh3 (changes requested):
Some minor changes and simplifications need nere.

src/XrdOssMirage/XrdOssMirage.cc
diogomattioli: done!

src/XrdOssMirage/README.md
diogomattioli: done!

src/XrdOssMirage/XrdOssMirageFile.hh
diogomattioli: Good one! 

I've removed those `Clone` methods due being already implemented with returning `-ENOTSUP` in the base class.

For the remaining methods like `Fchmod`, `FSync` and AIO methods in which the base class returns `-EISDIR` I'm keeping the implementation, even because we don't have support for directories in the plugin.

src/XrdOssMirage/XrdOssMirage.cc
diogomattioli: done! fixed on `XrdOssMirageFile` as well.

abh3 (approved):
Looks good; thanks!

PR #2819, opened 2026-06-08
[Python] Add xattr convenience helpers

lobis:
## Summary

- Add mapping-style xattr helpers on `FileSystem` and `File`.
- Preserve the existing low-level `get_xattr()` / `list_xattr()` APIs while adding direct helpers for common metadata use cases.
- Document the xattr APIs and add focused sync/callback tests.

## Motivation

Downstream users such as DIRAC and Rucio consume xattrs as metadata. The existing bindings expose the low-level tuple-shaped xattr response, which is useful for preserving per-attribute status, but it makes callers repeatedly unpack response tuples when they only need a metadata mapping or one attribute value.

This is especially relevant for GFAL2 replacement work:

- DIRAC currently lists all xattr names and then fetches each value into a dictionary: https://github.com/DIRACGrid/DIRAC/blob/517d12a1900558f1d778f9d4d0a599850e43895c/src/DIRAC/Resources/Storage/GFAL2_StorageBase.py#L1532-L1541
- Rucio currently fetches a single space-token xattr for space usage: https://github.com/rucio/rucio/blob/2262ce73fffff1b163aa3bd56d756f4051d7496a/lib/rucio/rse/protocols/gfal.py#L592-L594

These helpers do not add new protocol behavior. They provide a small Python convenience layer over the existing xattr binding methods.

## Before / After

Before, callers that wanted all xattrs had to unpack the low-level response shape and inspect each per-attribute status manually:

```python
status, attrs = fs.list_xattr(path)
if not status.ok:
    raise RuntimeError(status.message)

metadata = {}
for name, value, attr_status in attrs:
    attr_status = client.responses.XRootDStatus(attr_status)
    if not attr_status.ok:
        raise RuntimeError(attr_status.message)
    metadata[name] = value
```

After this change, the common metadata case is direct:

```python
status, metadata = fs.xattrs(path)
if not status.ok:
    raise RuntimeError(status.message)
```

Before, callers that wanted a single xattr had to pass a list and unpack the first tuple:

```python
status, attrs = fs.get_xattr(path, ["spacetoken.description?" + space_token])
if not status.ok:
    raise RuntimeError(status.message)

name, value, attr_status = attrs[0]
attr_status = client.responses.XRootDStatus(attr_status)
if not attr_status.ok:
    raise RuntimeError(attr_status.message)
```

After this change:

```python
status, value = fs.xattr(path, "spacetoken.description?" + space_token)
if not status.ok:
    raise RuntimeError(status.message)
```

## Tests

- `python3 -m py_compile python/libs/client/utils.py python/libs/client/filesystem.py python/libs/client/file.py python/tests/test_xattr_helpers.py`
- `PYTHONPATH=build/lib.macosx-10.13-universal2-cpython-313 pytest -q python/tests/test_xattr_helpers.py`
- `git diff --check upstream/master...HEAD`

### DIFF
--- a/python/libs/client/file.py
+++ b/python/libs/client/file.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import, division, print_function
 from pyxrootd import client
 from XRootD.client.responses import XRootDStatus, StatInfo, VectorReadInfo
 from XRootD.client.utils import CallbackWrapper
+from XRootD.client.utils import _xattr_mapping, _xattr_value
 
 class File(object):
   """Interact with an ``xrootd`` server to perform file-based operations such
@@ -373,6 +374,58 @@ class File(object):
 
     status, response = self.__file.list_xattr(timeout)
     return XRootDStatus(status), response
+
+  def xattrs(self, timeout=0, callback=None):
+    """Get all extended file attributes as a mapping.
+
+    :returns:     tuple containing :mod:`XRootD.client.responses.XRootDStatus`
+                  object and dict mapping xattr names to values
+    """
+    if callback:
+      def handle_xattrs(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_mapping(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.list_xattr(timeout, handle_xattrs)
+
+    status, response = self.list_xattr(timeout)
+    if not status.ok:
+      return status, None
+    item_status, response = _xattr_mapping(response)
+    if item_status:
+      return item_status, None
+    return status, response
+
+  def xattr(self, attr, timeout=0, callback=None):
+    """Get one extended file attribute value.
+
+    :param attr:  extended attribute name
+    :type  attr:  string
+    :returns:     tuple containing :mod:`XRootD.client.responses.XRootDStatus`
+                  object and the attribute value
+    """
+    if callback:
+      def handle_xattr(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_value(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.get_xattr([attr], timeout, handle_xattr)
+
+    status, response = self.get_xattr([attr], timeout)
+    if not status.ok:
+      return status, None
+    item_status, response = _xattr_value(response)
+    if item_status:
+      return item_status, None
+    return status, response
 
   def clone(self, locs, timeout=0, callback=None):
     """Duplicate ranges from other files into this file by using range based cloning.
--- a/python/libs/client/filesystem.py
+++ b/python/libs/client/filesystem.py
@@ -26,7 +26,7 @@ from __future__ import absolute_import, division, print_function
 from pyxrootd import client
 from XRootD.client.responses import XRootDStatus, StatInfo, StatInfoVFS
 from XRootD.client.responses import LocationInfo, DirectoryList, ProtocolInfo
-from XRootD.client.utils import CallbackWrapper
+from XRootD.client.utils import CallbackWrapper, _xattr_mapping, _xattr_value
 from XRootD.client.flags import AccessMode
 
 class FileSystem(object):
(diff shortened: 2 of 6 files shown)


amadio:
Before doing this type of work, I think we should ensure we can run the tests of the Python bindings in the CI. What do you think?

amadio:
I'd prefer not to duplicate existing interfaces if we can avoid it. I understand the interfaces you are adding are more ergonomic, but usually offering many ways of doing the same thing is not a good idea. Let's bring back this idea after the tape REST API and command line tools are ready. At that point we can start thinking about the longer term evolution for the Python bindings, and prepare ahead of time for XRootD 7.0, by adding the new interfaces and deprecating the old ones after XRootD 7.0 is out.

python/libs/client/filesystem.py
```diff
@@ -437,3 +435,57 @@ def list_xattr(self, path, timeout=0, callback=None):
...
+        callback(status, response, hostlist)
+      return self.get_xattr(path, [attr], timeout, handle_xattr)
+
+    status, response = self.get_xattr(path, [attr], timeout)
+    if not status.ok:
+      return status, None
+    item_status, response = _xattr_value(response)
+    if item_status:
+      return item_status, None
+    return status, response
```
amadio, line 491: These two functions are almost entirely a duplication of one another, including a nested implementation of `handle_xattr` within each one. This pattern is also repeated in other parts of this same pull request. This is in general a problem for maintainability, because it adds a lot more code without adding much more in terms of new functionality. In my opinion, the extra tech debt added here is not worth the small improvements in ergonomics.

python/libs/client/file.py
```diff
@@ -374,6 +375,58 @@ def list_xattr(self, timeout=0, callback=None):
...
+    if callback:
+      def handle_xattrs(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_mapping(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.list_xattr(timeout, handle_xattrs)
```
amadio, line 393: This is also another duplication of the same pattern.

python/libs/client/file.py
```diff
@@ -374,6 +375,58 @@ def list_xattr(self, timeout=0, callback=None):
...
+    if callback:
+      def handle_xattr(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_value(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.get_xattr([attr], timeout, handle_xattr)
```
amadio, line 420: And another duplication of the same pattern.

python/libs/client/filesystem.py
```diff
@@ -437,3 +435,57 @@ def list_xattr(self, path, timeout=0, callback=None):
...
+    if callback:
+      def handle_xattrs(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_mapping(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.list_xattr(path, timeout, handle_xattrs)
```
amadio, line 455: And another duplication of the same pattern.

python/libs/client/filesystem.py
```diff
@@ -437,3 +435,57 @@ def list_xattr(self, path, timeout=0, callback=None):
...
+    if callback:
+      def handle_xattr(status, response, hostlist):
+        if status.ok:
+          item_status, response = _xattr_value(response)
+          if item_status:
+            status, response = item_status, None
+        else:
+          response = None
+        callback(status, response, hostlist)
+      return self.get_xattr(path, [attr], timeout, handle_xattr)
```
amadio, line 483: I think you get the idea by now!

amadio (changes requested):
This adds a few ergonomic interfaces, but at a very high cost in terms of future maintenance needs. I think the costs outweight the benefits here.

amadio:
I'm closing this, I think it's better to rethink this after XRootD 6.2 is released, and certainly the amount of code duplication must be reduced before this can be merged into a release.

PR #2821, opened 2026-06-09
[CI] Run Python binding tests

lobis:
## Summary

- register the Python binding pytest suite as a CTest test using the existing `XRootD::noauth` fixture
- build the test extension as a normal CMake target and stage its build-tree package layout for pytest
- let CMake track the declared binding sources and `XrdCl`/`XrdUtils` target graph instead of maintaining a recursive glob, stamp, and nested `pip install`
- keep the post-install import checks, but drop the separate explicit pytest CI step now that standard CTest runs the suite
- add pytest to the relevant CI and package test dependencies
- allow the Python test endpoint to be overridden with `PYXROOTD_SERVER_URL`

## Rationale

The Python CI previously built and imported the bindings, but did not run the pytest suite through the normal CTest workflow. Running pytest as a CTest entry lets the existing test driver manage the noauth server fixture and keeps Python tests aligned with the rest of CI.

The existing `python/src` CMake target already declares the extension sources and links the `XrdCl` and `XrdUtils` targets. Reusing that target gives the test build complete, native dependency tracking without a second CMake build inside `pip`, a recursive source glob, or a hand-maintained stamp dependency list. Pure Python modules are exposed through the staged package layout, so their changes are picked up directly without rebuilding the extension.

The remote public-data glob tests remain excluded from the CTest invocation so CI does not depend on external EOS availability.

## Validation

- full local CMake build
- `ctest --test-dir build -R '^PyXRootD::pytest$' --output-on-failure`
  - `XRootD::noauth::setup`, `PyXRootD::pytest`, and `XRootD::noauth::teardown` passed
  - pytest result: `55 passed, 4 deselected`
- immediate verbose rebuild of the `client` target performed no compile, link, or `pip` step
- `python3 -m flake8 python/tests/env.py`
- `git diff --check upstream/master...HEAD`
- local YAML parser check for `.github/workflows/CI.yml`

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -10,6 +10,51 @@ else()
   configure_file(setup.py setup.py)
   file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/VERSION "${XRootD_VERSION_STRING}")
 
+  if(BUILD_TESTS AND ENABLE_SERVER_TESTS)
+    execute_process(COMMAND ${Python_EXECUTABLE} -m pytest --version
+      RESULT_VARIABLE PYTEST_STATUS OUTPUT_QUIET ERROR_QUIET)
+
+    if(PYTEST_STATUS EQUAL 0)
+      set(PYXROOTD_TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/test-install)
+
+      add_subdirectory(src)
+      set_target_properties(client PROPERTIES
+        LIBRARY_OUTPUT_DIRECTORY ${PYXROOTD_TEST_DIR}/pyxrootd)
+
+      file(MAKE_DIRECTORY ${PYXROOTD_TEST_DIR}/pyxrootd)
+      file(CREATE_LINK ${CMAKE_CURRENT_SOURCE_DIR}/libs
+        ${PYXROOTD_TEST_DIR}/XRootD SYMBOLIC)
+      configure_file(src/__init__.py
+        ${PYXROOTD_TEST_DIR}/pyxrootd/__init__.py COPYONLY)
+
+      set(PYXROOTD_TEST_ENV
+        PYTHONPATH=${PYXROOTD_TEST_DIR}
+        PYTHONDONTWRITEBYTECODE=1
+        PYXROOTD_SERVER_URL=root://localhost:3094/)
+
+      if(APPLE)
+        list(APPEND PYXROOTD_TEST_ENV
+          DYLD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      else()
+        list(APPEND PYXROOTD_TEST_ENV
+          LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      endif()
+
+      add_test(NAME PyXRootD::pytest
+        COMMAND ${Python_EXECUTABLE} -m pytest -v
+          ${CMAKE_CURRENT_SOURCE_DIR}/tests
+          -k "not test_glob_remote and not test_multiple_glob_with_url_params and not test_folder_glob_with_url_params")
+
+      set_tests_properties(PyXRootD::pytest
+        PROPERTIES
+          FIXTURES_REQUIRED XRootD::noauth
+          ENVIRONMENT "${PYXROOTD_TEST_ENV}"
+          RUN_SERIAL TRUE)
+    else()
+      message(STATUS "pytest not found; Python binding tests disabled")
+    endif()
+  endif()
+
   option(INSTALL_PYTHON_BINDINGS "Install Python bindings" TRUE)
 
   if(INSTALL_PYTHON_BINDINGS)
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -63,6 +63,7 @@ jobs:
           py3-setuptools \
           py3-wheel \
           python3-dev \
+          pytest \
           readline-dev \
           sudo \
           util-linux-dev \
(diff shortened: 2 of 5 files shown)


python/CMakeLists.txt
```diff
@@ -10,6 +10,71 @@ else()
...
+
+      add_custom_command(OUTPUT ${PYXROOTD_TEST_STAMP}
+        COMMAND ${CMAKE_COMMAND} -E rm -rf ${PYXROOTD_TEST_INSTALL_DIR}
+        COMMAND ${Python_EXECUTABLE} -m pip install --no-deps --upgrade
+          --target ${PYXROOTD_TEST_INSTALL_DIR} .
+        COMMAND ${CMAKE_COMMAND} -E touch ${PYXROOTD_TEST_STAMP}
+        DEPENDS ${PYXROOTD_TEST_INPUTS}
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMENT "Building Python bindings for tests"
+        VERBATIM)
```
amadio: I think it's possible to do this in a simpler way without having to perform the `file(GLOB ...)` and dependency tracking by hand. I'd like to try this first. If I don't manage, we can merge this like it is here later.

lobis: improved implementation

python/CMakeLists.txt
```diff
@@ -10,6 +10,51 @@ else()
...
+        PYTHONDONTWRITEBYTECODE=1
+        PYXROOTD_SERVER_URL=root://localhost:3094/)
+
+      if(APPLE)
+        list(APPEND PYXROOTD_TEST_ENV
+          DYLD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      else()
+        list(APPEND PYXROOTD_TEST_ENV
+          LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      endif()
```
amadio, line 41: Why is this required? In my branch I had a detailed comment explaining it, but it has been removed here.

python/CMakeLists.txt
```diff
@@ -10,6 +10,51 @@ else()
...
   file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/VERSION "${XRootD_VERSION_STRING}")
 
+  if(BUILD_TESTS AND ENABLE_SERVER_TESTS)
+    execute_process(COMMAND ${Python_EXECUTABLE} -m pytest --version
+      RESULT_VARIABLE PYTEST_STATUS OUTPUT_QUIET ERROR_QUIET)
+
+    if(PYTEST_STATUS EQUAL 0)
+      set(PYXROOTD_TEST_DIR ${CMAKE_CURRENT_BINARY_DIR}/test-install)
+
+      add_subdirectory(src)
```
amadio, line 20: You cannot have `add_subdirectory(src)` inside `if(PYTEST_STATUS EQUAL 0)`, which is itself inside `if(BUILD_TESTS AND ENABLE_SERVER_TESTS)` because that will break installation of the Python bindings when the server tests are disabled (e.g when XRootD is built as the `root` user).

python/CMakeLists.txt
```diff
@@ -10,6 +10,51 @@ else()
...
+      set_target_properties(client PROPERTIES
+        LIBRARY_OUTPUT_DIRECTORY ${PYXROOTD_TEST_DIR}/pyxrootd)
```
amadio: Have you tried to install the project after this change? It seems to me that this will also place libraries at an unexpected location, which will be missing by the regular install procedure later.

python/CMakeLists.txt
```diff
@@ -10,6 +10,51 @@ else()
...
+          DYLD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      else()
+        list(APPEND PYXROOTD_TEST_ENV
+          LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
+      endif()
+
+      add_test(NAME PyXRootD::pytest
+        COMMAND ${Python_EXECUTABLE} -m pytest -v
+          ${CMAKE_CURRENT_SOURCE_DIR}/tests
+          -k "not test_glob_remote and not test_multiple_glob_with_url_params and not test_folder_glob_with_url_params")
```
amadio, line 46: Rather than disabling tests, they need to be ported to not need eospublic, and remain enabled in the CI.

python/CMakeLists.txt
```diff
@@ -10,6 +10,51 @@ else()
...
+
+      add_subdirectory(src)
+      set_target_properties(client PROPERTIES
+        LIBRARY_OUTPUT_DIRECTORY ${PYXROOTD_TEST_DIR}/pyxrootd)
+
+      file(MAKE_DIRECTORY ${PYXROOTD_TEST_DIR}/pyxrootd)
+      file(CREATE_LINK ${CMAKE_CURRENT_SOURCE_DIR}/libs
+        ${PYXROOTD_TEST_DIR}/XRootD SYMBOLIC)
+      configure_file(src/__init__.py
+        ${PYXROOTD_TEST_DIR}/pyxrootd/__init__.py COPYONLY)
```
amadio, line 28: This is kinda hackish, it will break if we have changes to the layout of the repository, and will not catch installation problems. I think it's better to build the Python bindings as part of the regular build, rather than during the install like we have now, then run the tests against a real installation of the built wheel.

amadio (changes requested):
I'm going to revive my old branch, fix the problems I left unaddressed there with Claude, and create a new pull request. I will add you as reviewer there.

amadio:
Closing, as I will submit a different pull request to address the problems in my old branch, plus the ones identified here.

PR #2822, opened 2026-06-09
[Tests] Use private Kerberos test state

lobis:
## Summary

- select the build-tree Kerberos database module for the test realm
- keep the KDC incremental propagation log in the build tree as well

## Rationale

The Kerberos fixture already defines a private database path under `[dbmodules]`, but the `XROOTD.ORG` realm did not select that module. In CI logs, `kdb5_util` can therefore fall back to system KDC database paths such as `/var/lib/krb5kdc/principal`, making the fixture depend on runner state instead of the private test directory.

Selecting the database module explicitly keeps the KDC database under the build tree, matching the existing fixture cleanup. Setting `iprop_logfile` avoids another KDC state file falling back to a system-derived default.

## Validation

- `git diff --check -- tests/krb5/kdc.conf`
- checked that the `XROOTD.ORG` realm contains both `database_module = XROOTD.ORG` and the build-tree `iprop_logfile`

### DIFF
--- a/tests/krb5/kdc.conf
+++ b/tests/krb5/kdc.conf
@@ -12,6 +12,7 @@ XROOTD.ORG = {
 [realms]
 
 XROOTD.ORG = {
+  database_module = XROOTD.ORG
   acl_file = ""
   allow_pkinit = yes
   max_life = 1h 0m 0s
@@ -22,6 +23,7 @@ XROOTD.ORG = {
   kadmind_listen  = 50749
   kpasswd_listen  = 50464
 
+  iprop_logfile   = @CMAKE_CURRENT_BINARY_DIR@/kdc/db.ulog
   key_stash_file  = @CMAKE_CURRENT_BINARY_DIR@/kdc/.k5.XROOTD.ORG
   pkinit_anchors  = FILE:@CMAKE_CURRENT_BINARY_DIR@/kdc/cacert.pem
   pkinit_identity = FILE:@CMAKE_CURRENT_BINARY_DIR@/kdc/kdc.pem,@CMAKE_CURRENT_BINARY_DIR@/kdc/kdckey.pem
--- a/tests/krb5/kerberos.sh
+++ b/tests/krb5/kerberos.sh
@@ -5,6 +5,7 @@ set -e
 export KRB5CCNAME="${PWD}/krb5cc"
 export KRB5_CONFIG="${PWD}/krb5.conf"
 export KRB5_KDC_PROFILE="${PWD}/kdc.conf"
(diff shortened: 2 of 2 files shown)


amadio:
:+1: Looks reasonable at a glance, but have to cross-check with the docs to make sure.

tests/krb5/kerberos.sh
```diff
@@ -11,6 +11,7 @@ export PATH="${PATH}:/usr/sbin"
 export KRB5CCNAME="${PWD}/krb5cc"
 export KRB5_CONFIG="${PWD}/krb5.conf"
 export KRB5_KDC_PROFILE="${PWD}/kdc.conf"
+export KRB5_DB_NAME="${PWD}/kdc/db"
```
amadio, line 8: I believe this is not needed, as we already set it:

https://github.com/xrootd/xrootd/blob/57ff09c37883f85dd97c55b9ff99b3bcc8efad18/tests/krb5/kdc.conf#L6-L10

The man page of `kdb5_util` says:
```
       -d dbname
              specifies the name under which the principal database is stored;
              by default the database is that listed in kdc.conf.
```

tests/krb5/kdc.conf
```diff
@@ -12,6 +12,7 @@ XROOTD.ORG = {
 [realms]
 
 XROOTD.ORG = {
+  database_module = XROOTD.ORG
```
amadio, line 15: This seems to be the only thing really needed. I will try locally to check.

lobis: I believe the CI will fail if we just add this line that is why I ended up adding the rest but feel free to try I may not remember correctly.

amadio: You are right, it seems that it tries to initialize in `/var`:
```
123: [66105] 1781178427.760746: Retrieving K/M@XROOTD.ORG from FILE:/home/runner/xrootd/build/tests/krb5/kdc/.k5.XROOTD.ORG (vno 0, enctype 0) with result: 0/Success
123: Loading random data
123: Initializing database '/var/kerberos/krb5kdc/principal' for realm 'XROOTD.ORG',
123: master key name 'K/M@XROOTD.ORG'
```

amadio: This is after I added the line above to the config file. Looks like a bug in older versions, as the man page says that it should have used the value from the configuration.

amadio: Actually, it's a red herring. Even if the message says `/var/kerberos/...`, the file is actually created in the expected place. In any case, I think we should add the changes here anyway.

amadio:
@lobis Final thing, I guess this should be done both for v5.9.x and v6.x, no?

PR #2823, opened 2026-06-09
[Python] Expose stat ownership

lobis:
## Summary

- expose `owner` and `group` on Python `StatInfo` objects
- document the new fields in `XRootD.client.responses.StatInfo`
- add a binding test that verifies `File.stat()` provides both attributes

## Rationale

PR #2789 included this binding-side improvement as part of a larger native fsspec filesystem implementation. Since the fsspec filesystem code duplicates functionality maintained elsewhere, this PR extracts only the reusable Python binding piece needed by external filesystem adapters.

`XrdCl::StatInfo` already parses owner and group information, but the Python conversion did not expose those fields. Making them available avoids downstream adapters having to fill ownership fields with placeholder values.

## Validation

- `python3 -m py_compile python/libs/client/responses.py python/tests/test_file.py`
- `git diff --check -- python/src/Conversions.hh python/libs/client/responses.py python/tests/test_file.py`

### DIFF
--- a/python/libs/client/responses.py
+++ b/python/libs/client/responses.py
@@ -169,11 +169,21 @@ class StatInfo(Struct):
   """Status information for files and directories.
 
   :var         id: This file's unique identifier
+  :var       size: The file size (in bytes)
   :var      flags: Informational flags. An `ORed` combination of
                    :mod:`XRootD.client.flags.StatInfoFlags`
-  :var       size: The file size (in bytes)
-  :var    modtime: Modification time (in seconds since epoch)
-  :var modtimestr: Modification time (as readable string)
+  :var      mtime: Modification time (in seconds since epoch)
+  :var    modtime: Deprecated alias for ``mtime``
+  :var modtimestr: Deprecated modification time (as readable string)
+  :var      ctime: Change time (in seconds since epoch)
+  :var      atime: Access time (in seconds since epoch)
+  :var       mode: File mode
+  :var modeoctstr: File mode as a readable permissions string
+  :var      owner: File owner
+  :var      group: File group
+  :var   extended: Whether extended stat information is available
+  :var haschecksum: Whether checksum information is available
+  :var   checksum: File checksum
   """
   def __init__(self, info):
     super(StatInfo, self).__init__(info)
--- a/python/src/Conversions.hh
+++ b/python/src/Conversions.hh
@@ -125,13 +125,22 @@ namespace PyXRootD
   {
       static PyObject* Convert( XrdCl::StatInfo *info )
       {
-        return Py_BuildValue("{sNsNsNsNsN}",
+        return Py_BuildValue("{sNsNsNsNsNsNsNsNsNsNsNsNsNsNsN}",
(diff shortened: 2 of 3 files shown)


python/libs/client/responses.py
```diff
@@ -171,6 +171,8 @@ class StatInfo(Struct):
   :var         id: This file's unique identifier
   :var      flags: Informational flags. An `ORed` combination of
                    :mod:`XRootD.client.flags.StatInfoFlags`
+  :var      owner: File owner
+  :var      group: File group
```
amadio, line 183: There's more that we could expose to make the Python `StatInfo` complete relative to the C++ `StatInfo`:

https://github.com/xrootd/xrootd/blob/b72bd36ef5de200b2309ea43b2ae3bd5b3f89cca/src/XrdCl/XrdClXRootDResponses.cc#L208-L220

Since we are touching this code, we might as well expose all missing fields. What do you think?

lobis: Sure, I think its good idea to expose these as well, I will update the PR

lobis: Addressed, now we expose the full list of fields

python/src/Conversions.hh
```diff
@@ -125,13 +125,27 @@ namespace PyXRootD
...
+                                        info->GetModTimeAsString().c_str()),
+            "changetime", Py_BuildValue("k", info->GetChangeTime()),
+            "changetimestr", Py_BuildValue("s",
+                                           info->GetChangeTimeAsString().c_str()),
+            "accesstime", Py_BuildValue("k", info->GetAccessTime()),
+            "accesstimestr", Py_BuildValue("s",
+                                           info->GetAccessTimeAsString().c_str()),
+            "mode",       Py_BuildValue("s", info->GetModeAsString().c_str()),
+            "modeoctstr", Py_BuildValue("s",
+                                        info->GetModeAsOctString().c_str()),
```
amadio: Maybe make them all a single line? I think it's more readable than having some of the lines broken up. The lines are not that long in any case.

python/libs/client/responses.py
```diff
@@ -169,11 +169,22 @@ class StatInfo(Struct):
...
                    :mod:`XRootD.client.flags.StatInfoFlags`
-  :var       size: The file size (in bytes)
   :var    modtime: Modification time (in seconds since epoch)
   :var modtimestr: Modification time (as readable string)
+  :var changetime: Change time (in seconds since epoch)
+  :var changetimestr: Change time (as readable string)
+  :var accesstime: Access time (in seconds since epoch)
+  :var accesstimestr: Access time (as readable string)
+  :var       mode: File mode
+  :var modeoctstr: File mode as a readable permissions string
```
amadio: It's ok to expose just the field, without the `AsString()` variations. We could have a helper (or maybe one already exists in Python) to convert seconds since epoch to a string. That being said, the `mtime` has the bad precedent of exposing both... These are usually called `mtime`, `ctime`, `atime` everywhere, so I wonder if we should add `mtime` and leave `modtime`/`modtimestr` as deprecated things until we can remove them some time later.

PR #2824, opened 2026-06-10
[CI] alma8 gcc-10 bump

diogomattioli:


amadio:
After some discussion and trying the changes to `xrootd.spec` to enable GCC 15 usage on Alma 8, we realize that there is [no promise of ABI compatibility](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/developing_c_and_cpp_applications_in_rhel_8/additional-toolsets-for-development_developing-applications#cpp-compatibility-in-the-gcc-toolset-15_gcc-toolset-15) with devtoolset compilers. Therefore, while we can survive with C++17 and the little C++20 that GCC 8.5 supports, it's the safer option.

PR #2825, opened 2026-06-10
[XrdPfc] Plug the corner cases with orphaned cinfo or data file

osschar:
Under particular circumstances (loss of data file while cinfo remains, followed by a failed remote stat during an open call) the previous handling was observed to lead to cache erroneously assuming that the file size is supposed to be zero.

This fix is specific to 5.9.x branch as the file open path has diverged with respect to version 6. A separate review will be done there.

* XrdPfc.cc
  - In Cache::GetFile() do not attempt to open the file if initialStat (local or remote) failed.
  - In Cache::Prepare() require an OK state from both cinfo AND data stat calls for deferred open.

* XrdPfcFile.cc
  - In File::Open() truncate an orphaned data file.
  - In File::Open() properly pass the file_existed flag, taking into account possible purging and truncation during open.

* XrdPfcIOFile.cc
  - In IOFile::initialStat() require an OK state from both data AND cinfo stat calls for considering the local stat and file-size information to be current.

### DIFF
--- a/src/XrdPfc/XrdPfc.cc
+++ b/src/XrdPfc/XrdPfc.cc
@@ -431,9 +431,11 @@ File* Cache::GetFile(const std::string& path, IO* io, long long off, long long f
       int res = io->Fstat(st);
       if (res < 0) {
          errno = res;
+         filesize = -1;
          TRACE(Error, "GetFile, could not get valid stat");
       } else if (res > 0) {
          errno = ENOTSUP;
+         filesize = -1;
          TRACE(Error, "GetFile, stat returned positive value, this should NOT happen here");
       } else {
          filesize = st.st_size;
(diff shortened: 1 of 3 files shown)


amadio:
@osschar It would be good to cover the changes introduced here with tests to make sure things work as expected and to prevent regressions. Could you please add tests for this?

osschar:
OK, will do ... let me just do the 6.x version of this fix first. The test should then be good for both.

The test is potentially a bit complicated as the server needs to serve the file to the cach first and then, on the second attempt, return an error code on FS-stat. Hmmh, maybe I could just remove it from the server between the two cache-side file-opens.

abh3 (approved):
It looks like this implements the requirement that the cinfo and data files must exist in order to have a reuse of the file upon open. That does prevent the case of falsely assuming the file size should be zero. As I see it, it does not prevent an incorrect cinfo file to be paired with a data file. While not complex to guard against, it's unlikely that would ever happen. At least that is the supposition.

amadio:
@osschar Just a reminder, we are waiting for the tests in order to merge this into the next 5.9.x patch release.

osschar:
Sorry, can't do this now, still have to do the 6.0 fix and I'm not finding time for that, either.

PR #2826, opened 2026-06-12
[XrdHttp] Tests - Improve smuggling test for header parsing

ccaffy:
It reads the crafted request bytes from stdin, connects to the given
HOST PORT, sends the payload unchanged with sendall(), then reads from
the socket until it gets the first response line, the server closes the
connection, or a 10 second timeout expires.

http.sh just calls that script and can assert on values returned by it
like HTTP/1.1 Bad Request.

### DIFF
--- a/tests/XRootD/http.sh
+++ b/tests/XRootD/http.sh
@@ -3,7 +3,7 @@
 export XRD_PLUGINCONFDIR="${BINARY_DIR}/config"
 
 function setup_http() {
-	require_commands openssl curl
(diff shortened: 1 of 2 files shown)


amadio (approved):
Ok, with the Python script separate, looks ok. Thanks! Later on, we should convert all the integration tests we have now to use [bats](https://bats-core.readthedocs.io/en/stable/index.html). I will integrate it into the repo soon. Cheers,

PR #2827, opened 2026-06-12
[XrdApps] Prototype thin gfal2-util migration frontend

lobis:
## Summary

This intentionally replaces the earlier direct-implementation prototype with the first thin-frontend slice of #2863.

- Stack the current #2865 full-URL change so complete URLs can be delegated to `xrdfs` without splitting them again.
- Restrict that URL normalization to actual path operands, including URL-valued xattrs and prepare request IDs, and document the server-first-only `query` form.
- Add a provisional, build-only `xrd` frontend (working name) for `stat`, `ls`, `cat`, read-only `xattr`, and `copy`/`cp`.
- Translate supported gfal2-util option spellings, then use `execvp` to replace the frontend with `xrdfs` or `xrdcp`.
- Preserve the native implementation, streams, signals, logging, transfer behavior, and exit status; the frontend contains no storage, protocol, transfer, or output implementation.

The `xrd` name, syntax, packaging, and installation policy are **not final**. The target and draft man page are available for review and testing, but neither is installed or added to RPM/DEB packages.

The first commit is the current #2865 head and should be dropped when that PR merges. Follow-up commits contain operand-classification and test-portability fixes found while exercising the frontend.

## Prototype command surface

| Command | Native delegation | Initial compatibility |
| --- | --- | --- |
| `xrd stat URL` | `xrdfs stat URL` | Complete remote URL |
| `xrd ls URL` | `xrdfs ls URL` | `-a`, `-l`, `-H`, non-coloring modes |
| `xrd cat URL...` | `xrdfs cat URL...` | `-b` no-op; same-endpoint files |
| `xrd xattr URL [NAME]` | `xrdfs xattr URL list/get` | Read-only list/get; assignment rejected |
| `xrd copy SRC DST` | `xrdcp SRC DST` | `-f`, `-p`, `-r`, `-n`, `-T`, `-K`, `--from-file` |
| `xrd cp SRC DST` | `xrdcp SRC DST` | Alias for `copy` |

Common timeout, verbosity, certificate/key, IPv4/IPv6, and log-file options are translated to existing XRootD environment settings. Explicit certificate selection also overrides native GSI proxy discovery and inherited serialized XrdSec credentials. Unsupported GFAL-specific behavior fails before delegation with a focused diagnostic.

## Deliberate boundaries

- Output and exit codes are native `xrdfs`/`xrdcp` behavior, not exact gfal2-util formatting or POSIX errno compatibility.
- `sum` is deferred until a native command accepts a complete URL plus checksum algorithm without frontend URL splitting or output parsing.
- Multi-file `cat` currently requires one endpoint; gfal-cat can span endpoints.
- Copy accepts one source/destination pair, or `--from-file` plus one destination. Chained GFAL destinations are rejected because xrdcp would interpret them as multiple sources.
- No mutating namespace command is exposed in this slice. `xattr` is read-only.
- HTTPS delegates correctly when XrdClHttp is installed. Against EOS Public, `stat`, `cat`, and remote-to-local copy work; `ls` currently returns native `NOT_IMPLEMENTED` while `gfal-ls` succeeds. That listing gap belongs in XrdClHttp/`xrdfs`, not in this frontend.

The complete matrix and rationale are in `docs/xrd-gfal-compatibility.md`.

## Validation

- The final upstream Actions run is green: all 17 CI matrix jobs, ABI compatibility, and coverage passed.
- 28/28 focused CTest entries passed: 25 table-driven GTest cases, process-level delegation BATS, the `xrdfs` full-URL BATS suite, and a local-server read-only BATS suite.
- Both local-server BATS suites passed ten consecutive stress runs after making fixture cleanup idempotent.
- GCC 16 Release compiled the parser tests cleanly with `-O3 -Wall -Wextra -Wnull-dereference -Werror`; all 26 parser and process-delegation tests passed in that configuration.
- Strict standalone parser compile passed with `-Wall -Wextra -Werror`.
- ShellCheck, `git diff --check`, and man-page rendering checks passed (apart from existing man-title/version lint conventions).
- A clean configure confirmed that the provisional executable and man page are absent from install metadata.
- EOS Public `root://` validation used only `stat`, `ls`, `cat`, and download to a local temporary file: 2184 bytes, MD5 `93f402e24c6f870470e1c5fcc5400e25`.
- EOS Public HTTPS validation used only `stat`, `cat`, and download to a local temporary file with the same size and MD5; the native listing limitation above was reproduced.
- Reference gfal2-util behavior was checked read-only on lxplus, including successful `gfal-ls` of the same HTTPS directory.

No production storage was modified. Copy acceptance checks wrote only to local temporary destinations; mutation coverage uses only ephemeral local XRootD fixtures.

Refs #2863

### DIFF
--- a/docs/man/CMakeLists.txt
+++ b/docs/man/CMakeLists.txt
@@ -8,6 +8,10 @@ set(MAN1PAGES
   xrdmapc.1
 )
 
+# Keep the provisional migration frontend page renderable for review without
+# installing it before its final name and packaging policy are agreed.
+configure_file(xrd.1 xrd.1 @ONLY)
+
 set(MAN8PAGES
     cmsd.8
     frm_admin.8
--- a/docs/man/xrd.1
+++ b/docs/man/xrd.1
@@ -1,0 +1,223 @@
+.TH xrd 1 "@XRootD_VERSION_STRING@"
+.SH NAME
+xrd \- provisional thin migration frontend for XRootD client commands
+.SH SYNOPSIS
+.B xrd
+.I command
+[\fIoptions\fR] \fIarguments\fR
+.SH DESCRIPTION
+.B xrd
+provides a small command-oriented compatibility surface for operators moving
+from the GFAL command-line utilities to native XRootD tools.
+The executable name and packaging are provisional and remain subject to
+upstream review.
+.PP
+The frontend performs command and option translation only. It replaces itself
+with
+.BR xrdfs (1)
+for filesystem operations or
+.BR xrdcp (1)
+for transfers. It does not implement storage operations, protocol handling,
+output formatting, or transfer logic.
+.PP
+Standard output, standard error, signals, and the delegated command's exit
+status are preserved. The initial implementation therefore produces native
+XRootD output and exit codes; exact GFAL output and POSIX-errno compatibility
+are not yet provided.
+.SH COMMANDS
+.TP
+.BI "stat " URL
+Execute
+.BR "xrdfs stat " URL .
+.TP
+.BI "ls [" options "] " URL
+Execute
+.BR "xrdfs ls " URL
+with supported option translations.
+.TP
+.BI "cat [" options "] " URL...
+Execute
+.BR "xrdfs cat " URL...
+and concatenate one or more files from the same endpoint.
+.TP
+.BI "xattr " URL " [" ATTRIBUTE "]"
+Execute a read-only
+.B xrdfs xattr
+list or get operation. Attribute assignment is rejected.
+.TP
+.BI "copy [" options "] " SOURCE " " DESTINATION
+Execute
+.BR xrdcp (1)
+with supported option translations.
+.TP
+.BI "cp [" options "] " SOURCE " " DESTINATION
+Alias for
+.BR copy .
+.SH COMMON OPTIONS
+.TP
+.B \-h, \-\-help
+Show command help and exit.
+.TP
+.B \-V, \-\-version
+Show the XRootD version and exit.
+.TP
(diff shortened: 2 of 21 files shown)


abh3 (changes requested):
First, I am not in favor of adding CL11 dependency to the xrootd base. It's just adds another thing to worry about outside our scope. This is especially true in cases where the added benefit is small. Long term maintainability is a higher priority than convenience. 

Seconds, I question whether we want to use "xrd" as the command. The set of functions is small and somewhat esoteric. Attaching them to such a fundamental name seems quite odd and may be constraining in the long run. We already have xrdfs, perhaps these functions can happily coexist there, I know many people dislike xrdfs but we never even tried to figure out why, So, there is no guarantee they would like xrd any better.

The change request here is to remove the CLI11 dependency. We should have a discussion on whether xrd is the appropriate vehicle for these functions.

lmascett:
Ciao Andy, we got several comment on the usability of xrdfs and the discrepancy of handling full urls and similar. Ideally it would be great to have a coherent command line for the future as we would need to support additional functions/commands. We can also discuss these details in the next weeks in person, there are multiple way of addressing this.

abh3:
Would love to meet. Why don't we do it when I am at CERN in a couple of weeks, OK?

amadio:
This is likely obsoleted by the discussions in last collaboration meeting, where we agreed to first extend `xrdfs` and add the new command line syntax to it (i.e. with full URLs rather than the split `<endpoint command path>` style), and the new plan in #2863. However, you may simply force-push updates here based on the new plan rather than close this. The new plan looks good to me.

PR #2830, opened 2026-06-16
[Python] Add status exception helpers

lobis:
## Summary

Adds idiomatic Python error handling for unsuccessful `XRootDStatus` responses.

- exposes symbolic `error_name` values derived from the error constants on `XRootDStatus`
- provides dedicated exceptions for not-found, authorization, timeout, checksum, and generic operation failures
- classifies both native client status codes and XRootD server protocol errors, with a generic fallback for unmapped failures
- adds `XRootDStatus.exception()`, `XRootDStatus.raise_on_error()`, and a module-level `raise_on_error()` that also accepts raw status dictionaries
- aligns the Python status constants with the current client definitions while retaining the existing `errPipelineError` spelling as a compatibility alias
- exports the exception classes and helper from `XRootD.client` and documents the public API

## Testing

- adds focused coverage for symbolic names, native and server error mappings, raw dictionary normalization, and exception raising
- `python/tests/test_responses.py`: 9 tests passed

### DIFF
--- a/python/libs/client/__init__.py
+++ b/python/libs/client/__init__.py
@@ -16,5 +16,12 @@ from ._version import __version__
 from .env import EnvGetDefault
 from .env import SetLogLevel
 from .env import SetLogMask
+from .responses import XRootDError
+from .responses import XRootDNotFoundError
+from .responses import XRootDAuthorizationError
+from .responses import XRootDTimeoutError
+from .responses import XRootDChecksumError
+from .responses import XRootDOperationError
+from .responses import raise_on_error
 
 import XRootD.client.finalize
--- a/python/libs/client/responses.py
+++ b/python/libs/client/responses.py
@@ -19,6 +19,35 @@ from __future__ import absolute_import, division, print_function
 
 from XRootD.client.url import URL
 
+
+class XRootDError(RuntimeError):
+  """Base exception raised from unsuccessful :class:`XRootDStatus` objects."""
+
+  def __init__(self, status):
+    self.status = status
+    RuntimeError.__init__(self, str(status))
+
+
+class XRootDNotFoundError(XRootDError):
+  """The requested file or resource was not found."""
+
+
+class XRootDAuthorizationError(XRootDError):
+  """Authentication or authorization failed."""
+
+
+class XRootDTimeoutError(XRootDError):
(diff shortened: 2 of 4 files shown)


python/libs/client/responses.py
```diff
@@ -19,6 +19,80 @@
...
+  301: 'errNoMoreFreeSIDs',
+  302: 'errInvalidRedirectURL',
+  303: 'errInvalidResponse',
+  304: 'errNotFound',
+  305: 'errCheckSumError',
+  306: 'errRedirectLimit',
+  400: 'errErrorResponse',
+  401: 'errRedirect',
+  500: 'errResponseNegative',
+}
```
amadio: This should be part of the existing `class XRootDStatus(Struct):`, where the reverse map from error to error code resides. I wonder if it wouldn't be easier to reuse the C++ code from Python for these things to avoid them becoming out of sync.

python/libs/client/__init__.py
```diff
@@ -14,5 +14,12 @@
 from .env import EnvGetDefault as EnvGetDefault
 from .env import SetLogLevel as SetLogLevel
 from .env import SetLogMask as SetLogMask
+from .responses import XRootDError as XRootDError
+from .responses import XRootDNotFoundError as XRootDNotFoundError
+from .responses import XRootDAuthorizationError as XRootDAuthorizationError
+from .responses import XRootDTimeoutError as XRootDTimeoutError
+from .responses import XRootDChecksumError as XRootDChecksumError
+from .responses import XRootDOperationError as XRootDOperationError
+from .responses import raise_on_error as raise_on_error
```
amadio: Probably better to remove the `as foo` part in these `import foo as foo` lines.

amadio (commented):
The idea of making the errors better integrated into Python is good. Let's try to use this opportunity to wrap the C++ code better to reduce duplication, though, if that is possible. Then we don't have to manually keep them in sync.

python/libs/client/__init__.py
lobis: updated PR with this feedback

python/libs/client/responses.py
lobis: updated

PR #2831, opened 2026-06-16
[Python] Add filesystem query helpers

lobis:
## Summary

Adds Python filesystem helpers for common query-style operations:

- `ChecksumInfo` for structured checksum query responses
- `CapabilityInfo` for endpoint probing summaries
- `FileSystem.checksum()` wrapping `QueryCode.CHECKSUM`
- `FileSystem.mkdir_p()` wrapping recursive mkdir flags
- `FileSystem.probe()` combining ping, protocol, and config query results
- docs and tests for the new response objects and filesystem helpers

## Summary by CodeRabbit

- **New Features**
  - Added context-manager support for filesystem clients.
  - Added checksum retrieval with structured algorithm and value results.
  - Added recursive directory creation through `mkdir_p`.
  - Exported the new checksum response type for public use.

- **Documentation**
  - Expanded filesystem method documentation.
  - Documented checksum response details.

- **Tests**
  - Added coverage for checksum parsing, recursive directory creation, argument handling, and context-manager behavior.

### DIFF
--- a/python/libs/client/__init__.py
+++ b/python/libs/client/__init__.py
@@ -14,5 +14,7 @@ from ._version import __version__ as __version__
 from .env import EnvGetDefault as EnvGetDefault
 from .env import SetLogLevel as SetLogLevel
 from .env import SetLogMask as SetLogMask
+from .responses import ChecksumInfo as ChecksumInfo
+from .responses import CapabilityInfo as CapabilityInfo
 
 import XRootD.client.finalize
--- a/python/libs/client/__init__.py
+++ b/python/libs/client/__init__.py
@@ -24,5 +24,6 @@ from .responses import XRootDTimeoutError
 from .responses import XRootDChecksumError
 from .responses import XRootDOperationError
 from .responses import raise_on_error
+from .responses import ChecksumInfo
 
 import XRootD.client.finalize
--- a/python/libs/client/filesystem.py
+++ b/python/libs/client/filesystem.py
@@ -42,6 +42,14 @@ class FileSystem(object):
   def __init__(self, url):
     self.__fs = client.FileSystem(url)
 
+  def __enter__(self):
+    """Return this filesystem instance for use in a ``with`` statement."""
+    return self
+
+  def __exit__(self, exc_type, exc_value, traceback):
+    """Exit a ``with`` statement without suppressing exceptions."""
+    return None
+
   @property
   def url(self):
     """The server URL object, instance of :mod:`XRootD.client.URL`"""
--- a/python/libs/client/filesystem.py
+++ b/python/libs/client/filesystem.py
@@ -26,8 +26,9 @@ from __future__ import absolute_import, division, print_function
 from pyxrootd import client
 from XRootD.client.responses import XRootDStatus, StatInfo, StatInfoVFS
 from XRootD.client.responses import LocationInfo, DirectoryList, ProtocolInfo
+from XRootD.client.responses import ChecksumInfo, CapabilityInfo
 from XRootD.client.utils import CallbackWrapper
-from XRootD.client.flags import AccessMode
+from XRootD.client.flags import AccessMode, MkDirFlags, QueryCode
 
 class FileSystem(object):
   """Interact with an ``xrootd`` server to perform filesystem-based operations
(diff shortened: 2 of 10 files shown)


python/libs/client/filesystem.py
```diff
@@ -287,6 +321,29 @@ def protocol(self, timeout=0, callback=None):
...
+    if config_arg is None:
+      config_arg = self.__fs.url.hostid
+
+    config_status, config = self.query(QueryCode.CONFIG, config_arg,
+                                       timeout=timeout)
+    return ping_status, CapabilityInfo(ping_status=ping_status,
+                                       protocol_status=protocol_status,
+                                       protocol=protocol,
+                                       config_status=config_status,
+                                       config=config)
```
amadio: I understand the purpose of the other functions being added, but not really for this one. What's the purpose of this function? Just a convenience wrapper on `self.query()`? Why hard-code a timeout of 10? We saw from fsspec-xrootd that this is a bad idea.

python/libs/client/__init__.py
```diff
@@ -14,5 +14,7 @@
 from .env import EnvGetDefault as EnvGetDefault
 from .env import SetLogLevel as SetLogLevel
 from .env import SetLogMask as SetLogMask
+from .responses import ChecksumInfo as ChecksumInfo
+from .responses import CapabilityInfo as CapabilityInfo
```
amadio: Please remove redundant `as foo` from the imports.

python/libs/client/filesystem.py
```diff
@@ -197,6 +223,14 @@ def mkdir(self, path, flags=0, mode=0, timeout=0, callback=None):
     status, response = self.__fs.mkdir(path, flags, mode, timeout)
     return XRootDStatus(status), None
 
+  def mkdir_p(self, path, mode=0, timeout=0, callback=None):
+    """Create a directory and any missing parent directories.
+
+    This is a convenience wrapper around :meth:`mkdir` using
+    :data:`XRootD.client.flags.MkDirFlags.MAKEPATH`.
+    """
+    return self.mkdir(path, MkDirFlags.MAKEPATH, mode, timeout, callback)
```
amadio: This should likely still allow a `flags` variable to be passed in and just use `flags | MkDirFlags.MAKEPATH`. Or, we could consider just making `flags=MkDirFlags.MAKEPATH` the default in the regular `mkdir()` method.

python/libs/client/responses.py
```diff
@@ -165,6 +165,56 @@ class ProtocolInfo(Struct):
...
+class ChecksumInfo(Struct):
+  """Structured checksum response.
+
+  :var algorithm: checksum algorithm name, for example ``adler32``
+  :var value:     checksum value
+  """
+
+  def __init__(self, algorithm, value=None):
+    if value is None:
+      algorithm, value = self._parse(algorithm)
```
amadio: This looks awkward with `algorithm` being both passed in and assigned to in the same statement. Why is `_parse()` doing this?

python/libs/client/filesystem.py
```diff
@@ -287,6 +321,29 @@ def protocol(self, timeout=0, callback=None):
     if response: response = ProtocolInfo(response)
     return XRootDStatus(status), response
 
+  def probe(self, config_arg=None, timeout=10):
+    """Probe basic endpoint capabilities.
```
amadio: This is hard-coding a timeout of 10s without a good reason.

amadio (changes requested):
My comments from July 10 are still unanswered. The extra APIs for checksums are fine, but the `CapabilityInfo` bundling of several things together doesn't have a clear benefit, the `mkdir_p` wrapper doesn't allow to pass `flags` like the normal one does, and the hard-coded timeout of 10s is a blocker. The wrapping of multiple calls into `CapabilityInfo` likely is better located in `gfal` if that's the intended use for it.

amadio:
The filesystem context manager is fine, I might cherry-pick that even if the rest is not used.

lobis:
> My comments from July 10 are still unanswered. The extra APIs for checksums are fine, but the `CapabilityInfo` bundling of several things together doesn't have a clear benefit, the `mkdir_p` wrapper doesn't allow to pass `flags` like the normal one does, and the hard-coded timeout of 10s is a blocker. The wrapping of multiple calls into `CapabilityInfo` likely is better located in `gfal` if that's the intended use for it.

I addressed all comments and brought this old branch up to date

PR #2832, opened 2026-06-16
[Python] Add copy convenience helpers

lobis:
## Summary

Adds higher-level Python copy conveniences:

- `CopyProcess` can be used as a context manager
- `CopyProcess.copy()` runs a single copy job with common options such as mkdir, timeout, retry, checksum settings, progress handler, and cancellation via `should_cancel`
- `FileSystem.copy()` preserves its existing simple path, but routes advanced options through `CopyProcess.copy()`
- docs and server-backed test coverage for the new helper path

### DIFF
--- a/python/libs/client/copyprocess.py
+++ b/python/libs/client/copyprocess.py
@@ -62,6 +62,12 @@ class CopyProcess(object):
   def __init__(self):
     self.__process = client.CopyProcess()
 
+  def __enter__(self):
+    return self
+
+  def __exit__(self, exc_type, exc_value, traceback):
+    return False
+
   def parallel(self, parallel):
     """ Add a config job to the copy process in order to set the number of
         parallel copy jobs.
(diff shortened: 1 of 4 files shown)


python/libs/client/copyprocess.py
```diff
@@ -170,3 +176,56 @@ def run(self, handler=None):
       if 'status' in x:
         x['status'] = XRootDStatus(x['status'])
     return XRootDStatus(status), results
+
+  def copy(self, source, target, force=False, mkdir=False, timeout=0,
+           handler=None, retry=None, thirdparty='none', checksummode='none',
+           checksumtype='', checksumpreset='', posc=False, coerce=False,
+           cont=False, chunksize=None, parallelchunks=None):
+    """Run a single copy job with common high-level options.
```
amadio, line 184: This looks like a duplication of what `add_job` already provides. Please look at this alienpy example using it, maybe it's enough: https://github.com/adriansev/jalien_py/blob/47be38ee62b74cfd54bb64720d2ae3c2625d943b/alienpy/xrd_core.py#L829-L883

amadio:
This one is probably not needed, the code being added duplicates pre-existing code, as commented above.

PR #2834, opened 2026-06-16
[Python] Include py.typed marker

lobis:
## Summary

Marks the Python client package as typed for PEP 561-aware tooling:

- adds `python/libs/client/py.typed`
- includes the marker in both setup entry points via `package_data`


amadio (approved):
This looks good, let's add it to 6.2 (it's not really a bug fix).

PR #2835, opened 2026-06-16
[Python] Add recursive filesystem utilities

lobis:
## Summary

This adds a standalone `XRootD.client.fsutils` module with small high-level helpers for recursive filesystem workflows:

- `mkdir_p` for simple parent-directory creation semantics
- `list_tree` for recursive directory traversal with absolute remote paths
- `directory_size` for file count, byte total, and subdirectory accounting
- `remove_tree` for recursive deletion in the correct order: files first, child directories next, parent directory last

## Motivation

Storage clients such as Rucio and DIRAC can already call low-level XRootD operations like `dirlist`, `rm`, `rmdir`, and `mkdir`, but GFAL currently provides higher-level behavior that applications rely on when replacing GFAL-backed storage plugins with native XRootD bindings.

For DIRAC, this maps to GFAL storage behavior such as directory size/accounting, recursive directory removal, and pieces of recursive directory upload/download workflows. For Rucio, this avoids each integration having to reimplement the same recursive `xrdfs`-style traversal and cleanup logic on top of low-level calls.

The goal is to move this GFAL replacement glue into the Python bindings so clients can share one tested implementation while the lower-level API remains available for callers that need full control.

### DIFF
--- a/python/libs/client/fsutils.py
+++ b/python/libs/client/fsutils.py
@@ -1,0 +1,213 @@
+# -------------------------------------------------------------------------------
+# Copyright (c) 2026 by European Organization for Nuclear Research (CERN)
+# -------------------------------------------------------------------------------
+# This file is part of the XRootD software suite.
+#
+# XRootD is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# XRootD is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+# -------------------------------------------------------------------------------
+import posixpath
+import stat
+
+from XRootD.client.flags import DirListFlags, MkDirFlags, StatInfoFlags
+
+
+class DirectoryEntry:
+    """A normalized directory-listing entry with its absolute remote path."""
+
+    def __init__(self, parent, entry):
+        self.parent = parent
+        self.name = entry.name
+        self.path = remote_join(parent, entry.name)
+        self.hostaddr = getattr(entry, "hostaddr", None)
+        self.statinfo = entry.statinfo
+
+    @property
+    def is_directory(self):
+        return is_directory(self.statinfo)
+
+    @property
+    def is_file(self):
+        return is_file(self.statinfo)
+
+    @property
+    def size(self):
+        return file_size(self.statinfo)
+
+
+class DirectorySize:
+    """Directory accounting information.
+
+    :var files:      number of files
+    :var size:       summed file size in bytes
+    :var subdirs:    number of sub-directories
+    """
+
+    def __init__(self):
+        self.files = 0
+        self.size = 0
+        self.subdirs = 0
+
+    def add(self, entry):
+        if entry.is_directory:
+            self.subdirs += 1
+        elif entry.is_file:
+            self.files += 1
+            self.size += entry.size
+
+    def as_dict(self):
+        return {
+            "Files": self.files,
+            "Size": self.size,
+            "SubDirs": self.subdirs,
+        }
+
+
+class RemoveTreeResult:
+    """Recursive deletion accounting information."""
+
+    def __init__(self):
+        self.files_removed = 0
+        self.directories_removed = 0
+        self.size_removed = 0
+
+    def as_dict(self):
+        return {
+            "FilesRemoved": self.files_removed,
+            "DirectoriesRemoved": self.directories_removed,
+            "SizeRemoved": self.size_removed,
+        }
+
+
+def remote_join(parent, name):
+    """Join XRootD paths without using platform-specific path separators."""
+    if not parent:
+        parent = "/"
+    return posixpath.join(parent.rstrip("/") or "/", name)
+
+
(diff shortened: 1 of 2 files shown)


python/libs/client/fsutils.py
```diff
@@ -0,0 +1,199 @@
...
+  def __init__(self, parent, entry):
+    self.parent = parent
+    self.name = entry.name
+    self.path = remote_join(parent, entry.name)
+    self.hostaddr = getattr(entry, 'hostaddr', None)
+    self.statinfo = entry.statinfo
+
+  @property
+  def is_directory(self):
+    return is_directory(self.statinfo)
```
amadio: Please try to write new code following [PEP8](https://peps.python.org/pep-0008/#indentation). I suggest using 4 spaces for indentation. At some point we can then introduce a lint check for the coding style too.

abh3: You may wish to use flake8 which will tell you want needs to change to adhere to PEP8 specifications. Doing it this way avoids incremental stylistic changes and gets the pain over with in one go.

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,199 @@
...
+#
+# XRootD is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
+# -------------------------------------------------------------------------------
+from __future__ import absolute_import, division, print_function
```
abh3: Can we just stop supporting python 2 and drop this import? If not why not?

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,199 @@
...
+def is_directory(statinfo):
+    """Return ``True`` when a stat response describes a directory."""
+    return bool(statinfo and statinfo.flags & StatInfoFlags.IS_DIR)
+
+
+def is_file(statinfo):
+    """Return ``True`` when a stat response describes a regular file."""
+    if not statinfo:
+        return False
+    return not bool(statinfo.flags & (StatInfoFlags.IS_DIR | StatInfoFlags.OTHER))
```
abh3: This is wrong as it will return true for things that are not actual files. The correct way is:
return bool( statinfo.flags & StatInfoFlags.S_ISREG)

abh3 (changes requested):
A recommendation and a correction.

python/libs/client/fsutils.py
amadio: This is actually to support older versions of Python 3 (e.g. Python 3.6 in Alma 8), not to support Python 2.x. Support for Python 2.x has been already removed from the code a while ago in 0b3fe3a77422d1e9a9c059a0a0805195d1aeebcb.

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,213 @@
...
+    def is_directory(self):
+        return is_directory(self.statinfo)
+
+    @property
+    def is_file(self):
+        return is_file(self.statinfo)
+
+    @property
+    def size(self):
+        return file_size(self.statinfo)
```
amadio, line 45: This doesn't consider directories, so what happens when `foo.size()` is called on a `DirectoryEntry` which has `is_directory() == true`?

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,213 @@
...
+        elif entry.is_file:
+            self.files += 1
+            self.size += entry.size
+
+    def as_dict(self):
+        return {
+            "Files": self.files,
+            "Size": self.size,
+            "SubDirs": self.subdirs,
+        }
```
amadio, line 73: Is this in line with how fsspec does things? A class to represent directory size seems a bit overkill, no?

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,213 @@
...
+        self.files_removed = 0
+        self.directories_removed = 0
+        self.size_removed = 0
+
+    def as_dict(self):
+        return {
+            "FilesRemoved": self.files_removed,
+            "DirectoriesRemoved": self.directories_removed,
+            "SizeRemoved": self.size_removed,
+        }
```
amadio, line 89: Same here, I think we can do without the class for this.

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def remote_join(parent, name):
+    """Join XRootD paths without using platform-specific path separators."""
+    if not parent:
+        parent = "/"
+    return posixpath.join(parent.rstrip("/") or "/", name)
```
amadio, line 96: Why set `parent` to `"/"` above if you have `or "/"` here? The name `remote_join` doesn't make much sense when by default `parent` is `/`, that is, a local path...

python/libs/client/fsutils.py
```diff
@@ -0,0 +1,213 @@
...
+        status, entries = list_directory(filesystem, path, timeout)
+
+    result = DirectorySize()
+    for entry in entries:
+        result.add(entry)
+    return status, result
+
+
+def remove_tree(filesystem, path, timeout=0):
+    """Recursively remove a directory tree.
```
amadio, line 184: This timeout is deceptive, because a user may think it applies to `remove_tree`, but actually it applied to each individual operation inside it. Same for the other composite operations exposed here.

amadio (commented):
I'm not really convinced this should be in XRootD's bindings and not in the code using XRootD. Let me come back to it after holidays... For now, I left some comments.

lobis:
Closing this, if we really need this I will open a new PR but some of this changes are already part of existings PRs

PR #2836, opened 2026-06-16
[XrdCl] Add Tape REST client API

### LINKED ISSUE #2862: Add WLCG Tape REST API support to XRootD
### Feature Description

Track the work needed to expose the [WLCG Tape REST API](https://github.com/wlcg-storage/wlcg-tape-rest-api) through XRootD client and server surfaces.

The client goal is to let XRootD applications interact with tape-backed storage using the existing `XrdCl::FileSystem` model where it fits, rather than adding a disconnected client API. In particular:

- map stage / bring-online, cancel, and evict / release operations onto `FileSystem::Prepare`;
- map polling and metadata-style requests onto `FileSystem::Query` where possible;
- keep HTTP-specific discovery and WLCG Tape REST request translation inside `XrdClHttp`;
- expose the same functionality through the Python bindings;
- keep the path open for CLI integration in `xrd` / `xrdfs` without introducing a second implementation.

The server goal is to develop `XrdHttpTapeApi` as the production XrdHttp external handler for Tape REST requests. Its initial deterministic routes provide end-to-end CI coverage, but the component intentionally lives in the production plugin tree so it can evolve into the server-side bridge between WLCG Tape REST operations and XRootD storage/tape interfaces. It is not intended to remain a test-only mock.

### References

- WLCG Tape REST API specification: https://github.com/wlcg-storage/wlcg-tape-rest-api
- WLCG Tape REST API v1 reference PDF: https://github.com/wlcg-storage/wlcg-tape-rest-api/blob/main/WLCG_TapeRESTAPI_Reference_Document_v1.pdf
- XRootD discussion meeting: https://indico.cern.ch/event/1700187
- Initial implementation PR: https://github.com/xrootd/xrootd/pull/2836

### Current Implementation Direction

The first implementation is being developed in #2836. It adds Tape REST support to `XrdClHttp`, routes client-facing operations through `FileSystem::Prepare` / `Query`, adds Python bindings on top of that client surface, and introduces the production `XrdHttpTapeApi` handler.

The current direction is to keep the implementation aligned with existing XRootD interfaces:

- no new public Tape-specific C++ client API unless a gap is identified;
- `Prepare` handles stage, cancel, and release / evict style operations;
- `Query` handles stage polling and other read-style Tape REST operations;
- Tape REST HTTP calls run through the existing XrdClHttp curl worker infrastructure;
- `XrdHttpTapeApi` remains a production component, with deterministic behavior initially used to exercise the full client/server path in CI;
- tests cover native behavior, Python bindings, and an end-to-end CLI path through `xrdfs`.

### Open Follow-up Items

- Agree which Tape REST operations require new or extended `kXR_query` request codes.
- Finalize how archive information and stage deletion should be represented in the XRootD client model.
- Wire the functionality into the new `xrd` CLI / existing `xrdfs` surfaces once the client behavior is stable.
- Replace or extend the initial deterministic `XrdHttpTapeApi` behavior with translation to real XRootD storage and tape operations.

lobis:
## Summary

Add WLCG Tape REST API v1 support through the existing XrdCl interfaces.

Native applications use `FileSystem::Prepare` and `FileSystem::Query`; the HTTP
discovery, request translation, authentication, and response normalization
remain internal to `XrdClHttp`. Python users get a synchronous
`XRootD.client.TapeClient` wrapper over the same native operations.

The PR also adds `XrdHttpTapeApi`, a packaged XrdHttp external handler with a
deterministic filesystem-backed implementation for end-to-end integration
testing and future server-side development.

## Supported operations

| Operation | Tape REST endpoint | XrdCl interface | Python API |
|---|---|---|---|
| Discovery | `GET /.well-known/wlcg-tape-rest-api` | `QueryCode::Opaque` | `discover()` |
| Stage | `POST /stage` | `PrepareFlags::STAGE` | `stage()` |
| Stage status | `GET /stage/{request_id}` | `QueryCode::Prepare` | `stage_status()` |
| Cancel files | `POST /stage/{request_id}/cancel` | `PrepareFlags::CANCEL` | `stage_cancel()` |
| Delete request | `DELETE /stage/{request_id}` | `QueryCode::Opaque` | `stage_delete()` |
| Release files | `POST /release/{request_id}` | `PrepareFlags::EVICT` | `release()` |
| Archive information | `POST /archiveinfo` | `QueryCode::Opaque` | `archive_info()` |

## XrdClHttp client

The client:

- discovers and validates a version-one Tape REST endpoint before dispatching
  an operation;
- translates storage URLs and percent-encodes request identifiers and paths;
- supports HTTP/WebDAV and native XRootD storage URLs;
- normalizes stage-status and archive-information responses;
- validates timestamps, paths, localities, per-file errors, and endpoint
  consistency;
- maps failed HTTP responses through the existing `HTTPStatusConvert` logic,
  preserving standard XRootD error codes and problem-response details;
- negotiates checksum algorithms using the existing XrdClHttp checksum and
  header parsing infrastructure;
- rejects ambiguous prepare flags and malformed structured stage entries.

Bearer-token injection is handled in the common `CurlOperation` setup path
instead of being specific to Tape requests. It honors `XrdSecPROTOCOL`
ordering between bearer-token and X.509 credentials while preserving explicit
`Authorization` headers and `authz` URL parameters. The common hook is aligned
with the authentication work in #2856 to limit future reconciliation to token
discovery details.

Supporting URL joining and path normalization were moved into `XrdOucUtils`
and are also reused by the S3 and Macaroons implementations.

## Python API

Add `XRootD.client.TapeClient` and response wrappers for discovered endpoints,
stage requests, stage status, per-file status, and archive information.

`TapeClient.stage()` accepts scalar or sequence inputs, full storage URLs, and
structured dictionaries containing:

- `path` or `url`;
- `diskLifetime` as an ISO-8601 duration or non-negative seconds;
- `targetedMetadata` dictionaries.

The wrapper preserves the selected protocol and port, validates structured
inputs before dispatch, and returns native `XRootDStatus` objects alongside
typed Tape response objects.

## XrdHttpTapeApi handler

Add and package `XrdHttpTapeApi` as an XrdHttp external handler.

The initial implementation is a deterministic simulator, not a production
tape backend. It persists stage requests and models archive and disk replicas
under a configured state directory:

- `archive/` contains simulated tape replicas;
- `disk/` contains staged files served by XRootD;
- `requests/` contains persisted stage-request metadata.

Stage, cancel, release, deletion, and archive-information operations are
reproducible across requests and server restarts. Request bodies default to a
4 MiB limit, with an optional handler parameter for deployments requiring a
different limit.

## Tests

The PR adds:

- native unit tests for request construction, response normalization,
  authentication selection, checksum negotiation, malformed responses, and
  error propagation;
- unit tests for the filesystem-backed Tape API store;
- a BATS integration test covering the authenticated stage lifecycle,
  locality transitions, unsupported flags, and RFC 7807 error responses;
- Python tests for the complete `TapeClient` surface;
- expanded `xrdfs` end-to-end command coverage.

Related: #2862


.github/workflows/python.yml
```diff
@@ -83,7 +83,7 @@ jobs:
 
     steps:
     - name: Install dependencies
-      run: dnf install -y git krb5-devel libuuid-devel openssl-devel
+      run: dnf install -y git krb5-devel libcurl-devel libuuid-devel openssl-devel
```
amadio: This should probably be applied separately, to ensure that our binary wheels for XRootD 6.x contain the XrdClHttp plugin.

amadio (commented):
This has improved a lot relative to linking XrdCl with curl, but is still not going in the direction I believe is the most natural.

I think a deeper look at how existing things work may help you with the tape REST API implementation. I asked in deepwiki for an explanation of the code paths, which you can see here: https://deepwiki.com/search/please-explain-to-me-the-code_9332791c-f0bc-4b63-9fcb-d116354efc70?mode=deep

I repeat here the diagram of the code paths which are important for this:

Image

It seems to me that everything in this pull request is implemented in a way that is disconnected from existing interfaces. The manual inspection of JSON to check if something is a tape REST API call or not looks like a code smell for me.

In my opinion, most of what is the tape REST API could be implemented in terms of existing XRootD APIs. For example, `stage` (or `bringonline`), `evict`, and `cancel` fit into `XrdCl::FileSystem::Prepare`, while most of the rest could fit `XrdCl::FileSystem::Query`, as done for checksums (this is why I asked deepwiki for these two codepaths above). The natural way of implementing the tape REST API is to add a `Prepare` and `Query` implementation into `XrdClHttp` as a first step, as checksums are already supported as far as I can tell.

If something is truly missing in the XRootD APIs, which is likely the case for stage/archive polling, then we discuss with Andy how to extend `kXR_query` requests for these requests in a way that is easy to translate also into the HTTP world.

amadio (changes requested):
This still contains a lot of code duplication in the helper functions. When functionality is already available in other parts of the code, it should be reused instead of writing a new copy here.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -154,12 +156,69 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+    );
+    try
+    {
+        m_queue->Produce(std::move(tapeOp));
+    }
+    catch (...)
+    {
+        m_logger->Warning(kLogXrdClHttp, "Failed to add tape prepare operation to queue");
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errOSError);
+    }
```
amadio: It would be good to have an indication of which type of exception happened, otherwise this will make failures hard to debug in the future.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -154,12 +156,69 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidArgs,
+            0, "missing prepare file list");
+    }
+
+    const auto ts = XrdClHttp::Factory::GetHeaderTimeoutWithDefault(timeout);
+    std::unique_ptr<CurlTapeOp> tapeOp(
+        new CurlTapeOp(
+            handler, m_url.GetURL(), fileList, flags, ts, m_logger,
+            GetConnCallout(),
+            m_header_callout.load(std::memory_order_acquire)
```
amadio: Please use `std::make_unique` rather than `new`, same for other places where this pattern exists.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -154,12 +156,69 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+        try
+        {
+            m_queue->Produce(std::move(tapeOp));
+        }
+        catch (...)
+        {
+            m_logger->Warning(kLogXrdClHttp, "Failed to add tape query operation to queue");
+            return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errOSError);
+        }
+        return XrdCl::XRootDStatus();
```
amadio: There's some boiler plate code repeated on every operation with the `try/catch` block. I suggest to try to refactor into a switch, moving part of the boilerplace outside if possible.

src/XrdClHttp/XrdClHttpOps.cc
```diff
@@ -599,6 +586,42 @@ CurlOperation::Setup(CURL *curl, CurlWorker &worker)
...
+
+bool
+CurlOperation::SetupNextRequest(const std::string &url, CurlWorker &worker)
+{
+    if (!m_curl) return false;
+
+    curl_easy_reset(m_curl.get());
+    ConfigureHandle(m_curl.get(), false);
+
+    SetUrl(url);
```
amadio: This `SetUrl(url);` wrapper seems overkill, I suggest to just do
```
m_url = DavToHttp(url);
```
here directly.

src/XrdClHttp/XrdClHttpOps.hh
```diff
@@ -390,7 +401,7 @@ private:
 
 protected:
     void SetDone(bool has_failed) {m_done = true; m_has_failed.store(has_failed, std::memory_order_release);}
-    const std::string m_url;
+    std::string m_url;
```
amadio: Why is this change needed? Can `m_url` be set to `DavToHttp(url)` in the constructor to avoid removing the `const` here?

src/XrdClHttp/XrdClHttpOps.hh
```diff
@@ -608,6 +619,43 @@ public:
...
+    CurlTapeOp(XrdCl::ResponseHandler *handler, const std::string &url,
+        const std::vector<std::string> &file_list,
+        XrdCl::PrepareFlags::Flags flags, struct timespec timeout,
+        XrdCl::Log *logger,
+        CreateConnCalloutType callout, HeaderCallout *header_callout);
+
+    CurlTapeOp(XrdCl::ResponseHandler *handler, const std::string &url,
+        XrdCl::QueryCode::Code query_code, const XrdCl::Buffer &arg,
+        struct timespec timeout, XrdCl::Log *logger,
+        CreateConnCalloutType callout, HeaderCallout *header_callout);
```
amadio: Shouldn't these two be different derived classes of `CurlOperation`? They represent very different things, actually. At least, you could have a generic class to add the members you need, then derive from that for each of the two above.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,218 @@
...
+        Complete(status, response);
+        return;
+    }
+
+    m_request = std::move(nextRequest);
+    if(!m_worker || !SetupNextRequest(m_request.url, *m_worker)
+       || !ConfigureRequest() || !FinishSetup(m_curl.get()))
+    {
+        Complete(XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInternal,
+          0, "failed to configure the next Tape REST request"), "");
```
amadio: Again, it would be good to make an effort to give a more specific error to the user here than this, as it's not even clear which type of request failed. Also, why call `Complete(...)` on a failure, given that `CurlTapeOp::Fail()` exists?

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,218 @@
...
+size_t
+CurlTapeOp::WriteCallback(char *buffer, size_t size, size_t nitems, void *data)
+{
+    return static_cast<CurlTapeOp *>(data)->Write(buffer, size * nitems);
+}
+
+size_t
+CurlTapeOp::Write(char *buffer, size_t size)
+{
+    static constexpr size_t MaxResponseSize = 64 * 1024 * 1024;
```
amadio: This seems fairly arbitrary. Why fail for responses larger than 64MB?

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,218 @@
...
+
+    auto handle = m_handler;
+    m_handler = nullptr;
+
+    auto buffer = new XrdCl::Buffer();
+    buffer->FromString(response);
+
+    auto obj = new XrdCl::AnyObject();
+    obj->Set(buffer);
+    handle->HandleResponse(new XrdCl::XRootDStatus(), obj);
```
amadio: I see memory leaks here.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  }
+};
+
+struct TapeStageStatus
+{
+  std::string id;
+  std::vector<TapeStageFileStatus> files;
+  std::uint64_t createdAt = 0;
+  std::uint64_t startedAt = 0;
+  std::uint64_t completedAt = 0;
```
amadio: Since these are times, I suggest using `time_t`.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  bool HasStartedAtValue() const { return (flags & HasStartedAt) != 0; }
+  bool HasCompletedAtValue() const { return (flags & HasCompletedAt) != 0; }
+};
+
+std::string NormalizeTapePath(const std::string &path)
+{
+  std::string result = path.empty() ? "/" : path;
+  if(result.front() != '/') result.insert(result.begin(), '/');
+  return result;
+}
```
amadio: I think it makes sense to take `NormalizeSlashes` from `XrdMacaroons`, rename it to `NormalizePath` and put into `XrdOucUtils`, then reuse here to avoid code duplication.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+
+std::string JoinUrl(const std::string &base, const std::string &path)
+{
+  if(base.empty()) return path;
+  if(path.empty()) return base;
+  if(base.back() == '/' && path.front() == '/') return base + path.substr(1);
+  if(base.back() != '/' && path.front() != '/') return base + "/" + path;
+  return base + path;
+}
```
amadio: Same for this one, probably better placed in a common place to be used everywhere.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  {
+    return false;
+  }
+
+  char *end = nullptr;
+  const long number = std::strtol(value.c_str(), &end, 10);
+  if(end == value.c_str() || *end != '\0' || number < 0) return false;
+  parsed = static_cast<int>(number);
+  return true;
+}
```
amadio: This is another candidate for "utils" placement. It could be needed elsewhere later on.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+    return false;
+  }
+
+  std::string protocol = url.GetProtocol();
+  std::transform(protocol.begin(), protocol.end(), protocol.begin(),
+    [](unsigned char c) { return std::tolower(c); });
+  bool useUrlPort = true;
+  if(protocol == "davs") protocol = "https";
+  else if(protocol == "dav") protocol = "http";
+  else if(protocol == "root" || protocol == "xroot")
```
amadio: Not sure about this one, you probably will never end up here with `root://` or `xroot://` and this doesn't consider `roots://` and `xroots://`. I'd make it an error if you get anything other than the http/davs protocols in your urls rather than silently replace with https, as the port will most likely be wrong anyway.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+          if(problemDetail != json.end() && problemDetail->is_string())
+          {
+            detail += " - " + problemDetail->get<std::string>();
+          }
+        }
+      }
+    }
+    catch(const std::exception &)
+    {
+    }
```
amadio: We shouldn't be silently ignoring exceptions.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+    if(detail.empty()) detail = body;
+  }
+
+  constexpr std::size_t MaxProblemDetail = 4096;
```
amadio: Another magic number, why cap it at 4096? Is this in the specs? Can you please check what the RFC says on this and adapt the code to it?

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  std::ostringstream out;
+  out << "HTTP " << statusCode;
+  if(!detail.empty()) out << ": " << detail;
+  return out.str();
+}
+
+XrdCl::XRootDStatus ErrorStatus(uint16_t code, const std::string &message)
+{
+  return XrdCl::XRootDStatus(XrdCl::stError, code, 0, message);
+}
```
amadio, line 180: It's wrong to always use `0` for errors in the third argument. It should contain the equivalent `errno` value if one matches the problem being reported as an error.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+{
+  std::vector<std::string> lines;
+  XrdCl::Utils::splitString(lines, value, "\n");
+  return lines;
+}
+
+bool ContainsCarriageReturn(const std::string &value)
+{
+  return value.find('\r') != std::string::npos;
+}
```
amadio: Not sure these are really needed, are they used so frequently to need a wrapper for so little code? In any case, these are fairly generic things that belong in utils rather than here.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+        }
+        files.push_back(entry);
+        continue;
+      }
+      catch(const std::exception &ex)
+      {
+        return ErrorStatus(XrdCl::errInvalidArgs,
+          "malformed structured tape stage entry: "
+          + std::string(ex.what()));
+      }
```
amadio: Is malformed JSON the only way an exception will be raised above? That is not clear to me. Maybe a failed allocation could also raise an exception, in which case, the error message will be wrong. We should probably catch exceptions by each type and have a catch all returning `unknown exception` as the error. Then, if we do see `unknown exception` in the future, we should add a new case providing a better error message here.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  {
+    value = static_cast<std::uint64_t>(it->get<std::int64_t>());
+  }
+  else
+  {
+    return false;
+  }
+  hasValue = true;
+  return true;
+}
```
amadio: This should probably be in `XrdOucJson.hh`. Actually, I'd create a `XrdOucJsonUtils.hh` and include it at the end of `XrdOucJson.hh` to avoid making that header even bigger...

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+    status.files.push_back(fileStatus);
+  }
+  return XrdCl::XRootDStatus();
+}
+
+XrdCl::XRootDStatus EmptyResponseStatus(long statusCode,
+                                        const std::string &body,
+                                        const std::string &operation)
+{
+  if(statusCode != 200 && statusCode != 204)
```
amadio: What about `201 Created` and `202 Accepted`? Can they not happen here?

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1285 @@
...
+  if(strcasecmp(locality.c_str(), "disk") == 0) return TapeLocality::Disk;
+  if(strcasecmp(locality.c_str(), "tape") == 0) return TapeLocality::Tape;
+  if(strcasecmp(locality.c_str(), "disk_and_tape") == 0)
+    return TapeLocality::DiskAndTape;
+  if(strcasecmp(locality.c_str(), "lost") == 0) return TapeLocality::Lost;
+  if(strcasecmp(locality.c_str(), "none") == 0) return TapeLocality::None;
+  if(strcasecmp(locality.c_str(), "unavailable") == 0)
+    return TapeLocality::Unavailable;
+  return TapeLocality::Unknown;
+}
```
amadio: I suggest a for loop over a map from enum to names instead of these hard-coded loops. Then if/when the map changes, these functions don't have to. The code also becomes simpler. Something similar to the snippet below:

https://github.com/xrootd/xrootd/blob/4ec63a24e099f108f1b7af8cd0dd4e41ace6a8c5/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L324-L330

amadio (changes requested):
Alright, I have some comments, will continue later. Please take a look at the suggested changes.

lobis:
### Summary of Integration Testing: WLCG Tape REST API Implementation

We conducted end-to-end integration tests of the new XRootD WLCG Tape REST API client (`XrdClHttp`, `xrdfs`, and the Python `TapeClient` bindings) against a real EOS/CTA test instance configured with X.509 proxy authentication.

---

### 1. Operations Tested & Validated

#### A. Endpoint Discovery (`discover`)
- Verified automated discovery of the Tape REST API endpoint via `/.well-known/wlcg-tape-rest-api`.
- Successfully parsed and populated the site name, API version (`v1`), and base endpoint URI.

#### B. Archive Locality Queries (`archive_info`)
- Tested single and batch locality queries across various file states:
  - **Tape-only files**: Correctly returned locality `TAPE` (`user.status: NEARLINE`).
  - **Disk-resident files**: Correctly returned locality `DISK` (`user.status: ONLINE`).
  - **Dual-resident files**: Correctly returned locality `DISK_AND_TAPE`.
  - **Non-existent / Inaccessible files**: Correctly returned RFC 7807 problem details with per-file error descriptions.
- Verified both `https://`, `davs://`, and `root://` schemes (including custom storage ports).

#### C. Staging Lifecycle (`stage`, `stage_status`, `stage_cancel`, `release`, `stage_delete`)
- **Stage Submission (`stage`)**: Submitted staging requests with custom `disk_lifetime` and optional targeted metadata; successfully received server-generated UUID request identifiers.
- **Status Polling (`stage_status`)**: Polled active request states; verified accurate extraction of creation/start timestamps, file lists, and `onDisk` boolean availability flags.
- **Stage Cancellation (`stage_cancel`)**: Verified canceling file staging requests prior to completion.
- **Release / Eviction (`release`)**: Verified releasing pinned disk copies back to tape latency.
- **Request Cleanup (`stage_delete`)**: Successfully deleted stage request records from the server and verified that subsequent status lookups return HTTP 404, cleanly mapped to XRootD status `kXR_NotFound` (3011).

#### D. File Upload & Archival Queuing
- Uploaded fresh test files via `xrdcp` to verify proper initial registration on disk and transition into the tape archival queue.

---

### 2. Improvements & Fixes Implemented from Live Testing

1. **Safe Attribute Access in `TapeArchiveInfo`**:
   - Initialized `locality` and `error` attributes to default to `None`. This prevents `AttributeError` exceptions when inspecting mixed batches containing both valid and errored file items.

2. **Port Preservation in Scheme Translation**:
   - Updated `_filesystem_url()` so custom storage ports (e.g. non-default HTTPS/DAV ports) are preserved when converting `root://host:port/path` to HTTPS.

3. **Scalar Argument Convenience**:
   - Added support for single string URLs/paths and dictionaries across `stage()`, `stage_cancel()`, `release()`, and `archive_info()`, eliminating the need for mandatory list wrapping on single-item calls.

4. **Grid Middleware Integration (DIRAC)**:
   - Verified that the Python bindings can directly back native storage element plugins (`XROOTStorage` and `CTAStorage`), translating WLCG Tape REST localities into standard Grid metadata attributes (`Cached`, `Migrated`, `Accessible`, `user.status`) and replacing legacy GFAL2 dependencies.

---

### 3. Test Results
- **C++ Tests (`ctest` & BATS)**: 100% passing.
- **Python Unit Tests (`pytest`)**: 100% passing (26/26 tests).
- **Live Endpoint Integration Tests**: All read-only and staging lifecycle operations verified successfully.

amadio:
@lobis Can you please rebase on the couple of things I already merged into master and force-push? Thanks!

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -41,6 +47,38 @@ Filesystem::Filesystem(const std::string &url, std::shared_ptr<HandlerQueue> que
...
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errOSError,
+            EIO, ex.what());
+    }
+    catch(...)
+    {
+        m_logger->Warning(kLogXrdClHttp,
+            "Failed to add %s to queue: unknown exception", description);
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errOSError,
+            EIO, "unknown exception while queueing HTTP operation");
+    }
```
amadio: Only the error and string changes in these 3 `catch` blocks, can you maybe catch generically and use `ex.code()` to assign the error code and error message to avoid code duplication?

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,350 @@
+/******************************************************************************/
+/* Copyright (C) 2025, Pelican Project, Morgridge Institute for Research      */
```
amadio: This should use CERN rather than Pelican, no? Please fix in all related files.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,350 @@
...
+        Fail(XrdCl::errInternal, EIO,
+            "exception while preparing a Tape REST request: "
+            + std::string(ex.what()));
+    }
+    catch(...)
+    {
+        if(!m_curl) m_curl.reset(curl);
+        Fail(XrdCl::errInternal, EIO,
+            "unknown exception while preparing a Tape REST request");
+    }
```
amadio: Same thing. Can we reduce code duplication? A single `catch` with a `switch` on `ex.code()` inside to figure out what error message to print, and then a single call to `Fail()` outside the switch, same for `if(!m_curl) m_curl.reset(curl);`.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,350 @@
...
+
+bool
+CurlTapeOp::ConfigureRequest()
+{
+    if(!m_curl) return false;
+
+    m_response.clear();
+    m_headers_list.emplace_back("Accept", "application/json");
+    // CurlOperation::Setup configures the worker's X.509 credential.  A bearer
+    // token is added independently when one is available.
```
amadio: We should probably inspect and follow what's in `XrdSecPROTOCOL` for authentication, even within the plugin for other protocols. Users that set `XrdSecPROTOCOL=gsi,ztn` might expect their client to always use a certificate at higher preference.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,350 @@
...
+    {
+        Fail(XrdCl::errInternal, EIO,
+            "exception while processing " + RequestDescription()
+            + ": " + ex.what());
+    }
+    catch(...)
+    {
+        Fail(XrdCl::errInternal, EIO,
+            "unknown exception while processing " + RequestDescription());
+    }
```
amadio, line 219: This could probably be simplified a lot as well. Since this is a try/catch, you could setup an exception in the `try` block instead of mixing try/catch with regular error handling. It will be better in the end. You could throw a `std::logic_error` or `std::runtime_error` or somthing more specific depending on the error type. The catch block comments from before apply here too, the code is the same, only the exception is different, let's have a single catch block to avoid unnecessary code duplication.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,350 @@
...
+    {
+        return FailCallback(kXR_ServerError,
+            "exception while buffering a Tape REST response: "
+            + std::string(ex.what()));
+    }
+    catch(...)
+    {
+        return FailCallback(kXR_ServerError,
+            "unknown exception while buffering a Tape REST response");
+    }
```
amadio, line 288: This pattern is really everywhere, please fix them first, then I will come back and review again.

amadio (changes requested):
Too much code duplication in the error handling paths, and sometimes ugly mixing of normal error handling with exception throwing. Let's get these fixed, then another round of review. Thanks!

amadio:
@lobis Can you please click to resolve the comments you addressed?

python/libs/client/tape.py
```diff
@@ -0,0 +1,268 @@
...
+
+  def __init__(self, timeout=0):
+    self.timeout = timeout
+
+  def _filesystem_url(self, url):
+    parsed = urlparse(url)
+    scheme = parsed.scheme.lower()
+    if scheme in ('root', 'roots', 'xroot', 'xroots'):
+      if parsed.port and parsed.port != 1094 and parsed.hostname:
+        return 'https://%s:%d' % (parsed.hostname, parsed.port)
```
amadio: This changes the scheme from XRootD protocol to HTTP protocol, and may have funny behavior when XRootD is not running on the default port, since it uses the value as condition to return an https:// URL. I think it's important to understand why this code is needed.

amadio: When a client does a command with `xrdfs` using a `root://` URL, then the command should not change the scheme under the hood to another protocol (i.e. https). It should execute the command using the protocol the user asked for.

python/libs/client/tape.py
```diff
@@ -0,0 +1,268 @@
...
+    supported = ('root', 'roots', 'xroot', 'xroots',
+                 'dav', 'davs', 'http', 'https')
+    if scheme not in supported or not parsed.netloc:
+      return url
+
+    endpoint = self._filesystem_url(url)
+    path = parsed.path or '/'
+    if not path.startswith('/'):
+      path = '/' + path
+    return endpoint + path
```
amadio: Why is `root://` a supported protocol in `_http_url`?

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -41,6 +46,27 @@ Filesystem::Filesystem(const std::string &url, std::shared_ptr<HandlerQueue> que
...
+                           const char *description)
+{
+    try
+    {
+        m_queue->Produce(std::move(operation));
+    }
+    catch(...)
+    {
+        const auto ex = DescribeException(
+            std::current_exception(), "while queueing HTTP operation");
```
amadio: We can use `ex.code()` and `ex.what()` for most standard exception types, so we don't need to have this `DescribeException` wrapper to describe them.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,367 @@
...
+    {
+    }
+
+    uint16_t statusCode() const { return m_statusCode; }
+    uint32_t code() const { return m_code; }
+
+private:
+    uint16_t m_statusCode;
+    uint32_t m_code;
+};
```
amadio: `std::runtime_error` already carries a `ex.code()` and `ex.what()` which can be reused without deriving from it. I think reusing `std::runtime_error` is sufficient, so we don't need these classes.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -138,91 +153,101 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+}
+
+XrdCl::XRootDStatus Filesystem::Prepare(
+    const std::vector<std::string> &fileList,
+    XrdCl::PrepareFlags::Flags      flags,
+    uint8_t                         priority,
+    XrdCl::ResponseHandler         *handler,
+    time_t                          timeout)
+{
+    (void)priority;
```
amadio, line 166: This should be sent to the endpoint according to the tape REST API specs, otherwise FTS people will have a hard time debugging the lack of priority when using `xrdfs` later on.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -138,91 +153,101 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+        case XrdCl::QueryCode::Checksum:
         {
-            preferred = XrdClHttp::GetTypeFromString(iter->second);
-            if (preferred == XrdClHttp::ChecksumType::kUnknown)
+            const auto url = GetCurrentURL(arg.ToString());
+            m_logger->Debug(kLogXrdClHttp,
+                "XrdClHttp::Filesystem::Query checksum path %s", url.c_str());
+
+            XrdClHttp::ChecksumType preferred =
+                XrdClHttp::ChecksumType::kCRC32C;
```
amadio: Is this the right default? I believe `adler32` is what is used most with EOS. This may need to be discovered with the equivalent of `xrdfs query config checksum` as the default can vary server by server.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -138,91 +153,101 @@ XrdCl::XRootDStatus Filesystem::MkDir(const std::string        &path,
...
+            XrdCl::URL urlObj;
+            urlObj.FromString(url);
+            const auto iter = urlObj.GetParams().find("cks.type");
```
amadio, line 211: Another thing that is easy to check and ensure FTS doesn't have strange errors is to always query the server for supported checksums, and reject values in `cks.type` which are not supported (unless this is already clear from the existing error messages).

src/XrdClHttp/XrdClHttpOps.cc
```diff
@@ -311,47 +317,42 @@ CurlOperation::Redirect(std::string &target)
 
     auto location = m_headers.GetLocation();
     if (location.empty()) {
-        m_logger->Warning(kLogXrdClHttp, "After request to %s, server returned a redirect with no new location", m_url.c_str());
+        m_logger->Warning(kLogXrdClHttp, "After request to %s, server returned a redirect with no new location", m_request_url.c_str());
```
amadio, line 326: I'd like to understand why this change is needed.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -107,6 +113,109 @@ bool XrdClHttp::HTTPStatusIsError(unsigned status) {
...
+    return std::make_tuple(cert, key);
+}
+
+void XrdClHttp::SetClientX509(CURL *curl, const std::string &cert, const std::string &key,
+                              XrdCl::Log *log) {
+    if (cert.empty()) return;
+    if (log) log->Debug(kLogXrdClHttp, "Using client X.509 credential found at %s", cert.c_str());
+    curl_easy_setopt(curl, CURLOPT_SSLCERT, cert.c_str());
+    if (key.empty()) {
+        if (log) log->Error(kLogXrdClHttp, "X.509 client credential specified but not the client key");
```
amadio: This will log every proxy certificate with an error, because proxy certificates normally do not have matching keys.

amadio: Or maybe this doesn't work at all with a proxy cert, we should check with each mode cert+key and just a proxy without key.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -107,6 +113,109 @@ bool XrdClHttp::HTTPStatusIsError(unsigned status) {
...
+        env->GetString("HttpClientKeyFile", key);
+    }
+    return std::make_tuple(cert, key);
+}
+
+void XrdClHttp::SetClientX509(CURL *curl, const std::string &cert, const std::string &key,
+                              XrdCl::Log *log) {
+    if (cert.empty()) return;
+    if (log) log->Debug(kLogXrdClHttp, "Using client X.509 credential found at %s", cert.c_str());
+    curl_easy_setopt(curl, CURLOPT_SSLCERT, cert.c_str());
```
amadio: We should check the return value of `curl_easy_setopt`, as it may fail.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -107,6 +113,109 @@ bool XrdClHttp::HTTPStatusIsError(unsigned status) {
...
+    catch(const std::exception &ex)
+    {
+        return {EIO, "exception " + context + ": " + ex.what()};
+    }
+    catch(...)
+    {
+        return {EIO, "unknown exception " + context};
+    }
+    return {EIO, "unknown exception " + context};
+}
```
amadio: This can be removed when we move to just reuse the standard exceptions as is instead of wrapping. They already have the code and description.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -319,7 +428,9 @@ bool HeaderParser::Parse(const std::string &header_line)
 
     std::string header_name = header_line.substr(0, found);
     if (!Canonicalize(header_name)) {
-        return false;
+        // libcurl has already accepted and framed this response.  An invalid
+        // field name cannot be interpreted as response metadata, so skip it.
+        return true;
```
amadio: It's unclear why this change is needed, even with the comment. It would be good to understand this better.

src/XrdClHttp/XrdClHttpUtil.hh
```diff
@@ -55,12 +58,49 @@ bool HTTPStatusIsError(unsigned status);
...
+// whenever one is available.
+bool ShouldUseBearerToken(const std::string &protocols,
+                          bool hasX509Credential,
+                          bool hasBearerToken);
+
+struct ExceptionInfo
+{
+    uint32_t code;
+    std::string message;
+};
```
amadio: This is probably unnecessary, the standard exception types already have `.code()` and `.what()` with exactly this information available.

src/XrdHttpTapeApi/README.md
```diff
@@ -0,0 +1,15 @@
...
+The initial implementation provides a deterministic in-process Tape REST API
+surface used by the XrdClHttp Tape client integration tests. It is built and
+installed as a regular server plugin so the handler can evolve into a real
+server-side Tape REST API implementation.
+
+Example configuration:
+
+```conf
+http.exthandler xrdhttptapeapi libXrdHttpTapeApi.so
+```
```
amadio, line 25: Maybe this could be expanded a bit more?

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,310 @@
...
+}
+
+int TapeApiHandler::StageStatus(XrdHttpExtReq &req)
+{
+  if(req.verb != "GET") return SendError(req, 405, "expected GET");
+
+  return SendJson(req, 200,
+    R"({"id":"request-1","createdAt":1,"startedAt":2,)"
+    R"("completedAt":3,"files":[{"path":"/store/file",)"
+    R"("state":"COMPLETED","startedAt":2,"finishedAt":3}]})");
```
amadio: If this is just hard-coding a fixed answer, we need to change to a simple system that works. What I had discussed with Andy was to have a directory where we just dump the JSON into text files by id, and manage things by managing the files themselves.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,310 @@
...
+int TapeApiHandler::Release(XrdHttpExtReq &req, const std::string &body)
+{
+  return ProcessPathsRequest(req, body, "{}");
+}
+
+int TapeApiHandler::ArchiveInfo(XrdHttpExtReq &req,
+                                const std::string &body)
+{
+  return ProcessPathsRequest(req, body,
+    R"([{"path":"/store/file","locality":"DISK_AND_TAPE"}])");
```
amadio: Same here, we should not hard-code responses, but have a minimally functional tape REST API endpoint just using a directory and JSON files to provide the functionality.

tests/XrdClHttp/CMakeLists.txt
```diff
@@ -86,3 +105,13 @@ set_tests_properties(XrdClHttp::test
     ATTACHED_FILES_ON_FAIL
       "${CURL_TEST_LOGS}"
 )
+
+set_tests_properties(XrdClHttp::tape
+  PROPERTIES
+    FIXTURES_REQUIRED
+      XrdClHttp
+    ENVIRONMENT
+      "BATS_LIB_PATH=${PROJECT_SOURCE_DIR}/vendor/bats/lib/bats;BINARY_DIR=${CMAKE_BINARY_DIR}"
```
amadio: `BATS_LIB_PATH` should be set centrally in `common.bash` or such, so that all tests get it by default.

src/XrdClHttp/XrdClHttpFilesystem.cc
lobis: After some investigation, the conclusion is that this is a mess.

`priority` is not part of the wclg tape rest api spec so we will leave code as is and tackle this new issue (https://github.com/xrootd/xrootd/issues/2898) in a future PR.

src/XrdClHttp/XrdClHttpOps.cc
lobis: m_url is the immutable URL identifying the overall operation, while m_request_url tracks the URL of the HTTP request currently being executed.
Tape REST operations involve multiple requests. For example, a stage operation first requests:
https://eospilot.cern.ch/.well-known/wlcg-tape-rest-api
The discovery response may then direct the client to:
https://tape-api.cern.ch/api/v1/stage
If the latter returns a redirect without a Location header, logging m_url would incorrectly identify the original storage endpoint. Logging m_request_url identifies the endpoint that actually returned the malformed response.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,384 @@
...
+{
+  return status ? SendJson(req, status.code, body)
+                : SendError(req, status.code, status.message);
+}
+
+bool TapeApiHandler::ReadBody(XrdHttpExtReq &req, std::string &body,
+                              int &errorCode, std::string &error)
+{
+  body.clear();
+  if(req.length <= 0) return true;
```
amadio: This should only return true on `== 0`, and `false` on `< 0`, negative length makes no sense.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,384 @@
...
+  if(req.length > INT_MAX)
```
amadio: `INT_MAX` is too big here (2048MB), we should use a smaller number to protect the server against someone sending giant blobs into the tape REST API in parallel. A limit of a few MB that is configurable to be bigger if needed is better.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,384 @@
...
+
+  char *data = nullptr;
+  const int length = static_cast<int>(req.length);
+  if(req.BuffgetData(length, &data, true) != length || data == nullptr)
+  {
+    errorCode = 400;
+    error = "missing or invalid request body";
+    return false;
+  }
+  body.assign(data, static_cast<std::size_t>(length));
```
amadio: Why cast to `int` then cast again to `std::size_t`? Casting is expensive. This is also reading into a `char* data`, assigning to `body`, and likely leaking the data, as I don't see `data` being freed here after the copy. Better would be to read directly into the `body` string, without a copy.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
```diff
@@ -0,0 +1,384 @@
...
+  if(req.verb != "GET") return SendError(req, 405, "expected GET");
+
+  const auto hostHeader =
+    XrdOucTUtils::caseInsensitiveFind(req.headers, "host");
+  const std::string host =
+    hostHeader == req.headers.end() ? "" : hostHeader->second;
+  if(host.empty()) return SendError(req, 400, "missing Host header");
+
+  Json body;
+  body["sitename"] = "xrootd-ci";
```
amadio: This should use the result of `all.sitename`, not a hard-coded value. The XrdMacaroons plugin makes use of the value of `all.sitename`, maybe take a look there and do similar here.

src/XrdOuc/XrdOucJsonUtils.hh
```diff
@@ -0,0 +1,71 @@
...
+                                  const char *key,
+                                  Integer &value,
+                                  bool &present)
+  {
+    static_assert(std::is_integral_v<Integer>);
+    static_assert(!std::is_same_v<Integer, bool>);
+
+    present = false;
+    const auto item = json.find(key);
+    if(item == json.end()) return true;
```
amadio: This behavior is probaly going to lead to bugs, as you return `true` as if `value` initialized successfully, but in fact it may still be uninitialized.

src/XrdOuc/XrdOucJsonUtils.hh
```diff
@@ -0,0 +1,71 @@
...
+    std::uint64_t parsed = 0;
+    if(item->is_number_unsigned())
+    {
+      parsed = item->get<std::uint64_t>();
+    }
+    else if(item->is_number_integer())
+    {
+      const std::int64_t signedValue = item->get<std::int64_t>();
+      if(signedValue < 0) return false;
+      parsed = static_cast<std::uint64_t>(signedValue);
```
amadio: This also looks a bit fishy, I don't see the reasoning behind using a function called `GetOptionalUnsignedInteger` to read from a signed one (which seems to be the case when you read from/to a `time_t`).

src/XrdOuc/XrdOucJsonUtils.hh
```diff
@@ -0,0 +1,71 @@
...
+#include <limits>
+#include <type_traits>
+
+namespace XrdOucJsonUtils
+{
+  template<typename Integer>
+  bool GetOptionalUnsignedInteger(const nlohmann::json &json,
+                                  const char *key,
+                                  Integer &value,
+                                  bool &present)
```
amadio: Actually, I have only seen this function being used in a single place to load a timestamp, and in its own unit tests. Maybe this function can go, as `time_t` is actually a signed type in most places.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+  }
+  return body;
+}
+
+bool CopyOptionalTimestamp(const Json &input, const char *key,
+                           Json &output, bool &present)
+{
+  std::time_t value = 0;
+  if(!XrdOucJsonUtils::GetOptionalUnsignedInteger(
+       input, key, value, present))
```
amadio: Since `time_t` is often a signed type, and this is the only usage of `GetOptionalUnsignedInteger` I could find outside of unit tests, I suggest to drop that function and just read the timestamp here with more appropriate checks to validate a timestamp rather than an unsigned integer. If you add a static assert in your template that the `Integer` template argument has to be unsigned, this will likely stop compiling...

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+  {
+    output["state"] = input["state"];
+  }
+  if(input.contains("error") && !input["error"].is_string())
+  {
+    return ErrorStatus(XrdCl::errInvalidResponse,
+      "stage request response contains a non-string error field");
+  }
+  if(input.contains("error")
+     && !input["error"].get_ref<const std::string &>().empty())
```
amadio: This is a strange way to check that a string is non-empty. Isn't there a simpler way? The `input.contains("error")` check is also repeated unnecessarily, as the same check is done a few lines above.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+}
+
+XrdCl::XRootDStatus NormalizeStageFileStatusJson(
+  const Json &input, Json &output)
+{
+  output = Json::object();
+  if(input.contains("path") && input["path"].is_string())
+  {
+    output["path"] = NormalizeTapePath(input["path"].get<std::string>());
+  }
```
amadio: Why here you don't check for non-emptyness, but you do for `"error"` below? I suppose a `path` should always be non-empty. I guess this validation is already done when parsing the request, but if that is the case, then `input.contains("path")` is also unnecessary here? Also, you're passing a temporary to something that takes a `const std::string&`, so there's an unnecessary copy happening here.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+}
+
+std::string CanonicalLocality(const std::string &locality)
+{
+  for(const auto name : kTapeLocalities)
+  {
+    if(strcasecmp(locality.c_str(), name.data()) == 0)
+      return std::string(name);
+  }
+  return {};
```
amadio: Shouldn't this be `return "UNKNOWN";`?

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+      else result["error"] = error;
+    }
+    else
+    {
+      result["error"] = "error field is not a string";
+    }
+    return result;
+  }
+
+  if(!item->contains("locality") || !(*item)["locality"].is_string())
```
amadio: This does not ensure the locality is non-empty. Together with the `CanonicalLocality` above not flagging it, a real missing locality by being empty will not raise an error here.

src/XrdClHttp/XrdClHttpTape.cc
```diff
@@ -0,0 +1,1110 @@
...
+    result["error"] = "locality attribute missing";
+    return result;
+  }
+
+  const std::string locality = (*item)["locality"].get<std::string>();
+  const std::string canonical = CanonicalLocality(locality);
+  if(canonical.empty())
+  {
+    result["error"] = "file locality reported as \"" + locality + "\"";
+  }
```
amadio: Ok, an empty locality will show as `file locality reported as ""`, not `locality attribute missing`, but at least it will error out. Your choice how to handle it, but probably reporting an empty string as missing makes more sense.

src/XrdClHttp/XrdClHttpTape.hh
```diff
@@ -0,0 +1,88 @@
...
+#include <curl/curl.h>
+
+namespace XrdClHttp
+{
+  enum class TapeHttpMethod
+  {
+    Get,
+    Post,
+    Delete
+  };
```
amadio: Isn't there such an enum available somewhere shared already? If not, HTTP verbs should be created in a place usable by all of XrdClHttp instead of here.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -101,12 +106,130 @@ pid_t getthreadid() {
...
+bool DecodeBase64Checksum(std::string_view value, size_t decodedSize,
+                          std::array<unsigned char,
+                                     XrdClHttp::g_max_checksum_length> &output)
+{
+    const size_t encodedSize = 4 * ((decodedSize + 2) / 3);
+    return value.size() == encodedSize
+           && XrdClHttp::HeaderParser::Base64Decode(value, output);
+}
+
 }
```
amadio: We already have such helper functions elsewhere, let's reuse them here instead of creating code duplication.

src/XrdClHttp/XrdClHttpUtil.cc
```diff
@@ -101,12 +106,130 @@ pid_t getthreadid() {
...
 
 bool XrdClHttp::HTTPStatusIsError(unsigned status) {
      return (status < 100) || (status >= 400);
 }
 
+bool XrdClHttp::ClientX509Enabled(XrdCl::Env *env) {
+    if (!env) return false;
+    int disable_x509 = 0;
+    return env->GetInt("HttpDisableX509", disable_x509) && !disable_x509;
+}
```
amadio: The mechanism everyone is used to is `XrdSecPROTOCOL`. This and some changes below are also not really related to tape REST API, so it's probably better to work on this separately. @dynamic-entropy is also working in the same area and this will certainly lead to tough merge conflicts.

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,333 @@
...
+CurlTapeOp::GetVerb() const
+{
+    switch(m_request.method)
+    {
+        case TapeHttpMethod::Get: return HttpVerb::GET;
+        case TapeHttpMethod::Post: return HttpVerb::POST;
+        case TapeHttpMethod::Delete: return HttpVerb::DELETE;
+    }
+    return HttpVerb::GET;
+}
```
amadio, line 239: Yes, as I suspected, there's already `CurlOperation::HttpVerb`, so `TapeHttpMethod::*` are completely redundant. Please use `CurlOperation::HttpVerb` instead. This wrapper to convert then becomes unnecessary as well.

amadio (changes requested):
Now we are getting closer. After this last round, it will probably be good to merge.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
lobis: Fixed. A zero length is accepted as an empty body; negative lengths now return HTTP 400 and abort body processing.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
lobis: Fixed. The default maximum body size is now 4 MiB, with optional handler configuration using standard size suffixes. Requests above the configured limit return HTTP 413. Configuration is capped at INT_MAX only because BuffgetData() accepts an int length.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
lobis: BuffgetData() returns a borrowed pointer into protocol-owned storage, so it cannot read directly into a std::string and the pointer must not be freed by the handler. I documented that ownership, removed the redundant cast, validate the returned byte count, and copy exactly those bytes into body.

src/XrdHttpTapeApi/XrdHttpTapeApi.cc
lobis: Fixed. The handler now reads XRDSITE, which XRootD exports from all.sitename through XrdOucSiteName, and uses it in the discovery response. Initialization fails if it is unavailable, and the test configuration now sets all.sitename.

src/XrdOuc/XrdOucJsonUtils.hh
lobis: I first made the optional helper initialize the output when the key is absent, then removed the helper entirely when timestamp parsing was moved into the Tape response code.

src/XrdOuc/XrdOucJsonUtils.hh
lobis: Fixed. I first restricted the helper to unsigned output types, then removed it entirely; time_t is no longer read through an unsigned-integer helper.

src/XrdOuc/XrdOucJsonUtils.hh
lobis: Done. The single-use helper, its build integration, and its isolated unit test have been removed. Timestamp parsing now lives alongside Tape response normalization.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. The local timestamp parser accepts signed or unsigned JSON integers, rejects negative values and values above time_t's maximum, and only then converts to time_t. Tests cover negative top-level and per-file timestamps.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. The error member is now looked up once with an iterator, type-checked once, and accessed through a const string reference. It is copied into the result only when non-empty.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. The path member is looked up once, must be a non-empty string, and is passed by const reference without the temporary copy. I kept this response-side validation because the remote response is untrusted, and added an empty-path malformed-response test.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. Unrecognized locality values now canonicalize to UNKNOWN instead of producing a per-file error.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. An empty locality is explicitly detected and can no longer silently normalize to UNKNOWN.

src/XrdClHttp/XrdClHttpTape.cc
lobis: Done. Missing, non-string, and empty locality values now consistently report locality attribute missing. Valid locality strings are accessed through a const reference.

src/XrdClHttp/XrdClHttpTape.hh
lobis: Done. I moved the existing CurlOperation::HttpVerb definition into the shared XrdClHttpVerb.hh header and retained CurlOperation::HttpVerb as an alias. Tape now uses that same type and TapeHttpMethod is gone. The extraction avoids a circular include because XrdClHttpOps.hh already includes XrdClHttpTape.hh.

src/XrdClHttp/XrdClHttpUtil.cc
lobis: Done. I removed the duplicated TrimWhitespace, DecodeHex32, and DecodeBase64Checksum helpers and reused trim_view, XrdCksData::Set/GetChecksumLength, and HeaderParser::Base64Decode.

src/XrdClHttp/XrdClHttpOpTape.cc
lobis: Done. TapeHttpMethod has been removed, TapeHttpRequest uses the shared HttpVerb directly, and GetVerb() now simply returns m_request.method; the conversion switch is gone.

src/XrdClHttp/XrdClHttpUtil.cc
lobis: Addressed. Tape authentication now uses XrdSecPROTOCOL to choose between the configured GSI/X.509 and bearer-token (ztn) credentials. I also removed ClientX509Enabled, ClientX509CertKeyFromEnv, and SetClientX509 and restored the existing XrdClHttp certificate setup paths, so the unrelated general authentication refactor and overlap with @dynamic-entropy's work are no longer part of this PR.

src/XrdClHttp/XrdClHttpOps.hh
```diff
@@ -56,18 +59,7 @@ class ResponseInfo;
...
-        DELETE,
-        HEAD,
-        GET,
-        MKCOL,
-        OPTIONS,
-        PROPFIND,
-        PUT,
-        Count
-    };
+    using HttpVerb = XrdClHttp::HttpVerb;
```
amadio, line 62: If the tape operations derive from `CurlOperation`, shouldn't the enum class as it was before just fine to use in both cases? What am I missing?

src/XrdClHttp/XrdClHttpOpTape.cc
```diff
@@ -0,0 +1,327 @@
...
+            !std::get<0>(m_worker->ClientX509CertKeyFile()).empty();
+    }
+    const std::string token = GetBearerToken();
+    const char *configuredProtocols = std::getenv("XrdSecPROTOCOL");
+    if(ShouldUseBearerToken(
+         configuredProtocols ? configuredProtocols : "",
+         hasX509Credential, !token.empty()))
+    {
+        m_headers_list.emplace_back("Authorization", "Bearer " + token);
+    }
```
amadio: It would really be better if authentication was handled in a common way within XrdClHttp, not in different ways by different parts...

src/XrdClHttp/XrdClHttpOps.hh
lobis: Only CurlTapeOp derives from CurlOperation. The separate TapeOperation state machine creates TapeHttpRequest objects in XrdClHttpTape.hh, which is included by XrdClHttpOps.hh before CurlOperation is defined. Using CurlOperation::HttpVerb there would require referring to a nested type on an incomplete class or introducing a circular include. I therefore moved the existing enum to a small shared header and kept CurlOperation::HttpVerb as an alias.

src/XrdClHttp/XrdClHttpOpTape.cc
lobis: I will improve this in an upcoming commit so that eventual conflicts are easier to fix.

amadio:
@dynamic-entropy Can you please have a look at the authentication changes here and let me know what you think? I'd like to avoid tons of conflicts from this area later. The changes here look reasonable, but are not in the direction of having one place to handle authentication as in your pull request.

@lobis @dynamic-entropy I think for 6.2 we will have to keep things within XrdClHttp, and deduplicate and improve the authentication parts of XrdClHttp only in 6.3. What do you think?

src/XrdClHttp/XrdClHttpOps.hh
amadio: Ok, thanks for the explanation. For 6.3 we should have a discussion on how to improve the general architecture of XrdClHttp and reduce code duplication that exists now. For 6.2, I think it's getting too close, after a final set of changes from you, I will probably merge to master.

src/XrdClHttp/XrdClHttpOpTape.cc
lobis: https://github.com/xrootd/xrootd/pull/2836/commits/151801e2fae9b518c87496218ec6bc1c87e0c141 should make things better

lobis:
> @dynamic-entropy Can you please have a look at the authentication changes here and let me know what you think? I'd like to avoid tons of conflicts from this area later. The changes here look reasonable, but are not in the direction of having one place to handle authentication as in your pull request.
> 
> @lobis @dynamic-entropy I think for 6.2 we will have to keep things within XrdClHttp, and deduplicate and improve the authentication parts of XrdClHttp only in 6.3. What do you think?

I already pushed https://github.com/xrootd/xrootd/pull/2836/commits/151801e2fae9b518c87496218ec6bc1c87e0c141 which should make things better but I can remove it and do the deduplication and improvement in 6.3 with more time. your call.

amadio (approved):
Ok, I think this is now in good shape to go into XRootD 6.2! Thanks a lot for the work and the patience!

abh3:
Instead of adding a "fake" tape backend to test the API, I strongly suggest that requests be vectored to the prepare implementation via the bridge and kXR_prepare request. The native prepare implementation is very basic and won't exactly be able to test very much. However, it is completely replaceable by a plugin. There already exists an interface plugin, libXrdOfsPrepGPI.so which provides a generic driver for scripts to handle prepare requests (see the comments . These can be any language you want and can implement the basics of the expected behavior of the tape REST API for testing purposes. It would likely be the way most sites go to add tape staging support. In fact, one site already has done this. It would also provide much needed code coverage. XrdOfsPrepGPI.cc implements the plugin. The documentation can be found at https://xrootd.web.cern.ch/doc/dev61/ofs_config.htm#_Toc228291074 with details in the subsequent sections.

lobis:
> Instead of adding a "fake" tape backend to test the API, I strongly suggest that requests be vectored to the prepare implementation via the bridge and kXR_prepare request. The native prepare implementation is very basic and won't exactly be able to test very much. However, it is completely replaceable by a plugin. There already exists an interface plugin, libXrdOfsPrepGPI.so which provides a generic driver for scripts to handle prepare requests (see the comments . These can be any language you want and can implement the basics of the expected behavior of the tape REST API for testing purposes. It would likely be the way most sites go to add tape staging support. In fact, one site already has done this. It would also provide much needed code coverage. XrdOfsPrepGPI.cc implements the plugin. The documentation can be found at https://xrootd.web.cern.ch/doc/dev61/ofs_config.htm#_Toc228291074 with details in the subsequent sections.

I created https://github.com/xrootd/xrootd/issues/2922 in case we need to follow-up, hopefully it summarizes what you are proposing correctly.

PR #2837, opened 2026-06-16
[Python] Add lint checks

lobis:
## Summary
- add a dedicated lint workflow that currently runs the Python Ruff pre-commit hook
- add a root pre-commit configuration with YAML/TOML checks and scoped whitespace cleanup for config/workflow files
- configure Ruff for the Python bindings, package entrypoints, and tests

## Notes
- The lint workflow is structured as a general "Lint" workflow with a Python job, so C++ linting can be added later as a separate job.
- This PR intentionally does not modify existing Python source files. Historical whitespace and formatting cleanup can be handled separately.
- Ruff is initially scoped to python/libs, python/src, and python/tests. The legacy python/examples tree still contains Python 2 style snippets, so it is excluded from Ruff until that cleanup is handled separately.
- The initial Ruff rule set focuses on syntax and serious name errors, with historical Python binding issues ignored so the hook can be enabled incrementally.

## Validation
- pre-commit run --all-files --show-diff-on-failure
- pre-commit run ruff --all-files --show-diff-on-failure
- git diff --check origin/master...HEAD

## Summary by CodeRabbit

* **Chores**
  * Added automated lint checks for code changes and manual runs.
  * Added pre-commit checks for formatting, file consistency, and Python linting.
  * Configured standardized Python linting and formatting rules, including legacy-code exceptions.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,49 @@
+name: Lint
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - '**'
+    paths:
+      - .github/workflows/lint.yml
+      - .pre-commit-config.yaml
+      - pyproject.toml
+      - 'python/**'
+  pull_request:
+    paths:
+      - .github/workflows/lint.yml
+      - .pre-commit-config.yaml
+      - pyproject.toml
+      - 'python/**'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  python:
+    name: Python
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v6
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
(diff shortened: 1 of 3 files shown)


amadio:
For security, I want to stick to verified creators for our actions workflows. Also, until we decide to actually enforce a coding convention (currently there is no real consensus on this), I think lint checks are not really necessary.

@abh3 Please add your opinion here as well. I'm tagging you as reviewer.

abh3:
Correct. Historically, we have never had a coding convention and simply asked people to write understandable code. Generally, understandability and style are weakly connected. The most important part is to leave quality comments on what is happening in short code segments that are difficult to decipher unless you have intimate knowledge of the task being accomplished.  In such circumstances, even AI can lead you astray. Anyway, for C++ I am not in favor of a rigid coding style standard as it may make the code look pretty (to someone's ideal) but doesn't necessarily improve code quality.

For python, I am neutral since Python already enforces a "coding style" since syntax and structure are tied together. That said,  choosing a coding style is like choosing an editor; it almost borders on religion.

abh3 (commented):
I left a general comment here and mostly agree with @amadio. However, the ruff author is generally considered a verified creator as far as I can tell. So, I'm not sure what the hesitation is on that.

PR #2838, opened 2026-06-16
[XrdCl] Extract curl helper utilities

lobis:
## Summary
- add a reusable `XrdCl::CurlUtil` helper for curl status mapping, CA setup, X.509 env import, client credential lookup, and client credential application
- update `XrdClHttp` to consume the shared helper instead of carrying local copies of this setup logic
- keep the helper source under `src/XrdCl` so a later Tape REST implementation in XrdCl can reuse it without depending on the XrdClHttp plugin

### DIFF
--- a/src/XrdCl/XrdClCurlUtil.cc
+++ b/src/XrdCl/XrdClCurlUtil.cc
@@ -1,0 +1,340 @@
+/******************************************************************************/
+/*                                                                            */
+/*                    X r d C l C u r l U t i l . c c                         */
+/*                                                                            */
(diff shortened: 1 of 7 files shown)


amadio:
This has been obsoleted by work on other pull requests. The main client library, i.e. `libXrdCl.so`, should not link to `curl` directly, code supporting HTTP features must go into `libXrdClHttp.so` instead. The tape REST api and HTTP TPC work both are going in that direction. Therefore, I think this can be closed.

PR #2839, opened 2026-06-16
[CI] Add Uproot validation workflow

lobis:
## Summary
- add a manually triggered Uproot validation workflow for the XRootD Python wheel
- build a local XRootD wheel and XRootD server, install Uproot/fsspec-xrootd test dependencies, and run Uproot xrootd tests

### DIFF
--- a/.github/scripts/run-uproot-ci-local.sh
+++ b/.github/scripts/run-uproot-ci-local.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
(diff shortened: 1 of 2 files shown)


amadio:
We need to cover our testing within XRootD, without relying on external packages. I want to avoid creating entanglement with reverse dependencies (i.e. add dependencies in CI to packages that themselves depend on XRootD). If a regression happens in ROOT or uproot, we can create a new test in XRootD that ensures it will not happen again.

PR #2651, opened 2025-11-21
XrdCeph changes from STFC/RAL

Jo-stfc:
rebased PR of https://github.com/xrootd/xrootd/pull/2628

PR #2840, opened 2026-06-22
RAL XrdCeph changes

Jo-stfc:
- feature to add an Erasure Coding correction factor for space reporting
- XrdCeph Streamed Checksum Implementation
- Write-only buffers 
- Clear checksum on write completion to avoid storing a partial checksum if the checksum was requested during transfer

### DIFF
--- a/src/XrdCeph/CMakeLists.txt
+++ b/src/XrdCeph/CMakeLists.txt
@@ -13,6 +13,8 @@ else()
   endif()
 endif()
 
+include_directories( ${RADOS_INCLUDE_DIR} )
+
 set(BUILD_CEPH TRUE CACHE BOOL INTERNAL FORCE)
 
 add_library(XrdCephPosix SHARED
@@ -29,7 +31,7 @@ target_include_directories(XrdCephPosix
   PUBLIC ${RADOS_INCLUDE_DIR} $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
 
 set_target_properties(XrdCephPosix
-  PROPERTIES VERSION 0.0.1 SOVERSION 0)
+  PROPERTIES VERSION 1.0.0 SOVERSION 0)
 
 set(LIB_XRD_CEPH XrdCeph-${PLUGIN_VERSION})
 
@@ -50,7 +52,7 @@ add_library(${LIB_XRD_CEPH} MODULE
 )
 
 target_link_libraries(${LIB_XRD_CEPH}
-  PRIVATE XrdCephPosix XrdUtils XrdServer ${RADOS_LIBS})
+  PRIVATE ${XROOTD_LIBRARIES} XrdCephPosix XrdUtils XrdServer ${RADOS_LIBS})
 
 set(LIB_XRD_CEPH_XATTR XrdCephXattr-${PLUGIN_VERSION})
 
--- a/src/XrdCeph/XrdCephOss.cc
+++ b/src/XrdCeph/XrdCephOss.cc
@@ -422,8 +422,8 @@ int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
                g_logStreamedAdler32 = true;
 	       g_storeStreamedAdler32 = true;
            }
-        } 
-        if (!strcmp(var, "ceph.streamed-cks-logfile") ) {
+         } 
+	if (!strcmp(var, "ceph.streamed-cks-logfile") ) {
          var = Config.GetWord();
 	 if (var) { 
            g_cksLogFileName = strdup(var);
@@ -435,6 +435,7 @@ int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
  	   return 1;
          }
        }
+      }
      }
      // Now check if any errors occurred during file i/o
 
--- a/src/XrdCeph/XrdCephOss.cc
+++ b/src/XrdCeph/XrdCephOss.cc
@@ -115,7 +115,7 @@ ssize_t getNumericAttr(const char* const path, const char* attrName, const int m
   if (attrLen <= 0) {
     retval = -EINVAL;
   } else {
-    attrValue[attrLen] = '\0';
+    attrValue[attrLen] = (char)'\0';
     char *endPointer = (char *)NULL;
     retval = strtoll(attrValue, &endPointer, 10);
   }
--- a/src/XrdCeph/XrdCephOss.cc
+++ b/src/XrdCeph/XrdCephOss.cc
@@ -128,6 +128,9 @@ ssize_t getNumericAttr(const char* const path, const char* attrName, const int m
 
 }
 
+char *g_cksLogFileName;
+extern FILE *g_cksLogFile;
+
 extern "C"
 {
   XrdOss*
@@ -164,6 +167,9 @@ XrdCephOss::~XrdCephOss() {
 // declared and used in XrdCephPosix.cc
 extern unsigned int g_maxCephPoolIdx;
 extern unsigned int g_cephAioWaitThresh;
+extern bool g_calcStreamedAdler32;
+extern bool g_storeStreamedAdler32;
+extern bool g_logStreamedAdler32;
 extern double g_ECcorrectionFactor; // correction factor to apply to used space when EC pools are used, to get a better estimate of actual used space
 
 int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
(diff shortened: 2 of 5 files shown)


src/XrdCeph/CMakeLists.txt
```diff
@@ -29,7 +31,7 @@ target_include_directories(XrdCephPosix
   PUBLIC ${RADOS_INCLUDE_DIR} $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>)
 
 set_target_properties(XrdCephPosix
-  PROPERTIES VERSION 0.0.1 SOVERSION 0)
+  PROPERTIES VERSION 1.0.0 SOVERSION 0)
```
amadio, line 34: This cannot go into a patch release. Only major releases are allowed to bump SOVERSION of libraries.

src/XrdCeph/CMakeLists.txt
```diff
@@ -50,7 +52,7 @@ add_library(${LIB_XRD_CEPH} MODULE
 )
 
 target_link_libraries(${LIB_XRD_CEPH}
-  PRIVATE XrdCephPosix XrdUtils XrdServer ${RADOS_LIBS})
+  PRIVATE ${XROOTD_LIBRARIES} XrdCephPosix XrdUtils XrdServer ${RADOS_LIBS})
```
amadio, line 55: Why is this needed? The commit message should explain it. Given that `XrdServer` and `XrdUtils` are already both linked here, I wonder what you need in addition by linking `${XROOTD_LIBRARIES}`. Note that within the XRootD project itself, this variable will be actually blank, so I assume you are trying to build the plugin externally on top of a pre-built XRootD, right?

src/XrdCeph/XrdCephOss.cc
```diff
@@ -115,7 +115,7 @@ ssize_t getNumericAttr(const char* const path, const char* attrName, const int m
   if (attrLen <= 0) {
     retval = -EINVAL;
   } else {
-    attrValue[attrLen] = '\0';
+    attrValue[attrLen] = (char)'\0';
```
amadio, line 118: It's probably better to use `NUL` here to avoid a C-style cast, what do you think?

src/XrdCeph/XrdCephOss.cc
```diff
@@ -363,7 +364,25 @@ int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
...
+               Eroute.Emsg("Config", "Invalid value for ceph.ECcorrectionFactor in config file; enter a value between 0 and 1", configfn, var);
+               return 1;
+             }
+         } else {
+           Eroute.Emsg("Config", "Missing value for ceph.ECcorrectionFactor in config file. Setting default 8/11", configfn);
+	   g_ECcorrectionFactor = 0.727272; // default 8/11 EC correction factor
+           return 1;
+           }
+         }
+       }
```
amadio, line 385: It looks like you are mixing tabs and spaces here. The formatting looks off. In any case, new features are not allowed in patch releases. All new features should go into XRootD 6.x instead.

src/XrdCeph/XrdCephOss.cc
```diff
@@ -349,7 +349,7 @@ int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
            if (!Config.GetRest(parms, sizeof(parms)) || parms[0]) {
              Eroute.Emsg("Config", "readvalgname parameters will be ignored");
            }
-          m_configBufferIOmode = var; // allowed values would be aio, io
+          m_configBufferIOmode = var; // allowed values would be aio, io, write-only-io
```
amadio, line 352: Same as with the other change. This is a new feature, so it should go into a feature release, not into `v5.9.x`.

src/XrdCeph/XrdCephOss.cc
```diff
@@ -383,8 +389,53 @@ int XrdCephOss::Configure(const char *configfn, XrdSysError &Eroute) {
...
+           }
+        } 
+        if (!strcmp(var, "ceph.streamed-cks-logfile") ) {
+         var = Config.GetWord();
+	 if (var) { 
+           g_cksLogFileName = strdup(var);
+         } else {
+           const char *defLogFileName = "/tmp/checksums.log"; // To-DO: Move defLogFileName so it can also be used as fallback 
+	                                                      //  when attempt to open specified log file below fails
+           Eroute.Emsg("Config", "Missing value for ceph.streamed-cks-logfile in config file, setting to default = ", defLogFileName);
```
amadio, line 433: Hard-coded paths into `/tmp` are not a good solution here. This was also pointed out in a previous pull request [here](https://github.com/xrootd/xrootd/pull/2651#discussion_r2555029163).

src/XrdCeph/XrdCephPosix.cc
```diff
@@ -107,12 +149,20 @@ unsigned int g_nextCephFd = 0;
...
+/// mutex protecting checksum log file
+XrdSysMutex g_logAdler32;
 
 //JW Counter for number of times a given cluster is resolved.
 std::map<unsigned int, unsigned long long> g_idxCntr;
 
+//IJJ: Actions for Adler32 checksum
+bool g_calcStreamedAdler32;
+bool g_logStreamedAdler32;
+bool g_storeStreamedAdler32;
```
amadio, line 161: Please have a look at XRootD 6.1. Andy introduced streaming checksum support there in a way that should work for you, so this might no longer be necessary. And as with other changes, this is a new feature, which should go into a feature release.

src/XrdCeph/XrdCephPosix.cc
```diff
@@ -836,6 +836,10 @@ int ceph_posix_close(int fd) {
       lastAsyncAge = 1.0 * (now.tv_sec - fr->lastAsyncSubmission.tv_sec) 
               + 0.000001 * (now.tv_usec - fr->lastAsyncSubmission.tv_usec);
     }
+    //clear checksum if this is a file has been modified
+    if (fr->bytesWritten > 0){
+      ceph_posix_fremovexattr(fd,"XrdCks.adler32");
+    }
```
amadio, line 842: The commit message for this commit is too long. Please keep the first line to about ~50 characters, then add a blank line and explanation below.

amadio (changes requested):
There are several changes proposed here which are not suitable for a patch release. We only add new features into the current master branch. Patch releases should only contain bug fixes. I strongly recommend looking into 6.1, which was released two days ago, and try the new support for streaming checksums added there. If it doesn't satisfy your needs, please file a bug or feature request so that we can address any problems in 6.2, which should come out in September.

amadio:
I asked GitHub Copilot to review in addition. Below is what it says as well. Note that the memory leaks on `strdup()` were also raised as a concern in the [previous pull request](https://github.com/xrootd/xrootd/pull/2651#discussion_r2555043573). The global variables without adequate synchronization are also a real problem. I removed some other points which were not so relevant.
 
## Critical Concerns

### Global State Management & Thread Safety Issues

Multiple global variables introduced without adequate synchronization:

```cpp
char *g_cksLogFileName;
FILE *g_cksLogFile;
bool g_calcStreamedAdler32;
bool g_logStreamedAdler32;
bool g_storeStreamedAdler32;
double g_ECcorrectionFactor;
```

Only one mutex (`g_logAdler32`) protects the log file, but not all flag reads/writes.

**Risk**: Race conditions in multi-threaded scenarios when these globals are read/modified concurrently.

## High-Risk Issues

### Memory Leaks & Resource Management

`strdup()` calls without guaranteed cleanup:

```cpp
g_cksLogFileName = strdup(var);  // Line ~381
char *timeString = ts_rfc3339(); // Returns strdup'd memory
hexbytes2ascii();                 // Also returns strdup'd memory
```

No explicit deallocation visible in destructors or cleanup routines.

**Risk**: Long-running processes accumulate memory; file handles may not be properly released.

### Configuration File Handling Issues

In `XrdCephOss::Configure()`: Missing `return` statement after logging error:

```cpp
g_ECcorrectionFactor = 0.727272; // default 8/11 EC correction factor
return 1;  // This returns an error, but g_ECcorrectionFactor is set
```

Log file opening silently fails without propagating error up properly.

**Risk**: Configuration errors could be masked; defaults applied unexpectedly.

### Version Number Bump Without ABI/API Clarity

`XrdCephPosix` version changed from `0.0.1` to `1.0.0`.

CMakeLists adds `${XROOTD_LIBRARIES}` to link command.

**Risk**: This is a significant version bump suggesting API stability guarantees that may not be met; dependency chain changes could cause incompatibilities.

## Medium-Risk Issues

### Checksum Implementation Concerns

Checksum is **cleared on write completion**:

```cpp
if (fr->bytesWritten > 0) {
  ceph_posix_fremovexattr(fd, "XrdCks.adler32");
}
```

**Risk**: Could lose valid checksums if reads occur after partial writes; behavior may surprise users expecting persisted checksums.

### New Buffer Mode ("write-only-io") Incomplete Integration

Mode added to code paths but error messages still reference old options.

Configuration parsing doesn't validate this mode explicitly.

**Risk**: Silent fallback to wrong buffer strategy; could degrade performance or cause data correctness issues.

### File Descriptor Initialization

`fr.writingData = false` set in `ceph_posix_open()` then immediately reassigned.

`cksCalcadler32` allocated but only if `g_calcStreamedAdler32` is true (uninitialized initially).

**Risk**: Potential null pointer dereference if checksum update called before initialization.

### 9. Unchecked Type Casting

Multiple implicit casts: `(char)'\\0'`, casting `XrdCksData*` from `void*`.

**Risk**: Data corruption if caller passes unexpected value types.

## Lower-Risk Issues

### 10. Code Duplication

`maxOffsetWritten` calculation duplicated in `ceph_posix_pwrite()` (lines appear twice).

Multiple overlapping string comparisons for mode validation.

**Risk**: Maintenance burden; future inconsistencies.

Jo-stfc:
Thanks Guilherme, I'll test the 6.1 version of streamed checksums and resubmit without that change into the 6.0 branch If everything else looks fine.

src/XrdCeph/CMakeLists.txt
Jo-stfc: This was to use the XrdCks libraries. I'll test the upstream version to make sure if it works

amadio:
Closing due to lack of activity post-review. I strongly suggest to break the changes into smaller pieces which will be easier to review and integrate, and to ensure that the changes are well tested and safe for production use in the next iteration (feel free to reopen this pull request when you update the branch, but remember to reopen before pushing, otherwise GitHub will not allow it).

src/XrdCeph/XrdCephOss.cc
Jo-stfc: fixed in following commit

src/XrdCeph/XrdCephOss.cc
Jo-stfc: not sure what you mean - this is used to manually terminate strings as ceph doesn't return null terminated strings by default. is there a different syntax in C++ for this?

PR #2841, opened 2026-06-22
Disallow relative path mkdirs and rmdir with davs

### LINKED ISSUE #2792: xrdfs mkdir/rmdir command target needs leading "/" only with xroot protocol (S3 storage backend)
### Operating System

AlmaLinux 9.7

### XRootD Version

XRootD 6.0.1

### Description

When running the `xrdfs` `mkdir` and `rmdir` commands, the target directory name requires a leading `/` when using the xroot protocol but not the dav protocol.  The following commands are deemed acceptable:
```
xrdfs dav://ADDRESS:PORT/ mkdir foo
xrdfs dav://ADDRESS:PORT/ rmdir foo
xrdfs xroot://ADDRESS:PORT/ mkdir /foo
xrdfs xroot://ADDRESS:PORT/ rmdir /foo
```
However, these are not:
```
> xrdfs xroot://45.113.232.90:1094/ mkdir foo
[ERROR] Server responded with an error: [3010] Creating relative path 'foo' is disallowed.

xrdfs xroot://45.113.232.90:1094/ rmdir foo
[ERROR] Server responded with an error: [3010] Removing relative path 'foo' is disallowed.
```
I understand the reason for the error in the xroot case, but would expect the same error to be reported when using the dav protocol.

1. Configure XRootD to use an S3 storage backend. See for example the configuration attached to 2754.
2. Start the XRootD server.
3. Run the commands given in the bug description.

### Additional Information

_No response_

dynamic-entropy:
Fixes https://github.com/xrootd/xrootd/issues/2792 

As part of refactoring BuildPath, I also took the the opportunity to add better error logging for other cases. 
Instead of returning InvalidArg, we now log the "operation" and "error".

### DIFF
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -63,7 +63,15 @@ XRootDStatus BuildPath( std::string &newPath, Env *env,
   Log *log = DefaultEnv::GetLog();
 
   if( path.empty() )
-    return XRootDStatus( stError, errInvalidArgs );
+  {
+    std::string msg;
+    if( op )
+      msg = std::string( op ) + " requires a path.";
+    else
+      msg = "path is required.";
+    log->Error( AppMsg, "%s", msg.c_str() );
+    return XRootDStatus( stError, errInvalidArgs, 0, msg );
+  }
 
   int noCwd = 0;
   env->GetInt( "NoCWD", noCwd );
@@ -108,7 +116,15 @@ XRootDStatus BuildPath( std::string &newPath, Env *env,
     if( *it == ".." )
     {
       if( it == pathComponents.begin() )
-        return XRootDStatus( stError, errInvalidArgs );
+      {
+        std::string msg;
+        if( op )
+          msg = std::string( op ) + " path '" + path + "' escapes above root.";
+        else
+          msg = "path '" + path + "' escapes above root.";
+        log->Error( AppMsg, "%s", msg.c_str() );
+        return XRootDStatus( stError, errInvalidArgs, 0, msg );
+      }
       std::list<std::string>::iterator it1 = it;
       --it1;
       it = pathComponents.erase( it1 );
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -22,30 +22,33 @@
 // or submit itself to any jurisdiction.
 //------------------------------------------------------------------------------
 
-#include "XrdCl/XrdClFileSystem.hh"
-#include "XrdCl/XrdClFileSystemUtils.hh"
-#include "XrdCl/XrdClFSExecutor.hh"
-#include "XrdCl/XrdClURL.hh"
-#include "XrdCl/XrdClLog.hh"
-#include "XrdCl/XrdClDefaultEnv.hh"
+#include "XProtocol/XProtocol.hh"
 #include "XrdCl/XrdClConstants.hh"
-#include "XrdCl/XrdClUtils.hh"
 #include "XrdCl/XrdClCopyProcess.hh"
+#include "XrdCl/XrdClDefaultEnv.hh"
+#include "XrdCl/XrdClFSExecutor.hh"
 #include "XrdCl/XrdClFile.hh"
+#include "XrdCl/XrdClFileSystem.hh"
 #include "XrdCl/XrdClFileSystemOperations.hh"
+#include "XrdCl/XrdClFileSystemUtils.hh"
+#include "XrdCl/XrdClLog.hh"
 #include "XrdCl/XrdClParallelOperation.hh"
+#include "XrdCl/XrdClStatus.hh"
+#include "XrdCl/XrdClURL.hh"
+#include "XrdCl/XrdClUtils.hh"
+#include "XrdCl/XrdClXRootDResponses.hh"
 #include "XrdOuc/XrdOucPrivateUtils.hh"
 #include "XrdSys/XrdSysE2T.hh"
 
-#include <cstdlib>
-#include <cstdio>
-#include <iostream>
-#include <iomanip>
 #include <cmath>
+#include <cstdio>
+#include <cstdlib>
+#include <iomanip>
+#include <iostream>
 
 #ifdef HAVE_READLINE
-#include <readline/readline.h>
 #include <readline/history.h>
+#include <readline/readline.h>
 #endif
 
 using namespace XrdCl;
@@ -54,18 +57,31 @@ using namespace XrdCl;
 // Build a path
 //------------------------------------------------------------------------------
 XRootDStatus BuildPath( std::string &newPath, Env *env,
-                        const std::string &path )
+                        const std::string &path,
+                        const char *op = nullptr )
 {
+  Log *log = DefaultEnv::GetLog();
+
   if( path.empty() )
     return XRootDStatus( stError, errInvalidArgs );
 
   int noCwd = 0;
   env->GetInt( "NoCWD", noCwd );
 
-  if( path[0] == '/' || noCwd )
+  if( path[0] == '/' )
   {
     newPath = path;
     return XRootDStatus();
+  }
+  else if( noCwd )
+  {
+    std::string msg;
+    if( op )
+      msg = std::string( op ) + " relative path '" + path + "' is disallowed.";
+    else
+      msg = "relative path '" + path + "' is disallowed.";
+    log->Error( AppMsg, "%s", msg.c_str() );
+    return XRootDStatus( stError, errInvalidArgs, 0, msg );
   }
 
   std::string cwd = "/";
@@ -560,11 +576,9 @@ XRootDStatus DoMkDir( FileSystem                      *fs,
   }
 
   std::string newPath;
-  if( !BuildPath( newPath, env, path ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid path." );
-    return XRootDStatus( stError, errInvalidArgs );
-  }
+  XRootDStatus pathSt = BuildPath( newPath, env, path, "Creating" );
+  if( !pathSt.IsOK() )
+    return pathSt;
 
   //----------------------------------------------------------------------------
   // Run the query
@@ -601,11 +615,9 @@ XRootDStatus DoRmDir( FileSystem                      *query,
   }
 
   std::string fullPath;
-  if( !BuildPath( fullPath, env, args[1] ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid path." );
-    return XRootDStatus( stError, errInvalidArgs );
-  }
+  XRootDStatus pathSt = BuildPath( fullPath, env, args[1], "Removing" );
+  if( !pathSt.IsOK() )
+    return pathSt;
 
   //----------------------------------------------------------------------------
   // Run the query
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -202,11 +202,9 @@ XRootDStatus DoCache( FileSystem                      *fs,
   }
 
   std::string fullPath;
-  if( !BuildPath( fullPath, env, args[2] ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid cache path." );
-    return XRootDStatus( stError, errInvalidArgs, 0, "Invalid cache path." );
-  }
+  XRootDStatus pathSt = BuildPath( fullPath, env, args[2], "cache" );
+  if( !pathSt.IsOK() )
+    return pathSt;
 
   //----------------------------------------------------------------------------
   // Create the command 
@@ -260,11 +258,9 @@ XRootDStatus DoCD( FileSystem                      *fs,
   env->PutInt( "NoCWD", 0 );
 
   std::string newPath;
-  if( !BuildPath( newPath, env, args[1] ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid path." );
-    return XRootDStatus( stError, errInvalidArgs );
-  }
+  XRootDStatus pathSt = BuildPath( newPath, env, args[1], "cd" );
+  if( !pathSt.IsOK() )
+    return pathSt;
 
   //----------------------------------------------------------------------------
   // Check if the path exist and is not a directory
@@ -432,11 +428,9 @@ XRootDStatus DoLS( FileSystem                      *fs,
     env->GetString( "CWD", newPath );
   else
   {
-    if( !BuildPath( newPath, env, path ).IsOK() )
-    {
-      log->Error( AppMsg, "Invalid arguments. Invalid path." );
-      return XRootDStatus( stError, errInvalidArgs );
-    }
+    XRootDStatus pathSt = BuildPath( newPath, env, path, "ls" );
+    if( !pathSt.IsOK() )
+      return pathSt;
   }
 
   //----------------------------------------------------------------------------
@@ -576,7 +570,7 @@ XRootDStatus DoMkDir( FileSystem                      *fs,
   }
 
   std::string newPath;
-  XRootDStatus pathSt = BuildPath( newPath, env, path, "Creating" );
+  XRootDStatus pathSt = BuildPath( newPath, env, path, "mkdir" );
   if( !pathSt.IsOK() )
     return pathSt;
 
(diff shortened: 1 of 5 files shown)


src/XrdCl/XrdClFS.cc
```diff
@@ -30,6 +30,8 @@
 #include "XrdCl/XrdClDefaultEnv.hh"
 #include "XrdCl/XrdClConstants.hh"
 #include "XrdCl/XrdClUtils.hh"
+#include "XrdCl/XrdClXRootDResponses.hh"
+#include "XProtocol/XProtocol.hh"
```
amadio, line 25: The order of header inclusion is odd, can you move this one so that all client headers are included together? You may also alternatively reorder them alphabetically, but it's not needed.

src/XrdCl/XrdClFS.cc
```diff
@@ -54,14 +56,25 @@ using namespace XrdCl;
...
+                        const std::string &path,
+                        const char *op = nullptr )
 {
   if( path.empty() )
     return XRootDStatus( stError, errInvalidArgs );
 
   int noCwd = 0;
   env->GetInt( "NoCWD", noCwd );
 
+  if( noCwd && path[0] != '/' )
```
amadio: This `if` and the next test something quite close. Maybe it's better to combine them and have an `if (path[0] == '/')` and then have `else if (noCwd) { ... }` for the other side (since you then already know that `path[0] != '/'` in the `else`.

src/XrdCl/XrdClFS.cc
amadio: Since there's no better place, adding this here too: the commit message is split into two lines, but the separate line is relevant to the meaning. I suggest joining the split line into the main commit header, and adding an explanation why the change is needed, with explicit reference to the ticket it is trying to fix with `Issue: #2792` or `Fixes: #2792`.

src/XrdCl/XrdClFS.cc
```diff
@@ -60,7 +60,14 @@ XRootDStatus BuildPath( std::string &newPath, Env *env,
...
   if( path.empty() )
-    return XRootDStatus( stError, errInvalidArgs );
+  {
+    std::string msg;
+    if( op )
+      msg = std::string( op ) + " requires a path.";
+    else
+      msg = "path is required.";
+    return XRootDStatus( stError, errInvalidArgs, 0, msg );
+  }
```
amadio, line 70: This commit also has a strange commit message, please fix. For this one you can probably just remove the split line. You may want to pass in the proper `errno` code to `XRootDStatus` here just in case.

src/XrdCl/XrdClFS.cc
```diff
@@ -814,11 +823,9 @@ XRootDStatus DoChMod( FileSystem                      *fs,
   }
 
   std::string fullPath;
-  if( !BuildPath( fullPath, env, args[1] ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid path." );
-    return XRootDStatus( stError, errInvalidArgs );
-  }
+  XRootDStatus pathSt = BuildPath( fullPath, env, args[1], "Changing mode of" );
```
amadio: Looks like the error message from this may look like
```
Changing mode of requires a path
```
Given that `BuildPath` only appends its own stuff to whatever is passed in as `op` in the last argument. Not that great.

amadio (changes requested):
Please see more detailed comments. Also, I suggest adding tests so that you can check that the error messages don't look bad (see respective comment on that).

amadio:
Below is what GitHub Copilot says, FYI. I just asked it to review the changes.

---

### PR Summary

This PR refactors the `BuildPath` function in XrdClFS.cc to improve error handling and enforce security constraints around relative paths when the `NoCWD` (No Current Working Directory) mode is enabled, particularly for davs (WebDAV over SSL).

**Key Details:**
- **Purpose**: Fixes [issue #2792](https://github.com/xrootd/xrootd/issues/2792)
- **Files Changed**: 1 (src/XrdCl/XrdClFS.cc)
- **Scope**: 85 additions, 93 deletions (3 commits)
- **Status**: Open, mergeable but **blocked** — needs review approval
- **Review Status**: 1 review with **changes requested** (by @amadio)
- **CI**: ✅ All 19 checks passing across multiple platforms and configurations

---

### Core Changes

The PR introduces three main improvements to the `BuildPath` function:

1. **Enhanced Security Check**: Adds validation to reject relative paths when `NoCWD` mode is enabled, returning an appropriate error status instead of proceeding silently.

2. **Better Error Messages**: The `BuildPath` function now accepts an optional `op` parameter to describe the operation being performed, allowing for more contextual error messages (e.g., "Caching requires a path" vs generic "Invalid cache path").

3. **Consistent Error Handling**: Replaces repetitive error handling patterns across 15+ call sites with a cleaner, DRY approach—instead of checking the return status and logging separately at each call site, the operation name is passed to `BuildPath` which generates a descriptive message.

**Example of the refactoring** (shown in DoChMod function):

```diff
  std::string fullPath;
- if( !BuildPath( fullPath, env, args[1] ).IsOK() )
- {
-   log->Error( AppMsg, "Invalid path." );
-   return XRootDStatus( stError, errInvalidArgs );
- }
+ XRootDStatus pathSt = BuildPath( fullPath, env, args[1], "Changing mode of" );
+ if( !pathSt.IsOK() )
+   return pathSt;
```

---

### Other Changes

The PR adds two new includes to support the enhanced status messages:
```cpp
#include "XrdCl/XrdClXRootDResponses.hh"
```

This pattern is repeated across all affected functions: `DoCache`, `DoCD`, `DoLS`, `DoMkDir`, `DoRmDir`, `DoMv`, `DoRm`, `DoTruncate`, `DoChMod`, `DoLocate`, `DoStat`, `DoStatVFS`, `DoQuery`, `DoCat`, `DoTail`, and `DoXAttr`.

---

### Merge Readiness and Risk Assessment

⚠️ **Not ready to merge** — The review has identified several issues:

**Critical Issues Requiring Resolution:**

1. **Include Header Ordering** (Reviewer Comment 1)
   - The new headers `XrdCl/XrdClXRootDResponses.hh` and `XProtocol/XProtocol.hh` are inserted in an odd location, breaking the convention of grouping all client headers together.
   - **Action**: Move the includes to maintain consistent header organization or reorder alphabetically.

2. **Error Message Formatting** (Reviewer Comment 5)
   - Error messages may produce awkward results. Example: "Changing mode of requires a path" reads unnaturally.
   - **Root Cause**: The operation name is directly concatenated to the error message suffix without proper grammar handling.
   - **Action**: Consider restructuring the message template to produce grammatically correct output.

3. **Commit Message Clarity** (Reviewer Comment 3)
   - The commit messages are split across multiple lines inconsistently, and lack explicit reference to the issue being fixed.
   - **Action**: Consolidate commit messages and add `Fixes: #2792` reference.

4. **Missing Errno Context** (Reviewer Comment 4)
   - When returning error status for invalid paths, the code might benefit from passing the appropriate `errno` code to `XRootDStatus` for better error tracking.
   - **Action**: Consider whether passing `errno` would improve debuggability.

5. **Control Flow Optimization** (Reviewer Comment 2)
   - The new `noCwd` validation check and the existing path validation check test similar conditions; combining them could improve clarity.
   - **Action**: Consider restructuring the logic: `if (path[0] == '/') { ... } else if (noCwd) { ... }` to avoid redundant condition testing.

**Risk Assessment**: **MEDIUM**
- The logic is sound and addresses a real security concern (disallowing relative paths in davs mode).
- All CI checks pass, indicating no compilation or ABI compatibility issues.
- The refactoring is mechanical and low-risk, but the new validation logic needs to be tested.
- Main risk is UX/error messaging polish, not correctness.

---

### Possible Improvements

Beyond the reviewer's identified items:

1. **Add Unit Tests**: The reviewer suggests adding tests to verify error messages look reasonable. Consider adding test cases that exercise the new error paths, especially for the `noCwd` relative path rejection.

2. **Consider Early Validation**: The `noCwd` check is now buried in `BuildPath`. Consider whether this validation should happen at a higher level (in the shell command handlers) for earlier detection and clearer error context.

3. **Consistent Verb Tense**: Review all operation names passed to `BuildPath` for consistency. Some use gerunds ("Caching", "Copying") while others use full phrases ("Changing mode of", "Changing directory to"). Standardizing these would improve readability.

---

amadio:
BTW, if there are no conflicts, this is a good candidate to go both into 5.x and 6.x.

src/XrdCl/XrdClFS.cc
dynamic-entropy: I will reorder all

src/XrdCl/XrdClFS.cc
dynamic-entropy: Makes sense, will do

src/XrdCl/XrdClFS.cc
```diff
@@ -54,14 +56,25 @@ using namespace XrdCl;
...
   env->GetInt( "NoCWD", noCwd );
 
+  if( noCwd && path[0] != '/' )
+  {
+    std::string msg;
+    if( op )
+      msg = std::string( op ) + " relative path '" + path + "' is disallowed.";
+    else
+      msg = "relative path '" + path + "' is disallowed.";
+    return XRootDStatus( stError, errErrorResponse, kXR_NotAuthorized, msg );
```
dynamic-entropy: I wanted to make it similar to xrootd, but looking at it again, I should not use `errErrorResponse` since this is validation in the client :)

dynamic-entropy (commented):
Changed the "sentence like" framing to simple operations name to not have awkward completions. thanks for catching.

src/XrdCl/XrdClFS.cc
dynamic-entropy: Done

amadio: I still think some number other than zero should be in the third argument here, which is used to carry an `errno` code. In this case, it should probably be `EINVAL`. In the other cases, it should also be non-zero and appropriate to the `stError` type being returned.

src/XrdCl/XrdClFS.cc
```diff
@@ -186,11 +213,9 @@ XRootDStatus DoCache( FileSystem                      *fs,
...
 
   std::string fullPath;
-  if( !BuildPath( fullPath, env, args[2] ).IsOK() )
-  {
-    log->Error( AppMsg, "Invalid cache path." );
-    return XRootDStatus( stError, errInvalidArgs, 0, "Invalid cache path." );
-  }
+  XRootDStatus pathSt = BuildPath( fullPath, env, args[2], "cache" );
+  if( !pathSt.IsOK() )
+    return pathSt;
```
amadio, line 215:
```suggestion
  if (auto status = BuildPath(fullPath, env, args[2], "cache"); !status.IsOK())
      return status;
```

Same for the others. I wonder if there's a good way to reduce code duplication here as well, but this can be tackled later.

amadio: Is the removal of the log message intentional? I wonder if that should be left in, as
```cpp
log->Error(AppMsg, status.ToString());
```

dynamic-entropy: >   if (auto status = BuildPath(fullPath, env, args[2], "cache"); !status.IsOK())

I think this is neither good for readability nor for consistency across the file.



> ```c++
> log->Error(AppMsg, status.ToString());
> ```

I removed it since the log is essentially a duplicate of the error message sent to the client. No additional information. But I will reconsider adding central logging

amadio: > I think this is neither good for readability nor for consistency across the file.

This is the more idiomatic C++17 way of doing things, it only looks different because pre-existing stuff uses C++11 at best. The main practical difference is that the way I suggested limits the scope of the `status` variable to only within the `if` statement itself.

dynamic-entropy: Yeah, though scope limiting is a benefit of the form you suggest, here its just for aesthetics. Not worth mixing styles in the same file. I can change it, but I'm really not in favour of mixing things like this. 
They should be done during larger refactors and consistently across the module.

src/XrdCl/XrdClFS.cc
amadio: Yes, `errErrorResponse` is for when you get an error back from the server, and should not be used when the client wants to create an error of its own. Also the error in the third argument, where you placed an `kXR_NotAuthorized` is an `errno`, not an XRootD error code. See, for example, this exerpt:

https://github.com/xrootd/xrootd/blob/4ec63a24e099f108f1b7af8cd0dd4e41ace6a8c5/src/XrdCl/XrdClFileStateHandler.cc#L1231-L1233

src/XrdCl/XrdClFS.cc
amadio: This is not resolved, the diff above still shows `0` instead of `EINVAL` for the `errno` argument.

amadio:
This is still missing tests covering the changes being made. Can you please add? You may wait until the bash framework gets merged if you want.

amadio (changes requested):
Please re-read the review, there are items which are still not addressed by the current changes.

src/XrdCl/XrdClFS.cc
dynamic-entropy: That is because it would only add redundant "Invalid Arguments" text to the printed message.

Nothing consumes the errNo other than message printing in `st.ToStr()`, neither does this add any information. The structure of the errNo is such that any extra information can be added mostly in addition to `kXRErrorResponse`.

dynamic-entropy:
Answered the comment about using `EINVAL` instead of 0 and added tests. 

I do not wish to wait for the BATS framework.

src/XrdCl/XrdClFS.cc
amadio: This is what I see when I run a mkdir without the path before and after your changes:

**before**
```
gentoo xrootd $ env XRD_LOGLEVEL=Error xrdfs root://gentoo.cern.ch mkdir 
[2026-08-05 09:18:12.784525 +0200][Error  ][App               ] Too few arguments.
[ERROR] Invalid arguments
```
**after**
```
gentoo xrootd $ env XRD_LOGLEVEL=Error build/bin/xrdfs root://gentoo.cern.ch mkdir 
[2026-08-05 09:17:58.782078 +0200][Error  ][App               ] Too few arguments.
[ERROR] Invalid arguments
```
It would be nice to get the "too few arguments" not just in the log, but also after the `[ERROR] Invalid arguments` that the client sees in `stderr`.

src/XrdCl/XrdClFS.cc
```diff
@@ -54,19 +57,40 @@ using namespace XrdCl;
...
+  else if( noCwd )
+  {
+    std::string msg;
+    if( op )
+      msg = std::string( op ) + " relative path '" + path + "' is disallowed.";
+    else
+      msg = "relative path '" + path + "' is disallowed.";
+    log->Error( AppMsg, "%s", msg.c_str() );
+    return XRootDStatus( stError, errInvalidArgs, 0, msg );
+  }
```
amadio, line 89: Here is the before/after for the message above. The new message is not gramatically well formed. Can we try to keep the format closer to the original, just not needing to go to the server to get the error?
```sh
before $ env XRD_LOGLEVEL=Error xrdfs root://gentoo.cern.ch mkdir foo
[2026-08-05 09:27:41.318292 +0200][Error  ][App               ] Unable create directory foo: [ERROR] Server responded with an error: [3010] Creating relative path 'foo' is disallowed.
[ERROR] Server responded with an error: [3010] Creating relative path 'foo' is disallowed.

after $ env XRD_LOGLEVEL=Error build/bin/xrdfs root://gentoo.cern.ch mkdir foo
[2026-08-05 09:27:55.279923 +0200][Error  ][App               ] mkdir relative path 'foo' is disallowed.
[ERROR] Invalid arguments: mkdir relative path 'foo' is disallowed.
```

Same for the other messages (i.e. for rmdir, etc).

amadio (commented):
I'm not sure the new messages are better than the old ones. I'd try to fix the problem by just avoiding the round trip to the server, but keeping the error messages gramatically well formed.

src/XrdCl/XrdClFS.cc
dynamic-entropy: Now the error message reads 
```
"Creating relative path 'foo' is disallowed."
```

Also added assertions for this and others in tests with the latest commit.

PR #2842, opened 2026-06-22
[XrdHttpTPC] Fix TPC PULL curl's buffer exhaustion

### LINKED ISSUE #2108: XrdHttp-TPC: Issue with TPC multi-stream PULL
While implementing the packet marking for HTTP TPC transfers, I saw that there is an issue with multistream HTTP TPC PULL transfers.

The request:

```
export DST=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY
export SRC=http://xrootd-ccaffy-dev01.cern.ch:1096/tmp/BIGFILE_5GB

curl -v --capath /etc/grid-security/certificates -L -X COPY  -H "Source: $DST" -H "SciTag: 144" -H 'X-Number-Of-Streams: 3' "$SRC"
```

On the client-side:

```
* About to connect() to xrootd-ccaffy-dev01.cern.ch port 1096 (#0)
*   Trying 137.138.123.14...
* Connected to xrootd-ccaffy-dev01.cern.ch (137.138.123.14) port 1096 (#0)
> COPY /tmp/BIGFILE_5GB HTTP/1.1
> User-Agent: curl/7.29.0
> Host: xrootd-ccaffy-dev01.cern.ch:1096
> Accept: */*
> Source: http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY
> SciTag: 144
> X-Number-Of-Streams: 3
> 
< HTTP/1.1 201 Created
< Connection: Keep-Alive
< Server: XrootD/v5.5.5-361-g1c80707
< Content-Type: text/plain
< Transfer-Encoding: chunked
< 
Perf Marker
Timestamp: 1697721653
Stripe Index: 0
Stripe Bytes Transferred: 50331648
Total Stripe Count: 1
RemoteConnections: tcp:137.138.123.14:2001
End
* Connection #0 to host xrootd-ccaffy-dev01.cern.ch left intact
failure: Internal logic error led to early abort; current offset is 50331648 while full size is 5368709120
```

The logs on the pulling machine:

```
231019 15:00:27 31857 anon.0:34@xrootd-ccaffy-dev01 http_Req: Appended header fields to opaque info: 'scitag.flow=144'
231019 15:00:27 31857 TPC_PullRequest: event=PULL_START, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous); Starting a pull request
[New Thread 0x7fffdffff700 (LWP 31914)]
[Thread 0x7fffdffff700 (LWP 31914) exited]
231019 15:00:27 31857 TPC_PullRequest: event=SIZE_SUCCESS, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3; Successfully determined remote size for pull request: 5368709120
231019 15:00:28 31857 TPC_PullRequest: event=SIZE_SUCCESS, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3; Successfully determined remote size for pull request: 5368709120
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Starting chunked response
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Sending resp: 201 header len:138
231019 15:00:28 31857 http_Protocol: Sending 138 bytes
231019 15:00:28 31857 TPC_PullRequest: event=RESPONSE_START, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3; Initial transfer response sent to the TPC client
231019 15:00:28 31857 TPC_PullRequest: event=PERF_MARKER, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3, bytes_transferred=50331648
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Sending encoded chunk of size 154
231019 15:00:28 31857 http_Protocol: Sending 4 bytes
231019 15:00:28 31857 http_Protocol: Sending 154 bytes
231019 15:00:28 31857 http_Protocol: Sending 2 bytes
[New Thread 0x7fffdffff700 (LWP 31915)]
[Thread 0x7fffdffff700 (LWP 31915) exited]
231019 15:00:28 31857 TPC_CanStartTransfer: Unable to start transfers as no buffers are available.  Available buffers: 0, Active curl handles: 0, Transfers in progress: 0
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Beginning dump of stream buffers.
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Stream offset: 0
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Buffer 0: Offset=16777216, Size=16777216, Capacity=16777216
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Buffer 1: Offset=33554432, Size=16777216, Capacity=16777216
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Buffer 2: Offset=0, Size=16777216, Capacity=16777216
231019 15:00:28 31857 TPC_Stream::DumpBuffers: Finish dump of stream buffers.
231019 15:00:28 31857 TPC_StartTransfers: Unable to start transfers.
231019 15:00:28 31857 TPC_PullRequest: event=MULTISTREAM_IDLE, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3, bytes_transferred=50331648; No handles are able to run.  Streams=3, concurrency=3
231019 15:00:28 31857 TPC_PullRequest: event=MULTISTREAM_FAIL, local=/tmp/BIGFILE_5GB, remote=http://xrootd-ccaffy-dev01.cern.ch:2001/tmp/BIGFILE_5GB_COPY, user=(anonymous), streams=3, bytes_transferred=50331648, tpc_status=0; failure: Internal logic error led to early abort; current offset is 50331648 while full size is 5368709120
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Sending encoded chunk of size 106
231019 15:00:28 31857 http_Protocol: Sending 4 bytes
231019 15:00:28 31857 http_Protocol: Sending 106 bytes
231019 15:00:28 31857 http_Protocol: Sending 2 bytes
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Sending encoded chunk of size 0
231019 15:00:28 31857 http_Protocol: Sending 3 bytes
231019 15:00:28 31857 http_Protocol: Sending 2 bytes
231019 15:00:28 31857 http_Req:  XrdHttpReq request ended.
231019 15:00:28 31857 anon.0:34@xrootd-ccaffy-dev01 http_Protocol: Process is exiting rc:1
231019 15:00:28 31857 Xrd_Poll: Poller 0 enabled anon.0:34@xrootd-ccaffy-dev01
231019 15:00:28 31859 anon.0:34@xrootd-ccaffy-dev01 Xrd_Poll: Link 34 terminating: hangup
231019 15:00:28 31877 Xrd_Sched: running anon.0:34@xrootd-ccaffy-dev01 inq=0
231019 15:00:28 31877 http_Protocol:  Cleanup
231019 15:00:28 31877 http_Protocol:  Reset
231019 15:00:28 31877 http_Req:  XrdHttpReq request ended.
231019 15:00:28 31877 anon.0:34@xrootd-ccaffy-dev01 Xrd_Poll: Poller 0 removing FD 34
231019 15:00:28 31877 anon.0:34@xrootd-ccaffy-dev01 Xrd_Poll: FD 34 detached from poller 0; num=0
231019 15:00:59 31862 TLS_Refresh: CRL refresh will happen in 60 seconds.
231019 15:00:59 31869 TLS_Refresh: CRL refresh created new x509 store.
231019 15:00:59 31869 TLS_Refresh: CRL refresh will happen in 60 seconds.
```

ccaffy:
For multi-stream HTTP TPC pull, incoming ranges may arrive out of order. TPC::Stream buffers those chunks until they become contiguous with m_offset, then writes them to the local file and marks the buffer available again.

The fix makes Stream::Write() attempt to write a buffer whenever it accepts new bytes into it, not only when the accept was partial.
So an exactly-filled contiguous buffer is flushed immediately and returned to the available-buffer pool.

Fixes #2108

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcStream.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcStream.cc
@@ -10,12 +10,6 @@ using namespace TPC;
 
 Stream::~Stream()
 {
-    for (std::vector<Entry*>::iterator buffer_iter = m_buffers.begin();
-        buffer_iter != m_buffers.end();
-        buffer_iter++) {
-        delete *buffer_iter;
-        *buffer_iter = NULL;
-    }
     m_fh->close();
 }
 
@@ -29,12 +23,10 @@ Stream::Finalize()
     }
     m_open_for_write = false;
 
-    for (std::vector<Entry*>::iterator buffer_iter = m_buffers.begin();
-        buffer_iter != m_buffers.end();
-        buffer_iter++) {
-        delete *buffer_iter;
-        *buffer_iter = NULL;
-    }
+    // If there are outstanding buffers to reorder, finalization failed; the
+    // check has to happen before the buffers are released.
+    bool all_buffers_returned = m_avail_count == m_buffers.size();
+    m_buffers.clear();
 
     if (m_fh->close() == SFS_ERROR) {
         std::stringstream ss;
@@ -45,8 +37,7 @@ Stream::Finalize()
         return false;
     }
 
-    // If there are outstanding buffers to reorder, finalization failed
-    return m_avail_count == m_buffers.size();
+    return all_buffers_returned;
 }
 
 
@@ -72,7 +63,7 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
         return SFS_ERROR;
     }
     size_t bytes_accepted = 0;
-    int retval = size;
+    ssize_t retval = size;
     if (offset < m_offset) {
         if (!m_error_buf.size()) {m_error_buf = "Logic error: writing to a prior offset";}
         return SFS_ERROR;
@@ -94,50 +85,24 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
             return retval;
         }
     }
-    // Even if we already accepted the current data, always
-    // iterate through available buffers and try to write as
-    // much out to disk as possible.
-    Entry *avail_entry;
-    bool buffer_was_written;
-    size_t avail_count = 0;
+    // Even if we already accepted the current data, always iterate through the
+    // buffers and try to write as much out to disk as possible.
+    //
+    // Accepting data can complete a buffer, and flushing a buffer advances
+    // m_offset, which can in turn let another buffer accept more data or become
+    // writable.  Alternate between the two until neither makes progress.  When
+    // size == 0 we force a flush even if things are not MB-aligned.
+    ssize_t buffers_flushed;
     do {
-        avail_count = 0;
-        avail_entry = NULL;
-        buffer_was_written = false;
-        for (std::vector<Entry*>::iterator entry_iter = m_buffers.begin();
-             entry_iter != m_buffers.end();
-             entry_iter++) {
-            // Always try to dump from memory; when size == 0, then we are
-            // going to force a flush even if things are not MB-aligned.
-            int retval2 = (*entry_iter)->Write(*this, size == 0);
-            if (retval2 == SFS_ERROR) {
-                if (!m_error_buf.size()) {m_error_buf = "Unknown filesystem write failure.";}
-                return retval2;
-            }
-            buffer_was_written |= retval2 > 0;
-            if ((*entry_iter)->Available()) { // Empty buffer
-                if (!avail_entry) {avail_entry = *entry_iter;}
-                avail_count ++;
-            }
-            else if (bytes_accepted != size && size) {
-                size_t new_accept = (*entry_iter)->Accept(offset + bytes_accepted, buf + bytes_accepted, size - bytes_accepted);
-                    // Partial accept; buffer should be writable which means we should free it up
-                    // for next iteration
-                if (new_accept && new_accept != size - bytes_accepted) {
-                    int retval3 = (*entry_iter)->Write(*this, false);
-                    if (retval3 == SFS_ERROR) {
-                        if (!m_error_buf.size()) {m_error_buf = "Unknown filesystem write failure.";}
-                        return SFS_ERROR;
-                    }
-                    buffer_was_written = true;
-                }
-                bytes_accepted += new_accept;
-            }
-        }
-    } while ((avail_count != m_buffers.size()) && buffer_was_written);
-    m_avail_count = avail_count;
+        bytes_accepted += AcceptIntoBuffers(offset + bytes_accepted,
+                                            buf + bytes_accepted,
+                                            size - bytes_accepted);
+        buffers_flushed = FlushBuffers(size == 0);
+        if (buffers_flushed == SFS_ERROR) {return SFS_ERROR;}
+    } while ((buffers_flushed > 0) && (bytes_accepted != size));
 
-    if (bytes_accepted != size && size) {  // No place for this data in allocated buffers
+    if (bytes_accepted != size && size) {  // No place for this data in the buffers currently in use
+        Entry *avail_entry = FirstAvailableBuffer();
         if (!avail_entry) {  // No available buffers to allocate; logic error, should not happen.
             DumpBuffers();
             m_error_buf = "No empty buffers available to place unordered data.";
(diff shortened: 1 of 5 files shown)


tests/XrdHttpTpc/XrdHttpTpcTests.cc
```diff
@@ -2,8 +2,11 @@
 
 #include "XrdHttpTpc/XrdHttpTpcUtils.hh"
 #include "XrdHttpTpc/XrdHttpTpcTPC.hh"
+
 #include <exception>
 #include <gtest/gtest.h>
+
+#include <map>
```
amadio, line 9: Why is this include needed?

ccaffy: Because we use std::map in prepareOpenURLTest ;) And it worked because XrdHttpTpcUtils.hh includes XrdHttpExtHandler.hh which itself includes map. Now it is just cleaner ;)

src/XrdHttpTpc/XrdHttpTpcStream.cc
```diff
@@ -121,15 +121,17 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
...
+                    // with m_offset.  Flush it now so a completed transfer
+                    // returns its reorder buffer before the scheduler starts
+                    // looking for more available buffers.
                     int retval3 = (*entry_iter)->Write(*this, false);
                     if (retval3 == SFS_ERROR) {
                         if (!m_error_buf.size()) {m_error_buf = "Unknown filesystem write failure.";}
                         return SFS_ERROR;
                     }
-                    buffer_was_written = true;
+                    buffer_was_written |= retval3 > 0;
```
amadio: This is a bitwise operation, it looks a bit suspicious the way it is done above. For example, what if `retval3` is 0? This whole function in fact is quite hard to understand. Why do we need so many `retvaln` variables (e.g. `retval`, `retval2`, `retval3`, etc).

`Write(...)` is doing both IO and adjusting buffer bookkeeping (make buffer available again). Doing this while still in the middle of `Accept` semantics mixes concerns and makes the function hard to reason about and test. It's probably a good idea to do a small refactoring to make it clearer and less bug prone.

src/XrdHttpTpc/XrdHttpTpcStream.cc
```diff
@@ -121,15 +121,17 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
             }
             else if (bytes_accepted != size && size) {
                 size_t new_accept = (*entry_iter)->Accept(offset + bytes_accepted, buf + bytes_accepted, size - bytes_accepted);
-                    // Partial accept; buffer should be writable which means we should free it up
-                    // for next iteration
-                if (new_accept && new_accept != size - bytes_accepted) {
+                if (new_accept) {
```
amadio: Why is `new_accept != size - bytes_accepted` removed here? The `Accept` path now calls `(*entry_iter)->Write(...)` immediately when `new_accept` is non-zero (previously it did it only on partial accept).

src/XrdHttpTpc/XrdHttpTpcStream.cc
```diff
@@ -121,15 +121,17 @@ Stream::Write(off_t offset, const char *buf, size_t size, bool force)
...
+                    // looking for more available buffers.
                     int retval3 = (*entry_iter)->Write(*this, false);
                     if (retval3 == SFS_ERROR) {
                         if (!m_error_buf.size()) {m_error_buf = "Unknown filesystem write failure.";}
                         return SFS_ERROR;
                     }
-                    buffer_was_written = true;
+                    buffer_was_written |= retval3 > 0;
                 }
                 bytes_accepted += new_accept;
```
amadio: Shouldn't this have some relation to `retval3` as well? I'm new to this part of the code, but it looks to me like if `new_accept` is assigned, say, 1024 bytes, then `retval3 = (*entry_iter)->Write(*this, false);` writes less than that, say 512 bytes, the count above might be wrong about what was really accepted.

amadio (changes requested):
These changes look risky, I think it's a good idea to invest a bit of time to refactor the `Stream::Write` function to be safer. More concretely, the function is trying to juggle two responsibilities while iterating over the buffers and that is very error prone. I suggest refactoring into a two phase operation, one which accepts more data, and the other that flushes them out:

- *Accept phase*: iterate entries and call acceptIntoEntry for each. For each accept that returns >0, increment bytes_accepted and if the entry became ready-to-flush, append the entry pointer to a local vector flush_list.
- *Flush phase*: iterate flush_list and call flushEntry on each. Handle retval and adjust m_avail_count consistently in flushEntry. If flushEntry modifies container state, it's okay — we've already collected pointers (or better: indices or weak references) in flush_list. If Write may delete/free the entry object, store identifiers that remain valid (for example, a stable index or unique id) rather than raw pointers.

For reference, I asked GitHub Copilot to suggest how to do the refactor, this is the structure it proposes (after adding some extra helpers):

```cpp
// Pseudo-refactor sketch — adapt types/names to your codebase.
int Stream::Write(off_t offset, const char *buf, size_t size, bool force)
{
    std::vector<Entry*> to_flush;

    // --- Accept pass: only mutate accept state, collect completed buffers ---
    while (/* iterate entries as before */) {
        size_t accepted = (*entry_iter)->Accept(offset + bytes_accepted, buf + bytes_accepted, size - bytes_accepted);
        if (accepted == 0) continue;
        bytes_accepted += accepted;

        // If this entry is now full/contiguous with m_offset, mark for flush.
        if ((*entry_iter)->IsContiguousWith(m_offset) || (*entry_iter)->IsWritable()) {
            to_flush.push_back(*entry_iter);
        }

        // Note: do NOT call Write() here.
    }

    // If we created a fresh avail_entry earlier, add it to to_flush if ready:
    if (avail_entry && avail_entry->IsContiguousWith(m_offset)) {
        to_flush.push_back(avail_entry);
    }

    // --- Flush pass: perform writes outside the iteration context ---
    for (Entry* e : to_flush) {
        int rv = e->Write(*this, false);
        if (rv == SFS_ERROR) {
            if (!m_error_buf.size()) m_error_buf = "Unknown filesystem write failure.";
            return SFS_ERROR;
        }
        // Update bookkeeping consistently in one place:
        if (rv > 0) {
            // buffer was completed/written and returned to available pool
            m_avail_count++;
            buffer_was_written = true;
        }
    }

    // ... rest of function
}
```

If you want, I can pass on to you the whole feedback from Copilot, but with the refactor as shown above, I think it will be already much easier to reason about the changes.

src/XrdHttpTpc/XrdHttpTpcStream.cc
ccaffy: All gone after the refactoring you suggested

src/XrdHttpTpc/XrdHttpTpcStream.cc
ccaffy: Gone after the refactoring suggested

amadio:
@ccaffy Maybe this is also a good candidate to retarget to v5.9.x, and please replace `Co-authored-by` with `Assisted-by` in the commit message to credit Claude. Cheers,

ccaffy:
OK done, I also separated the two commits between the bugfix and the tests. They both seems to be OK in that branch so I can squash it back if you want.

amadio:
Did a first pass review, looks much better than before. Very nice. The tests added are quite complex, but while we don't have enough examples migrated to GMock+BATS, let's add these and refactor them later.

PR #2844, opened 2026-06-23
Use config in pluginconfdir to configure client plugins

### LINKED ISSUE #2787: S3 settings in client.plugins.d/s3.conf ignored
### Operating System

AlmaLinux 9.7

### XRootD Version

6.0.1

### Description

As far as I can tell, details about the S3 object store to be used by the XrdClS3 client (`XrdClS3Endpoint` for example) should be specified in the `client.plugins.d/s3.conf` file. However, if this is done the S3 client completely ignores the settings. The only way to pass these settings into XRootD is to define them in the Systemd unit file using `Environment` (or placing them in one of the files referenced in `EnvironmentFile` entries) and passing them to the binary to be run with `PassEnvironment` in the Systemd unit file.

The files under `client.plugins.d/` are read by `Utils::ProcessConfig()`, which parses all "KEY=VALUE" pairs in the file into a `std::map` object. That map object is passed to `LoadFactory()` which passes it to the plugin's XrdClGetPlugIn() function. In the case of the S3 client plugin, this function calls the Factory::Factory() constructor. However, the `std::map` object containing the configuration entries from the s3.conf file is not passed to the constructor. The constructor ultimately calls the `InitS3Config()` method which consults the shell environment for `XRD_...` equivalent variables. However, there's no way for the `s3.conf` contents to be accessible at this point. This explains why the S3 parameters in the `s3.conf` file have no effect on the fS3 client plugin.

If the `client.plugins.d/s3.conf` file is the intended place for S3 parameter definitions, any parameters set in that file need to be accessible to `InitS3Config()`.  Either the map needs to make its way to `InitS3Config()` so it can be consulted before the default environment, or the default environment needs to be populated with the map contents before `InitS3Config()` is called.

The same situation seems to be present http client plugin code: its `XrdClGetPlugIn()` function doesn't pass the map to the constructor.

1. Define `XrdClS3UrlStyle = foobar` in `client.plugins.d/s3.conf`.
2. Start xrootd.
3. Observe that xrootd does not exit with the "Server configuration has invalid setting for URL style" error, which it should do if the invalid setting was being read from `s3.conf`.

### Additional Information

_No response_

dynamic-entropy:
Fixes #2787 

I was using environment variables for now, from comments in https://github.com/xrootd/xrootd/issues/2787 I understood with systemd managing xrootd process, one can only use Environment directive since `xrd` prefix is not accepted in env by xrootd.  https://github.com/xrootd/xrootd/issues/2787

I think its best to allow configuring client plugins through the same file (though one can argue it mixes up plugin config with lib loading config). It seems much more intuitive to do it this way. 

If we come to a consensus on the approach, we can also implement this for HTTP plugin configuration, too.
The second commit simply makes the errors more explicit, something I realised while testing the changes.

### DIFF
--- a/src/XrdCl/XrdClEnv.cc
+++ b/src/XrdCl/XrdClEnv.cc
@@ -17,6 +17,7 @@
 //------------------------------------------------------------------------------
 
 #include <cstdlib>
+#include <algorithm>
 
 #include "XrdCl/XrdClEnv.hh"
 #include "XrdCl/XrdClDefaultEnv.hh"
@@ -25,6 +26,67 @@
 
 namespace XrdCl
 {
+  //----------------------------------------------------------------------------
+  // Normalize a configuration key
+  //----------------------------------------------------------------------------
+  std::string Env::UnifyKey( std::string key )
+  {
+    std::transform( key.begin(), key.end(), key.begin(), ::tolower );
+
+    static const char prefix[] = "xrd_";
+    if( key.compare( 0, sizeof( prefix ) - 1, prefix ) == 0 )
+      key = key.substr( sizeof( prefix ) - 1 );
+
+    return key;
+  }
+
+  //----------------------------------------------------------------------------
+  // Look up a value in a plug-in configuration map
+  //----------------------------------------------------------------------------
+  std::string Env::GetPluginConfigValue(
+      const std::map<std::string, std::string> &pluginConfig,
+      const std::string                         &key )
+  {
+    const auto unified = UnifyKey( key );
+    for( const auto &entry : pluginConfig )
+    {
+      if( UnifyKey( entry.first ) == unified )
+        return entry.second;
+    }
+    return "";
+  }
+
+  //----------------------------------------------------------------------------
+  // Resolve a string setting
+  //----------------------------------------------------------------------------
+  bool Env::ResolveString( const std::string                         &key,
+                           const std::string                         &shellKey,
+                           const std::map<std::string, std::string> *pluginConfig,
+                           std::string                               &value,
+                           const std::string                         &defaultValue )
+  {
+    value.clear();
+    PutString( key, "" );
+    if( !shellKey.empty() )
+      ImportString( key, shellKey );
+
+    if( GetString( key, value ) && !value.empty() )
+      return true;
+
+    if( pluginConfig )
+    {
+      value = GetPluginConfigValue( *pluginConfig, key );
+      if( !value.empty() )
+      {
+        PutString( key, value );
+        return true;
+      }
+    }
+
+    value = defaultValue;
+    return false;
+  }
+
   //----------------------------------------------------------------------------
   // Get string
   //----------------------------------------------------------------------------
--- a/src/XrdCl/XrdClEnv.cc
+++ b/src/XrdCl/XrdClEnv.cc
@@ -88,6 +88,48 @@ namespace XrdCl
   }
 
   //----------------------------------------------------------------------------
+  // Resolve an integer setting
+  //----------------------------------------------------------------------------
+  bool Env::ResolveInt( const std::string                         &key,
+                        const std::string                         &shellKey,
+                        const std::map<std::string, std::string> *pluginConfig,
+                        int                                       &value,
+                        int                                        defaultValue )
+  {
+    PutInt( key, defaultValue );
+    if( !shellKey.empty() && ImportInt( key, shellKey ) )
+    {
+      GetInt( key, value );
+      return true;
+    }
+
+    if( pluginConfig )
+    {
+      const auto strValue = GetPluginConfigValue( *pluginConfig, key );
+      if( !strValue.empty() )
+      {
+        char *endPtr = nullptr;
+        const int parsed = static_cast<int>( strtol( strValue.c_str(), &endPtr, 0 ) );
+        if( endPtr && *endPtr == '\0' )
+        {
+          PutInt( key, parsed );
+          value = parsed;
+          return true;
+        }
+
+        Log *log = DefaultEnv::GetLog();
+        log->Error( UtilityMsg,
+                    "Env: Unable to import %s as %s: %s is not a proper integer",
+                    strValue.c_str(), key.c_str(), strValue.c_str() );
+      }
+    }
+
+    value = defaultValue;
+    PutInt( key, defaultValue );
+    return false;
+  }
+
+  //----------------------------------------------------------------------------
   // Get string
   //----------------------------------------------------------------------------
   bool Env::GetString( const std::string &k, std::string &value )
--- a/src/XrdCl/XrdClEnv.hh
+++ b/src/XrdCl/XrdClEnv.hh
@@ -139,6 +139,76 @@ namespace XrdCl
       bool GetDefaultStringValue( const std::string &key, std::string &value );
 
       //------------------------------------------------------------------------
+      //! Normalize a configuration key for case-insensitive lookup.
+      //! The key is lowercased and a leading "xrd_" prefix is removed.
+      //! @param key : configuration key to normalize
+      //! @return    : normalized key suitable for map lookup
+      //------------------------------------------------------------------------
+      static std::string UnifyKey( std::string key );
+
+      //------------------------------------------------------------------------
+      //! Look up a value in a plug-in configuration map.
+      //! Keys are matched case-insensitively after applying UnifyKey().
+      //! @param pluginConfig : settings from a client.plugins.d entry
+      //! @param key          : configuration key to look up
+      //! @return             : value associated with @a key, or an empty
+      //!                       string if no matching entry exists
+      //------------------------------------------------------------------------
+      static std::string GetPluginConfigValue(
+          const std::map<std::string, std::string> &pluginConfig,
+          const std::string                         &key );
+
+      //------------------------------------------------------------------------
+      //! Resolve a string setting using shell environment, then plug-in
+      //! configuration, then @a defaultValue.
+      //!
+      //! The resolved value is stored in the environment under @a key.
+      //! @param key          : environment key to store the resolved value
+      //!                       under
+      //! @param shellKey     : shell environment variable to import from;
+      //!                       may be empty to skip shell lookup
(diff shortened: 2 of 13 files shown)


dynamic-entropy:
> It seems much more intuitive to do it this way.

Talking with @abh3 today, I rechecked this and realise that it was intended to be this way, i.e the plugin configuration to live alongside the library loading configuration used by `PluginManager`. 

https://github.com/dynamic-entropy/xrootd/blob/4ec63a24e099f108f1b7af8cd0dd4e41ace6a8c5/src/XrdCl/XrdClPlugInManager.hh#L114-L115

src/XrdCl/XrdClEnv.hh
```diff
@@ -138,6 +138,45 @@ namespace XrdCl
       //------------------------------------------------------------------------
       bool GetDefaultStringValue( const std::string &key, std::string &value );
 
+      //------------------------------------------------------------------------
+      //! Normalize a configuration key for case-insensitive lookup.
+      //! The key is lowercased and a leading "xrd_" prefix is removed.
+      //------------------------------------------------------------------------
+      static std::string UnifyKey( std::string key );
```
abh3, line 147: Now that this is public, it would be nice to hace Doxygen headers for this. The same goes for all the new functions added here. Not a strict requirement but a nice to have feature, especially since all the existing functions have it. Note, just adding the @return is not sufficient. We need the @params as well.

src/XrdCl/XrdClFileSystem.cc
```diff
@@ -1195,7 +1195,7 @@ namespace XrdCl
                                        time_t              timeout )
   {
     SyncResponseHandler handler;
-    Status st = DeepLocate( path, flags, &handler, timeout );
+    XRootDStatus st = DeepLocate( path, flags, &handler, timeout );
```
abh3, line 1198: Please explain why it is that you need all the information that XRootDStatus has when only the parent Status object information is sufficient. One could claim that throwing away unwanted information is an optimization. On the other hand textually aligning the return type is technically appealing. However, it is not a sufficient reason to warrant all these changes the are merely technically appealing. So, what are we missing here?

src/XrdCl/XrdClEnv.cc
```diff
@@ -25,6 +26,109 @@
...
+      if( UnifyKey( entry.first ) == unified )
+        return entry.second;
+    }
+    return "";
+  }
+
+  //----------------------------------------------------------------------------
+  // Resolve a string setting
+  //----------------------------------------------------------------------------
+  bool Env::ResolveString( const std::string                         &key,
```
abh3, line 62: OK, I get why you added these to ClEnv as opposed to extending the existing methods as to not break ABI. On the other hand, why do you want to ever inspect/set the default env? Plugins keep their own environment and I think we want to keep it that way. This would also mean there is no reason to pass a pointer to the plugin env and could be changed to a reference, which is preferable. If we go down this path, then would it not make more sense to create new class, say XrdClEnvPI.hh that inherits XrdClEnv. Then we simply need to make UnifyKey() protected (a much simpler change) and isolates the plugin env's from the general env that actually has no ned to know anything about the plugin env's. OK, I suppose you would say that would orphan the plugin configs from the python API but I think this is far too complicated to provide python API access to the plugin env's.

Having a XrdClEnvPI object would also allow that object to also hold the hpecific plugin's env map eliminating the need to pass the map address to any of its methods. A further simplifucation.

So, we need some rethinking here.

src/XrdClHttp/XrdClHttpFactory.cc
```diff
@@ -102,78 +102,65 @@ Factory::Initialize()
...
+        m_log->Debug(kLogXrdClHttp, "Using %d bytes/sec for the slow transfer rate threshold", slow_xfer_rate);
         XrdClHttp::CurlOperation::SetSlowRateBytesSec(slow_xfer_rate);
 
         // Determine the minimum header timeout.  It's somewhat arbitrarily defaulted to 2s; below
         // that and timeouts could be caused by OS scheduling noise.  If the client has unreasonable
         // expectations of the origin, we don't want to cause it to generate lots of origin-side load.
-        std::string val;
         struct timespec mct{2, 0};
-        if (env->GetString("HttpMinimumHeaderTimeout", val) && !val.empty()) {
+        if (!val.empty()) {
```
abh3: So, which val are you referencing? You just deleted the previous definition so you must be using some further away definition. Could you point out why val still has contextual sense here?

src/XrdClS3/XrdClS3Factory.cc
```diff
@@ -349,18 +340,24 @@ Factory::ExtractHostname(const std::string_view url) {
     return authority;
 }
 
+void
+Factory::SetPluginConfig(const std::map<std::string, std::string> &config)
```
abh3, line 344: I know the OSG developers love doing this but isn't the config going to be static in any case? So, why set a dynamic variable to a a static address as opposed to using it directly, even if you have multiple instances of this object and may be more obvious.

abh3 (commented):
This is pass 1 on the review with loads of questions. I won't say changes needed because I need to fully understand why this particular path was chosen. Once we resolve this, part 2 of the review will occur.

src/XrdCl/XrdClFileSystem.cc
dynamic-entropy: I think the parent status information is "not sufficient". As mentioned in the second commit message, this was the reason to change to `XRootDStatus`, inorder to preserve `pMessage` (which I am assuming is "wanted" information). Plugins are supposed to return custom error messages through this.

It is also not for technical appeal but to prevent object slicing of `XRootDStatus` into `Status` and losing `pMessage`.

src/XrdClHttp/XrdClHttpFactory.cc
dynamic-entropy: This is the `minimum_header_timeout`. I should name this better to `minimum_header_timeout`, thanks for catching.

https://github.com/dynamic-entropy/xrootd/blob/73f42ca649cd3daaf71c668a0bbf9ea113e10250/src/XrdClHttp/XrdClHttpFactory.cc#L105-L106

src/XrdCl/XrdClEnv.cc
dynamic-entropy: I do like this idea, and would be happy to implement. 

The reason we use `DefaultEnv` in the current PR is because we used it before. i.e this PR did not change the "store" for the configuration. 
It looks like it when we look at the S3 changes, however it only matches what we already do in HTTP. 
`env = XrdCl::DefaultEnv` in XrdHttpUtil.

src/XrdClS3/XrdClS3Factory.cc
dynamic-entropy: This is again for symmetry with HTTP (same reason as above). The config in HTTP is loaded later on `CreateFile ` as oposed to `Initialise` like in S3 plugin.

src/XrdCl/XrdClEnv.cc
abh3: OK, I think we agreed with go ahead and make the better change.

src/XrdClS3/XrdClS3Factory.cc
abh3: Same here. If he S3 changes work out well, we will extend it to XrdClHttp.

src/XrdCl/XrdClEnv.hh
abh3: I suppose we will keep this comment open and hope for some doxygen comments.

src/XrdClS3/XrdClS3File.cc
```diff
@@ -135,8 +135,9 @@ File::GetFileHandle(const std::string &s3_url) {
     }
 
     std::string https_url, err_msg;
-    if (!Factory::GenerateHttpUrl(s3_noslash_url.empty() ? s3_url : s3_noslash_url, https_url, nullptr, err_msg)) {
-        return std::make_tuple(XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidAddr, 0, err_msg), "", nullptr);
+    uint16_t err_code = 0;
+    if (!Factory::GenerateHttpUrl(s3_noslash_url.empty() ? s3_url : s3_noslash_url, https_url, nullptr, err_msg, &err_code)) {
+        return std::make_tuple(XrdCl::XRootDStatus(XrdCl::stError, err_code, 0, err_msg), "", nullptr);
```
amadio: The interface changes look a bit clunky. If you are going to change the interface, it's probably easier to return `XrdCl::XRootDStatus` instead of bool, so you don't have to have this mix of pointers and references in the argument list. Then if you get back a status which is not ok, you just relay it over to the user.

src/XrdClS3/XrdClS3Factory.hh
```diff
@@ -49,7 +50,11 @@ public:
     //
     // If "obj_result" is not nullptr, then it will be set to the object/key and the resulting HTTPS URL
     // will not include the key name.
-    static bool GenerateHttpUrl(const std::string &s3_url, std::string &https_url, std::string *obj_result, std::string &err_msg);
+    static bool GenerateHttpUrl(const std::string &s3_url,
+                                std::string &https_url,
+                                std::string *obj_result,
+                                std::string &err_msg,
+                                uint16_t *err_code = nullptr);
```
amadio: Defaulting arguments to `nullptr` is not a good idea, please remove this. We had a really hard to debug problem in the client because of this already. I think the better way is to return a client status instead of a `bool` for this function.

amadio (changes requested):

src/XrdClS3/XrdClS3File.cc
dynamic-entropy: Agree, that would be a much cleaner interface.

src/XrdCl/XrdClEnv.hh
dynamic-entropy: Done!

src/XrdClS3/XrdClS3File.cc
dynamic-entropy: Done

PR #2850, opened 2026-06-26
[XrdClS3] Fix use-after-move crash when synchronous read fails

### LINKED ISSUE #2790: Segfaults with some operations with S3 backend
### Operating System

AlmaLinux 9.7

### XRootD Version

6.0.1

### Description

While testing XRootD 6.0.1 with an S3 storage backend, I have encountered a number of operations which trigger segfaults in the XRootD server:
  * Write files with `xrdcp` via dav or xroot protocols
  * Request removal of a non-existent file with the `xrdfs` `rm` command via dav protocol
  * Request removal of an existing or non-existent directory with the `xrdfs` `rmdir` command via the dav protocol

The stack traces from these three actions are in the attachment[1]. They appear to be nearly identical, suggesting a common cause. They have similarities to the traces I observed from the XRootD 6.0.0 segfault described in #2776. The apparent sensitivity to the protocol (dav, xroot) seen by the `rm` and `rmdir` commands is interesting.

For what it's worth, using the equivalent gfal commands had the same effect on the server.

[1] [xrootd-stack_traces.txt](https://github.com/user-attachments/files/27625092/xrootd-stack_traces.txt)

1. Configure XRootD to use an S3 storage backend.  See for example the configuration attached to #2754.
2. Start the XRootD server
3. Run the commands included in the xrootd-stack_traces.txt attachment. They can be run on the same machine that's running the XRootD server.

### Additional Information

_No response_

dynamic-entropy:
Fixes: #2790

This is a partial fix to the issue reported in #2790. 
We fix the branch leading to the crash from segfault, but the reason for the Read failure that lead to this path still needs investigation.

### DIFF
--- a/src/XrdClHttp/XrdClHttpFile.cc
+++ b/src/XrdClHttp/XrdClHttpFile.cc
@@ -651,7 +651,9 @@ File::ReadPrefetch(uint64_t offset, uint64_t size, void *buffer, XrdCl::Response
         return std::make_tuple(XrdCl::XRootDStatus{}, false);
     }
     std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
-    if (m_prefetch_size == -1) {
+    // In full-download mode the transfer is a single GET whose end is detected on transfer completion 
+    // This is similar to a red to end case
+    if (m_prefetch_size == -1 && !m_full_download.load(std::memory_order_relaxed)) {
         m_logger->Debug(kLogXrdClHttp, "%sRead prefetch skipping due to unknown file size", isPgRead ? "Pg": "");
         m_prefetch_reads_miss.fetch_add(1, std::memory_order_relaxed);
         m_default_prefetch_handler->m_prefetch_enabled = false;
--- a/src/XrdClHttp/XrdClHttpFile.cc
+++ b/src/XrdClHttp/XrdClHttpFile.cc
@@ -651,8 +651,8 @@ File::ReadPrefetch(uint64_t offset, uint64_t size, void *buffer, XrdCl::Response
         return std::make_tuple(XrdCl::XRootDStatus{}, false);
     }
     std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
-    // In full-download mode the transfer is a single GET whose end is detected on transfer completion 
-    // This is similar to a red to end case
+    // In full-download mode the transfer is a single GET whose end is detected on transfer completion
+    // This is similar to a read to end case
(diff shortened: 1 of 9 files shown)


amadio:
I'll wait until CI is green before reviewing this. Please fix when you have time.

dynamic-entropy:
> I'll wait until CI is green before reviewing this. Please fix when you have time.

Well, there are no failures in XrootD tests. It's CDash failing to upload the test results.
This is becoming frequent lately. A rerun is what you need.

I will rebase on the current master to trigger one.

amadio:
Ok, but also I don't see tests to make sure the problem doesn't happen again. Can you please add?

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -651,7 +651,9 @@ File::ReadPrefetch(uint64_t offset, uint64_t size, void *buffer, XrdCl::Response
         return std::make_tuple(XrdCl::XRootDStatus{}, false);
     }
     std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
-    if (m_prefetch_size == -1) {
+    // In full-download mode the transfer is a single GET whose end is detected on transfer completion 
+    // This is similar to a red to end case
```
amadio: There is a typo above. Could you please fix?

amadio:
@dynamic-entropy I'm not a big fan of the `unique_ptr(this)` pattern, but it's not your fault, this pattern unfortunately comes from XrdCl itself from long ago. However, I still think that it would be worth adding a test to check that the fix indeed fixes the problem, and that it will not regress in the future. Could you please add a test?

amadio:
@dynamic-entropy Please rebase your branch on the current master (i.e. with 5.9.7 already merged). Thank you.

src/XrdClHttp/XrdClHttpFile.cc
```diff
@@ -651,8 +651,8 @@ File::ReadPrefetch(uint64_t offset, uint64_t size, void *buffer, XrdCl::Response
         return std::make_tuple(XrdCl::XRootDStatus{}, false);
     }
     std::unique_lock lock(m_default_prefetch_handler->m_prefetch_mutex);
-    // In full-download mode the transfer is a single GET whose end is detected on transfer completion 
-    // This is similar to a red to end case
+    // In full-download mode the transfer is a single GET whose end is detected on transfer completion
+    // This is similar to a read to end case
```
amadio, line 655: This should really have been fixed in the original commit... I was tricked by the fact that the fix is this unrelated commit...

PR #2851, opened 2026-07-01
[Tests] Vendored gTest and gMock

diogomattioli:
This PR introduces vendored support for GoogleTest (gtest) and GoogleMock (gmock) in the repository, pinning both dependencies to version **1.16.0**.

Vendoring these dependencies eliminates version inconsistencies across Linux distributions and MacOS that can lead to test and mock build failures. It also removes the requirement to install GoogleMock as a separate package on distributions where it is not bundled with GoogleTest, resulting in a more consistent and reproducible build environment across supported platforms.


tests/unit/basic/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(basic-tests-unit mock.cc
+        )
+
+target_link_libraries(basic-tests-unit GTest::gmock GTest::gtest GTest::gtest_main)
```
amadio: This is just a small inconsistency in formatting. Please pick which one you prefer and apply to both `add_executable` and `target_link_libraries`.

diogomattioli: done!

cmake/XRootDFindLibs.cmake
```diff
@@ -89,16 +89,16 @@ if( ENABLE_XRDOSSARC )
...
+
+  include(FetchContent)
+
+  FetchContent_Declare(
+      googletest
+      SOURCE_DIR "${CMAKE_SOURCE_DIR}/vendor/googletest"
+  )
+  FetchContent_MakeAvailable(googletest)
+
+  set( BUILD_TESTS TRUE )
```
amadio: I'd prefer to `add_subdir(${CMAKE_SOURCE_DIR}/vendor/googletest EXCLUDE_FROM_ALL)`, like done [here](https://github.com/root-project/veccore/blob/accd6393230ede07d7a5028bcd12e26ad90cacd4/cmake/Builtins.cmake#L20).

cmake/XRootDSummary.cmake
```diff
@@ -22,7 +22,7 @@ component_status( PYTHON    BUILD_PYTHON      Python_Interpreter_FOUND AND Pytho
 component_status( READLINE  ENABLE_READLINE   READLINE_FOUND )
 component_status( SCITOKENS ENABLE_SCITOKENS  BUILD_SCITOKENS )
 component_status( SERVER_TESTS ENABLE_SERVER_TESTS TRUE_VAR )
-component_status( TESTS     BUILD_TESTS       GTEST_FOUND )
+component_status( TESTS     BUILD_TESTS       TRUE )
```
amadio:
```diff
-component_status( TESTS     BUILD_TESTS       TRUE )
+component_status( TESTS     ENABLE_TESTS      BUILD_TESTS )
```

diogomattioli: done!

cmake/XRootDFindLibs.cmake
diogomattioli: done!

abh3:
I think the review from me rests on ther I concur that we lock the gTools version number or not. Given that we need to support a wide range of distributions with varying version levels, I don'd an reasonable option other that, for now, standardizing on the least comment denominator version. At least that will help us sift out our bugs with little contamination from the run-time environment.  So, yes, I don't see anther viable option other than locking the version number.

abh3 (approved):
See my lasr comment.

tests/XrdHttpTpc/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ add_library(XrdHttpTpcUtils
   ${PROJECT_SOURCE_DIR}/src/XrdHttpTpc/XrdHttpTpcUtils.cc
 )
 
+target_include_directories(xrdhttptpc-unit-tests PRIVATE
+    ${OPENSSL_INCLUDE_DIR}
+)
```
amadio: Why is this added? It should never be necessary. The `OpenSSL::SSL` target below should already provide the required include directories for OpenSSL.

amadio: Actually, we likely just need to change `PRIVATE` to `PUBLIC` below in `target_link_libraries(XrdHttpTpcUtils ...)`, or link `xrdhttptpc-unit-tests` explicitly with `OpenSSL::SSL` if it uses it directly itself.

PR #2852, opened 2026-07-02
Obuscate auth when logging

dynamic-entropy:
Remove log noise from dumping of host info from `FromString` function

### DIFF
--- a/src/XrdCl/XrdClLocalFileHandler.cc
+++ b/src/XrdCl/XrdClLocalFileHandler.cc
@@ -17,29 +17,30 @@
 // along with XRootD.  If not, see <http://www.gnu.org/licenses/>.
 //------------------------------------------------------------------------------
 #include "XrdCl/XrdClLocalFileHandler.hh"
+#include "XProtocol/XProtocol.hh"
 #include "XrdCl/XrdClConstants.hh"
+#include "XrdCl/XrdClFileSystem.hh"
+#include "XrdCl/XrdClMessageUtils.hh"
 #include "XrdCl/XrdClPostMaster.hh"
 #include "XrdCl/XrdClURL.hh"
-#include "XrdCl/XrdClMessageUtils.hh"
-#include "XrdCl/XrdClFileSystem.hh"
-#include "XProtocol/XProtocol.hh"
 
+#include "XrdOuc/XrdOucPrivateUtils.hh"
 #include "XrdSys/XrdSysE2T.hh"
-#include "XrdSys/XrdSysXAttr.hh"
 #include "XrdSys/XrdSysFAttr.hh"
 #include "XrdSys/XrdSysFD.hh"
+#include "XrdSys/XrdSysXAttr.hh"
 
-#include <string>
 #include <memory>
 #include <stdexcept>
+#include <string>
 
-#include <fcntl.h>
+#include <aio.h>
+#include <arpa/inet.h>
 #include <cstdio>
 #include <cstdlib>
-#include <unistd.h>
+#include <fcntl.h>
 #include <sys/stat.h>
-#include <arpa/inet.h>
-#include <aio.h>
+#include <unistd.h>
 
 namespace
 {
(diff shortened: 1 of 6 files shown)


src/XrdCl/XrdClMetalinkRedirector.cc
```diff
@@ -419,13 +419,26 @@ namespace XrdCl
...
       replica = URL( url );
-      if( !replica.IsValid() || replica.GetURL().size() > 4096 )
-        continue; // this is the internal limit (defined in the protocol)
+      if( !replica.IsValid() )
+      {
+        log->Debug( UtilityMsg, "Skipping invalid metalink replica URL: %s",
+                    obfuscateAuth( url ).c_str() );
+        continue;
+      }
+      if( replica.GetURL().size() > 4096 )
```
amadio: `} else if {`...

src/XrdCl/XrdClURL.cc
```diff
@@ -144,35 +141,15 @@ namespace XrdCl
 
     if( !ParseHostInfo( hostInfo ) )
     {
+      log->Debug( UtilityMsg, "Failed to parse host info from URL: %s",
+                  obfuscateAuth( url ).c_str() );
+      log->Dump( UtilityMsg, "Host info component: %s", obfuscateAuth( hostInfo ).c_str() );
```
amadio: The  `hostInfo` doesn't need the `obfuscateAuth` filter, does it? It has no query parameters, since it's only the host info. This should also be combined into a single message, at warning level and using only `hostInfo`, without calling `obfuscateAuth()`.

src/XrdCl/XrdClMetalinkRedirector.cc
```diff
@@ -419,13 +419,26 @@ namespace XrdCl
...
+      {
+        log->Debug( UtilityMsg, "Skipping metalink replica URL exceeding "
+                    "size limit: %s", obfuscateAuth( url ).c_str() );
+        log->Dump( UtilityMsg, "Replica URL length: %zu",
+                   replica.GetURL().size() );
```
amadio: Please combine these two into a single log message, also at warning level:
```cpp
log->Warning(UtilityMsg,
  "Metalink replica URL size exceeds the limit (%zu > 4096)", replica.GetURL().size());
```
I think it's best not to print the URL to not pay the penalty for `obfuscateAuth(url)` too.

src/XrdCl/XrdClMetalinkRedirector.cc
```diff
@@ -419,13 +419,26 @@ namespace XrdCl
...
     const char *url = 0;
     while( ( url = fileInfos[0]->GetUrl() ) )
     {
       replica = URL( url );
-      if( !replica.IsValid() || replica.GetURL().size() > 4096 )
-        continue; // this is the internal limit (defined in the protocol)
+      if( !replica.IsValid() )
+      {
+        log->Debug( UtilityMsg, "Skipping invalid metalink replica URL: %s",
+                    obfuscateAuth( url ).c_str() );
```
amadio: This message should probably be at warning level, not debug.

src/XrdCl/XrdClLocalFileHandler.cc
```diff
@@ -822,7 +823,11 @@ namespace XrdCl
 
     URL fileUrl( url );
     if( !fileUrl.IsValid() )
+    {
+      log->Error( FileMsg, "Trying to open invalid local file URL: %s",
+                  obfuscateAuth( url ).c_str() );
```
amadio: Local URLs do not need/use cgi for authentication, why use `obfuscateAuth(url)` here?

amadio (changes requested):
I think the log messages being added fit better as warnings. This allows people to use `env XRD_LOGLEVEL=Warning` to see problems, but with much less verbose output than `Debug`.

amadio:
@dynamic-entropy Maybe you forgot to push the newer changes here? I don't see anything after my review.

amadio (approved):

amadio:
@dynamic-entropy I merged a previous change from you, which caused merge conflicts with this. Could you please rebase on master and fix the conflicts? Then this can also be merged for 6.2. Thanks!

dynamic-entropy:
The merge conflict was not from my own changes, but from David's instead. 

Anyways, I have rebased it on the current master.

amadio:
Thanks!

PR #2856, opened 2026-07-06
[XrdClHttp] Read BEARER_TOKEN authorisation

### LINKED ISSUE #2740: Support `BEARER_TOKEN` environment variable for xrdcp (http)
### Feature Description

To match the behaviour of xrdcp with xrootd protocol, we should also support sending the `Authorization` header for http protocol whenever we find the `BEARER_TOKEN` env var set. 

Are there any caveats we need to take into consideration? Or can we simply check for the env var and set `CURLOPT_XOAUTH2_BEARER` in GetHandle?
@bbockelm

dynamic-entropy:
Currently, there is no way to pass a bearer_token to xrdclhttp without an external header callout. For e.g. a client performing `xrdfs https://host:port stat /` has no way to set Authorisation header for the HTTP call.
This commit adds reading token files similar to XrdSecProtocolztn to maintain parity with the xrootd protocol.

The header callouts can override the token set in InjectBearerToken. We skip setting authorisation header if it is present in the url parameters via authz= or if a "Authorization" header is already present.

Fixes: #2740

### DIFF
--- a/src/XrdOuc/XrdOucBearerToken.cc
+++ b/src/XrdOuc/XrdOucBearerToken.cc
@@ -1,0 +1,102 @@
+#include "XrdOuc/XrdOucBearerToken.hh"
+
+#include <fcntl.h>
+#include <sys/param.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include <array>
+#include <cerrno>
+#include <cstdio>
+#include <string>
+#include <utility>
+
+std::string XrdOucBearerToken::Strip(std::string_view token) {
+    constexpr std::string_view ws = " \t\n\r\f\v";
+    const auto start = token.find_first_not_of(ws);
+    if (start == std::string_view::npos) return {};
+    const auto end = token.find_last_not_of(ws);
+    return std::string(token.substr(start, end - start + 1));
+}
+
+XrdOucBearerToken::Result XrdOucBearerToken::FromEnvValue(const char* value, size_t maxSize) {
+    if (!value || !*value) return {};
+
+    auto token = Strip(value);
+    if (token.empty()) return {};
+
+    if (maxSize && token.size() > maxSize) return {Status::Error, {}, EMSGSIZE, value};
+
+    return {Status::Found, std::move(token), 0, {}};
+}
+
+XrdOucBearerToken::Result XrdOucBearerToken::ReadFile(const char* path, size_t maxSize) {
+    if (!path || !*path) return {};
+
+    struct stat st;
+    if (stat(path, &st) != 0) {
+        if (errno == ENOENT) return {};
+        return {Status::Error, {}, errno, path};
+    }
+
+    if (maxSize && static_cast<size_t>(st.st_size) > maxSize)
+        return {Status::Error, {}, EMSGSIZE, path};
+
+    // make sure the file is not accessible to anyone but the owner
+    if (st.st_mode & (S_IRWXG | S_IRWXO)) return {Status::Error, {}, EPERM, path};
+
+    int fd = open(path, O_RDONLY);
+    if (fd < 0) return {Status::Error, {}, errno, path};
+
+    std::string token(static_cast<size_t>(st.st_size), '\0');
+    ssize_t nread = read(fd, &token[0], static_cast<size_t>(st.st_size));
+    int readErr = errno;
+    close(fd);
+
+    if (nread != st.st_size) return {Status::Error, {}, nread < 0 ? readErr : EIO, path};
+
+    token = Strip(token);
+    if (token.empty()) return {};
+
+    return {Status::Found, std::move(token), 0, path};
+}
+
+XrdOucBearerToken::Result XrdOucBearerToken::TryEntry(std::string_view entry, size_t maxSize) {
+    if (entry.empty()) return {};
+
+    // absolute path to the token file
+    if (entry.front() == '/') {
+        char path[MAXPATHLEN + 8];
+        const std::string fmt(entry);
+        snprintf(path, sizeof(path), fmt.c_str(), static_cast<int>(geteuid()));
+        return ReadFile(path, maxSize);
+    }
+
+    const std::string envName(entry);
+    const char* env = getenv(envName.c_str());
+    if (!env || !*env) return {};
+
+    if (entry.size() >= 4 && !entry.compare(entry.size() - 4, 4, "_DIR")) {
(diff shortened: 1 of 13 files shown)


abh3:
So, what we have a problem because there now will be two versions of code that find the token using the WLCG discovery rules. Not good. What we really need is to abstract the token discovery in ztn into a callable utility function so that ztn and ClHttp can use the same logic implementation. Do you want to discuss how we should proceed with that effort? I'm here all week.

dynamic-entropy:
As per our discussions last week, I am done refactoring this.

amadio (changes requested):
I like the idea of deduplicating and fixing the token discovery mechanism. However, we need to ensure that the behavior stays exactly the same as it was before. Since you are changing significantly the implementation, this should come with tests which are created prior to the changes into master, then the refactor keeping the tests green. I'm not quite sure about the naming, though.

@abh3 Any suggestions?

dynamic-entropy:
@amadio Added black-box tests for XrdZtnToken

src/XrdOuc/XrdOucBearerToken.cc
```diff
@@ -0,0 +1,102 @@
...
+#include <string>
+#include <utility>
+
+std::string XrdOucBearerToken::Strip(std::string_view token) {
+    constexpr std::string_view ws = " \t\n\r\f\v";
+    const auto start = token.find_first_not_of(ws);
+    if (start == std::string_view::npos) return {};
+    const auto end = token.find_last_not_of(ws);
+    return std::string(token.substr(start, end - start + 1));
+}
```
amadio, line 20: Don't we already have functions in XrdOuc to trim strings somewhere else? Why add another one? There's `void XrdOucUtils::trim(std::string &str)`, please reuse what we have instead of adding code duplication.

src/XrdOuc/XrdOucBearerToken.hh
```diff
@@ -0,0 +1,95 @@
...
+#include <string_view>
+
+/**
+ * WLCG bearer-token discovery helpers shared by client components.
+ *
+ * Implements the discovery order defined by the WLCG Bearer Token Discovery
+ * specification and currently used by the ztn security protocol.
+ * https://github.com/WLCG-AuthZ-WG/bearer-token-discovery
+ */
+class XrdOucBearerToken {
```
amadio, line 15: I think a class is overkill, a single function (like the `getCredentials` it's going to replace from the `XrdSecztn` plugin should be enough.

amadio (changes requested):
Please let's try to avoid duplicating existing functionality and adding enums and classes for everything. This is a bad pattern in code generated by AI which I'd like to prevent from creeping up everywhere in the code. The proliferation of classes and enums only make maintenance harder in the future if any of this stuff becomes part of public interfaces.

amadio:
@dynamic-entropy @lobis's changes in #2836 and this pull request will certainly have merge conflicts. So, they cannot both be merged as is.

amadio:
This is going to have conflicts with the tape REST API, which is now good enough to be merged into 6.2. We can do the rearchitecting of authentication in XrdClHttp in 6.3.

dynamic-entropy:
We can close this if BearerTokens work for http in xrdcl from the other PR (i.e a token is discovered similar to xroot while using the xrootd client?) @lobis can you please confirm if this is the case, and I will close this.

lobis:
> We can close this if BearerTokens work for http in xrdcl from the other PR (i.e a token is discovered similar to xroot while using the xrootd client?) @lobis can you please confirm if this is the case, and I will close this.

It works, I just tested it and also have [tests](https://github.com/xrootd/xrootd/blob/36b7781945a185b5ef559f09628af3998cb80623/tests/XrdClHttp/TapeTest.cc#L83-L124) to check for this.

dynamic-entropy:
> It works, I just tested it and also have [tests](https://github.com/xrootd/xrootd/blob/36b7781945a185b5ef559f09628af3998cb80623/tests/XrdClHttp/TapeTest.cc#L83-L124) to check for this.

Doesn't look like the tests cover the wlcg token discovery mechanism. 

Are you able to store your token at `bt_<uid>` and use it with `xrdcp` for http?

lobis:
> > It works, I just tested it and also have [tests](https://github.com/xrootd/xrootd/blob/36b7781945a185b5ef559f09628af3998cb80623/tests/XrdClHttp/TapeTest.cc#L83-L124) to check for this.
> 
> Doesn't look like the tests cover the wlcg token discovery mechanism.
> 
> Are you able to store your token at `bt_<uid>` and use it with `xrdcp` for http?

No, perhaps I should have been more clear. In this commit we just implemente the support for `BEARER_TOKEN` and `BEARER_TOKEN_FILE`, so its not fully equivalent.

abh3:
As for token discovery, the WLCG mechanism is encoded in XrdSecztn method XrdSecProtocolztn::findToken(). That mechanism should not be duplicated everywhere it is needed. We don't want to chase all the implementations or have two possibly divergent mechanisms. So, that mechanism should be repackaged as a utility function that could be called by anyone needing to find a token (e.g. XrdOucUils).

dynamic-entropy:
> As for token discovery, the WLCG mechanism is encoded in XrdSecztn method XrdSecProtocolztn::findToken(). That mechanism should not be duplicated everywhere it is needed. We don't want to chase all the implementations or have two possibly divergent mechanisms. So, that mechanism should be repackaged as a utility function that could be called by anyone needing to find a token (e.g. XrdOucUils).

That is exactly what the PR does. The reason it's not a single function is to keep the use of the interface provided by `XrdSecProtocolztn` unchanged for the xrootd protocol. If the consensus is otherwise (and I am surprised that this was not the case in previous discussions), I can change it for both protocols.

It does not fit HTTP to use find and read separately, and I did not want to change xrootd's interface for token discovery. However, if the consensus is to get rid of (by not using them) XrdSecZtn::findToken and XrdSecZtn::readToken and have a single utility function, we can do that. 

The class allows me to have a container that can bundle all the helpers together; I can easily switch to a namespace, but I would still like to provide building blocks and not a single function. 
Having free functions that we would lose track of (both humans and LLMs). We see this happening all the time with utility functions; it's a structural issue, not because someone doesn't want to.

abh3:
Thanks Rahul. I haven't looked at the ztn code in quite a while and don't recall why the function needed to be split. Maybe it doesn't need to be split at all. Anyway, I am busy with the workshop so time is a bit short. Your patience is requested.

PR #2857, opened 2026-07-08
Make sure path in request is updated on retry. #2394

### LINKED ISSUE #2394: Wrong path when falling back to load-balancer
I am redirecting a client asking for /a/b/c to a local path /x/y/z. 

The local path does not exist, so the client falls back to the initial endpoint and sets the correct tried information:
```

[2024-12-20 11:06:30.866676 +0100][Debug  ][XRootD            ] 	0. Redirected from: root://host.cern.ch:1094//a/b/c to: file://localhost/x/y/z
[2024-12-20 11:06:30.866676 +0100][Debug  ][XRootD            ] 	1. Retrying: root://host.cern.ch:1094//a/b/c
```

But what happens is that the clients sends the request with the wrong path for retry:

```
func=open                     path=/x/y/z info=tried=localhost&triedrc=srverr`
```

This is probably what should happen here!

smithdh:


abh3 (approved):
This looks like it implements what it claims to do plus one more thing not claimed, it avoids tried processing when the URL is for a local file. This is correct and really should be added to the pull request description.

amadio:
@smithdh Could you please address the suggestion by @abh3, so we can merge this for 6.1.1? Thank you.

tests/cluster/ossredir/oss_url_redir_cluster_plugin.cc
```diff
@@ -0,0 +1,93 @@
...
+// URL-form SFS_REDIRECT handling by the XrdCl client. In particular
+// the tests are concerned with the behavior of the client if the redirection
+// target is file://, and/or if the open at the redirect target fails.
+// Additionally we may interested to see if a modified pathname is presented
+// back to any load-balancer or if the original pathname is used.
+//
+// For a file open (for write) with name matching kTrigger[12] we trigger the
+// ofs to issue a redirect, making use of -EDESTADDREEQ return code with
+// the "FileURL" key.
+//------------------------------------------------------------------------------
```
amadio: We already have `XrdOssTests` for testing purposes. I'd like to avoid the proliferation of oss plugins for tests. Can you please update `XrdOssTests` with the required functionality for this test instead of adding a new plugin?

amadio (changes requested):
Could you please update `XrdOssTests` with the functionality needed by this test, rather than implement another plugin?

tests/cluster/ossredir/oss_url_redir_cluster_plugin.cc
smithdh: yes, certainly. adding, along with abh3 comment-changes.

smithdh:
there are some tests failures; not certain if realted to the PR itself or not.  i'll check into them.

smithdh:
fixed tests after adopting use of XrdOssTest. (Needed extension of funcationity to allow Clone test to work when the test plugin is configured on one of the 'cluster' test data nodes, plus fix of missing library needed after I added XrdOucEnv.hh).

PR #2860, opened 2026-07-08
[Pfc] Fix snprintf handling in config printout

osschar:
Properly handle snprintf return values when putting together the final active configuration string that is reported in the log.

### DIFF
--- a/src/XrdPfc/XrdPfcConfiguration.cc
+++ b/src/XrdPfc/XrdPfcConfiguration.cc
@@ -21,14 +21,15 @@
 
 extern XrdSysXAttr *XrdSysXAttrActive;
 
-namespace XrdPfc
-{
-   const char *trace_what_strings[] = {"","error ","warning ","info ","debug ","dump "};
-}
-
-using namespace XrdPfc;
-
 XrdVERSIONINFO(XrdOucGetCache, XrdPfc);
+
+namespace XrdPfc {
+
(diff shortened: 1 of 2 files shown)


src/XrdPfc/XrdPfcConfiguration.cc
```diff
@@ -746,19 +790,19 @@ bool Cache::Config(const char *config_filename, const char *parameters, XrdOucEn
...
+         cfg_printf("       pfc.httpcc on\n");
       }
       if (m_configuration.m_qfsredir)
       {
-         loff += snprintf(buff + loff, sizeof(buff) - loff, "       pfc.qfsredir on\n");
+         cfg_printf("       pfc.qfsredir on\n");
       }
 
-      m_log.Say(buff);
+      m_log.Say(cfg_printf.c_str());
```
abh3, line 805: So, the big question is why are you collecting things in a big buffer only to spit this out in a single call to m_log.Say() (well, other than that is what previous config functiions have done). Why not just cann Say() for each line you generate then you wouldn't need any fancy functions, would you?

osschar: Well, it's not really a very big question :) I thought I will need the same wrapper for g-stream and other stuff.

amadio:
@osschar I verified that this indeed fixed the problems reported by CodeQL. Thank you!

Issue #2858, opened 2026-07-08
XrdClHttp: errors in complex vectored reads


ffurano:
### Operating System

All

### XRootD Version

v6.0.3-4-ga8fb957c3

### Description

The code that processes the curl chunks in response to a multipart request cannot properly handle header that fall at the boundary of the buffer, causing fatal internal errors

### Steps to Reproduce the Behavior

Look at CurlVectorReadOp::Write()
Craft a readV that puts the header on the boundary of 16KB (default buffer)
boom!

ffurano:
I think we have to reopen this.

After the fix (commit 8eac57c8cc7db11610f3f823d3b5747ab1b5fc67) my tests work

Some other commit broke them, today's master does not work, same issue described in this tkt

amadio:
@ffurano Can you please try to bisect to the problem commit? Also, if you can propose an extra regression test, it would be nice. The commit with the fix added a test, but apparently that was not enough. Thanks!   Just to confirm, 6.1.1 then works fine, right?

ffurano:
i'll try to do it, crossing fingers for the abi compatibility, as root takes a long time to compile

ffurano:
Mah, looks a bit strange to me

$ git bisect good
550c545ec2f9a171d4b864ad9e113c67377cfd84 is the first bad commit
commit 550c545ec2f9a171d4b864ad9e113c67377cfd84
Author: David Smith <david.smith@cern.ch>
Date:   Fri Jul 3 15:19:42 2026 +0200

    [Tests] Add redirect test with URL and a pathname change

 tests/XrdOssTests/CMakeLists.txt |  2 +-
 tests/XrdOssTests/XrdOssTests.cc | 46 +++++++++++++++++++++++++++++++++++++++-
 tests/cluster/srv1.cfg           |  2 ++
 tests/cluster/test.sh            | 37 ++++++++++++++++++++++++++++++++
 4 files changed, 85 insertions(+), 2 deletions(-)

amadio:
ABI is fine, you don't need to recompile ROOT. But the commit above could not have been what broke the test for you. You can try to revert and check. If you could paste the output of your failed check here, it could help too.

ffurano:
the test is a root script that draws an histogram that i jealously kept many years, difficult to reproduce with a copypaste here. At low level what it does is several complex, fat readV requests

ffurano:
[...]

PR #2861, opened 2026-07-09
XrdClHttp: fix processing of complex readv responses

ffurano:
Make the code able to correctly process readv responses where intermediate headers overlap the curl buffer

### DIFF
--- a/src/XrdClHttp/XrdClHttpOpReadV.cc
+++ b/src/XrdClHttp/XrdClHttpOpReadV.cc
@@ -33,7 +33,11 @@ CurlVectorReadOp::CurlVectorReadOp(XrdCl::ResponseHandler *handler, const std::s
         CurlOperation(handler, url, timeout, logger, callout, header_callout),
         m_vr(new XrdCl::VectorReadInfo()),
         m_chunk_list(op_list)
-    {}
+    {
+        m_remainderbuflen = 0;
+        m_doublebuf = NULL;
+        m_doublebufsz = 0;
+    }
 
 bool
 CurlVectorReadOp::Setup(CURL *curl, CurlWorker &worker)
@@ -145,6 +149,23 @@ CurlVectorReadOp::Write(char *orig_buffer, size_t orig_length)
 
     auto buffer = orig_buffer;
     auto length = orig_length;
+    
+    if (m_remainderbuflen) {
+        // Do we have unprocessed bytes from the previous chunk?
+        if (m_doublebufsz < length) {
+            if (!m_doublebufsz)
+                m_doublebuf = (char *)malloc(length+1024);
+            else
+                m_doublebuf = (char *)realloc(m_doublebuf, length+1024);
+            
+            m_doublebufsz = length;
+        }
+        memcpy(m_doublebuf, m_remainderbuf, m_remainderbuflen);
+        memcpy(m_doublebuf+m_remainderbuflen, buffer, orig_length);
+        buffer = m_doublebuf;
+        length += m_remainderbuflen;
+        m_remainderbuflen = 0;
+    }
 
     while (length) {
         // If we're in the middle of a response chunk, copy as much data as possible.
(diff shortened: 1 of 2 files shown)


ffurano:
Should fix #2858

ffurano:
Even after rebasing with master, I confirm that the Vector tests work for me (Ubuntu 24), while other tests that have nothing to do fail for unknown reasons. Maybe these unknown reasons are the same ones that make the vector tests fail in the CI.

```
The following tests passed:
        XrdClHttp::setup
        CurlVectorFixture.Test
        XrdClHttp::teardown

100% tests passed, 0 tests failed out of 3
```

I will need suggestions about how to debug this complex CI/Test environment

src/XrdClHttp/XrdClHttpOpReadV.cc
```diff
@@ -205,6 +226,17 @@ CurlVectorReadOp::Write(char *orig_buffer, size_t orig_length)
         // We are at the boundary between chunks; we must parse header lines to understand the
         // next thing to do.
 
+        // Attention! We could be close to the end of the incoming chunk, hence the
+        // separator/header we are looking for might have been truncated.
+        // If we are less than 1KB form the end of the curl chunk, save this
+        // less-than-one-kb
+        // to a temporary buffer. The next chunk will have to be appended to it
+        if (length && (length < 1024)) {
```
dynamic-entropy, line 234: Hello Fabrizio

The problem is the same as we discussed. The tests fail because the size of the entire response is smaller than 1KB. So, in this original test case we do not have the header split across boundary of two chunks, however because of this check, we accept all bytes from Curl without ever writing to the buffer of the original caller.

dynamic-entropy (commented):
if you run this with gdb remember to set these, otherwise `Produce` will timeout in 9 sec.
```
set environment XRD_HTTPDEFAULTHEADERTIMEOUT=3600s
set environment XRD_HTTPSTALLTIMEOUT=3600
```

src/XrdClHttp/XrdClHttpOpReadV.cc
ffurano: Maybe yes, maybe not. We should clarify:

- who are you
- why the same test works in my machine
- how to debug with the complex infrastructure
- why I am getting failed tests that have nothing to do with this

My activity is actually stuck because of the last 3 points

dynamic-entropy: > who are you

haha, the fact that we discussed this should narrow me down to one or two persons. I would leave this question open ended


> why the same test works in my machine

That has been bothering me too. Since this is a deterministic bug, most probably the test never ran on your machine with the changes included.


> how to debug with the complex infrastructure

Here is how I do it with gdb. I am open to better setups (at some point I shouuld do this with vscode)
From the build directory
```bash
ctest -R CurlVectorFixture -V -N
```
gives you the test commands and dependencies. 

Truncating irrelevant info and rearranging we end up with

Environment to set
```bash
247: Working Directory: /home/rchauhan/dev/xrootd/build/tests/XrdClHttp
247: Environment variables: 
247:  ENV_FILE=/home/rchauhan/dev/xrootd/build/tests/curl/setup.sh
247:  XRD_PLUGINCONFDIR=/home/rchauhan/dev/xrootd/build/tests/curl/client.plugins.d
```

setup
```bash
Added setup test XrdClHttp::setup required by fixture XrdClHttp
249: Test command: /home/rchauhan/dev/xrootd/tests/XrdClHttp/setup.sh "curl"
249: Working Directory: /home/rchauhan/dev/xrootd/build/tests/XrdClHttp
249: Environment variables: 
249:  OPENSSL_BIN=/usr/bin/openssl
249:  BINARY_DIR=/home/rchauhan/dev/xrootd/build
249:  SOURCE_DIR=/home/rchauhan/dev/xrootd
  Test #249: XrdClHttp::setup
```

test commands
```
Test command: /home/rchauhan/dev/xrootd/build/bin/xrdcl-http-test "--gtest_filter=CurlVectorFixture.Test"
```

teardown
```bash
Added cleanup test XrdClHttp::teardown required by fixture XrdClHttp
250: Test command: /home/rchauhan/dev/xrootd/tests/XrdClHttp/teardown.sh "curl"
250: Working Directory: /home/rchauhan/dev/xrootd/build/tests/XrdClHttp
250: Environment variables: 
250:  BINARY_DIR=/home/rchauhan/dev/xrootd/build
  Test #250: XrdClHttp::teardown
```

thus with gdb, you can do

```bash
 gdb --tui --args /home/rchauhan/dev/xrootd/build/bin/xrdcl-http-test   --gtest_filter=CurlVectorFixture.Test
 ```
You may or maynot like tui :)




> why I am getting failed tests that have nothing to do with this

it has to do with the change, the size check in you commit is not scoped to - "is the header split across two chunks" but is global. The assumption that "till we fill a 1KB buffer", do not process the data, is the reason.

ffurano:
Closing, a better solution was developed aside :-)

Issue #2863, opened 2026-07-10
[XrdApps] Provide gfal2-util-equivalent workflows through xrdfs and xrdcp


lobis:
## Motivation

The `gfal2-util` commands provide a convenient command-line interface for common storage operations. XRootD already provides much of the same functionality through `xrdfs`, `xrdcp`, and the XrdCl APIs, but the functionality is spread across different commands and the CLI conventions are not always consistent.

A prominent example is URL handling: `xrdcp` accepts complete URLs, while `xrdfs` normally requires the server and path to be split manually. This is already tracked in #1664.

This issue tracks making the native XRootD command-line tools a practical replacement for `gfal2-util` when using protocols supported by XRootD.

## Goal

Provide functional equivalents for the standard `gfal2-util` workflows through:

- `xrdcp` for data transfer operations;
- `xrdfs` for metadata and filesystem operations;
- existing XrdCl `FileSystem` and `File` APIs wherever possible;
- XrdClHttp for HTTP-specific functionality such as WLCG Tape REST and token operations.

The initial compatibility target is functional equivalence for protocols supported by XRootD, rather than byte-for-byte compatibility with every GFAL option and output format.

Existing `xrdfs` and `xrdcp` syntax must remain supported.

## Current design status

The migration frontend described below is the current agreed direction, but it is **not yet a final CLI or packaging design**.

The name `xrd` is used throughout this issue as a working name only. The final executable name, invocation syntax, compatibility package name, and installation policy remain open for upstream discussion.

## Non-goals

- GFAL library API compatibility.
- Support for SRM, GridFTP, LFC, or legacy catalog commands.
- Reimplementing existing XrdCl or `xrdcp` functionality in a new CLI layer.
- Adding a new command-line parsing dependency.
- Moving storage operations, protocol handling, or output reconstruction into the migration frontend.
- Treating the working name `xrd` as a settled naming decision.

## Thin migration frontend (provisional design)

Some operators and existing scripts are familiar with `gfal-*` commands but not with the current split between `xrdfs` and `xrdcp`. The project should therefore provide an easy migration path in addition to improving the native commands themselves.

The current proposal is a small command dispatcher, referred to here by the **working name** `xrd`:

```console
xrd copy SOURCE DESTINATION
xrd stat URL
xrd ls URL
xrd sum URL ADLER32
xrd mkdir -p URL
xrd rm URL
```

This would allow a simple mechanical migration:

```text
gfal-copy  -> xrd copy
[...]

PR #2865, opened 2026-07-10
[XrdCl] Accept full URLs in xrdfs

### LINKED ISSUE #1664: [XrdCl] Allow xrdfs commands to accept full URLs
## Motivation

`xrdcp` and `gfal2-util` accept complete storage URLs, while `xrdfs` currently requires users to split the endpoint from each path:

```console
xrdfs host[:port] stat /path
```

Users commonly already have a complete URL from a transfer, catalog, or job configuration. Splitting it in user code is cumbersome and error-prone. This is also the first compatibility improvement needed by #2863, which tracks providing gfal2-util-equivalent workflows through native XRootD commands.

## Desired behavior

Keep the current server-first syntax and interactive mode unchanged:

```console
xrdfs host[:port] stat /path
xrdfs host[:port]
```

Also accept command-first invocations where path operands are complete remote URLs:

```console
xrdfs stat root://host//path
xrdfs ls root://host//directory
xrdfs mv root://host//old root://host//new
```

This gives operators a gfal-style full-URL workflow while preserving all existing `xrdfs` usage.

## Scope of the first change

- Recognize command-first mode only when a path-oriented `xrdfs` command receives at least one valid complete remote URL.
- Infer the filesystem endpoint from the first complete URL.
- Pass the URL path and CGI/opaque parameters to the existing `xrdfs` command implementation.
- Require all complete URL operands in one command to use the same protocol, credentials, host, and port.
- Reject mixed endpoints clearly; operations such as `mv` remain same-storage operations.
- Preserve non-URL operands so existing command option parsing and multi-path behavior remain in the native command.
- Do not hardcode the parser to `root://`; accept valid remote URL schemes supported through XrdCl and its plugins.
- Keep commands whose operands are not normal filesystem paths, such as raw implementation-dependent queries, on the existing server-first form until their URL grammar is explicitly defined.

The implementation should only normalize the command line and then use the existing `FSExecutor`, `FileSystem`, and command handlers. It must not duplicate filesystem operations.

## Testing

Every behavior change should carry tests in the same pull request.

For this first slice:

- add GTest coverage for legacy parsing, command-first parsing, URL parameter preservation, options around URL operands, multiple paths, and mixed-endpoint rejection;
- keep the parser tests independent of a running server;
- add BATS end-to-end command coverage once the shared BATS integration topic is ready, rather than introducing another ad-hoc shell test framework.

## Acceptance criteria

- Existing server-first batch invocations behave exactly as before.
- Existing interactive mode behaves exactly as before.
- Command-first `stat`, `ls`, and other supported path-oriented commands accept complete remote URLs.
- Multiple complete URLs for the same endpoint are reduced to native path operands.
- Mixed endpoints fail before any operation is sent.
- URL CGI/opaque parameters are preserved on the path operand.
- Help and the `xrdfs(1)` man page document both forms.
- Focused parser tests pass through the existing CTest/GTest integration.

## Related work

- #2863 tracks the broader gfal2-util migration and thin compatibility frontend.
- #2827 contains prototype CLI behavior and tests that may be reused selectively, but the #1664 implementation should remain a focused change to `xrdfs`.

### LINKED ISSUE #2105: Add -4  and -6 options to xrdfs
Would be nice to be able select ipv4 or ipv6 network stack used when contacting a server.

lobis:
## Summary

- allow path-oriented `xrdfs` commands to accept complete remote URL operands in command-first form
- use `getopt_long` for global options and stop at the first subcommand so command-specific flags remain untouched
- infer one endpoint, preserve URL CGI parameters, and reject invalid remote URLs, mixed endpoints, or mixed credentials before dispatch
- preserve the existing server-first batch syntax and interactive mode
- keep storage operations in the existing `FSExecutor` and command handlers
- document both invocation styles and add process-level BATS coverage

Examples of the new form:

```console
xrdfs stat root://host//path
xrdfs ls root://host//directory
xrdfs mv root://host//old root://host//new
```

The existing form remains supported:

```console
xrdfs host stat /path
```

## Implementation

Global options are parsed with `getopt_long`. Parsing stops at the first
non-option argument, which lets existing subcommands continue to handle flags
such as `xrdfs ls -l ...` themselves.

For path-oriented commands, complete remote URL operands are normalized to
paths before being passed to the existing executor. All URL operands in one
command must use the same protocol, credentials, host, and port. URL path
parameters are preserved, while local `file` and `stdio` URLs are rejected.
URL-like extended-attribute values remain unchanged.

Raw implementation-dependent `query` commands remain on the explicit
server-first form so URL-like opaque payloads are not misclassified.

## Motivation

Operators often already have complete URLs from transfers, catalogs, or job
configuration. Requiring them to split the endpoint and path before invoking
`xrdfs` is cumbersome and prevents a straightforward migration from
`gfal2-util` workflows.

This is the first focused implementation slice of #2863. It only normalizes
the command line and delegates to the existing filesystem implementation.
Protocol capabilities remain unchanged and continue to depend on the native
transport or configured client plugin, including `XrdClHttp` for HTTP(S).

## Tests

- configured with `ENABLE_TESTS=ON` and `ENABLE_SERVER_TESTS=ON`
- built `xrdfs`, `xrdcp`, `xrootd`, and `xrdcl-unit-tests`
- `XrdCl::xrdfs-full-url` process test — 10 BATS cases passed
- XrdCl unit-test suites — 12/12 passed
- coverage includes legacy and command-first syntax, subcommand options, URL
  parameters, URL-like xattr values, multiple operands, mixed endpoints and
  credentials, local URLs, and raw-query rejection
- `git diff --check`

Closes #1664

### DIFF
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -2058,6 +2058,8 @@ XRootDStatus PrintHelp( FileSystem *, Env *,
 
   printf( "Available options:\n\n"                                            );
 
+  printf( "   -4, --ipv4          use only the IPv4 network stack\n"          );
+  printf( "   -6, --ipv6          use only the IPv6 network stack\n"          );
   printf( "   -d, --debug <level> set debug level: 0 off, 1 low, 2 medium,\n" );
   printf( "                       3 high\n"                                  );
   printf( "   -h, --help show this help\n"                                    );
@@ -2435,6 +2437,8 @@ int main( int argc, char **argv )
   XrdCl::FSExecutor::CommandParams params;
   enum { NoCwdOption = 256 };
   static const option options[] = {
+    { "ipv4",   no_argument,       0, '4' },
+    { "ipv6",   no_argument,       0, '6' },
     { "debug",  required_argument, 0, 'd' },
     { "help",   no_argument, 0, 'h' },
     { "no-cwd", no_argument, 0, NoCwdOption },
@@ -2443,12 +2447,29 @@ int main( int argc, char **argv )
 
   bool noCwd = false;
   int debugLevel = 0;
+  std::string networkStack;
   opterr = 0;
   int option = 0;
-  while( (option = getopt_long( argc, argv, "+d:h", options, 0 )) != -1 )
+  while( (option = getopt_long( argc, argv, "+46d:h", options, 0 )) != -1 )
   {
     switch( option )
     {
+      case '4':
+        if( networkStack == "IPv6" )
+        {
+          std::cerr << "xrdfs: -4 and -6 are mutually exclusive" << std::endl;
+          return 1;
+        }
+        networkStack = "IPv4";
+        break;
+      case '6':
+        if( networkStack == "IPv4" )
+        {
+          std::cerr << "xrdfs: -4 and -6 are mutually exclusive" << std::endl;
+          return 1;
+        }
+        networkStack = "IPv6";
+        break;
       case 'd':
       {
         char *end = 0;
@@ -2473,6 +2494,9 @@ int main( int argc, char **argv )
         return 1;
     }
   }
+
+  if( !networkStack.empty() )
+    DefaultEnv::GetEnv()->PutString( "NetworkStack", networkStack );
 
   if( debugLevel )
   {
--- a/src/XrdCl/XrdClFS.cc
+++ b/src/XrdCl/XrdClFS.cc
@@ -40,6 +40,10 @@
 #include "XrdOuc/XrdOucPrivateUtils.hh"
 #include "XrdSys/XrdSysE2T.hh"
 
+#include <algorithm>
+#include <cctype>
+#include <getopt.h>
+#include <iterator>
 #include <cmath>
 #include <cstdio>
 #include <cstdlib>
@@ -52,6 +56,107 @@
 #endif
 
 using namespace XrdCl;
+
+namespace
+{
+  enum URLCommandResult
+  {
+    NotURLCommand,
+    ValidURLCommand,
+    InvalidURLCommand
+  };
+
+  bool SupportsFullURLs( const std::string &command )
+  {
+    static const char *commands[] = {
+      "cache", "cat", "chmod", "locate", "ls", "mkdir", "mv",
+      "prepare", "rm", "rmdir", "spaceinfo", "stat", "statvfs",
+      "tail", "truncate", "xattr"
+    };
+
+    return std::find( std::begin( commands ), std::end( commands ), command )
+           != std::end( commands );
+  }
+
+  // XrdCl::URL also accepts bare hosts and local paths. Keep this small check
+  // to distinguish explicit URL operands from those shorthand forms and from
+  // values such as "name=https://example.org".
+  bool HasExplicitURLScheme( const std::string &argument )
+  {
+    std::string::size_type separator = argument.find( "://" );
+    if( separator == std::string::npos || separator == 0
+        || !std::isalpha( static_cast<unsigned char>( argument[0] ) ) )
+      return false;
+
+    for( std::string::size_type i = 1; i < separator; ++i )
+    {
+      unsigned char character = static_cast<unsigned char>( argument[i] );
+      if( !std::isalnum( character ) && character != '+'
+          && character != '-' && character != '.' )
+        return false;
+    }
+
+    return true;
+  }
+
+  URLCommandResult ParseURLCommand( FSExecutor::CommandParams &arguments,
+                                    URL &endpoint, std::string &error )
+  {
+    if( arguments.empty() )
+      return NotURLCommand;
+
+    if( arguments[0] == "query" )
+    {
+      for( std::size_t i = 1; i < arguments.size(); ++i )
+      {
+        if( HasExplicitURLScheme( arguments[i] ) )
+        {
+          error = "command-first full URLs are not supported for 'query'";
+          return InvalidURLCommand;
+        }
+      }
+      return NotURLCommand;
+    }
+
+    if( !SupportsFullURLs( arguments[0] ) )
+      return NotURLCommand;
+
+    bool foundURL = false;
+    for( std::size_t i = 1; i < arguments.size(); ++i )
+    {
+      if( !HasExplicitURLScheme( arguments[i] ) )
+        continue;
+
+      URL operand( arguments[i] );
+      if( !operand.IsValid() || operand.GetProtocol() == "file"
+          || operand.GetProtocol() == "stdio" )
+      {
+        error = "invalid remote URL operand";
+        return InvalidURLCommand;
+      }
+
+      if( !foundURL )
+      {
+        endpoint = operand;
+        endpoint.SetPath( "" );
+        endpoint.SetParams( URL::ParamsMap() );
+        foundURL = true;
+      }
+      else if( endpoint.GetProtocol() != operand.GetProtocol()
+               || endpoint.GetHostId() != operand.GetHostId() )
+      {
+        error = "all URL operands must use the same endpoint";
+        return InvalidURLCommand;
+      }
+
+      arguments[i] = operand.GetPathWithParams();
+      if( arguments[i].empty() || arguments[i][0] != '/' )
+        arguments[i].insert( arguments[i].begin(), '/' );
+    }
+
+    return foundURL ? ValidURLCommand : NotURLCommand;
+  }
+}
 
 //------------------------------------------------------------------------------
 // Build a path
@@ -1945,12 +2050,18 @@ XRootDStatus PrintHelp( FileSystem *, Env *,
                         const FSExecutor::CommandParams & )
 {
   printf( "Usage:\n"                                                          );
-  printf( "   xrdfs [--no-cwd] host[:port]              - interactive mode\n" );
-  printf( "   xrdfs            host[:port] command args - batch mode\n\n"     );
+  printf( "   xrdfs [options] host[:port]              - interactive mode\n"  );
+  printf( "   xrdfs [options] host[:port] command args - server-first batch\n" );
+  printf( "   xrdfs [options] command args-with-URLs   - command-first batch\n\n" );
+
+  printf( "The two batch forms cannot be mixed.\n\n"                          );
 
   printf( "Available options:\n\n"                                            );
 
-  printf( "   --no-cwd no CWD is being preset\n\n"                            );
+  printf( "   -d, --debug <level> set debug level: 0 off, 1 low, 2 medium,\n" );
+  printf( "                       3 high\n"                                  );
+  printf( "   -h, --help show this help\n"                                    );
+  printf( "   --no-cwd    do not preset a CWD in interactive mode\n\n"        );
 
   printf( "Available commands:\n\n"                                           );
 
@@ -2123,15 +2234,8 @@ FSExecutor *CreateExecutor( const URL &url )
 //------------------------------------------------------------------------------
 // Execute command
 //------------------------------------------------------------------------------
-int ExecuteCommand( FSExecutor *ex, int argc, char **argv )
+int ExecuteCommand( FSExecutor *ex, const FSExecutor::CommandParams &args )
 {
-  // std::vector<std::string> args (argv, argv + argc);
-  std::vector<std::string> args;
-  args.reserve(argc);
-  for (int i = 0; i < argc; ++i)
-  {
-    args.push_back(argv[i]);
-  }
   XRootDStatus st = ex->Execute( args );
   if( !st.IsOK() )
     std::cerr << st.ToStr() << std::endl;
@@ -2314,21 +2418,11 @@ int ExecuteInteractive( const URL &url, bool noCwd = false )
 //------------------------------------------------------------------------------
 // Execute command
 //------------------------------------------------------------------------------
-int ExecuteCommand( const URL &url, int argc, char **argv )
+int ExecuteCommand( const URL &url, const FSExecutor::CommandParams &args )
 {
-  //----------------------------------------------------------------------------
-  // Build the command to be executed
-  //----------------------------------------------------------------------------
-  std::string commandline;
-  for( int i = 0; i < argc; ++i )
-  {
-    commandline += argv[i];
-    commandline += " ";
-  }
-
   FSExecutor *ex = CreateExecutor( url );
   ex->GetEnv()->PutInt( "NoCWD", 1 );
-  int st = ExecuteCommand( ex, argc, argv );
+  int st = ExecuteCommand( ex, args );
   delete ex;
   return st;
 }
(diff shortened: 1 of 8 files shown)


amadio:
I'd like to settle on using [`getopt_long`](https://man7.org/linux/man-pages/man3/getopt_long.3.html) for command-line argument processing. It's part of the [Linux Standard Base](https://refspecs.linuxbase.org/LSB_5.0.0/LSB-Core-generic/LSB-Core-generic/baselib-getopt-long-3.html), and it's also available on [macOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getopt_long.3.html). It's not really necessary to create a class to parse options, also because in this case it's specialized for `xrdfs`, so we'd need to create another one for `xrdcp`, and so on. If `getopt_long` is not a good fit for `xrdfs` (and due to the sub-commands feature it may really not be), then I'd go for a generic utility that can be shared between the various commands. Have a look at how `git` does it's command parsing/executing in [git.c](https://github.com/git/git/blob/f60db8d575adb79761d363e026fb49bddf330c73/git.c#L832-L1012), and in [exec-cmd.h](https://github.com/git/git/blob/master/exec-cmd.h)/[exec-cmd.c](https://github.com/git/git/blob/master/exec-cmd.c). Something like that in C++ for XRootD's command line tools would be nice.

lobis:
@abh3 could you take a look? Thanks!

abh3 (approved):
Looks OK to me.

tests/end-to-end/xrdfs/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+set(XRDFS_BATS_ENV
+  "BATS_LIB_PATH=${PROJECT_SOURCE_DIR}/vendor/bats/lib/bats"
+  "DYLD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:$ENV{DYLD_LIBRARY_PATH}"
+  "LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:$ENV{LD_LIBRARY_PATH}"
```
amadio: These shouldn't need to be set, did it fail without?

tests/end-to-end/xrdfs/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+  "PATH=${CMAKE_BINARY_DIR}/bin:$ENV{PATH}"
```
amadio: Can this be set within the shell script rather than here? Then we can use just `xrdcp` in the tests rather than `${XRDCP}` which is uglier. I know the cluster tests use the latter, but in `tests/XRootD/*` things look much cleaner since `PATH` is set in `tests/XRootD/test.sh` to enforce that the command line tools always come from the build directory.

tests/end-to-end/xrdfs/xrdfs-full-url.bats
```diff
@@ -0,0 +1,90 @@
...
+bats_load_library 'bats-assert'
+
+load ../helper/common.bash
+
+setup() {
+    launch_xrootd xrdfs-full-url.cfg xrdfs-full-url
+    sleep 0.5
+
+    run bats_pipe -0 echo 'full URL test' \| "$XRDCP" - \
+        root://localhost:11965//examplefile
```
amadio: This is what I mean, if `PATH` is set within this script or (even better) in `common.bash`, then `"$XRDCP"` can be replaced with `xrdcp` everywhere.

amadio:
I also asked GitHub Copilot to check my review and add its own comments for things I may have missed. See below. I like the idea it suggested of adding a table with equivalent commands to the manual page. Also, I'm interested to know what happens when someone mixes "command-first" style with "host-first" style in the command line. Will things work reasonably in that case? Once these are settled, I think this is good to merge. Thanks!

---

Quick summary / recommendation
- This PR is a sensible, well-tested enhancement that makes xrdfs accept command-first invocations with full URLs. Tests and docs were added. Risk is medium because it changes CLI parsing and URL handling; please address the items below (documentation clarifications and a few edge-cases/portability/security checks) before merging.

What I did
- I loaded the PR reviewer guidance, read the PR description and the provided manpage diff, and inspected the changes described in the commit message and tests summary. Below I summarize the changes, call out potential problems I think were not addressed in the PR text, and list concrete review points and suggestions.

Concrete review points to address in the PR (these map to the docs/man/xrdfs.1 changes and should be checked in code/docs)


- docs/man/xrdfs.1 — Clarify that the new synopsis supports both global options and command-first URL operands (the current new synopsis line "xrdfs command [args containing complete URLs]" could be interpreted as the only supported form). Suggest rewording to explicitly show both supported invocations (e.g., show both `xrdfs [global-options] host[:port] command [args]` and `xrdfs [global-options] command [args containing complete URLs]`). This avoids confusing readers who use the old server-first form.
- docs/man/xrdfs.1 — The commands list now includes `cache` and `xattr`. Verify the implementation adds handlers (or forwards) for `cache` and `xattr` in the command-first path; if not, either remove from the man page or add the implementation. If these are forwarded to existing code paths, ensure their behavior is covered by tests.
- docs/man/xrdfs.1 — The examples use double slashes (root://root.example.org//store/file). Add a short note clarifying the semantics of the double slash versus single slash (i.e., the meaning of the authority + absolute path separator in xrootd URLs) and whether normalization removes/rewrites extra slashes. Make sure examples exactly reflect the canonical form your code accepts and prints in error messages.

Other important issues and edge cases I think the PR should explicitly handle (not all are necessarily addressed in tests)

- Mixed server-first and command-first operands: how does the program behave if the user supplies both a leading endpoint argument and later full-URL operands (e.g., `xrdfs host stat root://host//path`)? Either document the precedence or reject the mixture with a clear error. Current docs imply "endpoint may either be provided separately before the command or inferred from complete URL path operands", but the behavior when both are present should be explicit.
- Mixed local and remote operands: the PR text says local `file`/`stdio` URLs are rejected. Consider whether rejecting `file://` is desirable (some users may legitimately want to operate on local files); if rejection is intentional, add an explanatory note to docs and a clear error message suggesting the server-first form or a different tool for local paths.
- URL scheme support and plugin selection: confirm how the code selects protocol/transport when multiple schemes are possible (root://, http://, https://). The PR mentions preserving CGI params and that XrdClHttp remains the configured plugin — make sure the code still properly chooses the transport plugin based on scheme and that scheme normalization doesn't accidentally treat http(s) URLs as xrootd ones.
- IPv6 and host-literal handling: ensure URL parsing accepts IPv6 literal addresses (e.g., root://[2001:db8::1]/...) including userinfo and port parts. Add tests that exercise IPv6 addresses in command-first and server-first forms.
- URL-encoding/percent-decoding: confirm that percent-encoded path components and query components are preserved correctly and not double-decoded or accidentally normalized in a way that changes meaning. Add tests for operands with spaces and percent-encoded characters.
- Fragment handling (#): HTTP/URL fragments are client-side only and should not be sent to the server; document whether the fragment is preserved, stripped, or an error in xrdfs URL operands. Tests should ensure fragments don't get forwarded incorrectly.
- Credential precedence and security: the PR preserves credentials in URLs and rejects mixed credentials. Two items:
  - Document the security implications of passing credentials on the command line (they are visible to other users via process listings). Consider documenting a safer alternative (e.g., environment variable or config file) or at least add a warning in docs/manpage.
  - Verify precedence: if credentials are provided both as the leading host[:port] argument (server-first) and within a URL operand, clarify which wins, or reject the mix.
- Port defaults and authority matching: when comparing endpoints for "must refer to same endpoint", be precise about what equality means: is root://host and root://host:1094 same endpoint (port default)? Be sure the implementation canonicalizes host+port (including default port) and handles case-insensitive scheme/host comparisons, and document the rule.
- getopt_long portability: getopt_long is widely available but is not strictly POSIX; confirm the repository expects this dependency and that it builds on supported platforms (e.g., portability to non-glibc systems). If not already, add a configure-time check and a fallback or document the min required toolchain.
- Subcommand flags that look like global options: stopping parsing at the first non-option is correct, but ensure cases where a subcommand expects an optional argument that starts with a hyphen (e.g., a negative number) are still parsed as subcommand args. Tests should include such edge cases.
- Error messages and exit codes: add tests (or confirm existing tests) ensure clear, user-friendly error messages for the common failure modes: mixed endpoints, mixed credentials, invalid URL, local file URLs rejected, and ambiguous operand forms. Messages should mention suggested remedies (e.g., split endpoint and path, use server-first form).
- Logging and debugging: consider adding a verbose/debug mode flag (or reusing an existing one) that prints the normalized endpoint and operands so users can understand how their input was interpreted, especially useful during migration.

Possible improvements (concrete suggestions)

- docs/man/xrdfs.1 — Example tweak: add a small table or side-by-side example showing equivalent server-first and command-first invocations and what the normalized internal endpoint + path will be for each. This will reduce confusion for existing users migrating scripts.
- tests — Add explicit tests for: IPv6 brackets, percent-encoded paths, presence of fragments (#), mixing server-first + command-first (should be rejected or have a well-documented precedence), and command-line credentials security note (a test can't assert a doc note, so update docs).
- CLI UX — If not already present, consider an explicit and consistent error code for "mixed endpoints" vs "invalid URL" vs "local URL rejected" so scripts can detect the failure mode reliably.

Minor doc/text nits in docs/man/xrdfs.1
- The new synopsis line "xrdfs command [args containing complete URLs]" could be misread as the only form: make docs show both forms clearly.
- The sentence "All complete URLs in one command must refer to the same endpoint." is good but should state whether endpoint equality compares scheme, userinfo, host, and port, and whether it canonicalizes default ports.

---

lobis:
> I also asked GitHub Copilot to check my review and add its own comments for things I may have missed. See below. I like the idea it suggested of adding a table with equivalent commands to the manual page. Also, I'm interested to know what happens when someone mixes "command-first" style with "host-first" style in the command line. Will things work reasonably in that case? Once these are settled, I think this is good to merge. Thanks!
> 
> Quick summary / recommendation
> 
> * This PR is a sensible, well-tested enhancement that makes xrdfs accept command-first invocations with full URLs. Tests and docs were added. Risk is medium because it changes CLI parsing and URL handling; please address the items below (documentation clarifications and a few edge-cases/portability/security checks) before merging.
> 
> What I did
> 
> * I loaded the PR reviewer guidance, read the PR description and the provided manpage diff, and inspected the changes described in the commit message and tests summary. Below I summarize the changes, call out potential problems I think were not addressed in the PR text, and list concrete review points and suggestions.
> 
> Concrete review points to address in the PR (these map to the docs/man/xrdfs.1 changes and should be checked in code/docs)
> 
> * docs/man/xrdfs.1 — Clarify that the new synopsis supports both global options and command-first URL operands (the current new synopsis line "xrdfs command [args containing complete URLs]" could be interpreted as the only supported form). Suggest rewording to explicitly show both supported invocations (e.g., show both `xrdfs [global-options] host[:port] command [args]` and `xrdfs [global-options] command [args containing complete URLs]`). This avoids confusing readers who use the old server-first form.
> * docs/man/xrdfs.1 — The commands list now includes `cache` and `xattr`. Verify the implementation adds handlers (or forwards) for `cache` and `xattr` in the command-first path; if not, either remove from the man page or add the implementation. If these are forwarded to existing code paths, ensure their behavior is covered by tests.
> * docs/man/xrdfs.1 — The examples use double slashes (root://root.example.org//store/file). Add a short note clarifying the semantics of the double slash versus single slash (i.e., the meaning of the authority + absolute path separator in xrootd URLs) and whether normalization removes/rewrites extra slashes. Make sure examples exactly reflect the canonical form your code accepts and prints in error messages.
> 
> Other important issues and edge cases I think the PR should explicitly handle (not all are necessarily addressed in tests)
> 
> * Mixed server-first and command-first operands: how does the program behave if the user supplies both a leading endpoint argument and later full-URL operands (e.g., `xrdfs host stat root://host//path`)? Either document the precedence or reject the mixture with a clear error. Current docs imply "endpoint may either be provided separately before the command or inferred from complete URL path operands", but the behavior when both are present should be explicit.
> * Mixed local and remote operands: the PR text says local `file`/`stdio` URLs are rejected. Consider whether rejecting `file://` is desirable (some users may legitimately want to operate on local files); if rejection is intentional, add an explanatory note to docs and a clear error message suggesting the server-first form or a different tool for local paths.
> * URL scheme support and plugin selection: confirm how the code selects protocol/transport when multiple schemes are possible (root://, http://, https://). The PR mentions preserving CGI params and that XrdClHttp remains the configured plugin — make sure the code still properly chooses the transport plugin based on scheme and that scheme normalization doesn't accidentally treat http(s) URLs as xrootd ones.
> * IPv6 and host-literal handling: ensure URL parsing accepts IPv6 literal addresses (e.g., root://[2001:db8::1]/...) including userinfo and port parts. Add tests that exercise IPv6 addresses in command-first and server-first forms.
> * URL-encoding/percent-decoding: confirm that percent-encoded path components and query components are preserved correctly and not double-decoded or accidentally normalized in a way that changes meaning. Add tests for operands with spaces and percent-encoded characters.
> * Fragment handling (#): HTTP/URL fragments are client-side only and should not be sent to the server; document whether the fragment is preserved, stripped, or an error in xrdfs URL operands. Tests should ensure fragments don't get forwarded incorrectly.
> * Credential precedence and security: the PR preserves credentials in URLs and rejects mixed credentials. Two items:
>   
>   * Document the security implications of passing credentials on the command line (they are visible to other users via process listings). Consider documenting a safer alternative (e.g., environment variable or config file) or at least add a warning in docs/manpage.
>   * Verify precedence: if credentials are provided both as the leading host[:port] argument (server-first) and within a URL operand, clarify which wins, or reject the mix.
> * Port defaults and authority matching: when comparing endpoints for "must refer to same endpoint", be precise about what equality means: is root://host and root://host:1094 same endpoint (port default)? Be sure the implementation canonicalizes host+port (including default port) and handles case-insensitive scheme/host comparisons, and document the rule.
> * getopt_long portability: getopt_long is widely available but is not strictly POSIX; confirm the repository expects this dependency and that it builds on supported platforms (e.g., portability to non-glibc systems). If not already, add a configure-time check and a fallback or document the min required toolchain.
> * Subcommand flags that look like global options: stopping parsing at the first non-option is correct, but ensure cases where a subcommand expects an optional argument that starts with a hyphen (e.g., a negative number) are still parsed as subcommand args. Tests should include such edge cases.
> * Error messages and exit codes: add tests (or confirm existing tests) ensure clear, user-friendly error messages for the common failure modes: mixed endpoints, mixed credentials, invalid URL, local file URLs rejected, and ambiguous operand forms. Messages should mention suggested remedies (e.g., split endpoint and path, use server-first form).
> * Logging and debugging: consider adding a verbose/debug mode flag (or reusing an existing one) that prints the normalized endpoint and operands so users can understand how their input was interpreted, especially useful during migration.
> 
> Possible improvements (concrete suggestions)
> 
> * docs/man/xrdfs.1 — Example tweak: add a small table or side-by-side example showing equivalent server-first and command-first invocations and what the normalized internal endpoint + path will be for each. This will reduce confusion for existing users migrating scripts.
> * tests — Add explicit tests for: IPv6 brackets, percent-encoded paths, presence of fragments (#), mixing server-first + command-first (should be rejected or have a well-documented precedence), and command-line credentials security note (a test can't assert a doc note, so update docs).
> * CLI UX — If not already present, consider an explicit and consistent error code for "mixed endpoints" vs "invalid URL" vs "local URL rejected" so scripts can detect the failure mode reliably.
> 
> Minor doc/text nits in docs/man/xrdfs.1
> 
> * The new synopsis line "xrdfs command [args containing complete URLs]" could be misread as the only form: make docs show both forms clearly.
> * The sentence "All complete URLs in one command must refer to the same endpoint." is good but should state whether endpoint equality compares scheme, userinfo, host, and port, and whether it canonicalizes default ports.

Addressed these comments in 881e3a30712b2fcbdda2e335038d78f020b04f14 (will squash so this PR becomes 1 single commit but I leave it as a distinct commit to facilitate review):

Updated the URL handling to rely more directly on XrdCl::URL for parsing, endpoint canonicalization, credentials, effective ports, paths, and CGI parameters. I kept a small HasExplicitURLScheme() check because XrdCl::URL intentionally also accepts bare hosts and local paths; the CLI still needs to distinguish explicitly written URL operands from those shorthand forms and from values such as name=https://example.org.
Added the side-by-side server-first/command-first table to the manual page.
Clarified the synopsis, global-option placement, endpoint equality, default-port handling, the XRootD double-slash convention, rejection of local URLs, and why query remains server-first.

Explicitly reject mixed invocation styles rather than assigning precedence. For example:

xrdfs root://host stat root://host//path

now fails with a clear message asking the user to choose either server-first or command-first syntax.

Added BATS coverage for mixed-style rejection and for URL-looking parameters in server-first query commands.
Endpoint comparison now uses the protocol and GetHostId(), so an omitted default port compares equal to the same port specified explicitly.
cache and xattr continue to dispatch through their existing FSExecutor handlers; command-first xattr behavior is covered by the existing end-to-end test.

amadio:
@lobis Sorry, I thought I submitted all in one go, but some comments were still left pending. Could you please take a look?

lobis:
> @lobis Sorry, I thought I submitted all in one go, but some comments were still left pending. Could you please take a look?

Should be good to review now!

src/XrdCl/XrdClFS.cc
```diff
@@ -53,6 +57,107 @@
...
+      for( std::size_t i = 1; i < arguments.size(); ++i )
+      {
+        if( HasExplicitURLScheme( arguments[i] ) )
+        {
+          error = "command-first full URLs are not supported for 'query'";
+          return InvalidURLCommand;
+        }
+      }
+      return NotURLCommand;
+    }
```
amadio, line 119: It seems this bit needs a couple of extra tests. See coverage details here: https://my.cdash.org/builds/3948142/coverage/29693945

src/XrdCl/XrdClFS.cc
```diff
@@ -2338,40 +2430,84 @@ int ExecuteCommand( const URL &url, int argc, char **argv )
...
-  // Check the commandline parameters
-  //----------------------------------------------------------------------------
   XrdCl::FSExecutor::CommandParams params;
-  if( argc == 1 )
+  enum { NoCwdOption = 256 };
+  static const option options[] = {
+    { "help",   no_argument, 0, 'h' },
+    { "no-cwd", no_argument, 0, NoCwdOption },
+    { 0, 0, 0, 0 }
+  };
```
amadio, line 2446: I think it would be good to use this PR as an opportunity to resolve also #2105, as well as adding some extra options from `xrdcp` to `xrdfs`, namely, a `-d` option that allows us to set the log level to not need to use `env XRD_LOGLEVEL=Debug` all the time.

tests/end-to-end/xrdfs/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(XRDFS_BATS_ENV
+  "BATS_LIB_PATH=${PROJECT_SOURCE_DIR}/vendor/bats/lib/bats"
```
amadio: I think this should not be necessary, it should be done by default in some common place rather than in each test using BATS.

tests/end-to-end/xrdfs/xrdfs-full-url.bats
```diff
@@ -0,0 +1,104 @@
...
+@test "server-first and command-first syntax cannot be mixed" {
+    run xrdfs root://localhost:11965 stat \
+        root://localhost:11965//examplefile
+    assert_failure 1
+    assert_output --partial 'cannot mix a leading endpoint with full URL operands'
+}
+
+@test "command-first syntax accepts a full URL" {
+    run xrdfs stat root://localhost:11965//examplefile
+    assert_success
```
amadio: Rather than do as above, you can do just
```
run -0 xrdfs stat root://localhost:11965//examplefile
```

tests/end-to-end/xrdfs/xrdfs-full-url.bats
```diff
@@ -0,0 +1,104 @@
...
+
+@test "legacy server-first syntax remains supported" {
+    run xrdfs root://localhost:11965 stat //examplefile
+    assert_success
+}
+
+@test "server-first and command-first syntax cannot be mixed" {
+    run xrdfs root://localhost:11965 stat \
+        root://localhost:11965//examplefile
+    assert_failure 1
```
amadio: Here you can use `run -1 xrdfs ...`

src/XrdCl/XrdClFS.cc
lobis: I will bring coverage of `src/XrdCl/XrdClFS.cc` close to 100% in https://github.com/xrootd/xrootd/pull/2836

amadio (approved):

PR #2867, opened 2026-07-13
RAL XrdCeph changes

Jo-stfc:
branch with changes following review of https://github.com/xrootd/xrootd/pull/2840. 
XrdCeph level streaming checksum has been dropped due to being supported in the upstream implementation.
Checksum clearing has also been removed due to streamed checksums making this change redundant. 

Current feature list:
Write-only buffers
Erasure coding factor for space reporting


src/XrdCeph/XrdCephOss.cc
```diff
@@ -115,7 +115,7 @@ ssize_t getNumericAttr(const char* const path, const char* attrName, const int m
   if (attrLen <= 0) {
     retval = -EINVAL;
   } else {
-    attrValue[attrLen] = '\0';
+    attrValue[attrLen] = (char)'\0';
```
amadio: Please make sure to address comments from previous reviews...

Image

Jo-stfc: sorry, I'm confused on that one - it's already using the NUL character? or is there a seperate syntax for C++?

amadio: Nevermind, I just wonder why a C-style cast is needed there. I guess it's a `signed char` vs `unsigned char` problem, or `char` vs `int` problem. What platform gives you a warning here? Can you add it to the commit message?

Jo-stfc: that doesn't compile 
```/root/rpmbuild/BUILD/xro
otd/src/XrdCeph/XrdCephOss.cc:118:26: error: 'NUL' was not declared in this scope; did you mean 'NULL'?
  118 |     attrValue[attrLen] = NUL;
      |                          ^~~
      |                          NULL
```
possibly it needs a separate library? I don't see NUL used anywhere else in the codebase

amadio: No, I was wrong, it's a C thing, not C++... But the problem is not the syntax, but why a cast is needed to assign a `char` to another `char`...

Jo-stfc: because of strict sign conversion - '\0' is a char, the array element is an unsigned char
(or at least that's AI's explanation)

Jo-stfc: so on strict checking the compiler throws a warning for it

amadio: Yes, but the array is declared `char`, not `unsigned char`. Therefore my question.

Jo-stfc: I tried recompiling it without the char casting and the error went away... that's weird

amadio:
@Jo-stfc Please, can you make this pass the CI checks? When it's green I can review again. Thanks.

Jo-stfc:
@amadio tests pass now. Apologies for the delay.

PR #2870, opened 2026-07-15
[Tests] BATS integration into CMake/CTest

diogomattioli:
This PR integrates BATS (Bash Automated Testing System) with the CMake/CTest test framework, enabling shell-based tests to be built into the standard testing workflow. The integration allows BATS tests to be executed through `ctest`, providing a unified interface for running tests locally and in CI while improving consistency across the project's test infrastructure.


tests/end-to-end/helper/bats-format-cdash
```diff
@@ -0,0 +1,252 @@
...
+#
+# Usage:
+#     BATS_XML_DIR=.../build/Testing \
+#       bats --recursive --timing \
+#            --formatter /abs/path/to/bats-format-cdash tests/end-to-end
+#
+# The formatter protocol (bats_tap_stream_* callbacks + the parser) comes from
+# Bats' own formatter library, exactly like the bundled tap/junit formatters.
+
+trap '' INT
```
amadio, line 10: Don't we maybe need a cleanup function here? See what we do below in `test.sh` to prevent stuff from being left behind:

https://github.com/xrootd/xrootd/blob/64e6707db2d3f79f3b3190191f63f41657473801/tests/XRootD/test.sh#L186-L187

tests/end-to-end/helper/bats-format-cdash
```diff
@@ -0,0 +1,252 @@
...
+end_epoch=$(date +%s)
+end_human=$(date '+%b %d %H:%M %Z')
+elapsed_minutes=$(( (end_epoch - start_epoch) / 60 ))
+
+site_name=$(xml_escape "${BATS_XML_SITE:-$(hostname)}")
+build_name=$(xml_escape "${BATS_XML_BUILDNAME}")
+build_stamp=$(xml_escape "${BATS_XML_BUILDSTAMP}")
+
+output_file=$(resolve_output_file)
+mkdir -p "$(dirname "$output_file")" 2>/dev/null || true
```
amadio: I think it might be better to just post one output per bats file to CDash. This adds a lot of complexity. If later we want better reports, we can try to integrate the junit outputs into a summary, like what [ROOT does](https://github.com/root-project/root/pull/20272#issuecomment-3480450461).

tests/end-to-end/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+#     Each per-file BATS run produces its own numbered BatsTest<N>.xml.
+#   * otherwise: the default (tap) formatter, for readable local output.
+#-------------------------------------------------------------------------------
+
+find_program(BASH_PROGRAM bash)
+
+if(NOT BASH_PROGRAM)
+  message(WARNING "bash not found: end-to-end (BATS) tests are disabled")
+  return()
+endif()
```
amadio, line 6: Does `bats` really need `bash` or could it maybe run with plain `sh` (in which case we can assume it to be already available)?

tests/end-to-end/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+# per file (in parallel with the rest, like every other CTest test) and reports
+# each file individually. The @test cases inside a file are deliberately NOT
+# registered with CTest. Which BATS formatter is used depends on whether this
+# build submits to CDash:
+#
+#   * CDash builds (-DCDASH=ON, set by test.cmake): bats-format-cdash writes
+#     every test case into Testing/BatsTest<N>.xml, in CTest's own Test.xml
+#     format, which test.cmake then submits via ctest_submit alongside Test.xml.
+#     Each per-file BATS run produces its own numbered BatsTest<N>.xml.
+#   * otherwise: the default (tap) formatter, for readable local output.
```
amadio: We may be able to call bats with junit output from CTest to integrate it instead of doing it manually. Something we can consider as alternative to just not integrating anything.

tests/end-to-end/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+  "PATH=${CMAKE_BINARY_DIR}/bin:$ENV{PATH}"
+  "BATS_XML_SOURCE_ROOT=${PROJECT_SOURCE_DIR}"
+  "BATS_XML_TEST_ROOT=${CMAKE_CURRENT_SOURCE_DIR}")
+
+# Discover every .bats file and register each as its own CTest test, so BATS is
+# run once per file instead of once recursively over the whole tree.
+# CONFIGURE_DEPENDS re-globs at build time so newly added .bats files are picked
+# up without a manual re-run of CMake.
+file(GLOB_RECURSE BATS_FILES CONFIGURE_DEPENDS
+  RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/*.bats)
```
amadio: Like with C++ headers/sources, I'd prefer to list everything explicitly to avoid the case where a git add is forgotten on a bats file causing the tests to silently not being run.

Makefile
```diff
@@ -0,0 +1,36 @@
+# Convenience wrapper around the CMake build so the whole test suite — the
+# GoogleTest unit/integration tests and the BATS end-to-end tests — can be
+# built and run with a single command:
+#
+#     make test
```
amadio: This is what `ctest -V -S test.cmake` is for. Once you have `build/` directory inside the repository, you can `cd build/ && make test` already to run all the tests, so I don't understand why you need this `Makefile`. If it's needed/wanted, I'd add to the top directory, though.

tests/end-to-end/CMakeLists.txt
diogomattioli: The suggested approach will basically mix 2 different unrelated frameworks and bloat the test suite. This would literally require a full structure of `CMakeLists.txt` files on each folder that have no real relationship between `BATS -> CMAKE`.

BATS is a framework for executing command line programs, it can be executed independently (`./vendor/bats/bin/bats`) and has no dependency on CMake. In another words, even if a CMakeLists.txt is introduced, `.bats` files can be locally ran with `bats` command only.

For the scenario mentioned above, most likely `git status` will show that files were not added/tracked. 
On the other hand, what would happen if implementing the suggestion above, adding/tracking the file and forgetting to mention it on `CMakeLists.txt`? It wouldn't run anyway. 
And, as it can be ran manually with `bats` command, one can run it locally, forget to `git add` and forget to add on `CMakeLists.txt` as well.

Makefile
diogomattioli: I'll get rid of it. Misunderstood the requirement of a `make test` from project root.

tests/end-to-end/CMakeLists.txt
diogomattioli: Yes, it does need. `BATS` is `Bash Automated Testing System` and it uses `bash` to work.

https://github.com/xrootd/xrootd/blob/master/vendor/bats/bin/bats#L1

tests/end-to-end/helper/bats-format-cdash
diogomattioli: Each run of BATS produces one different output file.

There are 2 options here:
1 - Run `bats` only once passing `--recursive` and having all the `.bats` files ran in the same execution.
2 - Manipulate the output xml file to on each run merge the multiple generate outputs.

tests/end-to-end/CMakeLists.txt
amadio: I think you misunderstood what I said. I just want the `*.bats` to be listed explicitly instead of detected by `file(GLOB_RECURSE ...)`. The case I'm more concerned with is when the `.bats` file itself (or an update to it) is forgotten, in which case the everyone else will be missing the `.bats` file, and the CI will not detect the problem. If the files are listed explicitly in this `CMakeLists.txt`, it will fail in CI in that case. Dependency resolution can also break when inputs are detected with `file(GLOB_RECURSE ...)`, so changes to the `.bats` may not cause dependent actions to re-run, and so on.

tests/end-to-end/helper/bats-format-cdash
amadio: Let me rephrase it. I think that if it's possible to just make BATS output something in stdout that CDash understands, we can avoid posting all these XML files by hand as a manual step at the end. In that case, we'd just let the output in CDash be the stdout of BATS, like with the other tests.

tests/end-to-end/CMakeLists.txt
diogomattioli: I'm looking into that!

tests/end-to-end/CMakeLists.txt
diogomattioli: I agree that the dependency resolution won't work with `GLOB_RECURSE`, however we're trying to lower the barrier for creating tests.

Again, `BATS` and `CMake` serve for different purposes. Using `CMake` as a trigger or just having a single command line is fine. The problem is creating `CMake` files in each and every folder. Soon, the testing structure will be similar to what it is now.

I think perhaps it'd worth while some discussion and brainstorming about it.

tests/end-to-end/CMakeLists.txt
diogomattioli: Unfortunately `CDash` doesn't support `JUnit` format.

tests/end-to-end/helper/bats-format-cdash
diogomattioli: We can do that, but it will produce so many lines of log that on `CDash` one will be scrolling between 20k lines of logs every time a test fails.

I just refactored the formatter, for now it's separating `BATS` tests individually with their individual logs for each tests. All of that together with `GTest` unit tests.

Take a look at it, if it's not good we go for the plain output!

amadio:
Clicked to rebase on the current master to pick up the change that makes things more resilient against CDash post failures.

PR #2872, opened 2026-07-17
[XrdClHttp] Parse multipart headers split across chunks

### LINKED ISSUE #2858: XrdClHttp: errors in complex vectored reads

dynamic-entropy:
Fixes #2858 

@ffurano suggests that we try this fix in root.
Many thanks to fabrizio for explaning me the details of Range requests and multipart responses.

As fabrizio explained to me, 
the test in the PR incorrectly (unlike a valid response) starts the first chunk with a multipart boundary identifier `"\r\n--123456\r\n"`. Our code is still able to parse the response.

I intentionally divide the header in three parts instead of two, as originally reported, to have a much wider test case.

### DIFF
--- a/src/XrdClHttp/XrdClHttpOpReadV.cc
+++ b/src/XrdClHttp/XrdClHttpOpReadV.cc
@@ -177,6 +177,7 @@ CurlVectorReadOp::Write(char *orig_buffer, size_t orig_length)
                     m_response_idx++;
                     if (m_current_op.second == chunk.GetLength()) {
                         m_current_op.first = m_current_op.second = -1;
+                        m_multipart_boundary = true;
                     } else {
                         // We may need to skip the remaining bytes or, potentially, the server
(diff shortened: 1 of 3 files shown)


ffurano:
I think that this solution is simpler and more robust than the one that was proposed in #2861, so I vote for this one that is able to handle desperate cases as well. Good job!

ffurano:
Who can merge this?

abh3 (commented):
I will defer this to Fabrizio who is more in tune with this code than I.

PR #2875, opened 2026-07-23
[XrdHttpTPC] Share one parsed CA/CRL store across transfers

### LINKED ISSUE #2873: [XrdHttpTPC] High memory usage due to the CA/CRL parsing for each transfer
### Operating System

Alma Linux 9

### XRootD Version

6.1.0

### Summary

Every HTTP-TPC COPY creates a fresh libcurl easy handle, and `TPCHandler::ConfigureCurlCA()`
passes the CA bundle and CRL bundle **paths** to it (`CURLOPT_CAINFO` / `CURLOPT_CRLFILE`).
With OpenSSL 3, each handle therefore parses the entire bundle into its own private
`X509_STORE` at handshake time. With a typical grid CA directory (IGTF CAs + CRLs) this
costs **~50-60 MB of live heap per active transfer**, held for the whole transfer duration.

Nothing is shared between transfers, so server memory scales as
`concurrent_transfers × parsed_bundle_size`. On a storage server where transfers became slow
(and therefore piled up), this took the process to >200 GB RSS and the xrootd process was OOM-killed.

### Evidence (jemalloc heap profile from an affected server)

`jeprof --text --base=<start>.heap /opt/eos/xrootd/bin/xrootd <later>.heap`:

```
Total: 208931.5 MB
194857.3  93.3%  93.3% 194857.3  93.3% CRYPTO_malloc@@OPENSSL_3.0.0
  7123.2   3.4%  96.7%  24118.3  11.5% OSSL_STORE_close@@OPENSSL_3.0.0
  3465.3   1.7%  98.3%   3465.3   1.7% XrdBuffManager::Obtain
  2761.8   1.3%  99.7%   2761.8   1.3% std::vector::_M_default_append
  ...
    67.3   0.0%  99.9% 205102.6  98.2% curl_getdate            <- symbol artifact, static libcurl fn
     2.0   0.0% 100.0% 205379.2  98.3% TPC::TPCHandler::ProcessReq
```

- 98.3% of all live allocations are under `TPC::TPCHandler::ProcessReq`, and 93.3% of the
  total is flat `CRYPTO_malloc` (OpenSSL). The `OSSL_STORE_*` frames identify it as
  certificate-store loading.
- Cross-check on concurrency: `XrdBuffManager::Obtain` = 3465 MB ÷ 1 MiB per-connection
  XrdHttp buffer ⇒ ~3400 live connections at dump time.
- Per-transfer cost: 194 857 MB ÷ ~3400 ≈ **56 MB of parsed CA+CRL store per transfer**.
- At the OOM the process had ~236 GB anonymous RSS with 4000-5000 threads (one worker
  thread per in-flight TPC transfer).

### Root cause

- One curl easy handle per COPY: `TPCHandler::ProcessPullReq` / `ProcessPushReq`
  (`src/XrdHttpTpc/XrdHttpTpcTPC.cc`, `curl_easy_init()` per request).
- `TPCHandler::ConfigureCurlCA()` (`src/XrdHttpTpc/XrdHttpTpcTPC.cc:283-311`) sets
  `CURLOPT_CAINFO` to the `XrdTlsTempCA`-generated CA bundle and `CURLOPT_CRLFILE` to the
  CRL bundle, per handle.
- libcurl's OpenSSL backend loads/parses both into a per-`SSL_CTX` `X509_STORE` on every
  transfer. Parsed `X509`/`X509_CRL` objects are several times larger than the PEM input;
  grid CRL bundles are tens of MB of PEM.
- The parsed store lives as long as the transfer. Slow transfers ⇒ high concurrency ⇒
  memory scales without bound. (curl's CA-store cache, `CURLOPT_CA_CACHE_TIMEOUT`, does not
  help here because every transfer uses a fresh easy/multi handle.)

### Proposed fix: share one parsed X509_STORE across all transfers

`X509_STORE` is refcounted (`X509_STORE_up_ref`) and internally locked for concurrent
verification — it is designed to be shared. The plumbing for injection already exists:
`TPCHandler::ssl_ctx_callback` (`src/XrdHttpTpc/XrdHttpTpcTPC.cc:189-194`) is already
installed via `CURLOPT_SSL_CTX_FUNCTION` in some configurations.

Sketch:

1. Build the store **once per bundle rotation**, ideally inside `XrdTlsTempCA` (its
   maintenance thread already knows exactly when the CA/CRL files are regenerated):
   `X509_STORE_new()` + `X509_STORE_load_locations()` for the CAs, `X509_STORE_add_crl()`
   for the CRL bundle, `X509_STORE_set_flags(store, X509_V_FLAG_CRL_CHECK |
   X509_V_FLAG_CRL_CHECK_ALL)` to preserve current `CURLOPT_CRLFILE` semantics. Publish as
   a refcounted handle; in-flight transfers keep the previous store alive until their last
   `SSL_CTX` is destroyed, so rotation needs no synchronization with active handshakes.
2. In `ConfigureCurlCA()`, stop setting `CURLOPT_CAINFO`/`CURLOPT_CRLFILE`; always install
   `CURLOPT_SSL_CTX_FUNCTION` (+ `CURLOPT_SSL_CTX_DATA`), and in the callback:
   `SSL_CTX_set1_cert_store(ctx, shared_store)` — a refcount bump instead of a
   ~50 MB parse. Keep the existing `SSL_CTX_set_verify(..., verify_callback)` behavior for
   the missing-CRL tolerance path.

Expected effect: per-transfer TLS memory drops from ~56 MB to KBs; in the incident above
that is ~195 GB → <1 GB for the same workload.

On a server configured with a grid CA dir (CAs + CRLs), run N concurrent slow HTTP-TPC
pulls (e.g. rate-limit the sources). RSS grows by ≈ N × parsed-bundle-size and is released
only as transfers finish. A jemalloc/heap profile shows the allocations under
`ProcessReq → libcurl → OSSL_STORE/CRYPTO_malloc`.

### Additional Information

_No response_

ccaffy:
Every HTTP-TPC COPY was given the CA and CRL bundle paths through CURLOPT_CAINFO and CURLOPT_CRLFILE, so libcurl parsed both into a private X509_STORE for each connection and held it for the whole transfer. With a grid CA directory that costs tens of MB per in-flight transfer, and multi-stream pulls pay it once per stream. Memory therefore scaled with concurrency: on a busy server where thousands of transfers piled up this reached over 200 GB RSS and the process was OOM-killed.

libcurl's own CA store cache cannot help here, as it is keyed on the multi handle and every transfer creates a fresh one.

XrdTlsTempCA now parses the bundles into a single X509_STORE once per maintenance cycle and hands out reference counted copies. Transfers install it with SSL_CTX_set1_cert_store() from an SSL_CTX callback, so setting up a transfer costs a reference count instead of a full parse. In-flight transfers keep the previous store alive, so a reload never disturbs a running handshake.

The store is built with the same verification flags libcurl applies to these files, so chain building and CRL checking are unchanged: X509_V_FLAG_CRL_CHECK and X509_V_FLAG_CRL_CHECK_ALL when the CRL bundle holds at least one CRL, and X509_V_FLAG_PARTIAL_CHAIN when it does not.

A build of libcurl whose TLS backend does not implement CURLOPT_SSL_CTX_FUNCTION keeps the previous behaviour; this is detected once at configuration time and reported. A failure to build the store is instead treated as fatal, since silently letting each transfer load the bundles for itself would reintroduce the memory growth as soon as a reload went wrong.

Serialising the XrdTlsTempCA accessors also removes an unsynchronised read of the bundle filenames, which the maintenance thread can replace concurrently.

Fixes: #2873

Assisted-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

Memory usage before the fix (100 transfers in parallel with the ca-policy-egi-core CAs installed:
image

Memory usage after the fix:
image

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
@@ -185,10 +185,22 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
  * https://curl.se/libcurl/c/CURLOPT_SSL_CTX_FUNCTION.html
  */
 int TPCHandler::ssl_ctx_callback(CURL *curl, void *ssl_ctx, void *clientp) {
-    //TPCLogRecord * rec = (TPCLogRecord *)clientp;
+    TPCLogRecord * rec = (TPCLogRecord *)clientp;
     SSL_CTX* ctx = static_cast<SSL_CTX*>(ssl_ctx);
-    SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, verify_callback);
-    return CURL_SOCKOPT_OK;
+
+    if (rec && rec->ca_store) {
+        // Bumps the store's reference count instead of re-parsing the CA and CRL
+        // bundles for this connection.  libcurl runs this callback after it has
+        // applied its own TLS options, so this replaces whatever store it built.
+        SSL_CTX_set1_cert_store(ctx, rec->ca_store.get());
+    }
+    if (allowMissingCRL) {
+        // verify_callback only excuses X509_V_ERR_UNABLE_TO_GET_CRL, i.e. a CA in
+        // the chain for which no CRL could be found.  Every other verification rule
+        // still applies, including revocation itself whenever a CRL is present.
+        SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, verify_callback);
+    }
+    return CURLE_OK;
 }
 
 int TPCHandler::verify_callback(int preverify_ok, X509_STORE_CTX* ctx) {
@@ -252,9 +264,33 @@ std::string encode_xrootd_opaque_to_uri(CURL *curl, const std::string &opaque)
 /*           T P C H a n d l e r : : C o n f i g u r e C u r l C A            */
 /******************************************************************************/
   
-void
-TPCHandler::ConfigureCurlCA(CURL *curl)
+bool
+TPCHandler::ConfigureCurlCA(CURL *curl, TPCLogRecord &rec)
 {
+    // Preferred path: hand libcurl the CA/CRL store that XrdTlsTempCA already
+    // parsed, rather than the bundle filenames.  Passing filenames makes libcurl
+    // build a private X509_STORE per connection, which costs tens of MB for a grid
+    // CA directory and is held for the whole transfer; sharing one store makes that
+    // a reference count.  See https://github.com/xrootd/xrootd/issues/2873
+    //
+    // Skipped when m_cafile is set, so that the http.cafile precedence established
+    // at the bottom of this function is preserved.
+    if (m_ca_file && m_sslctx_supported && m_cafile.empty()) {
+        rec.ca_store = m_ca_file->CAStore();
+        if (!rec.ca_store) {
+            m_log.Log(Error, "TpcHandler", "No CA store is available; refusing to "
+                      "fall back to libcurl's default CA bundle");
+            return false;
+        }
+        // Stop libcurl loading its build-time default bundle, which the callback
+        // below would only discard; the callback supplies the trust anchors.
+        curl_easy_setopt(curl, CURLOPT_CAINFO, static_cast<char *>(nullptr));
+        curl_easy_setopt(curl, CURLOPT_CAPATH, static_cast<char *>(nullptr));
+        curl_easy_setopt(curl, CURLOPT_SSL_CTX_FUNCTION, ssl_ctx_callback);
+        curl_easy_setopt(curl, CURLOPT_SSL_CTX_DATA, &rec);
+        return true;
+    }
+
     auto ca_filename = m_ca_file ? m_ca_file->CAFilename() : "";
     auto crl_filename = m_ca_file ? m_ca_file->CRLFilename() : "";
     if (!ca_filename.empty() && !crl_filename.empty()) {
@@ -281,6 +317,7 @@ TPCHandler::ConfigureCurlCA(CURL *curl)
     if (!m_cafile.empty()) {
         curl_easy_setopt(curl, CURLOPT_CAINFO, m_cafile.c_str());
     }
+    return true;
 }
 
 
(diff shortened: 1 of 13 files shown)


ccaffy:
Also tested against EOS during some high load, I confirm this fixes the problem too.

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -187,10 +187,18 @@ int TPCHandler::closesocket_callback(void *clientp, curl_socket_t fd) {
...
     SSL_CTX* ctx = static_cast<SSL_CTX*>(ssl_ctx);
-    SSL_CTX_set_verify(ctx, SSL_VERIFY_PEER, verify_callback);
-    return CURL_SOCKOPT_OK;
+    if (rec && rec->ca_store) {
+        // Bumps the store's reference count instead of re-parsing the CA and CRL
+        // bundles for this connection.  libcurl runs this callback after it has
+        // applied its own TLS options, so this replaces whatever store it built.
+        SSL_CTX_set1_cert_store(ctx, rec->ca_store.get());
+    }
+    if (allowMissingCRL) {
```
abh3, line 197: A comment here that verify_callback will ignore the case where a CRL could not be found but otherwise will follow the remaining verification rules.

abh3 (approved):
A rather tour de force approach, not that I'm complaining. It looks like the tedious stuff has been taken care of. So, I just had a minor suggestion about a comment that would explain why you need a callback if you are ignoring CRL's.

StevenCTimm:
Three questions:  (A) What versions of xrootd were affected (b) what will be the new release which has the fix and (c) does this mean that EOS (both in eospublic.cern.ch and eosctapublic.cern.ch) is running a standalone xrootd daemon and the xrootd stuff is not built into the EOS software proper?

ccaffy:
> Three questions: (A) What versions of xrootd were affected (b) what will be the new release which has the fix and (c) does this mean that EOS (both in eospublic.cern.ch and eosctapublic.cern.ch) is running a standalone xrootd daemon and the xrootd stuff is not built into the EOS software proper?

Hi @StevenCTimm ,

A: All of them since 5.2.0 (year 2021).
B) @amadio can correct me if I'm wrong but I believe in 5.9.7 and 6.1.1.
C) EOS is an XRootD plugin. For HTTP TPC transfers, it just loads the vanilla XRootD HTTP TPC plugin. We fixed it in EOS because we do not rely on the official XRootD repository to create versions, we have our own one "eos-xrootd" which takes the XRootD tag we are interested in plus extra commits should we need to update quickly in production, like during this case ;)

Cheers,
Cedric

src/XrdHttpTpc/XrdHttpTpcTPC.cc
ccaffy: Absolutely, comment added! Thanks.

amadio:
Sorry, didn't look in detail at anything yet, but please retarget to v5.9.x branch, so I can merge this there and then merge v5.9.x into master later, so that the fix lands on both branches in the next patch releases. Thank you!

ccaffy:
I had to add two previous commits to make it work properly and not have conflicts.
The functionality allowing to validate a TLS handshake if CRLs are missing is therefore included. Indeed, the callback used to re-use the parsed CA/CRL store has been created during the implementation of "allow-missing-crl"

amadio:
@abh3 Any objection to backporting the `allow-missing-crl` feature to 5.9.x, so that the bugfix that depends on that can also be backported? I think the tradeoff is good for this, since no ABI breakage is introduced. Let me know what you think.

abh3:
Absolutely no objections. In fact, I encourage that performance issues be backported to the 5.9.x series as we all know that the 6.x series is not in the best shape and people will need to stay on 5.9. for a while longer.

PR #2876, opened 2026-07-27
Implement pseudo unload functionality for plugins

abh3:
The purpose of this modification is to provide a hook for shared library plugins that can be called prior to the xrootd server exiting due to a SIGTERM signal. This hook can complete any necessary processing prior to the termination of the process. This is especially useful for code coverage analysis that requires a finalization in each shared library in order to produce statistics for the library (i.e. a clean exit). This is what the default implementation does when ENABLE_COVERAGE is defined. However, additional processing may be added in the future for specific plugins using the FINISH_HANDLER macro.

This update is compatible with all previous versions and can technically be considered for a patch release. However, since it adds new functionality it likely needs to be restricted to a feature release.

### DIFF
--- a/src/XrdSys/XrdSysPlugin.cc
+++ b/src/XrdSys/XrdSysPlugin.cc
@@ -51,14 +51,14 @@
 #else
 #include "XrdSys/XrdWin32.hh"
 #endif
-  
+
 #include "XrdSys/XrdSysError.hh"
 #include "XrdSys/XrdSysHeaders.hh"
 #include "XrdSys/XrdSysPlatform.hh"
 #include "XrdSys/XrdSysPlugin.hh"
 #include "XrdVersion.hh"
 #include "XrdVersionPlugin.hh"
- 
+
 /******************************************************************************/
 /*                        S t a t i c   M e m b e r s                         */
 /******************************************************************************/
@@ -68,7 +68,7 @@ struct XrdSysPlugin::PLlist *XrdSysPlugin::plList = 0;
 /******************************************************************************/
 /*                            D e s t r u c t o r                             */
 /******************************************************************************/
-  
+
 XrdSysPlugin::~XrdSysPlugin()
 {
    if (libHandle) dlclose(libHandle);
@@ -78,7 +78,7 @@ XrdSysPlugin::~XrdSysPlugin()
 /******************************************************************************/
 /* Private:                   b a d V e r s i o n                             */
 /******************************************************************************/
-  
+
 XrdSysPlugin::cvResult XrdSysPlugin::badVersion(XrdVersionInfo &urInfo,
                                                 char mmv, int majv, int minv)
 {
@@ -98,11 +98,11 @@ XrdSysPlugin::cvResult XrdSysPlugin::badVersion(XrdVersionInfo &urInfo,
 
    return cvBad;
 }
-  
+
 /******************************************************************************/
 /* Private:                   c h k V e r s i o n                             */
 /******************************************************************************/
-  
+
 XrdSysPlugin::cvResult XrdSysPlugin::chkVersion(XrdVersionInfo &urInfo,
                                                 const char     *pname,
                                                 void           *lHandle)
@@ -219,7 +219,7 @@ XrdSysPlugin::cvResult XrdSysPlugin::chkVersion(XrdVersionInfo &urInfo,
 /******************************************************************************/
 /* Private:                      D L F l a g s                                */
 /******************************************************************************/
-  
+
 int XrdSysPlugin::DLflags()
 {
 #if    defined(__APPLE__)
@@ -234,7 +234,7 @@ int XrdSysPlugin::DLflags()
 /******************************************************************************/
 /* Private:                         F i n d                                   */
 /******************************************************************************/
-  
+
 void *XrdSysPlugin::Find(const char *libpath)
 {
    struct PLlist *plP = plList;
@@ -251,7 +251,7 @@ void *XrdSysPlugin::Find(const char *libpath)
 /******************************************************************************/
 /*                            g e t L i b r a r y                             */
 /******************************************************************************/
-  
+
 void *XrdSysPlugin::getLibrary(bool allMsgs, bool global)
 {
    void *myHandle;
@@ -381,11 +381,11 @@ void XrdSysPlugin::Inform(const char *txt1, const char *txt2, const char *txt3,
             }
       }
 }
-  
+
 /******************************************************************************/
 /* Private:                       l i b M s g                                 */
 /******************************************************************************/
-  
+
 XrdSysPlugin::cvResult XrdSysPlugin::libMsg(const char *txt1, const char *txt2,
                                             const char *mSym)
 {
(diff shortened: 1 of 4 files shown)


src/XrdSys/XrdSysPlugin.cc
```diff
@@ -248,10 +250,21 @@ void *XrdSysPlugin::Find(const char *libpath)
...
 
+/******************************************************************************/
+/*                                F i n i s h                                 */
+/******************************************************************************/
+
+void XrdSysPlugin::Finish()
+{
+// Call Finish() for every plugin in our loaded plugin set
+//
+   for (auto it = piSet.begin(); it != piSet.end(); it++) (*it)();
```
amadio: This could use a range-based for, like this:
```cpp
for (auto& plugin : piSet)
  plugin();
```
I think that `piSet` could get a better name, though, as it's not really a set of plugins, but a set of callbacks for termination. Maybe `plugin_unload_callbacks`? Then you'd have a nicer
```cpp
for (auto& callback : plugin_unload_callbacks)
  callback();
```

src/XrdSys/XrdSysPlugin.cc
```diff
@@ -319,9 +332,17 @@ void *XrdSysPlugin::getPlugin(const char *pname, int optional, bool global)
...
+// it's psuedo unload handler, if any.
 //
    if ((cvRC = chkVersion(urInfo, pname, myHandle)) == cvBad) return 0;
+      else {void (*pUnLoad)(void);
+            char nUnLoad[512];
+            if (snprintf(nUnLoad, sizeof(nUnLoad), "%s%s", XrdVERSIONINFOSFX,
+                         pname) < (int)sizeof(nUnLoad)
+                && (pUnLoad = (void (*)())dlsym(myHandle, nUnLoad))
+               )  piSet.insert(pUnLoad);
+           }
```
amadio: Why have this in an `else` clause? The `if` returns anyway, so this only adds extra indentation. I'd split the copying of the symbol name and the call to `dlsym` in separate code blocks, as well as add some code to handle unexpected results from `snprintf`, or make this a `GetUnloadCallback(pname, myHandle)` to make things easier to understand. Using a `std::string` would also make things more readable, as this is not a performance sensitive code path in any case.

src/XrdSys/XrdSysPlugin.hh
```diff
@@ -246,5 +255,6 @@ struct PLlist {PLlist  *next;
               };
 
 static PLlist *plList;
+static std::set<void (*)()> piSet;
```
amadio: This can be a file-level static and live entirely in the `XrdSysPlugin.cc` file, which would also allow removing `#include <set>` from `XrdSysPlugin.hh`.

src/XrdSys/XrdSysUtils.cc
```diff
@@ -194,10 +195,16 @@ bool XrdSysUtils::SigBlock()
...
 
 #ifdef ENABLE_COVERAGE
-   // Dump coverage information and exit upon receiving a TERM signal
-   signal(SIGTERM, [](int) { __gcov_dump(); _exit(EXIT_SUCCESS); });
+#define COVERAGE_CALL __gcov_dump();
+#else
+#define COVERAGE_CALL
 #endif
 
+// Perform a psuedo unload and then exit when receiving a SIGTERM signal
```
amadio, line 203: `s/psuedo/pseudo/`

src/XrdSys/XrdSysUtils.cc
```diff
@@ -194,10 +195,16 @@ bool XrdSysUtils::SigBlock()
...
+//
+   signal(SIGTERM, [](int) { COVERAGE_CALL XrdSysPlugin::Finish();
+                             _exit(EXIT_SUCCESS); });
```
amadio, line 206: Should we not avoid installing the signal handler when not running a coverage build? Do you think we will find uses for the callbacks outside of this? The `COVERAGE_CALL` above without a semicolon also looks like a syntax error, so I'd suggest to leave the `;` out of the macro definition and use `COVERAGE_CALL;` here instead.

amadio (changes requested):
This doesn't work, the coverage build that was run for this pull request still doesn't show the coverage for XrdMacaroons, for example: https://my.cdash.org/builds/3947788/coverage

abh3:
> This doesn't work, the coverage build that was run for this pull request still doesn't show the coverage for XrdMacaroons, for example: https://my.cdash.org/builds/3947788/coverage

Well, the link no longer works. However, I checked the library and the symbol is there. You can do the same. Did you get a report for other plugins?

src/XrdSys/XrdSysPlugin.cc
abh3: Certainly in the eye of the beholder :-) I changed it to be more appealing.

src/XrdSys/XrdSysPlugin.cc
abh3: Oh, that was a post-edit holdover. Indeed, the else, while perfectly valid, adds nothing. As for std::string, I considered it but decided it wasn't worth it. I can change my mind again.

src/XrdSys/XrdSysPlugin.hh
abh3: Yes, it was to put it into a known namespace so I could easily find it. I'll move it to an anonymous namespace.

src/XrdSys/XrdSysUtils.cc
abh3: Not really. If you look at how it was written (see XrdVersion.hh.in), every plugin can have a custom unload function that can be called whether or not coverage is enabled. Some people have asked for that in the past. In order to do that, SIGTERM handling has to always be enabled. I don't think there is any harm in that.

amadio:
> > This doesn't work, the coverage build that was run for this pull request still doesn't show the coverage for XrdMacaroons, for example: https://my.cdash.org/builds/3947788/coverage
> 
> Well, the link no longer works. However, I checked the library and the symbol is there. You can do the same. Did you get a report for other plugins?

The proof that the approach works will be that coverage for plugins becomes visible in the coverage build, which should be reflected in the percentage of coverage we see of the `XrdMacaroons` plugin in particular, since a lot of tests were recently added to test it as part of the security work I did. If the coverage build shows only the unit test coverage, it's because the changes introduced do not fix the problem yet. This is indeed a complicated thing to get working, because the symbol to dump coverage must reside in each plugin and must be called before the process terminates after receiving a `SIGTERM`. In addition to the changes here, it's possible that we may also need changes to the coverage build configuration as well, and that's why it doesn't work yet.

> Well, the link no longer works. However, I checked the library and the symbol is there. You can do the same. Did you get a report for other plugins?

I do not see any difference from before in terms of plugin coverage reported. At the very least, we should see all files from XrdMacaroons covered, as explained above.

abh3:
All I can say is that in my tests the unload function was called for each plugin (I also, as you may have noticed, added a message should a plugin not have an unload function). So, my suspicion is that one needs more than just calling __gcov_dump(), it could be that something has to be called in the shared library to get things started when it gets loaded.

amadio:
@abh3 It turns out that the real problem was that the linker scripts were hiding the coverage symbols. I created an alternative pull request with a working fix for the plugin coverage in #2919. If that is merged, then this one is no longer necessary. Please review the other pull request and let me know your thoughts. Cheers,

amadio:
As we discussed in person, with #2919 merged, we no longer need these changes. The flag `-fvisibility-inlines-hidden` can cause ABI breakage, so let's not change this beyond the plugins for now. We can revisit this topic close to XRootD 7.0's release. Thanks!

PR #2881, opened 2026-08-03
[XrdHttpTPC] Make low-speed timeout configurable

lobis:
## Summary

- add `tpc.low_speed <rate> [period]` configuration
- preserve the existing defaults of 10 KiB/s for two minutes
- allow `tpc.low_speed 0` to disable the low-speed check
- keep curl transfer policy in `TPCHandler`, separate from `State`

## Motivation

HTTP TPC transfers currently abort after remaining below 10 KiB/s for two minutes, with no operator-facing way to tune the policy. This may reject legitimate transfers over persistently slow links.

We needed a way to disable this feature for some tests but I thought it would be good to have this be a toggle instead of a hardcoded value.

## Validation

- built the `XrdHttpTPC-6` CMake target
- ran `xrdhttptpc-unit-tests` successfully
- ran `git diff --check`

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcConfigure.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcConfigure.cc
@@ -1,6 +1,7 @@
 
 #include "XrdHttpTpcTPC.hh"
 
+#include <climits>
 #include <dlfcn.h>
 #include <fcntl.h>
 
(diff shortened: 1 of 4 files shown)


amadio:
Rebased onto v5.9.x to put this change both on 5.x and 6.x branches.

PR #2883, opened 2026-08-05
Final real-time checksum implementation.

abh3:
This pull request completes the real-time checksum implementation by a) allowing the crc32 checksum family to efficiently handle out-of-order writes, b) treating the cks,type CGI element uniformly between on-demand and real-time checksums, c) allowing cks.type=default to be specified with consistent results, and d) providing a test regime that verifies that checksums, on demand or real-time, are properly computed.

## Summary by CodeRabbit

* **New Features**
  * Added checksum combination support for CRC-32 and CRC-32C, including combining checksums from adjacent data blocks.
  * Added end-to-end checksum validation for on-demand and real-time calculations.
* **Bug Fixes**
  * Selecting the `default` checksum type now consistently uses the configured default across checksum requests and real-time processing.
  * CRC-32 and CRC-32C results and combined values now maintain consistent byte ordering.
* **Tests**
  * Added comprehensive coverage for checksum calculation, combination, empty blocks, sequential segments, and supported checksum types.


src/XrdOfs/XrdOfs.hh
```diff
@@ -478,6 +478,7 @@ XrdOfsPrepare    *prepHandler;    // Plugin   prepare
 XrdCks           *Cks;            // Checksum manager
 XrdCksCalc       *CksRTCalc;      // Automatic realtime checksum calculator
 char             *CksRTName;      // Automatic realtime checksum cipher
+char             *CksRTDflt;      // The default r/t checksum, if any
```
amadio: This addition is why the ABI build has failed. This changes the size of `XrdOfs` class, which causes applications compiled with earlier versions to allocate the wrong size after XRootD is upgraded. Since EOS is the only affected project, and forces specific versions of XRootD on each build, we can accept this change into 6.2 without causing problems.

abh3: OK, but I want to raise the issue (again) that some times we may need to break ABI to fix things in private headers. We cannot be restricted only to feature releases just because of EOS. The EOS team is well aware that they may have to recompile to apply a patch because private headers can change at any time.

src/XrdCks/XrdCksCalccrc32.cc
```diff
@@ -2,10 +2,8 @@
 /*                                                                            */
 /*                    X r d C k s C a l c c r c 3 2 . c c                     */
 /*                                                                            */
-/* (c) 2011 by the Board of Trustees of the Leland Stanford, Jr., University  */
-/*                            All Rights Reserved                             */
-/*   Produced by Andrew Hanushevsky for Stanford University under contract    */
-/*              DE-AC02-76-SFO0515 with the Department of Energy              */
+/* Copyright (c) 2026 by European Organization of Nuclear Research (CERN)     */
+/* Produced by Lukasz Janyst <ljanyst@cern.ch>                                */
```
amadio: I suppose Lukasz did not write this one, right?

src/XrdCks/XrdCksCalccrc32.cc
```diff
@@ -29,150 +27,95 @@
...
-//
-   TotLen += reclen;
-   while(reclen-- > 0)
-        C32Result = (C32Result<<8) 
-                  ^ crctable[(unsigned char)((C32Result>>24)^*p++)];
+   pCheckSumTmp = crc32_combine(crc1, crc2, (z_off_t)DLen);
+
+#ifndef Xrd_Big_Endian
+   pCheckSumTmp = ntohl(pCheckSumTmp);
+#endif
```
amadio, line 94: We should verify that this really works on big endian architectures as intended.

tests/end-to-end/cks/XrdCksTestFile
```diff
@@ -0,0 +1,15 @@
...
+/*                                                                            */
+/* (c) 2004 by the Board of Trustees of the Leland Stanford, Jr., University  */
+/*   Produced by Andrew Hanushevsky for Stanford University under contract    */
+/*              DE-AC02-76-SFO0515 with the Department of Energy              */
+/*                                                                            */
+/* This file is part of the XRootD software suite.                            */
+/*                                                                            */
+/* XRootD is free software: you can redistribute it and/or modify it under    */
+/* the terms of the GNU Lesser General Public License as published by the     */
+/* F
```
amadio, line 15: This is a strange file, is it intended like this?

amadio:
@abh3 Please have a look at the CodeRabbit review comments. We should probably address at least some of the findings in the comments before merging.

tests/XrdCks/XrdCksTestOD.cf
```diff
@@ -0,0 +1,2 @@
+xrootd.chksum adler32 crc32 crc32c md5
+all.export /tmp
```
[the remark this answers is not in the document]
abh3: Correct, that was a transcription mistake. The test should match the export. The reason this worked for me originally is that my config file has os.localroot /tmp which fix the problem. That is another way of solving this problem.

tests/XrdCks/XrdCksTest.sh
```diff
@@ -0,0 +1,16 @@
...
+# When the test is for demand checksums, the server should be started
+# with the XrdCksTestOD.cf config file.
+
+# When the test is for r/t checksums, the server should be started
+# with the XrdCksTestRT.cf config file.
+#
+./TestCheckSum $test adler32 64fce8a2 XrdCksTestFile $target
+./TestCheckSum $test crc32   55bd9dea XrdCksTestFile $target
+./TestCheckSum $test crc32c  cf3aa257 XrdCksTestFile $target
+./TestCheckSum $test md5 b7018d4b11d10edbdba4a240e71d4976 XrdCksTestFile $target
```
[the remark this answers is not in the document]
abh3: OK, good point.

tests/XrdCks/XrdCksTest.sh
```diff
@@ -0,0 +1,16 @@
+target="xroot://sdfiana028.sdf.slac.stanford.edu//xdir/TestFile"
```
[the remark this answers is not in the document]
abh3: Well, someone else will have to change this because I have no clue how to do that in a test environment.

tests/end-to-end/cks/XrdCksTestFile
abh3: Yes, it is intended to be like this because it is arbitrary. You can make it whatever you want but then you will have to change all the tests to have the corresponding checksum in the test.

src/XrdCks/XrdCksCalccrc32.cc
abh3: Well, then point me to one, I don't have access to any such machines. However, data checksums have to be exported in big endian order. This statement simply makes sure that we do.

src/XrdCks/XrdCksCalccrc32.cc
abh3: No, I don't know how that happened that is why we stalled the merge.

src/XrdCks/XrdCksCalccrc32.cc
```diff
@@ -29,150 +27,95 @@
...
-     Use unsigned int instead of long to insure 32 bit values.
-     Include length bits at the end to correspond to the Posix 1003.2 spec.
-     Make this a C++ class.
-*/
-void XrdCksCalccrc32::Update(const char *p, int reclen)
+   uint32_t crc2 = getCS(Cksum);
+
+   pCheckSum = crc32_combine(pCheckSum, crc2, (z_off_t)DLen);
+
+   return (char *)&pCheckSum;
```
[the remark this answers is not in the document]
abh3: OK, you got it right for that one.

tests/XrdCks/XrdCksTest.cc
```diff
@@ -0,0 +1,261 @@
...
+   char buff[512];
+   uint64_t halfSZ = sizeof(buff)/2;
+
+   if (pread(lclFD, buff, sizeof(buff), 0) < 0)
+      Emsg(-errno, {"Unable to read the local file; ", strerror(errno)});
+
+   Status = clFile.Write(0, halfSZ, buff);
+   if (!Status.IsOK()) Fatal(Status, "Unable to write urlfile;");
+
+   FlushFile(halfSZ, lclFD, clFile);
```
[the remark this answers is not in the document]
abh3: Yeah, obviously I was thinking about something else at the time.

tests/XrdCks/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable(xrdcks-unit-tests
+  XrdCksTest.cc
+  )
+
+target_link_libraries(xrdcks-unit-tests
+  XrdCl
+  )
```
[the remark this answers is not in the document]
abh3: The person who knows how to integrate into unit test should do this then I will have a template for future tests. Right now there is no template.

tests/XrdCks/XrdCksTest.sh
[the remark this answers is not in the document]
amadio: I will do it today.

src/XrdCks/XrdCksCalccrc32.cc
abh3: OK, I figured it out and corrected it.

abh3:
OK, I have fixed things as needed. My test showed if the test is properly integrated (something I don't know how to do as there is no guide) it all works. Frustratingly, code rabbit refuses to let me resolve any of its bloody comments.

abh3:
@coderabbitai resolve

amadio:
@abh3 I connected the test to the build system, using Claude as you can see above. However, Claude noticed one thing that is important to address before merging. See it below, but I think you're already well aware of this.

---

**The `crc32` digest changes value.** The old `XrdCksCalccrc32` was the POSIX `cksum` algorithm: non-reflected polynomial `0x04C11DB7`, initial value 0, the file length appended as bytes, final XOR `0xFFFFFFFF`. The new one is the standard zlib CRC-32. On `tests/end-to-end/cks/XrdCksTestFile` (1024 bytes):

| source | value |
|---|---|
| old `XrdCksCalccrc32` | `d3d3ea32` |
| `cksum(1)` on the same file | `3553880626` = `0xd3d3ea32` |
| new `XrdCksCalccrc32` | `55bd9dea` |
| `python3 -c "import zlib; print(hex(zlib.crc32(open('XrdCksTestFile','rb').read())))"` | `0x55bd9dea` |

Every `crc32` value already stored in a `user.XrdCks.crc32` extended attribute therefore stops matching after an upgrade. No test caught this, since `tests/XRootD/noauth.sh` and `fuse.sh` check `crc32c`, not `crc32`.

Was this intended? I think the zlib definition is the better one, and the combine work needs it, but it needs a release note if we take it. I have not touched the algorithm.

**What I changed** (two commits on top of yours):

- `tests/XrdCks/` moves to `tests/end-to-end/cks/`, on the BATS harness, following the `tests/end-to-end/xrdfs` template. Two server configurations, `cks-ondemand.cfg` and `cks-realtime.cfg`, each running adler32, crc32, crc32c, and md5, so eight tests in total. The server writes inside `$BATS_TEST_TMPDIR` through `oss.localroot`, so the export path and the target path agree, and no host is hardcoded. This is the template you asked for.
- `tests/unit/XrdCks` and `tests/unit/XrdOuc` add GoogleTest coverage of the new combine code. On the big-endian question: the byte-order tests compare the four returned bytes against the big-endian decomposition of the reference value, so they hold on a little-endian host too. If the code emitted host order, they fail here.
- One small fix in `XrdOfsConfig.cc`: when `ofs.cksrt` names `default`, `CksRTName` was assigned the `CksRTDflt` pointer rather than a copy, so two members owned one allocation.

**Verification.** All 16 tests pass. To confirm the real-time suite really exercises `Combine` and is not falling back to an on-demand recompute, I perturbed the combine functions and re-ran:

- `crc32_combine` result XORed with 1: only `real-time crc32` fails; on-demand passes.
- `crc32c_combine` result XORed with 1: only `real-time crc32c` fails; on-demand passes.

---

@abh After your final review of the changes I added, and an answer for the above, I think this should be in good shape to be merged into XRootD 6.2.

tests/end-to-end/cks/cks-ondemand.bats
```diff
@@ -0,0 +1,56 @@
...
+}
+
+bats::on_failure() {
+    print_log_files
+}
+
+# The helper writes the local file to the server out of order, then asks
+# the server for the digest. On demand, the server reads the file back.
+
+check() {
```
diogomattioli, line 28: i'd avoid this type of helper. this masks the intent of the test.

call each test with the command necessary. be explicit, static, and avoid additional functions, loops, extra code.

put the `run -0 xrdckstest ...` on each single test

tests/end-to-end/cks/cks-ondemand.bats
```diff
@@ -0,0 +1,56 @@
...
+    local digest=$1 expected=$2
+    run -0 xrdckstest on-demand "${digest}" "${expected}" \
+        "${BATS_TEST_DIRNAME}/XrdCksTestFile" \
+        "root://localhost:${PORT}//TestFile.${digest}"
+}
+
+@test "on-demand adler32 digest" {
```
diogomattioli, line 35: better testing name. ask claude to write meaningful names.
`on-demand adler32 checksum digest for testfile should match`

on the other comment, where i address the use of the helper function, to write this 'example title' i had to look at the helper function and interpret what it is doing. that's the type of stuff we should avoid. we should make it explicit and static.

tests/end-to-end/cks/cks-realtime.bats
```diff
@@ -0,0 +1,68 @@
...
+bats::on_failure() {
+    print_log_files
+}
+
+# The helper writes the local file to the server out of order: the first
+# half block, then everything past the first block, then the hole. The
+# server must combine the segment digests while the file is written, and
+# must arrive at the same value the on-demand suite expects.
+
+check() {
```
diogomattioli, line 30: same thing

tests/end-to-end/cks/XrdCksTest.cc
```diff
@@ -1,6 +1,6 @@
 /******************************************************************************/
```
diogomattioli, line 1: `.cc` file inside the `e2e` folder?

what exactly does it do? is this a helper to just calculate checksums? is there any linux available command that does that? should we this binary in the bin folder when tests are created instead?

tests/unit/XrdCks/XrdCksCombineTests.cc
```diff
@@ -0,0 +1,229 @@
...
+
+namespace
+{
+using Digest = std::array<unsigned char, 4>;
+
+//------------------------------------------------------------------------------
+// Deterministic test data, so that a failure is reproducible.
+//------------------------------------------------------------------------------
+
+std::vector<char> TestData(size_t len)
```
diogomattioli, line 43: i don't know what it does, but already saw the same implementation in 3 different places.

these tests have more helper functions than tests.

should we test the helper functions? if they fail or have bugs, are the tests going to be affected?

tests/unit/XrdCks/XrdCksCombineTests.cc
```diff
@@ -0,0 +1,229 @@
...
+TEST(XrdCksCalccrc32CTest, FinalIsBigEndianCrc32C)
+{
+   const std::vector<char> data = TestData(dataLen);
+   const uint32_t expect = crc32c(0, data.data(), data.size());
+
+   XrdCksCalccrc32C calc;
+   EXPECT_EQ(Bytes(calc.Calc(data.data(), data.size())), BigEndian(expect));
+}
+
+TEST(XrdCksCalccrc32CTest, IsCombinable)
```
diogomattioli, line 154: as a person that haven't used this module nor understand what it does the name 'IsCombinable' does not help me to understand. i don't know what are the inputs, the expected output, etc.

with that i'd say that other functions follow the same structure. name does not appoint input, expected output, etc.

diogomattioli (commented):
Purely comments and suggestions, not a concrete review.
If you think they are not necessary or don't reflect the real intent with testing please just disconsider them.

diogomattioli:
> Purely comments and suggestions, not a concrete review. If you think they are not necessary or don't reflect the real intent with testing please just disconsider them.

One more thing I forgot to mention, the separation with 'mostly' one assert per test is excellent.
Is very easy to understand the separation between those tests and in case something fails there won't be the need to debug or interpret one single big function that has all the asserts on it.

tests/end-to-end/cks/cks-ondemand.bats
amadio: This is a simple wrapper function around the custom test command to avoid code duplication. Similar to the `get_macaroon` function. I guess it's a matter of taste, but I prefer the way it is now. The name could be better, indeed.

@abh3 which would you rather have? Calls to `xrdckstest` in each test? Or the wrapper function?

tests/end-to-end/cks/cks-ondemand.bats
amadio: I think with AI we need to choose between verbose or concise. For me the above is perfectly clear that this test checks the adler32 checksum when read back from the server matches against the known value.

tests/end-to-end/cks/XrdCksTest.cc
amadio: This is the source file for the custom command (`xrdckstest`) that executes a checksum test against the server by writing the file out of order. It's not used for anything else other than the end-to-end checksum tests, so it makes sense to keep it here. Where else would you move it to? It does go into `<binary_dir>/bin` when built. All binaries built by the project go there, actually.

tests/unit/XrdCks/XrdCksCombineTests.cc
amadio: See line 159 below. This test checks that `Combinable()` returns `true` for `crc32c`, as it's a combinable checksum type. No inputs/outputs are needed.

tests/unit/XrdCks/XrdCksCombineTests.cc
amadio: This function is very simple, just returning random data deterministically. It's used by tests below with different values, to cover more than 1, 2, or 3 pages (4k) worth of data, depending on what the tests need. We could use multiple hard-coded data vectors for this, but it would make the file a lot larger.

tests/end-to-end/cks/cks-ondemand.bats
abh3: I'm neutral as long as the test runs correctly.

tests/end-to-end/cks/cks-realtime.bats
abh3: Same here.

abh3:
I thin Claude is confusing zcrc32 with crc32. They are both based on crc32 but, as the comments say, crc32 needs to always be export in Big Endian order which zcrc32 need to be exported in little endian order. The proof of the pudding is whether the test file produces the correct checksum for each test case. I did not include a test case for zcrc32. So, I am totally mystified why Clade is complaining about this.

As for two variables pointing to the same allocation, that is perfectly normal as that variable is never deleted because it is not possible to delete it at the point it is assigned to the other variable.

amadio:
@abh3 Alright, I'm going to merge this. After connecting the tests to the test suite, they all pass. I don't want to delay the release any further than necessary.

PR #2884, opened 2026-08-05
[XrdHttpTpc] Do not let a destination close failure mask the transfer…

### LINKED ISSUE #2878: [XrdHttpTpc] Destination close failure masks the real transfer error (curl result / stall timeout) in single-stream pull TPC
### Operating System

Alma Linux 9

### XRootD Version

All

### Description

In TPCHandler::RunCurlWithUpdates() (single-stream path), State::Finalize() is called before the final error is composed, and it unconditionally overwrites m_error_code / m_error_buf. So whenever the close of
  the destination file fails, the error reported to the TPC client — and the only error logged — is the close failure. The actual reason the transfer stopped (the libcurl result code, or XrdHttpTpc's own
  stall-timeout message) is silently discarded and never appears anywhere.

  This is not a rare corner: when a pull is aborted mid-flight, the close is expected to fail on any storage backend that validates the expected file size, because XrdHttpTpc itself supplies that expected size.
  ProcessPullReq() appends oss.asize=<source Content-Length> to the open URL (XrdHttpTpcTPC.cc:1206), and a backend that enforces it will reject the short file at close. So on such a backend, every aborted pull
  reports the close error and never the abort reason.

  The net effect is that a whole class of HTTP-TPC failures is undiagnosable from either the server logs or the client-visible error.

  The multistream path (RunCurlWithStreams) does not have this problem — it calls Finalize() only in the success branch, so the ordering there is already correct.

Two things combine:

  1. Finalize() (and Flush(), which sets m_error_code = 2 at XrdHttpTpcState.cc:326) overwrite whatever error state was already recorded. They do not preserve a pre-existing error.
  2. The reporting chain tests state.GetErrorCode() at line 947 before res != CURLE_OK at line 955. So once Finalize() has set error code 3, the res != CURLE_OK branch is unreachable and curl_easy_strerror(res)
  is never evaluated, logged, or sent.

  The reasons that get lost include:

  - Any libcurl failure: CURLE_OPERATION_TIMEDOUT (28) from the hard-coded low-speed limit, CURLE_PARTIAL_FILE (18), CURLE_RECV_ERROR (56), etc.
  - XrdHttpTpc's own stall detector at XrdHttpTpcTPC.cc:828-834, which sets SetErrorCode(10) and the message "Transfer failed because no bytes have been received from the source (pull mode) in N seconds." — this
  message is composed and then thrown away by Finalize() a few lines later.

  The close failure that overwrites them is not itself informative — it is a downstream consequence of the abort, and on a backend that validates oss.asize it is the expected outcome of any aborted pull.
  Stream::Finalize() can also fail with no message at all (the m_avail_count == m_buffers.size() return at XrdHttpTpcStream.cc:49), in which case the client is told "Error when interacting with local filesystem:
  (no error message provided)".

_No response_

### Additional Information

_No response_

ccaffy:
… error

In the single-stream path, State::Flush() and State::Finalize() recorded their failures in the same error code and message as the transfer itself, and both ran before the final error was composed.  A failing close of the destination file therefore overwrote the reason why the transfer actually failed.  Since the reporting chain tested the state error code before the libcurl result, the libcurl branch became unreachable as well: timeouts, partial files and receive errors were all reported as "Error when interacting with local filesystem: <close error>".  This is typically hit with storage backends that validate the expected file size and reject the close of a pull that was aborted mid-transfer.

Flush() and Finalize() now record their failures separately, so they can no longer overwrite the transfer error.  Both the single-stream and the multistream paths now report, in order: the remote status code, the transfer error (stall detector or local write failure), the libcurl result, and finally the flush/close failure.  The flush/close failure is always logged, and is appended to the message sent to the client, so that neither error is lost.

The stall detector message is no longer prefixed with "Error when interacting with local filesystem", which did not describe it.

Two related problems are fixed in the multistream path: the result of MultiCurlHandler::Flush() was discarded, so a failure to flush the file was reported to the client as a success, and the remote status code failure was logged with an empty message.

The 1, 2, 3 and 10 error codes are now named in a State::ErrorCode enum; their values are unchanged.

Fixes: #2878

Assisted-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
@@ -124,16 +124,23 @@ public:
         return current_offset;
     }
 
-    int Flush() {
-        int last_error = 0;
+    // Flush every stream to the local filesystem.  Returns State::errNone and
+    // leaves error_msg untouched if all the streams could be flushed; otherwise
+    // returns the error code of the first failure encountered and sets error_msg
+    // to the corresponding error message.
+    int Flush(std::string &error_msg) {
+        int error_code = State::errNone;
         for (std::vector<State*>::iterator state_it = m_states.begin();
              state_it != m_states.end();
              state_it++)
         {
-            int error = (*state_it)->Flush();
-            if (error) {last_error = error;}
+            if (((*state_it)->Flush() == -1) && !error_code) {
+                error_code = State::errFlush;
+                error_msg = (*state_it)->GetFinalizeErrorMessage();
+                if (error_msg.empty()) {error_msg = "(no error message provided)";}
+            }
         }
-        return last_error;
+        return error_code;
     }
 
     off_t BytesTransferred() const {
@@ -324,7 +331,7 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
                 const char *log_prefix = rec.log_prefix.c_str();
                 bool tpc_pull = strncmp("Pull", log_prefix, 4) == 0;
 
-                mch.SetErrorCode(10);
+                mch.SetErrorCode(State::errTimeout);
                 std::stringstream ss;
                 ss << "Transfer failed because no bytes have been "
                    << (tpc_pull ? "received from the source (pull mode) in "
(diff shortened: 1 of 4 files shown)


amadio:
@ccaffy Can you please remove GitHub issue links from code comments? It's already in the commit message. I think it's not necessary to add also in the code. If the comments can be less verbose, that'd be good too. AI tends to overdo it. Thank you!

ccaffy:
> @ccaffy Can you please remove GitHub issue links from code comments? It's already in the commit message. I think it's not necessary to add also in the code. If the comments can be less verbose, that'd be good too. AI tends to overdo it. Thank you!

Yes with pleasure :)

PR #2885, opened 2026-08-06
[XrdHttpTpc] Report transferred bytes in multistream performance markers

### LINKED ISSUE #2874: [XrdHttpTpc] Multi-stream performance markers report scheduled bytes instead of transferred bytes
### Operating System

AlmaLinux 9

### XRootD Version

6.1.0

## Summary

For multi-stream HTTP TPC pulls, the performance markers sent back to the TPC client report
the **scheduling cursor** (how many bytes of Range requests have been *issued* to libcurl)
rather than the number of bytes actually received and written. The server therefore
over-reports progress by up to `concurrency × block_size` — 768 MiB with the default
settings and `streams=3`.

For any file smaller than that, the *very first* perf marker announces the full file size
before a single body byte has arrived, so the TPC client (FTS/gfal2) displays 100 % for the
entire duration of the real transfer.

The final `TRANSFER_SUCCESS` record is unaffected: it uses a different, correct source
(`MultiCurlHandler::BytesTransferred()`), so the bug is invisible unless you watch the
markers themselves.

## Evidence

Server-side log of a 5 GiB pull with `streams=3`:

```
14:20:15 TPC_PullRequest: event=PULL_START, local=/tmp/BIGFILE_5GB, ...
14:20:15 TPC_PullRequest: event=HEAD_SUCCESS, ..., streams=3; size=5368709120
14:20:15 TPC_PullRequest: event=RESPONSE_START, ...
14:20:15 TPC_PullRequest: event=PERF_MARKER, ..., bytes_transferred=805306368
14:20:20 TPC_PullRequest: event=PERF_MARKER, ..., bytes_transferred=2399141888
14:20:25 TPC_PullRequest: event=PERF_MARKER, ..., bytes_transferred=4177526784
14:20:30 TPC_PullRequest: event=PERF_MARKER, ..., bytes_transferred=5368709120
14:20:30 TPC_PullRequest: event=TRANSFER_SUCCESS, ..., bytes_transferred=5368709120
```

Two things give it away:

1. **Every reported value is an exact multiple of the 16 MiB block size** — 48, 143, 249 and
   320 blocks respectively. Across 48 concurrent range requests, the number of bytes actually
   received would essentially never land on a block boundary.
2. **The first marker is emitted in the same second as `PULL_START`/`HEAD_SUCCESS`**, before
   any body byte could have been received, and its value is exactly
   `concurrency × block_size` = `(3 streams × 16 pipelining) × 16 MiB` = 768 MiB
   (`TPCHandler::m_pipelining_multiplier`, `XrdHttpTpc/XrdHttpTpcTPC.hh:192`;
   `TPCHandler::m_block_size`, `XrdHttpTpc/XrdHttpTpcTPC.cc:40`).

## Root cause

`XrdHttpTpc/XrdHttpTpcMultistream.cc:317`:

```cpp
if (SendPerfMarker(req, rec, handles, current_offset)) {
```

`current_offset` is advanced by `MultiCurlHandler::StartTransfers()`
(`XrdHttpTpcMultistream.cc:105-125`) by `xfer_size` at the moment a request is handed to
libcurl, i.e. *before* any of that range has been received. `SendPerfMarker()` writes the
value straight into the wire format the client parses
(`XrdHttpTpc/XrdHttpTpcTPC.cc:694`):

```cpp
ss << "Stripe Bytes Transferred: " << bytes_transferred << crlf;
```

The correct figure is already available: `FinishCurlXfer()` accumulates the real received
count per completed request (`XrdHttpTpcMultistream.cc:78`) and then calls
`State::ResetAfterRequest()`, which zeroes `State::m_offset`
(`XrdHttpTpc/XrdHttpTpcState.cc:198-199`). So completed bytes and in-flight bytes are
disjoint and can simply be summed.

## Secondary issues in the same code path (same fix)

**a) The stall watchdog measures the wrong quantity.** `XrdHttpTpcMultistream.cc:313-316`
uses `current_offset` as its progress signal, while the error message it produces
(`:329-331`) claims *"no bytes have been received from the source (pull mode) in N seconds"*.
Those are not the same thing: when all handles are saturated mid-block the cursor is frozen
even though data is streaming in. With the default `m_timeout` of 60 s
(`XrdHttpTpcTPC.cc:400`) a spurious abort requires all 48 handles to stall simultaneously for
a full minute, so this is unlikely in practice — but it becomes correct for free once the
marker uses real received bytes.

**b) Misleading log message on the success path.** `XrdHttpTpcMultistream.cc:386-389` logs

```
event=MULTISTREAM_IDLE, ...; Unable to start new transfers; breaking loop.
```

whenever `current_offset == content_size && running_handles == 0` — i.e. the *normal*
end-of-transfer condition. In the log above it appears immediately before
`TRANSFER_SUCCESS` and reads like a failure. It should say something like
"All ranges have been scheduled and all handles are done; ending transfer loop."

## Proposed fix

The single-stream path is already the reference implementation for this — see below — so the
fix is to make `RunCurlWithStreamsImpl` compute the same quantity across all handles rather
than inventing new accounting:

```cpp
off_t bytes_received = mch.BytesTransferred();
for (auto *handle : handles) {bytes_received += handle->BytesTransferred();}
if (SendPerfMarker(req, rec, handles, bytes_received)) { ... }
```

and use the same value for the `last_advance_bytes` / `last_advance_time` watchdog just above
it, so that "no bytes received" means what it says. Reword the `MULTISTREAM_IDLE` message on
the success path.

## Affected versions

Present in 6.1.0 and in every release since multi-stream HTTP TPC was introduced (the code
predates the `XrdTpc` → `XrdHttpTpc` rename).

**Multi-stream only.** `streams` defaults to 1 (`XrdHttpTpcTPC.cc:1120`) and is only raised by
a client sending `X-Number-Of-Streams`; when it stays at 1 the request is dispatched to
`RunCurlWithUpdates` (`XrdHttpTpcTPC.cc:1191-1193`), which reports `state.BytesTransferred()`
— the real received count — both to the perf marker (`XrdHttpTpcTPC.cc:656`) and to the stall
watchdog (`XrdHttpTpcTPC.cc:774`). Single-stream also has no scheduling cursor to drift from,
since it issues one request for the whole file with no Range header. Only
`RunCurlWithStreamsImpl` uses the scheduling cursor, so in practice this affects FTS/gfal2-style
transfers that request multiple streams.

Pull a file smaller than `streams × 16 × 16 MiB` (e.g. 500 MiB with `streams=3`) over HTTP TPC,
sending `X-Number-Of-Streams: 3`, and watch the perf markers on the client side — or enable
`TPC_PullRequest` debug logging on the active server: the first `PERF_MARKER` reports the whole
file size instantly. Repeating the same transfer without the `X-Number-Of-Streams` header shows
correct, monotonically increasing values.

### Additional Information

_No response_

ccaffy:
The performance markers of a multi-stream transfer reported the scheduling cursor, i.e. the offset up to which range requests had been handed over to libcurl, instead of the number of bytes actually transferred.  The cursor runs ahead of the transfer by up to concurrency * block_size bytes, 768 MiB with the default block size and three streams, so progress was over-reported by that amount.  For any file smaller than that the very first marker announced the full file size before a single body byte had arrived, and the TPC client displayed 100% for the whole duration of the real transfer.  The final TRANSFER_SUCCESS record was unaffected, as it is computed from MultiCurlHandler::BytesTransferred().

The stall detector shared the same signal, while the error it produces states that no bytes have been received from the source, or transmitted to the destination in push mode.  Those are not the same quantity: the cursor stays frozen while all the handles are busy in the middle of their block, even though data is flowing.

MultiCurlHandler now exposes the bytes that have really been transferred, adding the bytes of the requests still in flight to those of the requests that already completed.  The two are disjoint, since FinishCurlXfer() accumulates the counter of a state before State::ResetAfterRequest() zeroes it.  Both the marker and the stall detector use that value.  The scheduling cursor keeps its role of deciding what is left to request and of checking, at the end, that the whole file was covered.

The single-stream path already reported State::BytesTransferred() in both places and is unchanged.

The MULTISTREAM_IDLE message logged when all the ranges have been scheduled and all the handles are done read as a failure, although it marks the normal end of the transfer loop, right before TRANSFER_SUCCESS.  It is reworded; the event name is kept.

Fixes: #2874

Assisted-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcMultistream.cc
@@ -147,6 +147,24 @@ public:
         return m_bytes_transferred;
     }
 
+    // Number of bytes that have actually been transferred so far: the bytes of
+    // the requests that already completed plus the bytes of the requests that
+    // are still in flight.  The two are disjoint: FinishCurlXfer() accumulates
+    // the counter of a state into m_bytes_transferred and then zeroes it via
+    // State::ResetAfterRequest(), so no byte is counted twice.
+    // Note this is not the same quantity as the scheduling offset maintained by
+    // StartTransfers(), which is advanced as soon as a range request is handed
+    // over to libcurl, hence before any byte of that range has been received.
+    off_t BytesInFlightAndTransferred() const {
+        off_t bytes = m_bytes_transferred;
+        for (std::vector<State*>::const_iterator state_iter = m_states.begin();
+             state_iter != m_states.end();
+             state_iter++) {
+            bytes += (*state_iter)->BytesTransferred();
+        }
+        return bytes;
+    }
+
     int GetStatusCode() const {
         return m_status_code;
     }
@@ -317,11 +335,16 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
         time_t now = time(NULL);
         time_t next_marker = last_marker + m_marker_period;
         if (now >= next_marker) {
-            if (current_offset > last_advance_bytes) {
-                last_advance_bytes = current_offset;
+            // Report - and watch for progress on - the bytes that have really
+            // been transferred, not the offset up to which the range requests
+            // have been scheduled: the latter runs ahead of the transfer by up
+            // to concurrency * m_block_size bytes.
+            const off_t bytes_transferred = mch.BytesInFlightAndTransferred();
+            if (bytes_transferred > last_advance_bytes) {
+                last_advance_bytes = bytes_transferred;
                 last_advance_time = now;
             }
-            if (SendPerfMarker(req, rec, handles, current_offset)) {
+            if (SendPerfMarker(req, rec, handles, bytes_transferred)) {
                 logTransferEvent(LogMask::Error, rec, "PERFMARKER_FAIL",
                     "Failed to send a perf marker to the TPC client");
                 return -1;
@@ -392,7 +415,7 @@ int TPCHandler::RunCurlWithStreamsImpl(XrdHttpExtReq &req, State &state,
                 }
             } else if (running_handles == 0) {
                 logTransferEvent(LogMask::Debug, rec, "MULTISTREAM_IDLE",
-                    "Unable to start new transfers; breaking loop.");
+                    "All the ranges have been scheduled and all the handles are done; ending the transfer loop.");
                 break;
             }
         }


src/XrdHttpTpc/XrdHttpTpcMultistream.cc
```diff
@@ -147,6 +147,24 @@ class MultiCurlHandler {
...
+    // the counter of a state into m_bytes_transferred and then zeroes it via
+    // State::ResetAfterRequest(), so no byte is counted twice.
+    // Note this is not the same quantity as the scheduling offset maintained by
+    // StartTransfers(), which is advanced as soon as a range request is handed
+    // over to libcurl, hence before any byte of that range has been received.
+    off_t BytesInFlightAndTransferred() const {
+        off_t bytes = m_bytes_transferred;
+        for (std::vector<State*>::const_iterator state_iter = m_states.begin();
+             state_iter != m_states.end();
+             state_iter++) {
```
amadio, line 162: Please use range-based for:
```cpp
for (const auto& state : m_states)
    bytes += state.BytesTransfered();
```
or similar.

amadio: Nevermind, I see the whole file uses this pattern. I will update that in a separate commit later.

PR #2886, opened 2026-08-07
[XrdClHttp] Third-party Copy support

diogomattioli:
This pull request adds third-party copy support for the HTTP and HTTPS protocols.

The implemented items are:

1. **[XrdCl]** Add third-party copy creation to the `PluginInterface` and connect it to the TPC fallback copy job. This implementation allows future client plugins to add third-party copy support with minimal changes.

2. **[XrdClHttp]** Add the HTTP/HTTPS third-party copy mechanism. The implementation supports:

   * Token injection
   * Multi-stream selection
   * TCP mode selection (pull or push)
   * Force-overwrite selection
   * PerfMarker feedback

3. **[XrdCl]** Add the `xrdcp` command-line option `-t push`. This option allows the user to select the transfer mode for HTTP third-party copy. The default mode is **pull**. The alternative mode is **push**.

4. **[XrdCl]** Add the `xrdcp` command-line option `--tpc-token-file`. This option allows the user to provide a token file that contains authentication tokens for the source and destination servers.

5. **[XrdClHttp]** Add a token file parser. The parser supports:

   * Plain text files, where the first line contains the source token and the second line contains the destination token.
   * JSON files with the properties `src` and `dst`.

## Summary by CodeRabbit

* **New Features**
  * Added source-driven third-party copy (`--tpc push`) for HTTP(S), while retaining pull as the default.
  * Added bearer-token authentication through `--tpc-token-file`, supporting plain-text and JSON formats, optional tokens, and `BEARER_TOKEN_FILE` fallback.
  * Added configurable transfer streams, progress reporting, overwrite handling, and operation timeouts for HTTP(S) copies.
  * Added support for HTTP(S) third-party copy operations through the client interface.

* **Documentation**
  * Updated `xrdcp` usage and option documentation with push mode and token-file details.

* **Tests**
  * Added comprehensive coverage for HTTP, HTTPS, authentication, progress, timeout, protocol, token handling, and large-file transfers.


src/XrdCl/XrdClPlugInInterface.hh
```diff
@@ -602,6 +602,18 @@ namespace XrdCl
...
+      //! @see FileSystem::ThirdPartyCopy
+      //------------------------------------------------------------------------
+      virtual XRootDStatus ThirdPartyCopy( [[maybe_unused]] const std::string  &source,
+                                           [[maybe_unused]] const std::string  &dest,
+                                           [[maybe_unused]] const PropertyList *properties,
+                                           [[maybe_unused]] ProgressHandler    *progress_handler,
+                                           [[maybe_unused]] time_t              timeout )
+      {
+        return XRootDStatus( stError, errNotImplemented );
+      }
```
amadio: @diogomattioli I discussed this with Andy on our Wednesday meeting. The easy solution to avoid the ABI issue caught by the build is to keep the TPC interface separate from the filesystem and plugin interfaces, in a separate class or simply a separate function or set of functions. The other path is to accept the ABI changes. If we go in that direction, I would like to ensure that at least ROOT compiled against 6.1.1 and preloaded with a later version containing this patch will keep working for both `root://` and `http(s)://` data access from EOS. Both options require extra work.  My preference would be to move the interfaces around a bit and/or use a separate class to remove the ABI problem, as it's the safer of the two.

diogomattioli: For the suggestion of use as a separate function: If the idea of implementing the third party copy on an interface is to expose it to a contract where every single shared object could be implement it, working around without interfaces wouldn't kill that purpose?

For the suggestion of moving around interfaces and moving it out of filesystem/plugin interfaces: Wouldn't it continue using another interface, thus leading to the same problem of breaking the ABI?

amadio:
I have rebased this branch on `master` to resolve conflicts in CMake with the tape REST API work. I intend to merge this as-is, then add a change to fix the ABI before the release next week. I could also squash the commits by hand, add the ABI change on top, and push here before merging.

amadio:
Looks like I messed up the rebase, let me do it again...

tests/unit/XrdClHttp/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOpPut.cc
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOpQuery.cc
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOpRead.cc
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOpReadV.cc
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOps.cc            ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOps.hh
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOpStat.cc
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOptionsCache.cc   ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpOptionsCache.hh
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpParseTimeout.cc   ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpParseTimeout.hh
+  ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpUtil.cc           ${PROJECT_SOURCE_DIR}/src/XrdClHttp/XrdClHttpUtil.hh
+)
```
amadio, line 35: @diogomattioli My rebase failed because of missing files here (which didn't cause conflicts during the rebase). Could you explain why you want to exclude `XrdClHttpFilesystem` here? I'm afraid this will break whenever a new file is added to `XrdClHttp` if the same source list needs to be kept in sync here. It would be better to just link the shared object library, like it's done here:

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/tests/XrdClHttp/CMakeLists.txt#L5-L14

The way it's done above can work even if the source list of XrdClHttp changes.

tests/unit/XrdClHttp/ParseTokenFileTests.cc
```diff
@@ -0,0 +1,193 @@
...
+
+#include <filesystem>
+#include <fstream>
+#include <string>
+
+using namespace std::string_literals;
+
+using namespace testing;
+
+#include "XrdClHttp/XrdClHttpFilesystem.cc"
```
amadio, line 18: I think I found out the reason. Can we not write the tests without including C++ source files?

amadio:
@diogomattioli For now, I've reverted the branch to where it was before I tried to rebase. Since the rebase is not that simple, could you try to resolve the problems I mentioned above. In the mean time I will review the changes (and comments by CodeRabbit), and try to get the streaming checksum pull request in share for the release as well. Then I will come back to this one.

tests/unit/XrdClHttp/CMakeLists.txt
diogomattioli: The reason for including the source code is to test the internal parts of `XrdClHttpFilesystem`. Without it you cannot test private methods, do namespace injection to inject mock types (which is being used here), etc.

Unfortunately using a shared object library would just duplicate the code (when including the source file).

One option to not break the build would be the use of `file(GLOB_RECURSE ...)` and then just removing `XrdClHttpFilesystem.cc` from the list. But as we know, we're not allowed to use `GLOB_RECURSE`.

Another possible option is to completely remove the tests, not testing the internal parts at all.

tests/unit/XrdClHttp/ParseTokenFileTests.cc
diogomattioli: Answer is in the other question.

tests/unit/XrdClHttp/CMakeLists.txt
amadio: Is it not possible to create a private header to expose the things you need to test, then link against just the object library? Let's look at the specifics and see if it's possible, otherwise whenever we add a source file to XrdClHttp, this test will break.

src/XrdCl/XrdClTPFallBackCopyJob.hh
```diff
@@ -43,7 +43,7 @@ namespace XrdCl
       //------------------------------------------------------------------------
       //! Destructor
       //------------------------------------------------------------------------
-      virtual ~TPFallBackCopyJob();
+      virtual ~TPFallBackCopyJob() = default;
```
amadio, line 46: Declaring a virtual destructor in a header file but defining it out-of-line in a source file (.cc) is used to pin the class's vtable and runtime type information (typeinfo) to a single shared object (.so or .dll)[¹](https://developercommunity.visualstudio.com/t/delete-calls-scalar-deleting-destructor-instead-of/346875) [²](https://github.com/InsightSoftwareConsortium/ITK/issues/6000), `libXrdCl.so` in this case. Without this technique, the codebase can suffer from binary bloat (multiple destructors in different translation units), slow linking, and potentially, subtle memory bugs across shared-object boundaries.

Please add the `= default;` into the .cc file to keep the vtable pinned to `libXrdCl.so`.

src/XrdCl/XrdClXRootDResponses.hh
```diff
@@ -1234,6 +1234,25 @@ namespace XrdCl
...
       static ResponseHandler* Wrap( std::function<void(XRootDStatus*, AnyObject*)> func );
   };
+
+  //----------------------------------------------------------------------------
+  //! Handle the progress of an asynchronous operation
+  //----------------------------------------------------------------------------
+  class ProgressHandler
+  {
+    public:
+      virtual ~ProgressHandler() = default;
```
amadio: Same here, please declare in the header, and use `= default` in the .cc file.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -24,11 +24,82 @@
...
+#include <string>
 
 using namespace XrdClHttp;
+using namespace std::string_literals;
+
+namespace
+{
+
+// Blocks the calling thread until the operation it is given to completes.
+class SyncResponseHandler final : public XrdCl::ResponseHandler
```
amadio, line 50: Isnt' this duplicating the existing `SyncResponseHandler` from `src/XrdCl/XrdClMessageUtils.hh`?

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/src/XrdCl/XrdClMessageUtils.hh#L43-L59

That also raises the issue that the destructor of that handler is not pinned...
If possible, I suggest to improve the existing handler.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +382,195 @@ Filesystem::Stat(const std::string      &path,
...
+// any other file holds the <src> token in the first line and the <dst> one
+// in the second line. An absent or empty token means the corresponding side
+// is given no token at all.
+//
+// Returns false if the file cannot be opened or holds no token at all, in
+// which case no header is appended.
+static bool ParseTokenFile( const std::string   &token_file,
+                            CurlCopyOp::Headers &src_hdrs,
+                            CurlCopyOp::Headers &dst_hdrs,
+                            XrdCl::Log          *log )
```
amadio, line 398: There are already some functions in `master` that can handle this, let's reuse them if possible:

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/src/XrdClHttp/XrdClHttpUtil.hh#L59-L75

If the fact that you need to handle two tokens gets in the way, then you can refactor things a bit to accomodate the needs in this branch.

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +382,195 @@ Filesystem::Stat(const std::string      &path,
...
+    return true;
+}
+
+// Returns true if the URL uses the http or the https protocol. A third party
+// copy is possible only between these two protocols.
+static bool is_http_url( const std::string &url )
+{
+    const std::string protocol = XrdCl::URL(url).GetProtocol();
+    return protocol == "http" || protocol == "https";
+}
```
amadio, line 461: Something to keep in mind for later, but in case it can be supported, `davs://` and `s3://` could be used as source/destination in the command, and converted to `https://` internally. This already happens in other places (i.e. in the `XrdHttpTpc` plugin on the server side):

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L372-L380

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +382,195 @@ Filesystem::Stat(const std::string      &path,
...
+                                                  mode, timespec{tpc_timeout, 0}, log, nullptr);
+    op_copy->SetProgressHandler(progress_handler);
+
+    try
+    {
+        m_queue->Produce(op_copy);
+    }
+    catch (...) {
+        log->Warning(kLogXrdClHttp, "Failed to add copy op to queue");
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInternal);
```
amadio, line 560: It would be good to know why things failed, by using `ex.code()` and `ex.what()` from the caught exception here.

src/XrdClHttp/XrdClHttpFilesystem.cc
amadio, line 461: Something to keep in mind for later, but in case it can be supported, `davs://` and `s3://` could be used as source/destination in the command, and converted to `https://` internally. This already happens in other places (i.e. in the `XrdHttpTpc` plugin on the server side):

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/src/XrdHttpTpc/XrdHttpTpcTPC.cc#L372-L380

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -24,11 +24,82 @@
...
+    bool wait(std::chrono::seconds timeout = std::chrono::seconds(0))
+    {
+        const auto start = std::chrono::steady_clock::now();
+
+        std::unique_lock<std::mutex> lock{mutex};
+
+        if (timeout.count() > 0)
+            is_ready_changed.wait_for(lock, timeout, [this]{ return is_ready; });
+        else
+            is_ready_changed.wait(lock, [this]{ return is_ready; });
```
amadio, line 87: `timeout == 0` means the default timeout applies, there's a specific configuration option in the client setting this, which should be used here instead of waiting indefinitely on `timeout == 0`.

amadio: Here's how this is handled on the XRootD protocol implementation:

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/src/XrdCl/XrdClThirdPartyCopyJob.cc#L710-L728

tests/end-to-end/tpc/http/macaroon.bats
```diff
@@ -0,0 +1,106 @@
...
+	generate_client_files $BATS_FILE_TMPDIR
+
+	openssl rand -base64 -out macaroon-secret 64
+
+	PORT=7094 launch_xrootd macaroon.cfg xrootd_src
+	PORT=7095 launch_xrootd macaroon.cfg xrootd_dst
+
+	sleep 0.5
+
+	src_token=$(curl -k --cert client.pem --key client.key -X POST -d '{ "caveats": [ "activity:READ_METADATA,UPDATE_METADATA,LIST,DOWNLOAD,UPLOAD,MANAGE,DELETE" ], "validity": "PT1H" }' -H 'Content-Type: application/macaroon-request' https://localhost:7094/ | jq -r .macaroon)
```
amadio, line 29: I think it's a good idea to use a shell function to avoid these long lines (I added one in the macaroons tests). If we place it in a shared script, like common.bats, then we can also share it between tests. Not perfect, but should work:

https://github.com/xrootd/xrootd/blob/599151f1d9016fcdf21f8ee43f1deba237bcda96/tests/XRootD/macaroons.sh#L40-L67

amadio:
I added two commits here, one to fix tests running in parallel, so they do not overwrite each others' files, and another to restore ABI compatibility.

amadio:
:+1: For the improved security.

tests/end-to-end/tpc/http/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+add_bats_test(XrdCl::tpc-http-bigfiles bigfiles.bats)
+add_bats_test(XrdCl::tpc-http-https https.bats)
+add_bats_test(XrdCl::tpc-http-macaroon macaroon.bats)
```
amadio: This one needs to be skipped if `ENABLE_MACAROONS=OFF`, as that's an optional feature.

tests/end-to-end/tpc/http/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
...
+add_bats_test(XrdCl::tpc-http-oauth oauth.bats)
+add_bats_test(XrdCl::tpc-http-plain plain.bats)
+add_bats_test(XrdCl::tpc-http-progressbar progressbar.bats)
+add_bats_test(XrdCl::tpc-http-protocol protocol.bats)
+add_bats_test(XrdCl::tpc-http-scitokens scitokens.bats)
```
amadio: Same here, support for SciTokens is optional, so this needs to be skipped. On Alpine, there's no macaroons or scitokens, same for macOS builds (we use these things only from the system in the CI).

amadio (changes requested):
A couple more things, tests need to run only when the feature is enabled for optional features.

tests/end-to-end/tpc/http/CMakeLists.txt
amadio: Example:

https://github.com/xrootd/xrootd/blob/36b7781945a185b5ef559f09628af3998cb80623/tests/XRootD/CMakeLists.txt#L19-L22

tests/end-to-end/tpc/http/bigfiles.bats
```diff
@@ -13,6 +13,8 @@ readonly XROOTD_DST="localhost:8459"
 setup() {
 	cd $BATS_TEST_TMPDIR
 
+	sleep 0.5
```
amadio, line 16: Why is this required? What if the tests run in parallel?

diogomattioli: this is required because on Ubuntu even killing the process it takes few milliseconds for the tcp port to be available again. If not using that I am getting errors about `port X is already in use`

src/XrdClHttp/XrdClHttpOpCopy.cc
```diff
@@ -20,21 +20,34 @@
...
+        const Headers &regular_hdrs  = mode == TpcMode::Pull ? dest_hdrs : source_hdrs;
+        const Headers &transfer_hdrs = mode == TpcMode::Pull ? source_hdrs : dest_hdrs;
+
+        if (mode == TpcMode::Pull)
+            m_headers_list.emplace_back("Source"s, source_url);
+        else
+            m_headers_list.emplace_back("Destination"s, dest_url);
+
+        std::copy(connection_hdrs.begin(), connection_hdrs.end(), std::back_inserter(m_headers_list));
+        std::copy(regular_hdrs.begin(),    regular_hdrs.end(),    std::back_inserter(m_headers_list));
```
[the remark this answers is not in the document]
diogomattioli: separate PR

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +384,195 @@ Filesystem::Stat(const std::string      &path,
...
+                                                time_t                        timeout )
+{
+    XrdCl::Log *const log = XrdCl::DefaultEnv::GetLog();
+    log->Debug(kLogXrdClHttp, "XrdClHttp::ThirdPartyCopy src %s dst %s", source.c_str(), dest.c_str());
+
+    if (!is_http_url(source) || !is_http_url(dest))
+    {
+        log->Error(kLogXrdClHttp, "Third party copy can only be done between http(s) protocols");
+        return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errInvalidArgs, 0,
+                                   "Third party copy can only be done between http(s) protocols");
```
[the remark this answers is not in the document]
diogomattioli: separate PR

tests/end-to-end/tpc/http/plain.bats
```diff
@@ -0,0 +1,74 @@
...
+	assert_equal "$(cat xrootd_dst/file_dst)" 'source content'
+}
+
+@test "push copy sends the Destination header to the source server" {
+	run -0 xrdcp -T push only http://localhost:7094//file_src http://localhost:7095//file_dst
+	run -0 grep "Destination: http://localhost:7095//file_dst" xrootd_src.log
+}
+
+@test "push copy without --force fails when the destination exists" {
+	skip "known bug: the push copy replaces the destination without --force"
```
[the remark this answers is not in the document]
diogomattioli: separate MR - `XrdHttpTpc` to be fixed

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +382,195 @@ Filesystem::Stat(const std::string      &path,
...
+        };
+
+        read_token("src", src_token);
+        read_token("dst", dst_token);
+    }
+    else
+    {
+        std::istringstream lines(content);
+        std::getline(lines, src_token);
+        std::getline(lines, dst_token);
```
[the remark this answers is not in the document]
diogomattioli: `trim` on PR https://github.com/xrootd/xrootd/pull/2888 already does that. will be used when merged.

src/XrdClHttp/XrdClHttpFilesystem.cc
diogomattioli: 2 tokens are necessary. Also the new function has improved logic to deal with JSON files are conditional missing tokens.

Mixing those functions that accept a `--tpc-token-file` param does not seem like a good idea. One should be definitely used only when TPC is requested.

src/XrdClHttp/XrdClHttpFilesystem.cc
diogomattioli: `dav` and `davs` will be included as well

diogomattioli: `dav` and `davs` were implemented. `s3` should not be used for now

src/XrdClHttp/XrdClHttpFilesystem.cc
```diff
@@ -311,6 +384,195 @@ Filesystem::Stat(const std::string      &path,
...
+
+    if (progress_handler)
+    {
+        try
+        {
+            const auto op_stat = std::make_shared<StatOp>(&rh, source, timespec{init_timeout,0}, log, true, nullptr, nullptr);
+            m_queue->Produce(op_stat);
+
+            if (!rh.wait(std::chrono::seconds(init_timeout)))
+                log->Warning(kLogXrdClHttp, "Failed to get source file size: Operation timed out. Continuing...");
```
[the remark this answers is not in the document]
diogomattioli: a `shared_ptr` was used to share `rh`.

tests/end-to-end/helper/ssl.bash
```diff
@@ -0,0 +1,28 @@
...
+}
+
+generate_host_files() {
+    local path=${1:-.}
+    local subject="/C=CH/ST=Geneva/L=Geneva/O=XRootD/OU=IT/CN=localhost"
+
+    openssl genrsa -out host.key 2048
+    openssl req -new -key host.key -out host.csr -subj "$subject"
+
+    openssl x509 -req -in host.csr -CA $path/ca.pem -CAkey $path/ca.key -CAcreateserial -out host.pem -days 825 -sha256 -subj "$subject"
```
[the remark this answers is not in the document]
diogomattioli: `tls.sh` is being used instead

tests/end-to-end/tpc/http/macaroon.bats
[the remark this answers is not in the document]
diogomattioli: using `tls.sh` instead

tests/end-to-end/tpc/http/scitokens.bats
```diff
@@ -0,0 +1,117 @@
...
+teardown() {
+	kill_pid_files
+}
+
+@test "pull copy without a token fails" {
+	run ! xrdcp -T only http://localhost:7094//file_src http://localhost:7095//file_dst
+}
+
+@test "pull copy with --tpc-token-file holding both tokens succeeds" {
+	run -0 xrdcp -T only --tpc-token-file token-file http://localhost:7094//file_src http://localhost:7095//file_dst
```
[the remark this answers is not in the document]
diogomattioli: `--tpc-token-file` can be used only over `https` or `davs` now

src/XrdCl/XrdClTPFallBackCopyJob.hh
diogomattioli: done!

src/XrdCl/XrdClXRootDResponses.hh
diogomattioli: done!

src/XrdClHttp/XrdClHttpFilesystem.cc
diogomattioli: done!

src/XrdClHttp/XrdClHttpOps.hh
```diff
@@ -894,16 +905,17 @@ public:
...
-    void SetCallback(std::unique_ptr<CurlProgressCallback> callback);
+    // Set the handler notified when a performance marker is received.
+    //
+    // The handler is not owned by this operation and must outlive it. Give
+    // nullptr to send no notification.
+    void SetProgressHandler(XrdCl::ProgressHandler *handler) noexcept;
 
     virtual HttpVerb GetVerb() const override {return HttpVerb::COPY;}
 
+    bool IsSentSucessfully() {return m_sent_success;}
```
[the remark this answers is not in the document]
diogomattioli: done!

tests/end-to-end/tpc/http/CMakeLists.txt
diogomattioli: done!

tests/end-to-end/tpc/http/CMakeLists.txt
diogomattioli: done!

tests/end-to-end/tpc/http/macaroon.bats
diogomattioli: i'd rather have this single query than creating additional functions or logic for that.

again, we should implement the most necessary helpers only, otherwise we'll end up creating logic and functions inside that tests that could fail and affect the tests.

src/XrdCl/XrdClTPFallBackCopyJob.cc
```diff
@@ -63,26 +93,31 @@ namespace XrdCl
...
+        progress_handler.emplace( pJobId, *progress );
+
+      st = ThirdPartyCopy(GetSource(), GetTarget(), pProperties,
+                          progress_handler ? &progress_handler.value() : nullptr);
+      if (st.IsOK())
+        return st;
+    }
+
     // check if we can fall back to streaming
     if( tpcFallBack && ( st.code == errNotSupported || st.code == errOperationExpired ) )
```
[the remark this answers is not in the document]
diogomattioli: done!

src/XrdCl/XrdClTPFallBackCopyJob.cc
```diff
@@ -63,26 +92,33 @@ namespace XrdCl
...
     {
       Log *log = DefaultEnv::GetLog();
       log->Debug( UtilityMsg, "TPC is not supported, falling back to streaming mode." );
 
-      delete pJob;
-      pJob = new ClassicCopyJob( pJobId, pProperties, pResults );
-      return pJob->Run( progress );
+      return ClassicCopyJob(pJobId, pProperties, pResults).Run(progress);
```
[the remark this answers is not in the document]
diogomattioli: done!

tests/end-to-end/tpc/http/oauth.bats
```diff
@@ -0,0 +1,110 @@
...
+
+	sed "s|issuer =.*|issuer = ${IAM_OAUTH_ISSUER}|" ${BATS_TEST_DIRNAME}/oauth-module.cfg > oauth-module.cfg
+
+	endpoint=$(curl -s "${IAM_OAUTH_ISSUER}/.well-known/openid-configuration" | jq -r .token_endpoint)
+
+	scope='storage.read:/ storage.modify:/ storage.create:/ storage.stage:/'
+	export TOKEN=$(curl -s -X POST --user "${IAM_OAUTH_CLIENT_ID}:${IAM_OAUTH_CLIENT_SECRET}" \
+		-H "Content-Type: application/x-www-form-urlencoded" \
+		--data "grant_type=client_credentials&scope=${scope}" \
+		${endpoint} | jq -r .access_token)
```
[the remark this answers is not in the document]
diogomattioli: done!

amadio:
@diogomattioli In the end, I decided to just remove the fallback end-to-end tests for now. They fail because the server returns true for `xrdfs query config tpc`, so there's no fallback in reality, and since TPC is not configured, the copies fail. The tests were also sharing the port with another set of tests which fail if run in parallel. We can add these tests later when we can make the server configure fine and return that TPC is not supported (I tried setting the program to /dev/null, but then the server doesn't start). Basically, the server will always return that it supports TPC, so for now fallback cannot be properly tested.

amadio:
It looks like the tests added here are going to fail randomly because of the hacky `sleep 0.5`. They failed in this job in the original repo: https://github.com/diogomattioli/xrootd/actions/runs/35961895960/job/107512050023

The job failed in test **451 – `XrdCl::tpc-http-macaroon`** while setting up the source XRootD server. Below is the explanation from GitHub Copilot:

```text
`PORT=${XROOTD_SRC##*:} launch_xrootd macaroon.cfg xrootd_src' failed
```

The test starts two servers on fixed ports (`8478` and `8479`) for every BATS case:

- [`macaroon.bats:11-22`](https://github.com/diogomattioli/xrootd/blob/becf9202c2cf915e07183b456734ec5e1cc59622/tests/end-to-end/tpc/http/macaroon.bats#L11-L22)
- [`common.bash:8-16`](https://github.com/diogomattioli/xrootd/blob/becf9202c2cf915e07183b456ec5e1cc59622/tests/end-to-end/helper/common.bash#L8-L16)

`setup()` launches the daemons and sleeps for only 500 ms. `teardown()` then force-kills them with `SIGKILL` but leaves the PID files and does not wait for the ports to become reusable:

```bash
find "$BATS_TEST_TMPDIR" -name '*.pid' ... -exec awk '{ print }' {} \; | xargs -r kill -9
```

This creates a startup race between BATS cases. Test 4 happens to encounter it, while the remaining cases pass.

## Recommended fix

Make server startup and teardown deterministic. Wait for each server to be ready, terminate it cleanly, wait for termination, and remove its PID file.

### `tests/end-to-end/helper/common.bash`

```bash
launch_xrootd() {
    local config=$1
    local name=$2
    local pidfile="$BATS_TEST_TMPDIR/$name.pid"

    pushd "$(pwd)" >/dev/null
    cd "$BATS_TEST_TMPDIR"

    BATS_TEST_DIRNAME="$BATS_TEST_DIRNAME" \
    BATS_SUITE_TMPDIR="$BATS_SUITE_TMPDIR" \
    NAME="$name" \
    xrootd -b -c "$BATS_TEST_DIRNAME/$config" -l "$name.log" -s "$pidfile"

    # Wait until xrootd has created the pid file and is accepting connections.
    for _ in {1..50}; do
        if [[ -s "$pidfile" ]] &&
           curl --silent --show-error --fail \
                --insecure \
                --cert "$BATS_SUITE_TMPDIR/client.crt" \
                --key "$BATS_SUITE_TMPDIR/client.key" \
                "https://localhost:${PORT}/" >/dev/null 2>&1; then
            popd >/dev/null
            return 0
        fi

        # Fail early if xrootd exited during startup.
        if [[ -s "$pidfile" ]] && ! kill -0 "$(cat "$pidfile")" 2>/dev/null; then
            break
        fi

        sleep 0.1
    done

    echo "xrootd failed to start: $name" >&2
    [[ -f "$name.log" ]] && cat "$name.log" >&2
    popd >/dev/null
    return 1
}

kill_pid_files() {
    while IFS= read -r pidfile; do
        pid=$(cat "$pidfile" 2>/dev/null || true)

        if [[ "$pid" =~ ^[0-9]+$ ]] && kill -0 "$pid" 2>/dev/null; then
            kill "$pid" 2>/dev/null || true

            for _ in {1..50}; do
                kill -0 "$pid" 2>/dev/null || break
                sleep 0.1
            done

            # Avoid leaking a daemon if graceful shutdown timed out.
            kill -9 "$pid" 2>/dev/null || true
        fi

        rm -f "$pidfile"
    done < <(find "$BATS_TEST_TMPDIR" -name '*.pid' -type f)

    print_log_files
}
```

The readiness check can instead use `xrdfs` or `nc` if HTTPS authentication makes a root request unsuitable. The important parts are:

1. Do not rely on a fixed `sleep 0.5`.
2. Verify that the process is alive and listening before continuing.
3. Shut down processes cleanly.
4. Remove stale PID files.
5. Wait for process termination before the next test starts.

A smaller minimal fix would be to update teardown as follows:

```bash
teardown() {
    while IFS= read -r pidfile; do
        pid=$(cat "$pidfile" 2>/dev/null || true)
        if [[ "$pid" =~ ^[0-9]+$ ]]; then
            kill "$pid" 2>/dev/null || true
            for _ in {1..50}; do
                kill -0 "$pid" 2>/dev/null || break
                sleep 0.1
            done
            kill -9 "$pid" 2>/dev/null || true
        fi
        rm -f "$pidfile"
    done < <(find "$BATS_TEST_TMPDIR" -name '*.pid' -type f)

    kill_pid_files
}
```

Also replace the setup sleeps in [`macaroon.bats:19-24`](https://github.com/diogomattioli/xrootd/blob/becf9202c2cf915e07183b456734ec5e1cc59622/tests/end-to-end/tpc/http/macaroon.bats#L19-L24) with a readiness-aware `launch_xrootd`. This addresses the actual failure rather than weakening the token test or marking the job as allowed to fail.

amadio:
Squashed all previous commits into one locally, and added a separate commit to fix the problem with the `sleep` commands.

amadio:
Added commits to address the problem that push without `--force` would overwrite the destination. Addresses CodeRabbit comments.

src/XrdClHttp/XrdClHttpOpCopy.cc
```diff
@@ -75,23 +87,35 @@ CurlCopyOp::CurlCopyOp(XrdCl::ResponseHandler *handler, const std::string &sourc
...
                 me->HandleLine(cur_line);
             } else {
                 me->m_line_buffer += cur_line;
                 me->HandleLine(me->m_line_buffer);
                 me->m_line_buffer.clear();
             }
+
+            if (end_line == str_data.size())
+                break;
+
```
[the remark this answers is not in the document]
amadio: Will be resolved in a separate pull request.

amadio (approved):

PR #2887, opened 2026-08-07
[CLI] Sort xrdmapc JSON managers and servers by name

dynamic-entropy:


amadio (approved):
Let's add to 6.2 due to the change in behavior (and since it will come soon), but it's fine otherwise.

PR #2888, opened 2026-08-10
[XrdCl] [XrdClHttp] Param --header -H to inject http headers

diogomattioli:
This PR adds `--header` functionality, allowing HTTP(S) headers to be injected into HTTP(S) transfers. Injected headers are supported only for HTTP(S) connections.

A blacklist filter is also implemented to prevent the injection of specific sensitive headers. All injected headers are appended to the end of the existing header list, ensuring that they do not replace headers added through other mechanisms.

This PR also repurposes the `-H` option, which was previously used to display the license, to maintain compatibility with `curl`.

Assisted-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/XrdApps/XrdCpConfig.cc
+++ b/src/XrdApps/XrdCpConfig.cc
@@ -76,11 +76,21 @@ namespace XrdCpConfiguration
 {
 static XrdSysLogger Logger;
 static XrdSysError  eDest(&Logger, "");
+
+// Returns true if the protocol is one that carries HTTP requests.
+//
+static bool isHttpProt(XrdCpFile::PType pType)
+                      {return pType == XrdCpFile::isHttp
+                           || pType == XrdCpFile::isHttps
+                           || pType == XrdCpFile::isS3;
+                      }
 };
+
+using XrdCpConfiguration::isHttpProt;
 
 XrdSysError  *XrdCpConfig::Log = &XrdCpConfiguration::eDest;
 
-const char   *XrdCpConfig::opLetters = ":C:d:D:EfFhHI:NpPrRsS:t:T:vVX:y:z:ZA";
+const char   *XrdCpConfig::opLetters = ":C:d:D:EfFhH:I:NpPrRsS:t:T:vVX:y:z:ZA";
 
 struct option XrdCpConfig::opVec[] =         // For getopt_long()
      {
@@ -90,6 +100,7 @@ struct option XrdCpConfig::opVec[] =         // For getopt_long()
       {OPT_TYPE "debug",          1, 0, XrdCpConfig::OpDebug},
       {OPT_TYPE "dynamic-src",    0, 0, XrdCpConfig::OpDynaSrc},
       {OPT_TYPE "force",          0, 0, XrdCpConfig::OpForce},
+      {OPT_TYPE "header",         1, 0, XrdCpConfig::OpHttpHeader},
       {OPT_TYPE "help",           0, 0, XrdCpConfig::OpHelp},
       {OPT_TYPE "infiles",        1, 0, XrdCpConfig::OpIfile},
       {OPT_TYPE "license",        0, 0, XrdCpConfig::OpLicense},
@@ -241,6 +252,9 @@ do{while(optind < Argc && Legacy(optind)) {}
           case OpZip:           OpSpec |= DoZip;
                                 if (zipFile) free(zipFile);
                                 zipFile = strdup(optarg);
+                                break;
+          case OpHttpHeader:    OpSpec |= DoHttpHeader;
+                                HttpHeaders.push_back(optarg);
                                 break;
           case OpHelp:          Usage(0);
                                 break;
@@ -430,6 +444,13 @@ do{while(optind < Argc && Legacy(optind)) {}
    if (isLcl && Opts & optNoLclCp)
       FMSG("All files are local; use 'cp' instead!", 1);
 
+// Injected headers only apply to http and https endpoints
+//
+   if (OpSpec & DoHttpHeader)
+      {if (!isHttpProt(srcFile->Protocol) && !isHttpProt(dstFile->Protocol))
+          UMSG("'--header' requires an http or https source or destination.");
+      }
+
 // Check for checksum spec conflicts
 //
    if (OpSpec & DoCksum)
@@ -465,7 +486,7 @@ do{while(optind < Argc && Legacy(optind)) {}
                          <<(numFiles != 1 ? " files." : " file."));
       }
 }
-  
+
 /******************************************************************************/
 /*                       P r i v a t e   M e t h o d s                        */
 /******************************************************************************/
@@ -930,8 +951,8 @@ void XrdCpConfig::Usage(int rc)
 
    static const char *Options= "\n"
    "Options: [--cksum <args>] [--coerce] [--continue]\n"
-   "         [--debug <lvl>] [--dynamic-src] [--force] [--help]\n"
-   "         [--infiles <fn>] [--license] [--nopbar] [--notlsok]\n"
+   "         [--debug <lvl>] [--dynamic-src] [--force] [--header <name: value>]\n"
+   "         [--help] [--infiles <fn>] [--license] [--nopbar] [--notlsok]\n"
    "         [--parallel <n>] [--posc] [--proxy <host>:<port>]\n"
    "         [--recursive] [--retry <n>] [--retry-policy <force|continue>]\n"
    "         [--rm-bad-cksum] [--server] [--silent] [--sources <n>]\n"
@@ -969,9 +990,28 @@ void XrdCpConfig::Usage(int rc)
    "-d | --debug <lvl>            sets the debug level: 0 off, 1 low, 2 medium, 3 high\n"
    "-Z | --dynamic-src            file size may change during the copy\n"
    "-f | --force                  replaces any existing output file\n"
+   "-H | --header <name: value>   adds the given header to the requests sent to an\n"
+   "                              http or https endpoint. May be repeated to add more\n"
+   "                              than one header. Requires an http or https source or\n"
+   "                              destination. The hop-by-hop and message framing\n"
+   "                              headers are reserved and cannot be overridden.\n"
+   "                              With --tpc, the headers go to the server the client\n"
+   "                              connects to: the destination server in 'pull' mode,\n"
+   "                              or the source server in 'push' mode. To send a header\n"
+   "                              to the other server, prepend 'TransferHeader' to the\n"
+   "                              header name. Example, in 'pull' mode:\n"
+   "                                -H \"X-Request-Id: <destination-id>\"\n"
+   "                                -H \"TransferHeaderX-Request-Id: <source-id>\"\n"
+   "                              The destination server gets the X-Request-Id header,\n"
+   "                              and sends the value of TransferHeaderX-Request-Id as\n"
+   "                              the X-Request-Id header of its request to the source.\n"
+   "                              If the option is not specified, the XRD_HTTPHEADERS\n"
+   "                              environment variable is used instead. It holds the\n"
+   "                              headers as a newline separated list of\n"
+   "                              '<name>: <value>' entries.\n"
    "-h | --help                   prints this information\n"
    "-I | --infiles <fname>        specifies the file that contains a list of input files\n"
-   "-H | --license                prints license terms and conditions\n"
+   "     --license                prints license terms and conditions\n"
    "-N | --nopbar                 does not print the progress bar\n"
    "     --notlsok                if server is too old to support TLS encryption fallback\n"
    "                              to unencrypted communication\n"
--- a/src/XrdClHttp/XrdClHttpHeaderBuilder.cc
+++ b/src/XrdClHttp/XrdClHttpHeaderBuilder.cc
@@ -1,0 +1,208 @@
+#include "XrdClHttpHeaderBuilder.hh"
+#include "XrdClHttpUtil.hh"
+
+#include <XrdCl/XrdClLog.hh>
+#include <XrdCl/XrdClDefaultEnv.hh>
+
+#include <algorithm>
+#include <array>
+#include <iterator>
+
+namespace {
+
+using namespace std::string_view_literals;
+
+// Headers the client must not inject: the RFC 7230 hop-by-hop and message
+// framing headers, and Host.  Overriding one permits request smuggling, cache
+// poisoning, or corrupt framing.
+//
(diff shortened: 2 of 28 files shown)


src/XrdApps/XrdCpConfig.cc
```diff
@@ -76,11 +76,20 @@ namespace XrdCpConfiguration
 {
 static XrdSysLogger Logger;
 static XrdSysError  eDest(&Logger, "");
+
+// Returns true if the protocol is one that carries HTTP requests.
+//
+static bool isHttpProt(XrdCpFile::PType pType)
+                      {return pType == XrdCpFile::isHttp
+                           || pType == XrdCpFile::isHttps;
```
abh3: Likely not necessary but it is allowed to add headers to s3 protocol so you may want to include that.

src/XrdApps/XrdCpConfig.cc
```diff
@@ -242,6 +252,9 @@ do{while(optind < Argc && Legacy(optind)) {}
                                 if (zipFile) free(zipFile);
                                 zipFile = strdup(optarg);
                                 break;
+          case OpHttpHeader:    OpSpec |= DoHttpHeader;
+                                HttpHeaders.push_back(optarg);
```
abh3, line 257: I suppose it's a matter of preference. Here you could at least check whether the specification is malformed and complain. As it stands, the check is relegated to the http plug-in which may or may not complain (i.e. it may simply ignore a bad one). A quick check didn't reveal any robust header validation.

src/XrdApps/XrdCpConfig.cc
```diff
@@ -969,9 +989,28 @@ void XrdCpConfig::Usage(int rc)
...
+   "                              and sends the value of TransferHeaderX-Request-Id as\n"
+   "                              the X-Request-Id header of its request to the source.\n"
+   "                              If the option is not specified, the XRD_HTTPHEADERS\n"
+   "                              environment variable is used instead. It holds the\n"
+   "                              headers as a newline separated list of\n"
+   "                              '<name>: <value>' entries.\n"
    "-h | --help                   prints this information\n"
    "-I | --infiles <fname>        specifies the file that contains a list of input files\n"
-   "-H | --license                prints license terms and conditions\n"
+   "     --license                prints license terms and conditions\n"
```
abh3, line 1014: Normally, we don't repurpose pre-existing options. In this case it seems rather benign.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,211 @@
...
+    "host"sv,
+    "keep-alive"sv,
+    "proxy-authenticate"sv,
+    "proxy-authorization"sv,
+    "proxy-connection"sv,
+    "te"sv,
+    "trailer"sv,
+    "transfer-encoding"sv,
+    "upgrade"sv
+};
```
abh3, line 34: This does not appear to be the full list of forbidden headers. A quick ask of Gemini revealed a few more that should be in the list. I think Claude can easily fill in the gaps.

abh3 (approved):
Didn't see anything that is considered a roadblock but some improvements should be considered. However, I'll approve it modulo addressing comments.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
amadio: Maybe it's better to have a list of allowed/supported headers instead of forbidden ones? The `--header` flag will most likely only be used with TPC to add authorization headers for source/destination, right?

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,211 @@
...
+
+} // namespace
+
+namespace XrdClHttp {
+namespace HeaderBuilder {
+
+bool
+IsForbiddenHeader(const std::string_view name)
+{
+    std::string lowered(name.size(), '\0');
```
amadio: If you are going to make a copy right away, you may as well take in a `std::string` instead of `std::string_view`. Rather than transform to lower, you may want to compare using a case-insensitive function (e.g. `stdcasecmp`) to avoid having to copy stuff, or do it like you did for `SameHeaderName` below.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
abh3: The problem with that approach is that it essentially forbids custom servers that accept weird headers from using our client since you need those added to the list. I don't think it's good to lock in a particular architecture relative to the client.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,211 @@
...
+    const auto discard = [strip](const char c) {
+        return strip.find(c) != std::string_view::npos;
+    };
+
+    const auto first = std::find_if_not(text.begin(), text.end(), discard);
+    const auto last = std::find_if_not(text.rbegin(),
+        std::make_reverse_iterator(first), discard).base();
+
+    return text.substr(first - text.begin(), last - first);
+}
```
amadio, line 62: We already have several `trim` implementations scattered in the code, I'd prefer to consolidate this type of function into a single place rather than add more variations. It makes maintenance harder, as you may be tricked by which `trim` is being actually used in some piece of code.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
abh3: Interestingly, I thought the same thing. But in my research I found that for this kind of stuff one should avoid using strcasecmp(); in fact that was eliminated via this pull request. The problem appears that strcasecmp() is required to follow the locale setting which may allow introduction of weird characters in the header and produce unexpected results. The advice is to use something that only allows POSIX characters, which this pull request does.

src/XrdClHttp/XrdClHttpHeaderBuilder.hh
```diff
@@ -0,0 +1,44 @@
...
+#include <vector>
+
+namespace XrdClHttp {
+
+// Builds the headers a client asks a request to carry.  The request is
+// described by the HttpHeaders environment setting, which xrdcp fills in from
+// its --header option.  These functions are the only place it is interpreted.
+namespace HeaderBuilder {
+
+using HeaderList = std::vector<std::pair<std::string, std::string>>;
```
amadio, line 16: Why use a `vector<pair<string,string>>` over a `map<string,string>` like other parts of the code do? As `std::strings` are normally on the heap anyway, so using `std::vector` doesn't buy us much in this case.

tests/end-to-end/http/header/option.bats
```diff
@@ -0,0 +1,31 @@
...
+	run ! xrdcp --header
+}
+
+#
+# options -H took over
+#
+
+@test "--license is still available after -H was reassigned" {
+	run xrdcp --license
+	assert_output --partial 'Copyright'
```
amadio: I think `assert_output --partial 'GNU Lesser General Public License'` (or `LGPL`) is more to the point here, but this is ok.

tests/end-to-end/http/header/env.bats
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
```
amadio, line 1: I don't see any of these tests being connected to CMake, so they are not running in the CI. They need to be connected.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
amadio: Ok, but then the comparison can still be done without a copy, like in `SameHeaderName` below...

amadio (changes requested):
The end-to-end tests are not running in the CI. Before merging, they need to run like the other tests. The other comments can be addressed post-merge if needed.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
abh3: Agreed.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
amadio: Ok, good point, that might also be a problem for Pelican now that I think of it more. So, better keep it as it is then.

src/XrdApps/XrdCpConfig.cc
diogomattioli: Sure. We're definitely aware that repurposing should be the last resort, but in this case wouldn't cause harm since `-H` only shows the license and the benefits are the similarity with curl.

src/XrdClHttp/XrdClHttpHeaderBuilder.hh
diogomattioli: In that scenario `vector` or `unordered_map` will basically work the same. The number of entries is insignificant to really weight any performance here.

However, even if there was a huge amount of entries, retrieving a header value by its header name (key) never happens here. Basically the set of headers is traversed searching for forbidden headers, while we consider the string case. So, getting the forbidden headers and querying the set is not feasible (due the case), and getting the headers set list is exactly the same as using a vector, the difference is that the vector is contiguous.

For appending headers to the set, both would have `O(1)`.

src/XrdApps/XrdCpConfig.cc
diogomattioli: I chose to have more cohesion and leave all the header checking under `XrdClHttp` plugin. Thus, not spreading functionality over different parts of the system. I'd prefer to maintain like that unless anybody has a strong reason not to.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
diogomattioli: This trim version seems to be more complete than the already offered ones as it's possible to trim by specific characters. My suggestion would be to move/replace the existing implementation by this one in a near future.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
diogomattioli: My suggestion would be to keep this list as it is for now, and in the near future have some more research on the total forbidden headers in order to not impact our customers.

diogomattioli:
> The end-to-end tests are not running in the CI. Before merging, they need to run like the other tests. The other comments can be addressed post-merge if needed.

All comments were addressed. If there're no further changes, feel free to merge.

src/XrdClHttp/XrdClHttpHeaderBuilder.hh
amadio: It's a matter of preference, but a map might be more "ergonomic" when searching, inserting, etc. When I see code with a lot of `.first`, `.second`, etc, I try to think of ways to eliminate it, it's not totally a code smell, but if you manage to write code without, it's in general more readable. Anyway, it's a matter of preference, this is fine if you want to keep it.

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
diogomattioli: done!

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
amadio: Ok, for the sake of time, let's merge this as is, but we have several `trim` implementations in the code, and I'd like to centralize them and use the same implementation everywhere later on, to be decided when exactly, but hopefully in 6.3 or 6.4.

tests/end-to-end/http/header/env.bats
diogomattioli: done!

src/XrdApps/XrdCpConfig.cc
diogomattioli: done!

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,208 @@
...
+        const auto value = trim(entry.substr(colon + 1), ows);
+        if (value.empty()) {
+            log->Error(kLogXrdClHttp, "Requested header %s holds no value",
+                name.c_str());
+            usable = false;
+            continue;
+        }
+        if (value.find('\r') != std::string_view::npos) {
+            log->Error(kLogXrdClHttp, "Requested header %s holds a carriage"
+                " return in its value", name.c_str());
```
amadio: Breaking strings in two lines breaks `grep` (i.e. grepping for the error message in the code to find it, for example, `grep "carriage return"` would break), I think it's better to tolerate a slightly longer line here.

diogomattioli: done!

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,202 @@
...
+    })
+        != forbidden_headers.end();
+}
+
+bool
+CompareIgnoreCase(const std::string_view lhs, const std::string_view rhs)
+{
+    return std::equal(lhs.begin(), lhs.end(), rhs.begin(), rhs.end(),
+        [](const unsigned char a, const unsigned char b) {
+            return std::tolower(a) == std::tolower(b);
```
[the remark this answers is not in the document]
diogomattioli: done!

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,202 @@
...
+            continue;
+        }
+
+        // The value may itself contain colons, so split on the first one only.
+        // Without a colon the entry carries no value, so the log can show the
+        // whole entry without showing a value the user keeps private.
+        const auto colon = entry.find(':');
+        if (colon == std::string_view::npos) {
+            log->Error(kLogXrdClHttp, "Requested header %s holds no colon; each header must read \"<name>: <value>\"",
+                std::string(entry).c_str());
```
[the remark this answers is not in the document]
diogomattioli: done!

src/XrdClHttp/XrdClHttpHeaderBuilder.cc
```diff
@@ -0,0 +1,202 @@
...
+
+        requested.emplace_back(name, value);
+    }
+
+    if (!usable) {
+        return false;
+    }
+
+    if (requested.empty()) {
+        return true;
```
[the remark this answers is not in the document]
diogomattioli: done!

amadio (approved):

PR #2893, opened 2026-08-11
Http fullurl redirect

dynamic-entropy:
`XrdHttpReq::Redir()` mishandles XRootD URL-form redirects (`port < 0`, host field is a full `scheme://host[:port]/path URL`).

It always prefixes http(s):// and appends the original resource (special-casing only file://), so other schemes get a doubled scheme and a duplicated path, e.g.:
```
https://ds.example:443//pfn/f.root
 ->  Location: https://https://ds.example:443//pfn/f.root/store/lfn
```

It worked internally, since `XrdCmsRedirLocal` stripped the prefix; this PR removes the bug and the workaround.
Additionally, the same block also picks xrdhttptk from http. desthttps instead of the destination URL scheme, and hashes the wrong path for the token.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved HTTP redirects to correctly handle absolute destination URLs, including their schemes, ports, paths, and query strings.
  - Prevented duplicated URL schemes and resource paths in redirect locations.
  - Corrected redirects for local resources and cleartext destinations.
  - Normalized repeated slashes in redirected resource paths while preserving encoded content.
  - Improved handling of file URLs, root paths, explicit ports, and destinations with empty or invalid paths.

### DIFF
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -852,18 +852,9 @@ void XrdHttpReq::parseResource(char *res) {
     // to the resource string. Here we choose to ignore it as a protection measure
     sanitizeResourcePfx();
 
-    std::string resourceDecoded = decode_str(resource.c_str());
-    resource = resourceDecoded.c_str();
+    const std::string resourceDecoded = decode_str(resource.c_str());
     resourceplusopaque = resourceDecoded.c_str();
-
-
-    // Sanitize the resource string, removing double slashes
-    int pos = 0;
-    do {
-      pos = resource.find("//", pos);
-      if (pos != STR_NPOS)
-        resource.erase(pos, 1);
-    } while (pos != STR_NPOS);
+    resource = httpCollapseSlashes(resourceDecoded).c_str();
 
     return;
   }
@@ -877,15 +868,7 @@ void XrdHttpReq::parseResource(char *res) {
   // to the resource string. Here we choose to ignore it as a protection measure
   sanitizeResourcePfx();
 
-  resource = decode_str(resource.c_str()).c_str();
-
-  // Sanitize the resource string, removing double slashes
-  int pos = 0;
-  do {
-    pos = resource.find("//", pos);
-    if (pos != STR_NPOS)
-      resource.erase(pos, 1);
-  } while (pos != STR_NPOS);
+  resource = httpCollapseSlashes(decode_str(resource.c_str())).c_str();
 
   resourceplusopaque = resource;
   // Whatever comes after is opaque data to be parsed
--- a/src/XrdHttp/XrdHttpReq.cc
+++ b/src/XrdHttp/XrdHttpReq.cc
@@ -603,7 +603,6 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
 
 
 
-  char buf[512];
   char hash[512];
   hash[0] = '\0';
 
@@ -622,20 +621,15 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
     return keepalive;
   }
 
-  if (prot->isdesthttps)
-    redirdest = "Location: https://";
-  else
-    redirdest = "Location: http://";
-
-  // port < 0 signals switch to full URL
-  if (port < 0)
-  {
-    if (strncmp(hname, "file://", 7) == 0)
-    {
-      TRACE(REQ, " XrdHttpReq::Redir Switching to file:// ");
-      redirdest = "Location: "; // "file://" already contained in hname
-    }
+  // port < 0 means XRootD URL-form redirect: hname is a full scheme://… URL
+  // (host+port+path).
+  // Do not prefix http(s):// and do not append the original resource
+  // otherwise, that would produce https://https://…/pfn/lfn.
+  const bool fullUrl = (port < 0);
+  if (fullUrl) {
+    TRACE(REQ, " XrdHttpReq::Redir using absolute URL: " << hname);
   }
+
   // Beware, certain Ofs implementations (e.g. EOS) add opaque data directly to the host name
   // This must be correctly treated here and appended to the opaque info
   // that we may already have
@@ -643,25 +637,36 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
   char *vardata = 0;
   if (pp) {
     *pp = '\0';
-    redirdest += hname;
     vardata = pp+1;
     int varlen = strlen(vardata);
 
     //Now extract the remaining, vardata points to it
     while(*vardata == '&' && varlen) {vardata++; varlen--;}
+  }
 
-    // Put the question mark back where it was
+  // If the destination speaks cleartext HTTP, it needs the
+  // secret hash to reconstruct authorization info.
+  const bool destHttpCleartext =
+      fullUrl ? (strncmp(hname, "http://", 7) == 0) : (!prot->isdesthttps);
+
+  std::string urlHashPath;
+  {
+    const std::string encodedResource =
+        fullUrl ? std::string() : encode_str(resource.c_str());
+    redirdest = httpBuildRedirectLocation(prot->isdesthttps, port, hname,
+                                          encodedResource.c_str()).c_str();
+
+    // Derive the token path while hname is still truncated at '?', so the CGI
+    // stays out of the hash and both ends digest the same string.
+    if (fullUrl && destHttpCleartext) {
+      if (const char *urlPath = httpPathFromAbsoluteUrl(hname))
+        urlHashPath = httpCollapseSlashes(decode_str(urlPath));
+    }
+  }
+
+  // Restore '?' so later uses of hname see the original string.
+  if (pp)
     *pp = '?';
-  }
-  else
-    redirdest += hname;
-
-  if (port > 0) {
-    sprintf(buf, ":%d", port);
-    redirdest += buf;
-  }
-
-  redirdest += encode_str(resource.c_str()).c_str();
 
   // Here we put back the opaque info, if any
   if (vardata) {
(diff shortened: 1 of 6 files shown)


smithdh:
(I'm setting up a local test, e.g. with a cmsredirlocal enabled-redirector, to better review this)

src/XrdHttp/XrdHttpUtils.cc
```diff
@@ -685,3 +685,43 @@ std::string httpStatusToString(int status) {
       }
   }
 }
+
+const char *httpPathFromAbsoluteUrl(const char *hname) {
```
abh3, line 689: Sort of a weird name for this function. Does it not return "host:port<resource>" or some variation of it? That is not a path, not clear what to call it.

abh3 (commented):
I didn't find anything particularly wrong but the whole hashing scheme seems mysterious and need more explanation of why it is even needed in this context. I suppose it's because a token may reference the url path but even if it does how would you actually know at this point.

I'm just commenting and will wait until @smithdh creates a real test for this to see if it works.

src/XrdHttp/XrdHttpUtils.cc
dynamic-entropy: It does not return `host:port`, but the "path" (file path, lfn , pfn, whatever the redirect points to). 
This is called when `hname` is a full (absolute) URL (`scheme://host:port/path`), it returns the path starting at first `/`.

dynamic-entropy:
@smithdh Thankyou for looking at this. However, with a `xrdcmsredirlocal ` setup, one can only test that there was no behaviour change, which is supposed to be the case.  `XrdCmsRedirLocal` redirect with `file://` redirects are simply no longer a special case and http is treated alike. 

The actual fix is that URL-form (port < 0) redirects can now send a full URL (http(s)://host[:port]/path) without doubling the scheme or appending the original request path.

I have just realised that there is no single point in code where this contract is documented. 
A host port form with a `SFS_REDIRECT` has the following meaning:
```
1. port > 0 : host-port form 
2. port = 0 : default port
3. port < 0 : URL-form 
```

A good description https://github.com/xrootd/xrootd/blob/36dcf492764a6d2b309fbf82b27857f142b09463/src/XrdXrootd/XrdXrootdRedirHelper.hh#L111-L118 
@abh3 
Token handling is unchanged for `host:port` redirects (port >= 0): we still hash the original request resource. For URL-form `http://` redirects, we hash the path taken from that URL and not the original LFN.

The destination HTTP server recomputes the HMAC over the path it actually received. If we hashed the original resource while Location pointed at a different path, that check would fail.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -622,46 +621,52 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
...
-  else
-    redirdest += hname;
 
-  if (port > 0) {
-    sprintf(buf, ":%d", port);
-    redirdest += buf;
+  // If the destination speaks cleartext HTTP, it needs the
+  // secret hash to reconstruct authorization info.
+  const bool destHttpCleartext =
+      fullUrl ? (strncmp(hname, "http://", 7) == 0) : (!prot->isdesthttps);
```
dynamic-entropy, line 650: Hello @abh3 

David and I discussed this today. One of the changes this introduces is not to enforce the `isdesthttps` as documented during a full URL redirect. 

We wanted to know your opinion on this. Which direction should we take: augment the documentation describing the full URL redirect case or force redirection to `http` or `https` based on the value set for the directive?

```
http.desthttps {no | yes}

Function
Specify whether or not HTTPS is to be used for redirections.

Parameters
no       A redirector will always redirect a client using http. The word false and the number 0 are synonyms.
yes      A redirector will always redirect a client using https. The word true and the number 1 are synonyms.

Defaults
http.desthttps no

```
 https://xrootd.web.cern.ch/doc/dev6/xrd_config.html#_Toc220357742

abh3: Well, before I add my thoughts, I'd like to know the motivation for removing it. Presumably, if the URL contains authorization information (i.e. a token) you would want to use https and, in fact, you really should. If it doesn't then it does not matter. If we ignore 'desthttps no' (which is the default) then the notion of using a hash to reconstruct authorization is no longer needed. Is code simplification another motivation here?

dynamic-entropy: I am not aware of who uses "http" (clear text) redirects. 

In this PR, we only ignore 'desthttp no' when the scheme in a full-url redirect ( `http(s)://host:port`) is `https`, thus, we do not force it to become `http` and vice versa based on the `desthttps` configuration.

The situation remains the same for `host-port` redirects, where the scheme is determined by `desthttp yes|no` setting, and the redirector does not convey any particular protocol "scheme".

The inconsistency this change brings in the doc is about the use of the word "always".

abh3: Thanks you, that clarifies this. Now could you explain why you need the hashing scheme to carry forward authorization. How does it work and why is it necessary?

dynamic-entropy: I am not aware of a deployment using this today. However, if I have to guess (somewhat analogous to what EOS does, or a presigned URL), a setup where the clients connect to a HTTPS gateway, which then redirects to data servers that themselves do not terminate TLS.

Since the original TLS credentials cannot be shared over HTTP, the gateway signs the context of the request with a shared "secretkey" (between the dataservers an gateway). Thus, appending this hash, i.e `xrdhttptk`, along with some parameters from SecEntity with the cgi.
The dataserver can then recompute the HMAC and validate the integrity of the request.

smithdh: Hi. I've reviewd and tried to test the patch. i.e. using an xrd http(s) server with a plugin that sets various redirects and checked which "Location" (redirect) is issued. Looks ok.

+ The the non-url style, i.e. Resp.setErrInfo(port,hostname), it acts as before. i.e. issues a http or https (depending on http.desthttps setting) to the specified hostname, using the original filename. This is as previously. If the conneciton was https but it is redirected to http the url is signed. I checked a destination server does validate the signature as expected.

+ The url style. i.e. Resp.setErrInfo(-1, url), issues the "Location" header using the speified url. It is up to whatever calls setErrInfo to choose the scheme, hostname and filename etc. (http.desthttps is not consulted inthe xrd http server code, but presumably whatever calls setErrInfo with the redirect should have consulated it). As a special case if the scheme is "http" and the conneciton was "https" the xrd http server add the url signing. I checked a server does validate it as expected, even if the redirect filename is different from the original one. It might be worth saying I don't think we have any client which will be able to handle a 'root' to 'http(s)' or vice-versa redirect. The "file" redirect also works (ie.g. using CmsRedirectLocal plugin), but it now has to include the filename. (This was a change in behavior, and indeed the patch includes a change in CmsRedirectLocal plugin to do that).

dynamic-entropy: Thank you, David. Much appreciated. 

@abh3 have we answered your queries? Please let me know if you need more information.
Cheers

amadio:
@dynamic-entropy Can you please check the AI review comments, and if relevant, fix the reported issues? One of them (about http://) you can ignore, but the other one might be worth a look. Thanks.

dynamic-entropy:
@amadio Checked. 
They are all wrong. I have used this opportunity to rebase the changes on master.

src/XrdHttp/XrdHttpReq.cc
abh3: I think you have.

abh3 (approved):
I'll approve at this point but it still needs to resolve the issues that code rabbit found.

src/XrdHttp/XrdHttpReq.cc
```diff
@@ -622,46 +621,52 @@ bool XrdHttpReq::Redir(XrdXrootd::Bridge::Context &info, //!< the result context
...
+    const std::string encodedResource =
+        fullUrl ? std::string() : encode_str(resource.c_str());
+    redirdest = httpBuildRedirectLocation(prot->isdesthttps, port, hname,
+                                          encodedResource.c_str()).c_str();
+
+    // Derive the token path while hname is still truncated at '?', so the CGI
+    // stays out of the hash and both ends digest the same string.
+    if (fullUrl && destHttpCleartext) {
+      if (const char *urlPath = httpPathFromAbsoluteUrl(hname))
+        urlHashPath = httpCollapseSlashes(decode_str(urlPath));
```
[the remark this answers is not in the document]
amadio: @dynamic-entropy This one needs to be addressed, for the case when `urlPath` is `nullptr`. `decode_str` takes a reference to string, you can't pass in a null pointer to it.

amadio (changes requested):
Please replace `char*` with `std::string` in some of your helper functions. Right now, you are returning `nullptr` in `httpPathFromAbsoluteUrl`, then passing it into `decode_str` which takes a `std::string&`, so it will cause crashes.

dynamic-entropy:
> Please replace `char*` with `std::string` in some of your helper functions. Right now, you are returning `nullptr` in `httpPathFromAbsoluteUrl`, then passing it into `decode_str` which takes a `std::string&`, so it will cause crashes.

Yes, this is what the AI review said, but I do not know of a case where there is no path but an `SFS_REDIRECT`.

Having said that I am not against make this change, it does not effect the logic.

src/XrdHttp/XrdHttpReq.cc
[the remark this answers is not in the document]
dynamic-entropy: Is it not within a `if` statement?

dynamic-entropy (commented):
Looking at this again, there is nothing I see that needs to be done. 

The only situation that requires thought, is redirecting from `https` to `http` such that the redirected path is `/`, this would mean the redirected `xrdhttptk` has the hash from the original resource instead of `/`. 

I will add two lines for that to satisy you, while noting that this is not an actual issue.

amadio (approved):
Tests could have been a bit more granular, but now this is good to go. I will merge, we will work on refaactoring the whole test suite later in any case. Thanks!

src/XrdHttp/XrdHttpReq.cc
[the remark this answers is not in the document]
amadio: Yes, but better to prevent bugs and avoid using `char*` in general. I think the final version is definitely an improvement. Thanks!

PR #2894, opened 2026-08-18
[Tests] Improve cluster test robustness and retry resilience

lobis:
### Summary

This PR improves the reliability of the `XRootD::cluster::test` test suite in CI:

1. **Bounded Random File Sizes**: Caps the generated random file sizes in `tests/cluster/test.sh` to 1–128 KB (instead of unbounded `$RANDOM` values producing up to 33.5 MB). Large payloads can saturate socket write buffers during HTTP PUT redirects through the cluster meta-manager (`metaman`), triggering transient `ECONNRESET` / `curl (55)` failures before the 307 redirect is processed. A dedicated large-file test (`2GB.dat`) already exists in `setup.sh`.
2. **Curl Retry Flags**: Adds `--retry 3 --retry-all-errors --retry-delay 1` to `curl` upload and MOVE requests in `test.sh` to tolerate transient network connection resets during multi-process cluster startup under heavy CI load.
3. **CTest Retries**: Adds `REPEAT UNTIL_PASS:2` test property to `XRootD::cluster::test` in CMake to protect against occasional scheduling/port race conditions in CI.

## Summary by CodeRabbit

* **Tests**
  * Updated generated reference files to use a bounded random size between 1 and 128 KiB, improving consistency during test runs.

### DIFF
--- a/tests/cluster/test.sh
+++ b/tests/cluster/test.sh
@@ -182,7 +182,7 @@ ${XRDFS} ${HOST_METAMAN} spaceinfo /
 # create local files with random contents using OpenSSL
 
 for host in "${!hosts[@]}"; do
-       ${OPENSSL} rand -out "${LCLDATADIR}/${host}.ref" $((1024 * ($RANDOM + 1)))
+       ${OPENSSL} rand -out "${LCLDATADIR}/${host}.ref" $((1024 * ($RANDOM % 128 + 1)))
 done
 
 # upload local files to the servers in parallel


tests/cluster/CMakeLists.txt
```diff
@@ -21,7 +21,10 @@ set_tests_properties(XRootD::cluster::start
 add_test(NAME XRootD::cluster::test
   COMMAND sh -c "${CMAKE_CURRENT_SOURCE_DIR}/test.sh")
 set_tests_properties(XRootD::cluster::test
-  PROPERTIES ENVIRONMENT "${XRDENV}" FIXTURES_REQUIRED XRootD_Cluster)
+  PROPERTIES
+    ENVIRONMENT "${XRDENV}"
+    FIXTURES_REQUIRED XRootD_Cluster
+    REPEAT UNTIL_PASS:2)
```
amadio: No retries, please. The point of the test suite is to catch errors, and this just hides them.

amadio:
I fixed recently two problems that were noticed by sporadic test failures. See b00acdefa954c3abc61a107f7f229020a6bc85d8, and #2901. If we add retries, these things will be hidden and never fixed.

lobis:
> I fixed recently two problems that were noticed by sporadic test failures. See [b00acde](https://github.com/xrootd/xrootd/commit/b00acdefa954c3abc61a107f7f229020a6bc85d8), and #2901. If we add retries, these things will be hidden and never fixed.

Without retry logic this PR becomes a single line change (which I still think would improve the testing).

As described in the body of this PR, by using smaller files we should be less likely to hit this failure mode.

amadio:
From the commit message my take is that even this one line should be converted into a fix for the described problem. The test is just fulfilling its purpose of finding a problem. If the connection resets are real, they must be happening in CMS AAA, so we should try to understand better the underlying issue and fix that instead of making the test files smaller.

PR #2896, opened 2026-08-19
XrdFfs/xrootdfs upgrade from FUSE 2.6 to FUSE 3.x

### LINKED ISSUE #2423: Migrate Fuse module to Fuse 3
xrootd-fuse currently uses Fuse 2 which is unmaintained. Consider migrating to Fuse 3.

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1094464 for more information.

sarahyang0:
Closes #2423

Migrates the xrootdfs FUSE client from FUSE 2.6 to FUSE 3.x API (getattr/chmod/chown/utimens/readdir/init/rename signature changes, ftruncate merged into truncate, invalid mount option removed) and updates CMakeLists.txt to detect libfuse3 via pkg-config.

Tested end-to-end in Docker: mounted against a real xrootd server and confirmed ls, df, cp, and rm work through the FUSE 3.x mount.

### DIFF
--- a/src/XrdFfs/CMakeLists.txt
+++ b/src/XrdFfs/CMakeLists.txt
@@ -24,17 +24,23 @@ target_link_libraries(XrdFfs
 install(TARGETS XrdFfs LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
 # FUSE is only supported on Linux and GNU/FreeBSD
+#
+# Converted from FUSE 2.x to FUSE 3.x: Xrootd's find_package(fuse) module
+# (cmake/FindFUSE.cmake) resolves to libfuse 2.x and there is no bundled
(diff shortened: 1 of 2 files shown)


wyang007:
thanks Sarah !

amadio (changes requested):
Hi, this pull request is only adding quite minimal changes to the existing FUSE 2.x `xrootdfs`, ignoring all the new things (i.e. new parameters and functionality in FUSE 3 API). This is not really how we'd like to go about it. There are existing tests in `tests/XRootD/fuse.{cfg,sh}` which are disabled because they currently fail, and with these minimal changes, it will continue to fail. What I'd like to have is an `xrootdfs` that works well enough that I can use it to mount EOS in the CI infrastructure and publish new releases directly there, so I can finish up the job of release automation. So, a test in Docker is not enough, the new `xrootdfs` will have to pass the existing tests as well as more tests before it will be accepted into the main branch.

PR #2897, opened 2026-08-21
[License] Normalize XRootD copyright headers

lobis:
Normalize license usage


amadio:
We can't change copyright, only the original contributor (i.e. Brian) can re-assign that. Let's close this one.

PR #2899, opened 2026-08-24
[Tests] Make the cluster test re-runnable

amadio:
XRootD::cluster::test failed on every run after the first against the same build tree, deterministically. CI never saw it: test.cmake wipes build/ with ctest_empty_binary_directory() before every configure and calls ctest_test() once, so the suite only ever runs on a virgin tree. It bites a developer running ctest --test-dir build twice, which is the normal inner loop.

Three separate reasons, all in the same file.

The redirect test uploaded cluster_redirfile{1,2,3} with no -f and nothing ever removed them, so a successful run guaranteed the next one died at set -e with "Invalid operation: file exists". Everything else was cleaned inline at the end of the happy path only: cleanup() was wired as `trap "cleanup; exit 1" ABRT`, and ABRT is not what set -e, an explicit exit 1 from a failed assertion, a ctest timeout or Ctrl-C sends, so one failure left both localdata/ and the servers dirty and every later run failed at the first upload or download -- neither of which passed -f either. It could not recover on its own.

Move the cleanup into a trap on EXIT INT TERM QUIT (EXIT rather than ERR because the assertions here exit explicitly), widen it to cover every path the test writes, and pass -f to the uploads and downloads. The -f is what actually guarantees re-runnability: a SIGKILLed or timed-out run executes no trap at all, so tolerating leftovers matters more than not leaving them. The generated fixtures under data/<srv>/data are deliberately untouched -- setup.sh spends a 2GB openssl rand and 4000 uuidgen calls on them and caches them for the life of the build tree.

/tmp/err.txt, where the redirect assertions captured stderr, is also written by tests/badredir/test.sh. The two sit behind different CTest fixtures with no RUN_SERIAL, so ctest runs them concurrently under -j and each clobbered the other's capture before grep read it. That was the one genuinely random failure in here; it is now an mktemp.

Both `query config` loops iterated "${!hosts[@]}" twenty-five lines above `declare -A hosts`. Bash expands an unset array to nothing, so all 28 version, role and sitename assertions across the seven hosts have silently not run since the file was added in 2fd4e6773b. Declaring the tables first revives them; they pass.

setup.sh gets the matching treatment: generate() keyed its "already built" check on ${DATAFOLDER}/${i} with ${i} unset, i.e. on ./data existing, so a generation interrupted part way was cached as complete forever -- it now writes a sentinel last and rebuilds when it is missing. stop() killed pids from pidfiles blindly under set -e, so a crashed instance made the fixture cleanup itself fail; it now checks liveness the way tests/XRootD/test.sh does. And start() calls stop() first, as tests/authenticated_cluster/setup.sh already did, so a half-killed cluster does not collide on the ports.

Verified: three consecutive XRootD::cluster runs pass; a run planted with the exact wedging leftovers (cluster_redirfile*, srvdata/*.ref, localdata/*.dat) passes and leaves nothing behind; a run SIGKILLed part way leaves state that the next run recovers from; and the full suite passes 679/679 twice in a row, which it did not before. tests/authenticated_cluster and tests/TPCTests were checked for the same defect and pass twice in a row unmodified.

Assisted-by: Claude:Opus 5

### DIFF
--- a/tests/cluster/setup.sh
+++ b/tests/cluster/setup.sh
@@ -61,11 +61,16 @@ formatfiles() {
 
 generate(){
 
-       # check if files are in the data directory already...
-       if [[ -e ${DATAFOLDER}/${i} ]]; then
+       # The fixtures below cost a 2GB openssl rand, 4x1000 uuidgen calls and
+       # ten 16MB files, so they are generated once and cached for the life of
+       # the build tree. The sentinel is written last, which is the point: an
+       # interrupted generation leaves no sentinel and is redone, where testing
+       # for the data directory itself would treat a half-built tree as done.
+       if [[ -e ${DATAFOLDER}/.generated ]]; then
               return
        fi
 
+       rm -rf ${DATAFOLDER}
        mkdir -p ${TMPDATAFOLDER}
 
        for i in ${filenames[@]}; do
@@ -144,9 +149,14 @@ generate(){
        done
 
        rm -rf ${TMPDATAFOLDER}
+
+       touch ${DATAFOLDER}/.generated
 }
 
 start(){
+       # A previous run killed part way through leaves its servers holding the
+       # ports and its instance directories on disk.
+       stop
        generate
        set -x
        # start for each component
@@ -162,11 +172,20 @@ start(){
        sleep 1
 }
 
+# Signal only pids that are still alive. A crashed or SIGKILLed instance leaves
+# a stale pidfile behind, and killing it blindly under `set -e` would abort the
+# script -- which, since this is the fixture *cleanup*, reports the whole
+# cluster teardown as failed and leaves the instance directories in place.
 stop() {
 	for i in "${servernames[@]}"; do
 		if [[ -d "${i}" ]]; then
-			kill -s TERM $(cat ${i}/cmsd.pid)
-			kill -s TERM $(cat ${i}/xrootd.pid)
+			for pidfile in "${i}"/cmsd.pid "${i}"/xrootd.pid; do
+				test -s "${pidfile}" || continue
+				pid="$(ps -o pid= "$(cat "${pidfile}")" || true)"
+				if test -n "${pid}"; then
+					kill -s TERM "${pid}"
+				fi
+			done
 			rm -rf "${i}"
 		fi
 	done
--- a/tests/cluster/test.sh
+++ b/tests/cluster/test.sh
@@ -43,6 +43,121 @@ done
 
 set -e
 
+# The host tables have to be declared before anything iterates them: bash
+# expands an unset array to nothing, so a loop placed above these silently does
+# not run at all (which is what happened to the query-config checks below).
+
+# hostname-address pair, so that we can keep track of files more easily
+declare -A hosts
+hosts["metaman"]="${HOST_METAMAN}"
+hosts["man1"]="${HOST_MAN1}"
+hosts["man2"]="${HOST_MAN2}"
+hosts["srv1"]="${HOST_SRV1}"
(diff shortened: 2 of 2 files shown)


tests/cluster/setup.sh
```diff
@@ -162,11 +172,20 @@ start(){
...
+# script -- which, since this is the fixture *cleanup*, reports the whole
+# cluster teardown as failed and leaves the instance directories in place.
 stop() {
 	for i in "${servernames[@]}"; do
 		if [[ -d "${i}" ]]; then
-			kill -s TERM $(cat ${i}/cmsd.pid)
-			kill -s TERM $(cat ${i}/xrootd.pid)
+			for pidfile in "${i}"/cmsd.pid "${i}"/xrootd.pid; do
+				test -s "${pidfile}" || continue
+				pid="$(ps -o pid= "$(cat "${pidfile}")" || true)"
```
lobis, line 184: The text below has been redacted by AI and reviewed by me:

This check only proves that *some* process currently owns the numeric PID. The recovery case described here is a SIGKILL, which leaves the pidfile behind indefinitely; once that PID is reused, the new `start()` calls `stop()` and can send `TERM` to an unrelated developer process.

Please verify that the process belongs to this fixture before killing it (expected `xrootd`/`cmsd` executable plus this instance's `-n`/configuration arguments). The race where it exits between validation and `kill` should also be non-fatal.

Since `stop()` now runs immediately before starting replacements, it should additionally wait for each verified process to exit, with a bounded timeout. `TERM` is asynchronous, so deleting the instance directory and launching immediately can still race with the old daemon holding its port.

tests/cluster/test.sh
```diff
@@ -104,29 +112,74 @@ assert_xrdmapc_json() {
...
+       done
+
+       for file in "${!redirfiles[@]}"; do
+              ${XRDFS} ${redirfiles[$file]} rm /${file} 2>/dev/null || :
        done
+
+       ${XRDFS} ${HOST_METAMAN} rmdir ${RMTDATADIR} 2>/dev/null || :
 }
-trap "cleanup; exit 1" ABRT
+trap cleanup EXIT INT TERM QUIT
```
lobis: The text below has been redacted by AI and reviewed by me:

The `EXIT` trap is appropriate, but attaching `cleanup` directly to `INT`, `TERM`, and `QUIT` replaces Bash's default terminating behavior. After the handler returns, the shell can continue; `set -e` only helps when the interrupted command itself returns a relevant non-zero status.

For example:

```bash
set -e
state=$(mktemp -d)
cleanup() { echo cleanup; rm -rf "$state"; }
trap cleanup EXIT INT TERM QUIT

touch "$state/test-state"
kill -TERM "$$"

echo "continued after TERM"
test ! -d "$state"
```

This prints `continued after TERM`, runs `cleanup` twice, and exits successfully. The test could therefore continue after its local and remote state has already been removed.

Could we keep `trap cleanup EXIT` and make the signal handlers exit explicitly, for example:

```bash
trap cleanup EXIT
trap 'exit 130' INT
trap 'exit 143' TERM
trap 'exit 131' QUIT
```

The `EXIT` trap will then perform cleanup once while preserving a failure status appropriate for the received signal.

tests/cluster/setup.sh
amadio: As discussed in person, there's no easy way to prevent this from killing a reused process id. The only case where the pid is owned by the test suite is when everything works. Otherwise, on a re-run, the pid is no longer owned by the ctest parent process anymore.

tests/cluster/test.sh
amadio: Done.

lobis (approved):
LGTM!

PR #2901, opened 2026-08-25
[XrdClHttp] Rebase PUT deadline on each client write

amadio:
`CurlWriteFixture.ParallelTest` [failed intermittently in CI](https://my.cdash.org/tests/460170329) with `"[ERROR] Operation expired"`.

A chunked `PUT` is one curl operation that spans all `Write()`s the client issues. However, it was given a single absolute deadline computed at the first write and never re-based. `File::Write derives ts = GetHeaderTimeout(timeout)`, which is `min(header timeout, caller's timeout)` -- 9.5s by default -- and the `CurlPutOp` constructor turns that into an absolute deadline assigned only once. `Setup()` re-bases every other clock but not that one, and `ContinueHandle()` does no re-bases. Writes 2..N reached `QueueWrite()`, where ts was computed, but not used.

The origin sends no response header until the full body is complete, and m_received_header is never set for these uploads, so `HeaderTimeoutExpired()` stays armed from the first write to the final response. The whole upload of one file was therefore capped at 9.5s of wall clock however much data was still flowing. Ten 40KB writes 1.5s apart -- every gap far inside its own 10s timeout -- die at 10.5s with "Origin did not respond within timeout", and move to 3s exactly when `XRD_HTTPDEFAULTHEADERTIMEOUT=3s` is set.

Thread each write's own timeout through `QueueWrite()` to a new `CurlOperation::ExtendDeadline()`, which moves the deadline forward for each new `Write()`. All five call sites now pass it, including `Close()`'s final flush -- which is what makes the header timeout mean what it says: the time from the last body byte to the response headers. A wedged transfer is still bounded, by the stall and slow-rate detectors, which this doesn't change.

`GetOperationExpiry()` had the same defect one level down: it fell back to the header deadline whenever `m_last_xfer` was unset, without checking `m_received_header` the way `HeaderTimeoutExpired()` does, so the queue sweeper and `Produce()` could fail an operation on a deadline already known to be moot. It now anchors on `m_header_lastop + m_stall_interval` in that case.

`m_header_expiry` becomes atomic: it was write-once-before-publish, and `ExtendDeadline()` adds a client-thread writer against a worker-thread reader. The `BeginCallout()` sites take the expiration by non-const reference and now pass a local copy.

TransferTest printed `rv.ToString()`, which gives only the error-code name and drops the detail message -- seven distinct sites in the plugin raise `errOperationExpired` with seven distinct messages, so the CI log could not say which one fired. Use `rv.ToStr()`, and give the assertions that carried no message at all one.

`CurlWriteFixture.SlowClientWriteTest` is the regression guard: five writes a second apart against a per-file xrdclhttp.timeout of 3s (parsed as 2s), so the writes span twice the header timeout while every gap stays well inside it.

Also correct `m_default_header_timeout from {9, 5}` -- nine seconds and five nanoseconds -- to `{9, 500'000'000}`, matching what `Factory::Initialize` sets. It was harmless only because the factory overwrites it.

Assisted-by: Claude:Opus 5

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected HTTP header timeout handling.
  - Improved timeout behavior for delayed and queued PUT writes, helping prevent premature failures.
  - Preserved each queued write’s timeout settings and extended deadlines when additional data is pending.

- **Tests**
  - Added coverage for slow sequential writes and queued asynchronous writes.
  - Improved transfer-test failure messages with operation and file details.

### DIFF
--- a/src/XrdClHttp/XrdClHttpFile.cc
+++ b/src/XrdClHttp/XrdClHttpFile.cc
@@ -195,7 +195,7 @@ private:
 // Note: these values are typically overwritten by `CurlFactory::CurlFactory`;
 // they are set here just to avoid uninitialized globals.
 struct timespec XrdClHttp::File::m_min_client_timeout = {2, 0};
-struct timespec XrdClHttp::File::m_default_header_timeout = {9, 5};
+struct timespec XrdClHttp::File::m_default_header_timeout = {9, 500'000'000};
 struct timespec XrdClHttp::File::m_fed_timeout = {5, 0};
 
 
@@ -412,7 +412,7 @@ File::Close(XrdCl::ResponseHandler *handler,
             m_logger->Debug(kLogXrdClHttp, "Flushing final write buffer on close");
             auto put_handler = m_put_handler.load(std::memory_order_acquire);
             if (put_handler) {
-                return put_handler->QueueWrite(std::make_pair(nullptr, 0), handler);
+                return put_handler->QueueWrite(std::make_pair(nullptr, 0), handler, GetHeaderTimeout(timeout));
             } else {
                 m_logger->Error(kLogXrdClHttp, "Internal state error - put operation ongoing without handle");
                 return XrdCl::XRootDStatus(XrdCl::stError, XrdCl::errOSError);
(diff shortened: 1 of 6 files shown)


amadio:
@bbockelm Please have a look at the above, it may also need to be corrected in `xrdcl-pelican`. Cheers,

PR #2902, opened 2026-08-26
[Tests] Prefer bundled GoogleTest headers

lobis:
Keep the bundled GoogleTest headers ahead of include paths exported by other dependencies. On Homebrew systems, nlohmann_json exposes the common package prefix and may otherwise select incompatible system GoogleTest headers while linking the bundled libraries.

Make the ordering apply centrally to every test target so individual test directories do not need their own workaround.

### DIFF
--- a/vendor/googletest/googlemock/CMakeLists.txt
+++ b/vendor/googletest/googlemock/CMakeLists.txt
@@ -103,12 +103,27 @@ else()
 endif()
 
 string(REPLACE ";" "$<SEMICOLON>" dirs "${gmock_build_include_dirs}")
-target_include_directories(gmock SYSTEM INTERFACE
-  "$<BUILD_INTERFACE:${dirs}>"
-  "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_INCLUDEDIR}>")
-target_include_directories(gmock_main SYSTEM INTERFACE
-  "$<BUILD_INTERFACE:${dirs}>"
-  "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_INCLUDEDIR}>")
+if(NOT MSVC AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  # Keep the build-tree headers ahead of unrelated system prefixes while
+  # retaining system-header diagnostics behavior for Google Mock itself.
+  target_include_directories(gmock BEFORE INTERFACE
+    "$<BUILD_INTERFACE:${dirs}>")
+  target_include_directories(gmock SYSTEM BEFORE INTERFACE
+    "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_INCLUDEDIR}>")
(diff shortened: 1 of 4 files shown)


tests/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ endif()
 
 include(GoogleTest)
 
+# Keep the headers paired with the bundled GoogleTest libraries ahead of any
+# system installation exposed by another dependency's include path.
+include_directories(BEFORE SYSTEM "${gtest_SOURCE_DIR}/include")
```
amadio: This is quite hacky. In "modern CMake", include directories should come from their respective targets, not added by hand. I knew I was going to regret vendoring stuff... This is a drawback that we have to just either accept, or fix within the vendored copy of GoogleTest, not here.

amadio (changes requested):
I have not had issues with this, even if I have GoogleTest installed in the system. It would be good to make the problem visible in CI, so that we can fix it and ensure it stays fixed. That probably means installing gtest with Homebrew on macOS in any case, and maybe also Alma 8, where the system one is broken, and have the CI still pass.

vendor/googletest/googlemock/CMakeLists.txt
```diff
@@ -103,10 +103,10 @@ else()
 endif()
 
 string(REPLACE ";" "$<SEMICOLON>" dirs "${gmock_build_include_dirs}")
-target_include_directories(gmock SYSTEM INTERFACE
+target_include_directories(gmock INTERFACE
   "$<BUILD_INTERFACE:${dirs}>"
   "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/${CMAKE_INSTALL_INCLUDEDIR}>")
-target_include_directories(gmock_main SYSTEM INTERFACE
+target_include_directories(gmock_main INTERFACE
```
amadio: This makes warnings in GoogleTest headers appear in CI, which I can see from the other changes you needed elsewhere. It's like a whack-a-mole game :-/  I believe if you use `target_include_directories(gmock_main BEFORE SYSTEM INTERFACE ...)` then this will come before the headers that are really from the system, but also not make warnings visible in our own builds.

lobis: Thanks for the tip, I had to do a bit extra work to get it to work on the mac but seems to be working now

amadio:
@lobis Did you check if CI fails without the fix and works after? I mean, fails if we only install gtest with Homebrew or in Alma 8, and gets fixed by the current patch.

lobis:
> @lobis Did you check if CI fails without the fix and works after? I mean, fails if we only install gtest with Homebrew or in Alma 8, and gets fixed by the current patch.

I tested on my machine and this patch properly fixes the issue I have with my GoogleTest v1.18 vs the vendored GoogleTest v1.16, so this is a proper fix other developers using macos will encounter.

This is however not reproduced in the CI (https://github.com/xrootd/xrootd/actions/runs/33190116660/job/98913153362) because here the version homebrew installs is 1.17 which does not present the same incompatibility.

amadio (approved):

PR #2912, opened 2026-08-28
Build Python bindings during regular build and run Python tests with CTest like the rest of the test suite

amadio:
This pull request is a replacement for #2821.

I revived my old branch from a couple of years ago and used Claude to fix the issues I had pending, which kept me from merging that work into `master` before. The problems were:

- The Python build would always run again with `make`. This is because `add_custom_target` always considers the custom target out of date (see CMake's own [documentation](https://cmake.org/cmake/help/latest/command/add_custom_target.html)). The usual solution for this is to wrap the build into an `add_custom_command()` instead, and add a custom target that depends on an output of the custom command.
- The Python bindings were not properly rebuilt on source changes. There were two choices for handling this problem, by adding the proper dependencies. Either integrating the extension client library into the CMake build, or keeping `pip` launching a CMake sub-build. In the end, using `pip` to launch a sub-build had a silent problem that some definitions were not being passed to the compilation of the extension and that led to a mismatch between the main and the Python module builds. This was fixed by the integration of the client library into the main build, and using `pip` to only bundle stuff into the wheel.
- A requirement I gave to Claude was that all 3 build modes (standalone, embedded, and native) should keep working, so an extra check was added in CI to cover the case which was missing (and we did have recent problems with that lack of coverage, as can be seen from #2816). Now we will detect when something breaks that build.
- Finally, I used the opportunity to convert the tests which depended on EOS public to use the new server configuration created just for the Python bindings tests.

Claude made some mistakes along the way, which I had to fixup in later commits. If you're curious how I used Claude for this, I am attaching the exported Markdown of my session below.

[xrootd-python-bindings.md](https://github.com/user-attachments/files/31558383/xrootd-python-bindings.md)

I still intend to make a few adjustments on top of what Claude did, I will take this out of draft when I'm done.

*Note: I intend to squash this on merge, the commit history is just to have a trace of the process in the commit messages, which should remain visible here afterwards.*

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -10,6 +10,37 @@ else()
   configure_file(setup.py setup.py)
   file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/VERSION "${XRootD_VERSION_STRING}")
 
+  add_custom_target(PyXRootD ALL
+    ${Python_EXECUTABLE} -m pip install --upgrade --target . .
+    BYPRODUCTS ${CMAKE_CURRENT_BINARY_DIR}/XRootD DEPENDS XrdCl
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+  if(NOT BUILD_TESTS OR NOT ENABLE_SERVER_TESTS)
+    return()
+  endif()
+
+  # We need to set a library path because we need to build the bindings C++
+  # libraries with their final install RPATH. We need to build with install
+  # RPATH because the libraries are installed by pip instead of CMake, so
+  # CMake cannot fix the RPATH during installation as with other libraries.
+
+  if(NOT APPLE)
+    list(APPEND ENVIRON "LD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+  else()
+    list(APPEND ENVIRON "DYLD_LIBRARY_PATH=${CMAKE_LIBRARY_OUTPUT_DIRECTORY}")
+  endif()
+
+  list(APPEND ENVIRON "PYTHONPATH=${CMAKE_CURRENT_BINARY_DIR}")
+
+  foreach(test copy file filesystem glob threads url)
+    add_test(NAME PyXRootD::${test}
+      COMMAND ${Python_EXECUTABLE} -m pytest ${CMAKE_CURRENT_SOURCE_DIR}/tests/test_${test}.py)
+    set_tests_properties(PyXRootD::${test}
+      PROPERTIES FIXTURES_REQUIRED XRootD::noauth ENVIRONMENT "${ENVIRON}" RUN_SERIAL TRUE)
+  endforeach()
+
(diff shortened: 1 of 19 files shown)


amadio:
I guess I will need to fix macOS myself. This readiness probe Claude added is probably what's breaking the tests on macOS.

amadio:
Hopefully without the custom readiness probe it will be ok on macOS. If not, I will do it by hand.

lobis:
> Hopefully without the custom readiness probe it will be ok on macOS. If not, I will do it by hand.

I will run this locally on my mac to verify and submit a review

xrootd.spec
```diff
@@ -49,6 +49,7 @@ BuildRequires:	systemd-rpm-macros
 BuildRequires:	systemd-devel
 BuildRequires:	python3-devel
 BuildRequires:	python3-pip
+BuildRequires:	python3-pytest
```
lobis: pytest is now a valid build-time test dependency because %ctest invokes the Python tests. However, could this BuildRequires be moved into the existing %if %{with tests} block? As written, rpmbuild --without tests still requires python3-pytest even though it will never be used.

python/CMakeLists.txt
```diff
@@ -4,24 +4,217 @@ project(PyXRootD LANGUAGES CXX)
...
+
+# The tests start their own server via conftest.py, so they need to know
+# where to find it. Keep the source tree clean of caches while at it.
+
+list(APPEND ENVIRON "XROOTD=$<TARGET_FILE:xrootd>")
+list(APPEND ENVIRON "XRDFS=$<TARGET_FILE:xrdfs>")
+list(APPEND ENVIRON "PYTHONDONTWRITEBYTECODE=1")
+
+foreach(test copy file filesystem glob threads url)
+  add_test(NAME PyXRootD::${test}
```
lobis, line 230: Since enabling the regular CMake tests now implicitly requires pytest, could configuration check that ${Python_EXECUTABLE} -m pytest is available, or could this new prerequisite be documented? Currently configuration succeeds and the missing dependency is only reported later by CTest.

amadio:
@lobis I told Claude to replace its broken probe with an xrdfs command, now the CI is happy. After the review comments I will update and squash things in preparation for the merge. You may want to take a branch with Python tests and rebase on this one to try things out for reviewing.

I wanted to compare pytest and BATS so I also told Claude to use a server fixture with pytest itself handling the server startup/shutdown.

xrootd.spec
amadio: Done.

python/CMakeLists.txt
amadio: Done.

amadio:
@lobis I fixed what you mentioned, and also added another commit to fix the annoying problem that XrdClHttp tests pick up a proxy from /tmp if it's there. Will merge this to master soon. I think I will merge this first, then things that have Python tests, to ensure Python tests actually pass.

amadio:
Alright, this is finally good to go. I ran also DEB/RPM builds on this branch to ensure packaging there is still working, and it's all good. Merging!

PR #2919, opened 2026-09-02
[Plugins] Replace version scripts with hidden visibility to be able to get plugin test coverage information

amadio:
This is an alternative to #2876, with the intent to fix plugin coverage of our plugins. The diagnostic of the problem is below. We need to drop linking scripts and use just hidden visibility instead, otherwise we hide also the coverage symbols, and coverage dumping then doesn't work properly.

Plugins linked with `--version-script (local: *)` localize the statically linked gcov runtime, giving each plugin a private `__gcov_master` chain. The process-wide `__gcov_dump()` called on `SIGTERM` cannot reach those chains, and `_exit()` skips the DSO finalizers that would flush them, so XrdMacaroons, XrdHttpTPC, and XrdOssStats produced no coverage data from server runs. Unit tests and the client-side plugins were saved by normal process exit running the finalizers.

Delete the `export-lib-symbols` scripts and compile the five plugins with hidden visibility by default instead, marking entry points with a new `XrdEXPORT` macro. `XrdVERSIONINFO` now applies `XrdEXPORT` to the version information symbol so it stays reachable via `dlsym()`. Precompiled `libgcov.a` members keep default visibility, so each plugin's `__gcov_master` is interposed to the executable's copy and its counters join the process-wide chain dumped on `SIGTERM`.

The XrdClHttpTesting shim existed only to re-export symbols hidden by the version script; the tests now link XrdClHttpObj directly, as the XrdClS3 tests already do.

Assisted-by: Claude:Fable 5

## Summary by CodeRabbit

* **Build and Compatibility**
  * Improved plugin symbol handling across HTTP, S3, authorization, and storage modules.
  * Plugin entry points and version information now remain explicitly available for supported integrations across platforms.
  * Reduced unintended exposure of internal library symbols, improving binary isolation and portability.
* **Tests**
  * HTTP tests now use the production library target directly, providing more representative validation.


amadio:
@abh3 I'm trying out CodeRabbitAI for code reviews, if it's too annoying, we can turn it off. I think it's good to have an extra pair of eyes in case we miss anything in our regular reviews. Also, it's free for open source projects.

amadio:
Coverage results are available at https://my.cdash.org/builds/4025827/coverage. Now we see all files from the `XrdMacaroons` plugin, which was my smoke test that things actually are working.

PR #2868, opened 2026-07-14
[XrdCl] Extend gfal2-util CLI compatibility

lobis:
## Scope

This draft makes the existing XRootD applications easier to use for common **gfal2-util read-only, token, and namespace workflows**. It remains a thin compatibility layer:

- `xrdfs` normalizes complete-URL operands and selected gfal-style options, then dispatches to the existing command handlers and XrdCl operations.
- `xrdcp` remains the copy engine for remote-to-local downloads.
- XrdClHttp continues to provide HTTP/WebDAV behavior and carries token requests through the existing connection and credential stack.
- No `xrd` frontend, wrapper, or second filesystem/copy implementation is introduced.

The eventual migration frontend and its final name remain separate work tracked by #2863.

## Supported mappings

| gfal2-util workflow | XRootD invocation | Existing implementation reused |
| --- | --- | --- |
| `gfal-stat URL` | `xrdfs stat URL` | `FileSystem::Stat` |
| `gfal-ls -lH URL` | `xrdfs ls -lH URL` | existing stat/list and formatting |
| `gfal-ls -l --xattr user.status URL` | `xrdfs ls -l --xattr user.status URL` | stat/xattr queries |
| `gfal-cat -b URL...` | `xrdfs cat -b URL...` | existing copy-to-stdout path |
| `gfal-sum URL ADLER32` | `xrdfs sum URL ADLER32` | checksum query |
| `gfal-xattr URL xroot.cksum` | `xrdfs xattr URL xroot.cksum` | checksum/space/stat adapters |
| `gfal-token URL` | `xrdfs token URL` | `QueryCode::Visa` through XrdClHttp |
| issuer-based token request | `xrdfs token --issuer ISSUER URL` | RFC 8414/OIDC discovery and the existing HTTPS stack |
| `gfal-mkdir -p -m 0755 URL` | `xrdfs mkdir -p -m 0755 URL` | `FileSystem::MkDir` |
| `gfal-chmod 0750 URL` | `xrdfs chmod 0750 URL` | `FileSystem::ChMod` |
| `gfal-rename OLD NEW` | `xrdfs mv OLD NEW` | `FileSystem::Mv` on one ROOT endpoint |
| `gfal-rm URL...` | `xrdfs rm URL...` | `FileSystem::Rm` |
| `gfal-rm -r URL` | `xrdfs rm -r URL` | iterative CLI-local use of `Rm`, `DirList`, and `RmDir` |
[...]

PR #2920, opened 2026-09-03
[XrdCl] Extend xrdfs command options

lobis:
## Summary

Extend existing `xrdfs` subcommands with option and operand forms used by common storage workflows:

- `cat`: accept byte-count options and the `--` separator
- `ls`: accept grouped short options and long aliases, including human-readable, directory, all, one-per-line, and color-disabled forms
- `sum`: query a selected checksum algorithm
- `xattr`: expose virtual checksum, space, status, and storage-token attributes while retaining native operations
- `mkdir`: accept long and separated options, octal modes, and multiple paths
- `chmod`: accept octal mode-first syntax while preserving path-first forms

The commands continue to use the existing XrdCl operations and support both server-first and complete-URL syntax.

## Motivation

This extracts the focused CLI-option changes from draft #2868 so they can be reviewed independently from the token, WebDAV, Python, Tape, and removal work in that aggregate PR. Keeping one commit per subcommand also makes each behavior change easier to review.

This work implements part of #2863 and builds on the complete-URL support tracked in #1664.

## Validation

- [Full GitHub Actions CI matrix](https://github.com/lobis/xrootd/actions/runs/33717857837)
- [ABI check](https://github.com/lobis/xrootd/actions/runs/33717857725)
- End-to-end coverage for the modified commands and complete-URL forms

### DIFF
--- a/docs/man/xrdfs.1
+++ b/docs/man/xrdfs.1
@@ -73,10 +73,13 @@ path. A single slash is also accepted and normalized to an absolute command
 path.
 
 .SH COMMANDS
-\fBchmod\fR \fIpath\fR \fI<user><group><other>\fR
+\fBchmod\fR \fIpath mode\fR
+.br
+\fBchmod\fR \fIoctal-mode path\fR
 .RS 3
-Modify permissions of the \fIpath\fR. Permission string example:
-\fIrwxr-x--x\fR
+Modify permissions of the \fIpath\fR. The path-first form accepts either a
+nine-character symbolic mode, such as \fIrwxr-x--x\fR, or an octal mode no
+greater than \fI0777\fR. The mode-first form accepts an octal mode.
 
 .RE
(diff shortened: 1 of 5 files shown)


docs/man/xrdfs.1
```diff
@@ -9,7 +9,7 @@ xrdfs - xrootd file and directory meta-data utility
 \fBxrdfs\fR \fI[global-options]\fR \fIcommand\fR \fI[args containing complete URLs]\fR
 
 \fBcommand\fR: help, cache, chmod, ls, locate, mkdir, mv, stat, statvfs, query,
-           rm, rmdir, truncate, prepare, cat, tail, spaceinfo, xattr
+           rm, rmdir, truncate, prepare, cat, tail, spaceinfo, sum, xattr
```
amadio: I'd suggest naming the command `cksum`, after the [`cksum`](https://man7.org/linux/man-pages/man1/cksum.1.html) command line tool (I know gfal has `gfal-sum`, but `cksum` is more explicit).

@abh3 What do you think?

amadio: I suggest to keep the syntax also close to the `cksum` command line utility, unless that makes the gfal compatibility too hard to implement later on.

amadio:
Should we move to `getopt_long` instead of hand-crafted option parsing?

docs/man/xrdfs.1
lobis: Agree, I will change to `cksum` and try to follow its syntax.

PR #2921, opened 2026-09-03
[XrdCl] Add storage and issuer token requests to xrdfs

lobis:
## Summary

Add `xrdfs token` for direct storage-issued and issuer-directed access-token requests over HTTPS and DAVS.

- request storage-issued macaroons with read/write defaults, explicit activities, and configurable validity
- discover issuer metadata through RFC 8414 and OpenID Connect
- support SciTokens client credentials and scoped OAuth macaroon flows
- fall back to direct storage issuance when issuer discovery is unavailable
- keep one deadline across the workflow, restrict redirects to HTTPS, bound response parsing, and avoid exposing secrets in diagnostics
- accept `scope` and `scopes` in XrdMacaroons requests

The implementation includes focused unit tests for URL handling, token responses, queueing, issuer discovery, and macaroon scope parsing, plus end-to-end command coverage.

## Motivation

This extracts the token work from draft #2868 into a standalone pull request so its API and protocol behavior can be reviewed independently from the CLI-option, WebDAV, Python, Tape, and removal work in that aggregate PR. This pull request is opened as a draft while that review is in progress.

This work implements the token workflow tracked in #2863. It is independent of the separate `xrdfs` option changes.

## Validation

- [Full GitHub Actions CI matrix](https://github.com/lobis/xrootd/actions/runs/33719277195)
- [ABI check](https://github.com/lobis/xrootd/actions/runs/33719277262)
- Unit and end-to-end coverage for direct storage, issuer discovery, SciTokens, and macaroon requests

## Summary by CodeRabbit

* **New Features**
  * Added the `xrdfs token` command for requesting storage access tokens.
  * Supports token requests using SciTokens, OAuth, OpenID Connect, and macaroons.
  * Added issuer discovery, access-mode selection, validity periods, activities, and complete-URL syntax.
  * Added HTTPS/DAVS validation and secure handling of token responses.

* **Bug Fixes**
  * Improved OAuth compatibility for both `scope` and `scopes` parameters.
  * Corrected handling of percent-encoded values at input boundaries.

* **Tests**
  * Added comprehensive coverage for token requests, issuer workflows, validation, timeouts, and queue behavior.

### DIFF
--- a/docs/man/xrdfs.1
+++ b/docs/man/xrdfs.1
@@ -9,7 +9,7 @@ xrdfs - xrootd file and directory meta-data utility
 \fBxrdfs\fR \fI[global-options]\fR \fIcommand\fR \fI[args containing complete URLs]\fR
 
 \fBcommand\fR: help, cache, chmod, ls, locate, mkdir, mv, stat, statvfs, query,
-           rm, rmdir, truncate, prepare, cat, tail, spaceinfo, xattr
+           rm, rmdir, truncate, prepare, cat, tail, spaceinfo, token, xattr
 .fi
 .br
 .ad l
@@ -27,6 +27,9 @@ The \fBquery\fR command remains server-first because its implementation-dependen
 parameters may themselves look like URLs.
 Local \fBfile://\fR and \fBstdio://\fR operands are not remote endpoints and are
 rejected in command-first form.
+The command-first complete-URL form is supported by cache, cat, chmod, locate,
+ls, mkdir, mv, prepare, rm, rmdir, spaceinfo, stat, statvfs, tail, token,
+truncate, and xattr.
 Command help is available by invoking \fBxrdfs\fR with no command
 line options or parameters and then typing "help" in response to the
 input prompt.
@@ -71,6 +74,10 @@ In an XRootD URL, the double slash after the host consists of the slash
 separating the authority from the path and the leading slash of the absolute
 path. A single slash is also accepted and normalized to an absolute command
 path.
+.PP
+.nf
+\fBxrdfs token --validity 30 https://storage.example.org/store/file DOWNLOAD\fR
+.fi
 
 .SH COMMANDS
 \fBchmod\fR \fIpath\fR \fI<user><group><other>\fR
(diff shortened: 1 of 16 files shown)


amadio:
I think we will need more than one command for getting tokens, for macaroons and OpenID Connect at least, so that this can be used to get OAuth2 tokens from Google/GitHub as well as macaroons from the storages. @apeters1971 was adding a script for this in #2809, so we should avoid duplicating the functionality between that PR when it is re-created and this one.

Let's chat about how to make a `xrdfs token` command work for obtaining both macaroons and OAuth2 tokens. It could be `xrdfs token -m` (macaroon), `xrdfs token -i` (id token), and `xrdfs token -s` (scitoken), with equivalent long-form options, or some other form that makes sense. It could also be a separate command line tool altogether (i.e. `xrdtoken`), which is what I believe Andreas implemented in his PR.

abh3:
Nit clear to me that xrdfs is the correct place to put token request handling. It seems that xrdfs is quickly becoming a kitchen sink of functionality loosely related but so loose that it starts to make little sense. Normally, we name commands after their primary function, xrdfs gives you commands that involve a filesystem. Tokens are really orthogonal to that concept. Yes, you may need one but not necessarily just for xrdfs operations. Frankly, the common way of doing this is to have a gettoken command or something similar that does just that kind of function whose result can be used by xrdfs. Right now xrdfs is just getting way too bloated for its own good.

lobis:
> Nit clear to me that xrdfs is the correct place to put token request handling. It seems that xrdfs is quickly becoming a kitchen sink of functionality loosely related but so loose that it starts to make little sense. Normally, we name commands after their primary function, xrdfs gives you commands that involve a filesystem. Tokens are really orthogonal to that concept. Yes, you may need one but not necessarily just for xrdfs operations. Frankly, the common way of doing this is to have a gettoken command or something similar that does just that kind of function whose result can be used by xrdfs. Right now xrdfs is just getting way too bloated for its own good.

I dont have a particular preference to use `xrdfs`. As long as this cli is available on the pip client install we should be good from the new gfal point of view.

Let me know if you or anyone else have a concrete proposal.

abh3:
I think that I and @amadio agree that it should be a function-specific tool placed in XrdApps with th assumption that it won't require multiple source files We both suggest the name "xrdtoken" with the source named XrdToken.hh/cc which implements the various token operations. If it does need multiple source files to do it properly, then it should go under the directory, src/XrdsecToken.

lobis:
> I think that I and @amadio agree that it should be a function-specific tool placed in XrdApps with th assumption that it won't require multiple source files We both suggest the name "xrdtoken" with the source named XrdToken.hh/cc which implements the various token operations. If it does need multiple source files to do it properly, then it should go under the directory, src/XrdsecToken.

I also think this is the best direction. I will keep this PR as is and work on a new PR implementing this new command, then close this PR after the new direction is established.

I also opened https://github.com/xrootd/xrootd/issues/2946 to address the fact that these cli tools do not come from the python client install, which is a requirement for the new gfal to work (at least right now).

amadio:
I agree with @abh3, let's do token requests with a dedicated command. The command should come *after* the OAuth2 support from @apeters1971 is already merged (or as a part of it), and should support scitokens, OAuth2 tokens, and macaroons. The command should be implemented in C++, not Python. The rationale is that in many distros there is more than one Python version available, and a command line tool may not find and use the right one and fail because of that (with XRootD module not found).

As for source placement, we need to centralize token handling for other parts of the code in any case, so a new module for token handling sounds like a reasonable approach. Let's come up with the final name before implementing things, though. This has strong potential to clash with work for OAuth2 support as well, so may be folded into that.

abh3:
Just to keep things clean, we would prefer you close the pull request the moment it is no longer relevant, as in this case. Plowing a lot of pull requests, some of which will never be used, is becoming time consuming.

PR #2926, opened 2026-09-08
[XrdHttpTPC] Do not accept Source AND Destination header in COPY

ccaffy:
Prevents users from submitting a COPY request with both Source and Destination headers.
The server will return 400 Bad Request if that is the case

Assisted-By Claude:Opus-5

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
@@ -379,9 +379,18 @@ int TPCHandler::ProcessReq(XrdHttpExtReq &req) {
             return req.SendSimpleResp(400, NULL, NULL, "COPY requestd an unsupported Credential type", 0);
         }
     }
-    header = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
-    if (header != req.headers.end()) {
-        std::string src = PrepareURL(header->second);
+    auto srcHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
+    auto dstHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"destination");
+    // A Source header asks for a pull, a Destination header for a push; asking
+    // for both at once is ambiguous, so the request is rejected instead of
+    // arbitrarily honouring one of the two.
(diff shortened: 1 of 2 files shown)


amadio:
@ccaffy Please add the issue number to the commit message, so it links the issue here to close automatically. The changes look good to me, thanks!

amadio:
As is, I guess this could actually target `v5.9.x` and get merged into `6.x` after the fact, no?

src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -406,25 +422,43 @@ int TPCHandler::ProcessReq(XrdHttpExtReq &req) {
...
+        return req.SendSimpleResp(400, NULL, NULL, error_both, 0);
+    }
+    if (srcHeader != req.headers.end()) {
+        std::string src = PrepareURL(srcHeader->second);
         if (!IsAllowedScheme(src)) {
             const char *error_src = "COPY rejected: disallowed scheme in source URL";
             m_log.Emsg("ProcessReq", error_src, src.c_str());
             return req.SendSimpleResp(400, NULL, NULL, error_src, 0);
         }
+        if (IsCopyOntoItself(src, req)) {
```
dynamic-entropy: This prevents us from fixing the underlying issue of self-copying. 

You do not need all these, the presence of `Source` and `Destination` is enough to respond with a "Bad Request" response.

dynamic-entropy (changes requested):
The more I look at this, the more problems I see. The objective seems completely misplaced.

We do not care about the values; checking their equality for self copy is useless. Its actually only a couple line change. 

Further, this is only a patch for the recent incident, but not at all fix for "self copy".

src/XrdHttpTpc/XrdHttpTpcTPC.cc
ccaffy: I'm not sure I understand your feedback :D I pushed this with the intention of saving my current work but it's not the last state of course.

The intention was: we get a Source header or a Destination header, which are strings. How do we ensure that the destination is not the source of the file being transferred? Especially on a redirector setup. (`COPY srv1/path/file.txt Source: manager/path/file.txt`)

PR #2931, opened 2026-09-09
[XrdHttpTPC] Do not accept Source AND Destination header in COPY

### LINKED ISSUE #2930: [XrdHttpTPC] Return 400 Bad request if Source AND Destination headers are passed with COPY
### Operating System

Alma Linux 9

### XRootD Version

All

### Description

We need to return a 400 Bad Request error to a client contacting an XRootD server with a HTTP COPY request with both Source and Destination headers passed.

_No response_

### Additional Information

_No response_

ccaffy:
Prevents users from submitting a COPY request with both Source and Destination headers.
The server will return 400 Bad Request if that is the case

Fixes #2930
Assisted-By Claude:Opus-5

### DIFF
--- a/src/XrdHttpTpc/XrdHttpTpcTPC.cc
+++ b/src/XrdHttpTpc/XrdHttpTpcTPC.cc
@@ -379,9 +379,18 @@ int TPCHandler::ProcessReq(XrdHttpExtReq &req) {
             return req.SendSimpleResp(400, NULL, NULL, "COPY requestd an unsupported Credential type", 0);
         }
     }
-    header = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
-    if (header != req.headers.end()) {
-        std::string src = PrepareURL(header->second);
+    auto srcHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
+    auto dstHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"destination");
+    // A Source header asks for a pull, a Destination header for a push; asking
+    // for both at once is ambiguous, so the request is rejected instead of
+    // arbitrarily honouring one of the two.
+    if (srcHeader != req.headers.end() && dstHeader != req.headers.end()) {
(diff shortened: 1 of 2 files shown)


src/XrdHttpTpc/XrdHttpTpcTPC.cc
```diff
@@ -379,25 +379,33 @@ int TPCHandler::ProcessReq(XrdHttpExtReq &req) {
...
     }
-    header = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
-    if (header != req.headers.end()) {
-        std::string src = PrepareURL(header->second);
+    auto srcHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"source");
+    auto dstHeader = XrdOucTUtils::caseInsensitiveFind(req.headers,"destination");
+    // A Source header asks for a pull, a Destination header for a push; asking
+    // for both at once is ambiguous, so the request is rejected instead of
+    // arbitrarily honouring one of the two.
+    if (srcHeader != req.headers.end() && dstHeader != req.headers.end()) {
```
dynamic-entropy, line 387: Okay,  given this, I guess you did understand the comment on the other PR. https://github.com/xrootd/xrootd/pull/2926

Let me know if that is not the case.

dynamic-entropy (commented):
However, again, this is not enough to refuse a request that would result in a trimming the source file. 

Simply imagine a case where both headers are not specified but only the unintended one (source when destination was intended and vice versa).

src/XrdHttpTpc/XrdHttpTpcTPC.cc
ccaffy: No I did not get it at all :D I think you got confused because I had two fixes in mind/pushed for a PR that was badly named. Hence why all this rework.
There are two orthogonal fixes:
- Return 400 if both headers are passed (it is this PR) or the first commit of the one I closed
- Prevent source = destination TPC to delete the file (a new PR to be opened)

amadio (approved):

Issue #2747, opened 2026-03-14
Expand test suite to cover XCache use cases (single cache server, clustered cache)


amadio:
We have only a basic test exercising XCache functionality. We need to expand the test suite to cover at least 50% of `XrdPfc` and `XrdPss` to make sure new developments do not introduce regressions.

Issue #2366, opened 2024-10-31
gstream hit and miss numbers are wrong


ddavila0:
The hit and miss data reported by xcache (gstream) is misleading. For example, requesting 4K bytes[1] from a file, that is NOT in the cache, is reported as 100% hit[2]

[1]
```
curl --cert /tmp/x509up_u31717 --cacert /tmp/x509up_u31717 --capath /etc/grid-security/certificates/ -r 0-4194304 https://xcache-10.t2.ucsd.edu:1094//store/user/ddavila/1G -o /tmp/4M
```

[2]
```
[root@xcache-10 ddavila]# xrdpfc_print -v 1G.cinfo 
FILE: 1G.cinfo
version 4, created Thu Oct 31 21:35:04 2024; cksum net
file_size 1048576 kB, buffer_size 512 kB, n_blocks 2048, n_downloaded 18, state incomplete [0.879%]
printing 2048 blocks:
               1         2         3         4         5         6
     0123456789012345678901234567890123456789012345678901234567890123
   0 xxxxxxxxxxxxxxxxxx..............................................
  64 ................................................................
...
Record Attach         Detach         Duration     N_ios N_mrg    B_hit[kB]   B_miss[kB] B_bypass[kB]
1      241031:213504  241031:213534  0:00:30          1     0         4096            0            0
 ```

osschar:
In this particular case, you seem to have been helped by prefetching, note that the file has 18 512kB blocks downloaded after your interaction with the cache. So, cache had the data when you asked for it. How do you want to account for prefetching? Do you want a separate field in g-stream?

Second, we know there is a problem when async reading is used and the layer above PFC chunks up read requests into 64kB pieces -- and PFC sends out a request to the remote to get a full block (512kB in this case) and would concequently, when the other 64kB async requests come in, mark them as already available in the cache. We are looking into this case with @abh3.

[ Note, with async fully supported since 5.5, you really want to reduce the size of pfc.blocksize, default is now 128kB, you might want to go down to 64kB --- unless there is a real reason why it needs to be large, e.g., when fronting an object store. ]

ddavila0:
How could the prefetcher know what to prefetch if that file was brand new and that was the first request for such file?

[...]

PR #2932, opened 2026-09-09
[Tests] Expand XrdPfc (XCache) test coverage

osschar:
***Disclaimer:*** This is mostly Claudine -- I have only done one small test before so I was a poor advisor to it -- but I did have it scan all other tests and match them in format and style. If things are off, please advise and I'll give it another spin.

Towards #2747. Covers the single-cache-server scenario; the clustered one is not addressed here.

Before this, XrdPfc had two unit tests over the path tokenizer, one integration test that copied a file through a cache and diffed it, and the xcachewithcsi test for `pfc.cschk cache`. The integration test passed equally well against a plain proxy: `cache.cfg` carried no `pfc.*` directives and nothing checked that anything had been cached.

### What the tests assert on

`b_todisk` and `b_prefetch` from the pfc g-stream `file_close` record, which are counted per block as blocks are written out and so are exact and independent of the client's request size, plus the `.cinfo` via `xrdpfc_print` for the state left on disk.

Deliberately *not* `b_hit`/`b_miss`. A block is scored as missed only for the request that created it, so a second request landing on the same in-flight block reads as a hit. Two things make that bite: the client's single large pgread reaches the cache as 64 kB segments, giving an exact 50/50 split at a 128 kB block; and with prefetching on, a cold read of a file that is not in the cache at all reports `b_hit == size, b_miss == 0` -- the symptom of #2366,
still reproducible. `b_todisk` states it correctly in both cases.

### Coverage

Seven ctest configurations under `tests/XRootD/`, each fronting the existing `XRootD::host` fixture as its origin:

| Config | What it pins down |
|---|---|
| `cache` | cold fetch / warm serve; a 4 kB read pulls exactly one block and only once; a vector read fetches only the blocks its chunks land in, checked against the block map; four concurrent readers of the same blocks fetch them once between them; `stat` reports the origin size for a partly cached file; writes refused cleanly; a missing file leaves no state |
| `cacheevict` | only-if-cached on an absent, a partial and a complete file; `xrdfs cache evict` removes data and cinfo and leaves the origin alone |
| `cacheprefetch` | a small read drags in more than its block and the excess is reported as prefetch; prefetch never runs past EOF |
| `cacheblacklist` | a path `pfc.decisionlib` refuses is proxied, never cached, and produces no `file_close` record at all -- `Cache::Attach` returns the upstream IO, so no `File` exists |
| `cachespaces` | `pfc.spaces`: symlink targets land in the right space, neither space holds the other's half |
| `cacherestart` | a complete and a partial file survive a server restart; the partial one keeps exactly its blocks, then fetches only what was missing |
| `cachepurge` | overruns a `files` budget and asserts the LRU ordering of what purge removed, and that a purged file is gone and refetchable |

Unit tests in `tests/XrdPfcTests/XrdPfcInfoTests.cc` cover the `.cinfo` format:
block accounting, the write/read round trip, access-history capping, and refusal of damaged files. The corruption test sweeps every byte position and requires all of them to be caught.

### New test-only helpers

- `tests/XRootD/utils/xrdpfc-read.cc` -- reads exactly the ranges given, as separate reads, one vector read, or from several threads. `xrdcp` only fetches whole files.
- `tests/XRootD/utils/gstream_recv.py` -- collects the pfc g-stream.
- `tests/XRootD/cachelib.sh` -- shared shell helpers over both oracles.

### Notes for reviewers

- `XRootD::cachepurge::test` takes about two minutes and is labelled `slow`.
  The purge check runs on a fixed 60 s cycle, recently opened files are purge-protected, and there is no way to request a purge, so convergence needs more than one pass.
- `pfc.hdfsmode` is not covered: `ConfigXeq` reports it unsupported and fails the config before doing anything.
- `XrdPfcInfo.cc` is compiled into the gtest binary rather than linked from the plugin, the way `xrdpfc_print` already builds it.
- No non-test source is touched.

### DIFF
--- a/tests/XRootD/CMakeLists.txt
+++ b/tests/XRootD/CMakeLists.txt
@@ -1,6 +1,22 @@
 
 list(APPEND XROOTD_CONFIGS noauth gsi host unix sss cache posix diglib mirage)
 
+# XCache (XrdPfc) configurations. Each fronts the XRootD::host fixture as its
+# origin and exercises one aspect of the cache; the shared shell helpers are in
+# cachelib.sh.
+list(APPEND XROOTD_CONFIGS cacheevict cacheprefetch cacheblacklist cachespaces cacherestart)
+
+# cachepurge is slow on purpose: the purge check runs on a fixed 60 s cycle and
+# there is no way to trigger one, so the test has to wait for at least one pass.
+list(APPEND XROOTD_CONFIGS cachepurge)
+
+# Range-reading client for the cache tests. xrdcp only does whole files, and
+# the interesting XCache behaviour is block-granular. Threads::Threads because
+# the concurrent-reader mode uses std::thread, which needs pthread linked
+# explicitly on glibc older than 2.34.
+add_executable(xrdpfc-read utils/xrdpfc-read.cc)
+target_link_libraries(xrdpfc-read XrdCl XrdUtils Threads::Threads)
+
 if(ENABLE_FUSE_TESTS)
   list(APPEND XROOTD_CONFIGS fuse)
 endif()
@@ -29,9 +45,11 @@ if(BUILD_SCITOKENS AND HAVE_SCITOKEN_CONFIG_SET_STR)
   list(APPEND httpnoclient_FIXTURES SciTokens)
 endif()
 
-# The cache test requires the XRootD::host fixture to be running to
+# The cache tests require the XRootD::host fixture to be running to
 # act as an origin
-list(APPEND cache_FIXTURES XRootD::host)
+foreach(CACHE_CONFIG cache cacheevict cacheprefetch cacheblacklist cachespaces cacherestart cachepurge)
+  list(APPEND ${CACHE_CONFIG}_FIXTURES XRootD::host)
+endforeach()
 
 list(APPEND gsi_FIXTURES TLS)
 
@@ -81,3 +99,6 @@ if(BUILD_SCITOKENS AND HAVE_SCITOKEN_CONFIG_SET_STR)
       FIXTURES_REQUIRED SciTokens
   )
 endif()
+
+# Waiting out purge cycles, see cachepurge.sh.
+set_tests_properties(XRootD::cachepurge::test PROPERTIES TIMEOUT 600 LABELS slow)
--- a/tests/XRootD/cache.cfg
+++ b/tests/XRootD/cache.cfg
@@ -21,4 +21,23 @@ pss.cachelib libXrdPfc.so
 pss.setopt DebugLevel 4
 pss.origin localhost:5094
 
+# Cache parameters, pinned down rather than left to the defaults because
+# cache.sh asserts on them. The block size fixes how many blocks a file of a
+# given size has; prefetching is off so that the cold read fetches only what the
+# client asked for and b_prefetch stays a fixed 0. Keep PFC_BLOCKSIZE in
+# cache.sh in step with pfc.blocksize.
+
+pfc.blocksize 128k
+pfc.prefetch 0
+
+# Publish the pfc g-stream, which cache.sh reads. b_todisk (bytes fetched from
+# the origin and written to disk) and b_prefetch are only reported here -- they
+# are not in the cinfo -- and b_todisk is the only exact statement of how much
+# came over the network, since the b_hit/b_miss split depends on request
+# granularity. 'json insthdr' makes the datagram JSON throughout, with no
+# binary header to unpack. Keep the port in step with PFC_GSTREAM_PORT in
+# cache.sh.
+
+xrootd.mongstream pfc use flush 1s maxlen 24000 send json insthdr localhost:7097
+
 continue $src/common.cfg
--- a/tests/XRootD/cache.sh
+++ b/tests/XRootD/cache.sh
@@ -1,18 +1,46 @@
 #!/usr/bin/env bash
+#
+# Core XCache read path: whole-file mode, prefetching off, one origin.
+# What this asserts is where the bytes came from, not just that they arrived --
+# returning the right bytes is what a plain proxy would also do.
 
 export XrdSecPROTOCOL=host
 
+# shellcheck source=/dev/null
+source "$(dirname "${BASH_SOURCE[0]}")/cachelib.sh"
+
+# Must match pfc.blocksize in cache.cfg.
+PFC_BLOCKSIZE=$((128 * 1024))
+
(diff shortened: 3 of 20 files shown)


amadio:
@osschar I will have a thorough look at this when I have the time, but this looks like a good start! Thank you very much! You can see the coverage here: https://my.cdash.org/builds/4058354/coverage and add more tests if you see important parts are missing coverage in your tests. Also, of course the tests have to pass :-)

osschar:
@amadio , when this converges (seems to be close, one timeout issue on alpine, ubuntu broken in CI) ... do you want me to squash-rebase into one commit over master?

amadio:
> @amadio , when this converges (seems to be close, one timeout issue on alpine, ubuntu broken in CI) ... do you want me to squash-rebase into one commit over master?

@osschar I can squash on merge, so feel free to push the commits in any form you prefer. Rebasing is also doable directly on GitHub, or I can rebase manually at the end. So, just ensuring the tests are in good shape and cover well the features of XCache is sufficient. Thanks!

osschar:
A few notes from the CI safari, in case they save someone the same afternoon.

- Ubuntu was red for four runs in `Install Python bindings`, no wheel built. Not this branch -- the tree was byte-identical before and after -- it simply stopped when the runner image rolled from `20260831.293.1` to `20260907.300.1`. May well bite again on an older image.
- `-- Python support: yes` in the configure summary is `BUILD_PYTHON`, i.e. "a python was found". `add_subdirectory(python)` is gated on `ENABLE_PYTHON` (`CMakeLists.txt:53`). So the log happily says yes while no wheel is built anywhere. This is what fooled us longest.
- When that step fails, `post-install.sh`, `check-headers.sh` and the standalone-bindings build are all skipped -- three checks gone, one red mark.
- macOS failed once with 30 XrdCl fixture tests (none of the new ones) on a tree that had passed 20 minutes earlier. Re-ran clean.

Claudine had several excellent theories about the first one, of which none survived contact with `git merge-tree`. I mostly just kept saying it looked strange.

amadio:
@osschar Yes, I think the Python problem bit other people too, I will look at getting that fixed before more people get tricked by it. And indeed, some tests can be flaky (often due to problems in GitHub CI runners or latent bugs that need fixing).

osschar:
Apparently this only happens on Ubuntu -- other distros never touch python stuff. Maybe you just need to un-install python3-dev from Ubuntu CI machines / containers ;)

I have two PRs stacked on this, the first one closing 2 issues, and the second one fixing things that I noticed during this work.

amadio:
What I think is happening is that the Ubuntu images sometimes have a newer Python than the default one installed, which causes the wrong one to be picked up by the build. We do want to keep the Python tests running, so I will fix things by preferring the default Python, even if newer ones are installed. That's something easy to do, we just need `-DPython3_FIND_UNVERSIONED_NAMES=FIRST` in the CMake configuration (see [`FindPython3`](https://cmake.org/cmake/help/latest/module/FindPython3.html) docs).

amadio (approved):
Thanks a lot for the new tests! I will merge this, then you can rebase the fix branch on master afterwards (the one for 6.x ony).

amadio:
Hi @osschar, these tests are randomly failing in pretty much all pull requests, so unfortunately we'll have to temporarily revert this while the random failures are sorted out. CDash is unfortunately down as well, so we have to dig logs in GitHub, and with the release pending my time to debug this is quite short. Please debug it and resubmit after a few consecutive runs pass on your fork (it seems that tests are sharing state, so they fail when run in parallel). Cheers,

osschar:
Crap, sorry. Is it python stuff again? I thought it went away with the image bump on Ubuntu. 

There were some failures in the first incarnation with UDP messages from g-stream not reaching the test collector, but these were supposedly also hardened against. Maybe higher load on test machines has something to do with the failures and UDP package drops, sigh.

amadio:
FYI, this is what failed my own pull request:
```
2026-09-21T16:21:59.6338503Z 369: --- a small read pulls more than it asked for, and it is prefetch
2026-09-21T16:21:59.6339263Z 369: xrdpfc-read read root://f6112da4ccc1:7102///pfc-prefetch.dat pf-small.dat 0:4096
2026-09-21T16:21:59.6455756Z 369: dd if=prefetch.dat of=pf-small.ref bs=4096 count=1 status=none
2026-09-21T16:21:59.6462846Z 369: cmp pf-small.dat pf-small.ref
2026-09-21T16:22:10.1797107Z 354: error: timed out after 20 s waiting for g-stream file_close of /pfc-hitmiss.dat at access_cnt 1; collector alive: yes, log holds 0
2026-09-21T16:22:10.1797738Z 354: 0 file_close records
2026-09-21T16:22:10.1816039Z 354: error: cold read should fetch the whole file from the origin: expected 4194304 but received 
```
Maybe you recognize it, could be the Python script that handles the g-stream, I don't know. In any case, I will try to submit the tests running serially before reverting. If that works, I will keep the tests enabled. But in any case, we should understand and fix the failures.

amadio:
I also see this:
```
2026-09-21T16:22:10.4179796Z 369: --- prefetch never fetches past the end of the file
2026-09-21T16:22:10.4187181Z 369: dd if=/dev/urandom of=prefetch-full.dat bs=131072 count=32 status=none
2026-09-21T16:22:10.4311231Z 369: xrdcp -fs prefetch-full.dat root://localhost:5094//pfc-prefetch-full.dat
2026-09-21T16:22:10.4463528Z 369: xrdcp -f root://f6112da4ccc1:7102///pfc-prefetch-full.dat pf-full.dat
2026-09-21T16:22:10.4634782Z 369: cmp prefetch-full.dat pf-full.dat
2026-09-21T16:22:11.5049782Z 372: --- the blacklisted file never became a cache file at all
2026-09-21T16:22:11.5051816Z 372: /__w/xrootd/xrootd/tests/XRootD/cacheblacklist.sh: line 80: assert_no_gstream_close: command not found
```

amadio:
Reverted in f87cad0c3332b30fbc383cd10fa5178a4dd49dbb. After XRootD 6.2 is out, let's add these tests back, I really would like to have them!

osschar:
Thank you! Sorry for the trouble.

It seems to be the gstream biting left and right again, sigh. Let's revisit after the release.

amadio:
No problem! Once I get the release out I will look into debugging things too. If there's a way to test without enabling the g-stream, however, maybe it helps with stability. Cheers,

PR #2933, opened 2026-09-09
[XrdHttpTPC] Files can be deleted if src = destination

### LINKED ISSUE #2925: [XrdHttpTPC] Files can be deleted if src = destination
### Operating System

Alma Linux 9

### XRootD Version

5.9.7

### Description

If transfer source and destination are local and same path, return a 400 Bad Request error

The latter avoids data loss in case one triggers a COPY PULL or PUSH with same source and destination

_No response_

### Additional Information

_No response_

ccaffy:
Do not merge yet. I am still reviewing it.

This is an attempt to properly fix the deletion of files that are COPYied on themselves.
Not that easy as we need to check the Source header or the Destination header and check if the destination of the COPY will be the source.

There are checks on the Source/Destination header input, but also a HEAD request being done for PUSH.

A new TPCClusterTest has been added to cover the HTTP TPC using cmsd.
It allowed to see that currently, a HTTP TPC from a manager would not work due to that stalling in OpenWaitStall that breaks the sleeping loop instead of retrying to open(). To be fixed later.

Due to that issue, few tests have been commented out the time we find a solution to this stall loop.

Fixes #2925


abh3:
This sounds very suspicious. Are we trying to "fix" the case where the source "abc_host:/tmp/file" and the destination is "abc:_host:/tmp/file"? This should not be allowed and certainly xroot would reject this kind of operation because you cannot have a file open for reading by one party and open for writing by another party (unless you explicitly disable this check via config option).

Also, is it not the case that Raul is correct that you cannot have a source and destination headers in the same request at the same time? If so, that is trivial to check as he points out.

As for the cmsd could you be more specific? The redirector simply sends requests to where the request should be executed. It should not be executing the request at all and no attempt should be made to do that. So, this sounds like a misunderstanding of the architecture.

amadio:
@abh3 I think there were some discussions between you and @ccaffy after your last comment above. Could you please review this again considering the latest information about it? Thanks!

amadio:
@abh3 @ccaffy I'm tentatively moving this to 6.3, as it will require careful consideration of new configuration options, etc. If it happens to be ready before 6.2 is done, then we may reconsider and add it to 6.2 instead.

PR #2936, opened 2026-09-16
[XrdClHttp] Fix multipart parsing when curl splits CRLF

dynamic-entropy:
@ffurano's test script found another issue with the multipart parsing. This is a fix for that; a CRLF could be split at the curl write boundary.


ffurano (approved):
Looks OK to me, will be also part of my upcoming deeper test round

amadio:
@ffurano I thought you checked that this fixed the problem you reported. Does it? I'd like to know that before merging.

PR #2937, opened 2026-09-17
[XrdOssMirage] Speed up multi-character string pattern reads

diogomattioli:
## Summary

`XrdOssMirageFile::Read` generated multi-character patterns byte-by-byte with `std::generate_n` and a modulo operation on every element, which was slow for large reads. This change replaces that with a pre-expanded pattern buffer:

- On first read of a multi-character pattern, `std::call_once` builds a `pattern_buffer` by repeating the pattern until it is at least as large as the requested read size.
- Subsequent reads copy directly out of that buffer with two `std::copy` calls (main chunk plus wraparound), avoiding per-byte modulo arithmetic.
- `pattern_buffer` and `pattern_buffer_once` are added as new members of `XrdOssMirageFile`.

## Tests

Adds a new bats end-to-end suite (`tests/end-to-end/mirage/`) covering:
- Basic stat/truncate behavior on Mirage files.
- Empty, single-character, and multi-character pattern content correctness.
- Correctness under small/misaligned chunk sizes, and under concurrent/parallel reads from one or multiple clients with different chunk sizes, exercising the `std::call_once`-guarded buffer initialization under concurrency.

Assisted-By: Claude Sonnet 5


src/XrdOssMirage/XrdOssMirageFile.cc
```diff
@@ -112,10 +112,21 @@ ssize_t XrdOssMirageFile::Read(void *buffer, off_t offset, size_t size)
         std::fill_n(static_cast<char *>(buffer), num_bytes, entry->pattern.front());
 
     if (entry->pattern.size() > 1)
-        std::generate_n(static_cast<char *>(buffer), num_bytes, [i = offset % entry->pattern.size(), this] () mutable {
-            return entry->pattern[i++ % entry->pattern.size()]; 
+    {
+        std::call_once(pattern_buffer_once, [this, size] () {
+            while (pattern_buffer.size() < size)
+                pattern_buffer.insert(pattern_buffer.end(), entry->pattern.begin(), entry->pattern.end());
         });
```
amadio, line 119: What's the reasoning behind using `std::call_once` here? I'm curious. For clarity, I wonder why not use something like shown below

https://github.com/xrootd/xrootd/blob/bcc2407dee51edfde305845f7c1cc360cf87046c/src/XrdOuc/XrdOucUtils.cc#L1639-L1650

where the lambda is declared and called at the same time. (I'm implicitly trying to understand if we should use `std::once` above as well, instead of what I did.)

src/XrdOssMirage/XrdOssMirageFile.hh
```diff
@@ -5,13 +5,18 @@
...
 #include <variant>
+#include <vector>
 
 class XrdOssMirageFile : public XrdOssDF {
 private:
     XrdOssMirage &oss;
     std::variant<XrdOssMirageEntry, XrdOssMirageEntryPtr> entry_storage;
     XrdOssMirageEntry *entry;
+    
+    std::vector<char> pattern_buffer;
```
amadio, line 18: Why `std::vector<char>` and not `std::string`? Is it because of locale? Or performance?

src/XrdOssMirage/XrdOssMirageFile.cc
amadio: I also have another question here. What happens if one tries to read 1TB (or more than the amount of RAM) from Mirage? The `pattern_buffer` might become too big, especially if there are several files with different patterns around.

amadio (commented):
I suggest trying to calculate `n` such that on the `read` you need `n` full copies of the pattern + some tail, do the `n` copies directly into the user's buffer to avoid using a lot of memory, then copy over the tail. In other words, I'd try to limit memory usage to the biggest chunk size rather than fill `pattern_buffer` to the full size of the `read`, to avoid memory issues when large reads are involved.

Issue #2922, opened 2026-09-03
[XrdHttpTapeApi] Route Tape REST stage operations through native prepare handling


lobis:
## Context

PR #2836 added WLCG Tape REST API client support and a deterministic,
filesystem-backed `XrdHttpTapeApi` handler for integration testing.

Following the post-merge suggestion in
https://github.com/xrootd/xrootd/pull/2836#issuecomment-5520908036,
the server-side handler should be reconsidered as a protocol adapter rather
than implementing its own simulated tape backend.

Currently, Tape REST requests are handled directly by
`XrdHttpTapeApiStore`, which maintains request state and copies files between
simulated `archive/` and `disk/` directories.

## Proposed design

Route the stage-related Tape REST operations through the XrdHttp/XRootD bridge
and the existing native prepare machinery:

| Tape REST operation | Native XRootD operation |
|---|---|
| Submit stage request | `kXR_prepare` with `kXR_stage` |
| Cancel stage files | `kXR_prepare` with `kXR_cancel` |
| Release/evict files | `kXR_prepare` with `kXR_evict` |
| Poll stage status | `kXR_query` with `kXR_QPrep` |

This would lead requests through `XrdXrootdProtocol`, `XrdOfs::prepare`, and
any configured `XrdOfsPrepare` implementation.

For integration testing, configure `libXrdOfsPrepGPI.so` with a test program
implementing `stage`, `query`, `cancel`, and `evict`. The program may use the
same filesystem-based model as the current store, but the behavior would live
behind the standard prepare interface.

A production deployment could replace that program or prepare plugin with its
site-specific tape integration without changing the REST handler.

## Goals

- Make `XrdHttpTapeApi` translate between Tape REST and the existing XRootD
  prepare interface.
- Exercise the bridge, native request handling, OFS authorization, request-ID
  handling, and prepare plugin interfaces in the integration tests.
- Allow native XRootD prepare requests and Tape REST requests to use the same
  backend implementation.
- Retain deterministic end-to-end testing without embedding a separate tape
  backend in the HTTP handler.
- Add integration coverage for `libXrdOfsPrepGPI.so`.

## Design considerations

The following details need to be investigated as part of the implementation:

- External HTTP handlers currently run before bridge login and do not expose a
  general asynchronous bridge request API. The handler lifecycle may need to
  be extended so an external handler can safely issue and receive
  `kXR_prepare` and `kXR_query` operations.
[...]

PR #2939, opened 2026-09-21
[XrdHttpTapeApi] Route Tape REST requests through native prepare with durable persistence

lobis:
## Overview

This is **Part 3 of a 3-layer series** preparing the XRootD server for native durable prepare and WLCG Tape REST API routing:
1. **Part 1 (#2940)**: GPI prepare plugin hardening and authorization fixes (~130 LOC)
2. **Part 2 (#2941)**: Durable prepare persistence wrapper (`libXrdOfsPrepPersist.so`)
3. **Part 3 (this PR, #2939)**: HTTP Tape REST API routing via native bridge

## Summary

Addresses xrootd/xrootd#2922 and builds on xrootd/xrootd#2836.

This PR turns `XrdHttpTapeApi` from a filesystem-backed tape simulator into a protocol adapter over XRootD's existing prepare machinery. Tape REST and native XRootD clients now reach the same configured prepare service; the actual tape integration remains a site-supplied plugin or program.

It also adds **`XrdOfsPrepPersist`**, an opt-in `XrdOfsPrepare` wrapper that provides persistent request identity, ownership, ordered operation intent, restart recovery, cached status, and retention. The simulator moves out of the production HTTP handler into a deterministic, test-only program behind `libXrdOfsPrepGPI.so`.

The central contract is:

> **A successful stage response means durable local admission. Backend acknowledgement and file staging completion are separate events.**

There is no tape-protocol implementation in the HTTP handler or persistence layer. Sites remain responsible for recalls, cancellation, residency policy, authoritative locality, and safe deduplication of replayed operations.

## Architecture

The following shows a deployment with the durable profile enabled. The script and compiled-plugin routes are alternatives, not two backends used simultaneously.

```mermaid
flowchart TB
    accTitle: Tape REST and native clients share a durable prepare service
    accDescr: HTTP requests pass through the Tape REST adapter and authenticated native bridge. Both HTTP and native clients reach OFS and the optional persistence wrapper. A local registry stores metadata while a worker invokes a compiled backend or GPI script. The site backend, not the REST handler, performs tape and disk operations.

    REST["Tape REST client"]
    NATIVE["Native XRootD client"]

    HTTP["XrdHttpTapeApi<br/>WLCG JSON + authenticated RunNative bridge"]
    CORE["Native prepare path<br/>XrdXrootdProtocol → XrdOfs::prepare"]

    PERSIST["XrdOfsPrepPersist<br/>Durable admission + ownership<br/>Background reconciliation + cached status"]
    STORE[("Local JSON registry<br/>Manifest + intent + observed status")]
    PLUGIN["Compiled XrdOfsPrepare plugin"]
    GPI["GPI + site program<br/>XrdOfsPrepGPI → script"]
    SERVICE["Site tape and disk service<br/>Recall / residency / locality / deduplication"]

    REST --> HTTP --> CORE
    NATIVE --> CORE
    CORE -->|authorized prepare / query| PERSIST
    PERSIST <-->|persistent metadata| STORE
    PERSIST -->|plugin route| PLUGIN
    PERSIST -->|script route| GPI
    PLUGIN --> SERVICE
    GPI --> SERVICE

    classDef client fill:#F1F5F9,stroke:#64748B,color:#0F172A;
    classDef core fill:#FFFFFF,stroke:#64748B,color:#0F172A;
    classDef added fill:#E8F3FC,stroke:#2773A5,color:#103D5B,stroke-width:2px;
    classDef durable fill:#E5F3EF,stroke:#29816E,color:#123E34,stroke-width:2px;
    classDef site fill:#FFF3DF,stroke:#A3752C,color:#573A13;
    class REST,NATIVE client;
    class CORE core;
    class HTTP,PERSIST added;
    class STORE durable;
    class PLUGIN,GPI,SERVICE site;
```

| Component | Responsibility |
| --- | --- |
| `XrdHttpTapeApi` | Validate WLCG requests, translate them to native prepare/query operations, and convert results into HTTP/JSON responses. |
| Opt-in XrdHttp bridge extension | Establish bridge login and keep the native request payload and response callback alive until completion. |
| `XrdXrootdProtocol` / `XrdOfs::prepare` | Reuse native request handling, per-path authorization, and prepare-plugin dispatch. |
| `XrdOfsPrepPersist` | Admit requests durably, enforce ownership and membership, persist mutations, and reconcile with the backend. |
| `XrdOfsPrepStore` | Store bounded, versioned request records in a sharded local filesystem layout. |
| Site `XrdOfsPrepare` backend | Implement the durable backend profile, directly in a compiled plugin or through GPI and a site program. |

The wrapper uses the existing **`XrdOfsAddPrepare` / `ofs.preplib ++`** mechanism. It does not introduce a parallel tape-plugin hierarchy. Existing prepare deployments do not gain a persistent registry simply by upgrading; the wrapper must be configured explicitly.

The HTTP adapter itself now requires the durable profile. A legacy prepare plugin needs to implement that profile before it can be used underneath the wrapper.

## Native HTTP dispatch

An additive, opt-in external-handler bridge interface allows the Tape REST handler to issue native requests through `RunNative()` and receive deferred completion. Existing external handlers retain their existing lifecycle unless they opt in. Tape REST discovery remains a direct, public HTTP response.

| Tape REST operation | Native operation / profile convention |
| --- | --- |
| `POST /api/v1/stage` | `kXR_prepare` with `kXR_stage` |
| `GET /api/v1/stage/<id>` | `kXR_query` with `kXR_QPrep` |
| `POST /api/v1/stage/<id>/cancel` | `kXR_prepare` with `kXR_cancel` |
| `POST /api/v1/release/<id>` | Native eviction, with `xrd.prepare.request=<id>` carried in per-file opaque data |
| `DELETE /api/v1/stage/<id>` | Cancellation using the reserved profile request ID `@xrdprep-v1:delete:<id>` |
| `POST /api/v1/archiveinfo` | Prepare query using `@xrdprep-v1:archiveinfo`, with the requested paths |

Deletion and archive information have no exact existing native prepare equivalent. Their reserved IDs are **provisional backend-profile conventions**, not new wire opcodes or semantics automatically supported by every existing plugin. Request context is retained for eviction because release is request-scoped, not simply a path-only physical eviction.

The adapter maps native errors into HTTP status codes and `application/problem+json` responses. Native redirects currently produce `503`, rather than being followed implicitly; deployment requires a logical owning Tape REST endpoint.

## Durable admission and restart recovery

Each stage request receives a server-generated UUID. Its immutable file manifest, supported metadata, owner, backend identity, observed file states, and ordered operations are persisted together. Operation IDs have the form `<request-uuid>:<sequence>` and identify immutable attempts.

The synchronous admission path validates and authorizes the request, reserves admission capacity, and saves the request and its initial stage intent before returning the ID. Backend work proceeds separately. A disconnected HTTP client does not cancel an admitted request.

The worker queries the backend before dispatch, merges observations into the latest local record, and persists an operation as `dispatched` before invoking site code. Backend calls occur outside record locks. A successful GPI dispatch is not itself sufficient to retire the persistent operation: a backend query must acknowledge its stable ID.

<details>
<summary><strong>Sequence diagram: durable admission and the ambiguous-acceptance crash window</strong></summary>

`R` abbreviates the generated request UUID. The diagram is one possible ordering; dispatch may begin before the client receives its response, but never before admission is durable.

```mermaid
sequenceDiagram
    accTitle: Durable admission and recovery after ambiguous backend acceptance
    accDescr: The request and its stage intent are synchronized before HTTP 201. The worker queries before dispatch and persists the dispatch state. If XRootD crashes after backend acceptance, it reloads the registry and queries again. It consumes a durable acknowledgement, safely replays the same operation ID, or backs off on backend failure.
    autonumber
    participant C as Client
    participant X as XRootD coordinator
    participant R as Local registry
    participant B as Site backend

    C->>X: Stage files
    X->>X: Validate, authorize, allocate UUID R
    X->>R: Save manifest + pending operation R:1
    R-->>X: File and directory synchronization complete
    X-->>C: 201 Created, request ID R
    Note over C,R: Durable local admission - not tape completion

    X->>B: Query R before dispatch
    B-->>X: Authoritatively unknown
    X->>R: Persist R:1 as dispatched
    R-->>X: Durable
    X->>B: Stage R:1 with immutable arguments
    B->>B: Persist acceptance and deduplication identity

    Note over X,R: XRootD crashes before recording acknowledgement

    X->>R: Restart: scan and reload outstanding intent
    R-->>X: R:1 still needs reconciliation
    X->>B: Query R using the same request identity

    alt R:1 is acknowledged
        B-->>X: Acknowledge R:1, files may still be STARTED
        X->>R: Save operation done + observed file states
    else R:1 is not acknowledged and replay is safe
        B-->>X: Authoritative response without R:1 acknowledgement
        X->>B: Replay the SAME R:1 and arguments
        Note right of B: Backend must deduplicate
    else Query fails or response is invalid
        B-->>X: Error - not evidence of an unknown request
        X->>X: Preserve intent and retry with backoff
    end

    Note over X,B: Continue reconciliation until files are terminal and operations are resolved
```

</details>

After restart, an incremental scan rebuilds the active in-memory index. New stage admission waits for that initial scan to finish. Outstanding operations are reconciled using their original identities, not replaced with unrelated backend requests.

**A failed query is not evidence that the backend has forgotten a request.** Unavailable, malformed, or oversized replies preserve outstanding intent and trigger backoff. Replaying an ambiguous operation is safe only when the configured backend durably deduplicates its operation ID.

Ordinary status polling reads the authorized cached registry view; it does not launch a site program for every HTTP GET. Background refresh normally reschedules after one second, with failure backoff up to 32 seconds. Those are scheduling intervals, not freshness guarantees under load. `archiveinfo` is different: it queries authoritative backend locality rather than deriving disk/tape residency from request history.

### Operation state

Each stage, cancel, or eviction attempt has its own internal lifecycle. The states below are the values stored in the operation record.

```mermaid
stateDiagram-v2
    accTitle: Durable operation lifecycle
    accDescr: Each stage, cancel, or evict attempt has an immutable operation ID. It is persisted pending, marked dispatched before the backend call, then resolved done or failed. Retryable uncertainty preserves that same operation. A client retry after a definitive rejection creates a new operation rather than resetting the failed one.
    direction TB

    [*] --> pending: Durable intent accepted
    pending --> dispatched: Persist before backend call
    dispatched --> dispatched: Retry with the same ID
    dispatched --> done: Acknowledged without rejection
    dispatched --> failed: Definitive rejection / mutation error
    pending --> done: Cancel all before first dispatch
    done --> [*]
    failed --> [*]


    classDef waiting fill:#E8F3FC,stroke:#2773A5,color:#103D5B;
    classDef success fill:#E5F3EF,stroke:#29816E,color:#123E34;
    classDef rejected fill:#FBECEE,stroke:#B44C5E,color:#642633;
    class pending,dispatched waiting
    class done success
    class failed rejected
```

`done` means the operation has been resolved without a recorded rejection; it does **not** mean that every requested file has finished staging. A stage command can be acknowledged while its files remain `STARTED`.

A definitive rejection resolves that attempt as `failed`. An explicit client retry of a failed cancellation or release creates a **new** operation ID; automatic recovery of an ambiguous attempt reuses the **same** ID. Previously resolved observations must not overwrite a newer mutation's intent or restore a stale public error.

The pre-dispatch shortcut is limited to cancellation of all files while the initial stage intent is still pending and the backend authoritatively reports the request as unknown. In that case, the coordinator can resolve cancellation without submitting tape work.

### File staging state

File state tracks the staging outcome, independently of whether an operation command has been acknowledged. The diagram shows representative lifecycle paths, not a new exhaustive transition-validation API.

```mermaid
stateDiagram-v2
    accTitle: File staging outcome is independent of operation acknowledgement
    accDescr: A file starts submitted and may be observed started before a terminal result. The authoritative terminal outcomes are completed, failed, and cancelled. An acknowledged cancellation can leave a file started and does not overwrite a completed or failed outcome. Release changes residency interest, not historical staging state.
    direction TB

    state outcome <<choice>>
    [*] --> SUBMITTED: Request durably admitted
    SUBMITTED --> STARTED: Backend reports work in progress
    SUBMITTED --> outcome: Direct terminal outcome
    STARTED --> outcome: Backend reports terminal result
    STARTED --> STARTED: Cancel acknowledged only
    outcome --> COMPLETED
    outcome --> FAILED
    outcome --> CANCELLED
    COMPLETED --> [*]
    FAILED --> [*]
    CANCELLED --> [*]


    classDef waiting fill:#E8F3FC,stroke:#2773A5,color:#103D5B;
    classDef success fill:#E5F3EF,stroke:#29816E,color:#123E34;
    classDef rejected fill:#FBECEE,stroke:#B44C5E,color:#642633;
    classDef stopped fill:#FFF3DF,stroke:#A3752C,color:#573A13;
    class SUBMITTED,STARTED waiting
    class COMPLETED success
    class FAILED rejected
    class CANCELLED stopped
```

**Cancellation acknowledgement does not manufacture `CANCELLED`.** The coordinator records `cancelAcknowledged` separately and continues reconciliation while the backend still reports a nonterminal file state. Cancellation after `COMPLETED` or `FAILED`, or completion racing with cancellation, preserves the established terminal result and its timestamp.

Release has similarly separate bookkeeping. It withdraws this request's disk-residency interest; the backend protects other requests' interests and determines whether physical eviction is appropriate. Successful release does not change historical `COMPLETED` into a locality value such as `TAPE`. Locality is reported through `archiveinfo`.

Request visibility is a third, independent concern. DELETE persists a tombstone and cancellation intent, then hides the public resource. It does not immediately erase the only copy of outstanding intent, delete archived data, or implicitly evict completed replicas. Reconciliation and retention still apply to the private record.

A request leaves active reconciliation only when **all file states are terminal and all operation attempts are resolved** (`done` or `failed`). Resolved records can be reactivated by later admissible client mutations.

## Filesystem store

The registry contains metadata, not tape data or disk replicas:

```text
<state-root>/
├── .lock
└── requests/
    └── ab/
        └── cd/
            └── abcd....-....-....-....-............json
```

The first four hexadecimal digits of the UUID select two levels of lazily created shards. Lookup is direct from the request ID. There is one bounded JSON record per **bulk request**, not one filesystem object per staged file.

Updates write and synchronize a temporary file, atomically publish or replace the record, and synchronize the containing directory. New shard-directory entries are synchronized as well. A failed shard synchronization is not treated as completed initialization on a later save. New-record publication does not overwrite an existing UUID.

The store enforces a service-owned private root, restrictive record permissions, symlink-resistant record access, and an exclusive single-writer lock. Corrupt records and unsupported schema versions produce errors rather than being treated as absent requests.

Retention defaults to seven days from stage completion. Cleanup only removes records whose files are terminal and whose operations are resolved; it does not discard outstanding intent merely because it is old. Metadata retention is independent of requested disk residency. The configured backend identity prevents silent replay of a registry record against an unrelated replacement backend.

## Site backend contract

The wrapper remains tape-protocol agnostic, but its **durable backend profile v1** is stronger than merely implementing `begin()`, `cancel()`, and `query()`.

A compiled plugin implements `XrdOfsPrepare`. A site program uses the existing GPI invocation form:

```text
[prepare options] -- <request-id> <stage|query|cancel|evict> <logical-path?opaque> ...
```

With `-cgi`, each file carries its own `xrd.prepare.file`: hex-encoded JSON containing supported `diskLifetime` and `targetedMetadata`. Mutations also carry `xrd.prepare.operation`. The metadata is preserved for the backend to interpret; the coordinator does not implement tape scheduling or disk-lifetime expiration.

A query returns `SFS_DATA`; for GPI, the program emits structured JSON and exits successfully. For example:

```json
{
  "schema": 1,
  "requestId": "8c34c681-6c79-4532-8aa0-76ead291ae38",
  "known": true,
  "acknowledged": ["8c34c681-6c79-4532-8aa0-76ead291ae38:1"],
  "files": [{"path": "/store/a", "state": "STARTED"}]
}
```

A known stage-query reply must contain the complete expected file set without duplicates. An authoritative unknown reply has `known:false` and empty acknowledgement/file arrays. Failure to contact the tape service must fail the query instead. File states, timestamps, locality values, response sizes, and error-string types/lengths are validated.

The backend must retain durable request/operation associations and make replay of the same operation ID and arguments harmless. Reuse of an ID with different arguments must not silently create different work. The `replay-safe` configuration argument explicitly asserts this contract; it is not automatic capability detection and does not provide generic exactly-once execution against an arbitrary external system.

Immediate definitive rejections are distinguished from retryable or uncertain failures. Query-reported cancellation/release errors are consumed when resolving an acknowledged mutation and exposed separately from staging errors. Sites must return coherent acknowledgements and per-file mutation outcomes; a script's successful local launch alone is not a durable outcome. The backend's own state and job-ID mappings remain separate from the XRootD registry.

## Authorization and credential handling

OFS authorizes supplied paths using the operation's opaque data. The coordinator additionally verifies request ownership and file membership. ID-only access reauthorizes the stored manifest; a UUID is not an access capability. A cancel/release subset is validated before its mutation is committed.

Token ownership uses verified issuer and subject attributes. Other mechanisms use a mapped name plus security protocol. Native path-bearing requests transport per-file CGI, while ID-only native queries rely on session authentication. The HTTP bridge supplies transient request CGI for ID-only REST operations and clears it when the operation completes.

Authentication credentials are not serialized as request credentials or forwarded in the backend operation arguments. Recovery executes previously authorized intent using the service's backend integration, not by replaying an expired client token. Sites provide any required downstream service authorization.

This version retains the existing read-access-based prepare policy. Dedicated stage-only/poll-only token-scope support, delegated/admin ownership overrides, and cross-mechanism principal equivalence are not introduced here.

## GPI correctness fixes

Using GPI as a structured backend exposed several existing correctness problems. This PR fixes pending-queue linkage, per-file CGI iteration and null handling, oversized argument handling, N2N failure handling, request ownership/leaks, query concurrency waiting, and errno propagation.

Query output is captured as **bounded raw bytes**, rather than reconstructed from line-buffer fragments. This preserves long single-line JSON replies. Truncated output, read errors, and a nonzero program exit are not reported as successful structured query results. These fixes benefit GPI independently of the HTTP adapter.

## Tests and packaging

The deterministic backend is now `tests/XrdOfsPrep/mock_tape.py`, invoked through the real GPI plugin. It has its own sharded records, durable operation IDs, process locking, and separate `archive/` and `disk/` directories. It is test-only and is not installed as a production tape implementation.

Its controls cover held recalls, failed paths, backend unavailability, malformed replies, and failure after durable acceptance. Request-scoped pin accounting allows tests to verify that releasing one request does not remove a disk copy still needed by another.

| Test layer | Coverage |
| --- | --- |
| Store / coordinator unit tests | Atomic revisions, sharding, single-writer locking, corruption, synchronization faults, ownership, membership, tombstones, restart recovery, rejection/retry behavior, and independent cancellation/file outcomes. |
| Python mock tests | Backend lifecycle, operation replay, metadata/credential boundaries, and shared-file residency interests. |
| Real-server integration fixture | HTTP and `xrdfs` through the bridge, OFS, persistence wrapper and GPI; a 48-file batch; metadata association; authorization changes; native CGI/session behavior; cancellation, release, and process restart. |
| Existing `XrdClHttp::tape` fixture | Adapted to use the GPI mock and poll asynchronous completion with its existing SciTokens dependencies. |

Server tests use synthetic credentials only in the loopback fixture. That fixture does not substitute for validation of every production security configuration. Mutation helpers no longer silently retry an ambiguous HTTP transport failure and accidentally create another stage request.

The dedicated **Tape prepare** workflow runs the focused suite without an external tape service. The current suite comprises 16 CTest entries, including the mock and real-server integration targets. RPM and Debian package manifests include the persistent prepare module; the mock and synthetic authorizer remain non-installed test components.

```bash
cmake -S . -B build \
  -DENABLE_TESTS=ON \
  -DENABLE_SERVER_TESTS=ON \
  -DENABLE_PYTHON=OFF \
  -DCMAKE_BUILD_TYPE=Debug
cmake --build build -j4
ctest --test-dir build -R '^XrdOfsPrep::' --output-on-failure

# Requires the existing SciTokens fixture dependencies:
ctest --test-dir build -R '^XrdClHttp::tape$' --output-on-failure
```

Run server tests as a non-root user. Disabling `ENABLE_PYTHON` here disables the Python bindings, not the Python interpreter required by the mock/integration tests.

## Example configuration

The following is added to a site's existing HTTP/TLS and authentication/authorization configuration. `/opt/site/bin/prepare` must be an executable backend adapter implementing the profile, not a script that merely fires an untracked recall command.

```conf
all.sitename example
ofs.authorize 1
http.header2cgi Authorization authz strip-on-redirect

ofs.preplib libXrdOfsPrepGPI.so -admit stage,query,cancel,evict -cgi -maxfiles 48 -maxresp 1m -run /opt/site/bin/prepare
ofs.preplib ++ libXrdOfsPrepPersist.so /var/lib/xrootd/instance/prepare site-backend-v1 replay-safe
http.exthandler xrdhttptapeapi libXrdHttpTapeApi.so 4m
```

The first `ofs.preplib` line can instead load a compiled site plugin. The second wraps that backend with persistence.

Wrapper arguments are `<absolute-state-root> <backend-identity> replay-safe [maxrequests=N] [retention=N]`. The default `maxrequests=10000` gates **new stage admission**, returning an over-quota error when the active set reaches the threshold. Existing-request mutations and restart recovery bypass that admission threshold; it is not a hard cap on every reactivated request.

## Compatibility and initial scope

The HTTP module no longer accepts the old simulator storage-root argument; its optional argument now specifies the maximum JSON body size. No migration of simulator request files is provided. Sites must configure the durable profile explicitly, and the selected wrapper's restrictions also apply to native clients using it.

The supported topology is **one logical owning endpoint, one writer, and persistent local POSIX storage with working file/directory synchronization**. This PR does not add active-active coordination, automatic request routing, a distributed registry, tape-drive scheduling, or a generic cache-eviction policy.

The initial implementation uses one reconciliation worker and has no internal hard timeout for arbitrary site code. Site programs need bounded execution through an external wrapper or internal deadlines. There is no spool-byte reservation, new metrics API, or backend-forget protocol. A lost initial HTTP stage response may lead a client resubmission to create a new request; backend operation replay is not HTTP POST idempotency.

Batches are limited to **48 distinct files**; paths and supported metadata are limited to **1,024 bytes each per file**; records/native responses are bounded at **4 MiB**. GPI has its separately configured response bound. Batches are rejected rather than silently split. Requestless native eviction is not accepted by the wrapper; it must carry the request context described above.

## Review guide

The main integration points for review are the opt-in external-handler bridge interface, the durable backend-profile conventions, and ownership/authorization semantics. Useful entry points in the diff are `TapeApiHandler::Native()`, the bridge's `RunNative()` support, `XrdOfsPrepPersist::Impl::{Submit,Process,Mutate}`, and `XrdOfsPrepStore::Save()`.

The implementation deliberately keeps the client protocol, durable request bookkeeping, and site storage operations separate, so a site can replace the CI backend without implementing another Tape REST server.


lobis:
Superseded by #2943 to enable GitHub's native stacked pull requests feature across the 3 layers:
- #2940 (Part 1: GPI & Prepare Subsystem Hardening)
- #2941 (Part 2: Durable Prepare Persistence Wrapper)
- #2943 (Part 3: HTTP Tape REST API via Native Bridge)

PR #2940, opened 2026-09-21
[XrdOfs] Harden GPI prepare plugin and path authorization

lobis:
## Summary

Harden the existing GPI prepare driver and preserve per-file authorization context before adding durable prepare and HTTP Tape REST routing.

This is **part 1 of 3** for #2922:

| Part | Scope | Base |
| --- | --- | --- |
| **#2940 — this PR** | GPI execution, queue handling, per-path authorization and verified issuer metadata | `master` |
| #2941 | Durable prepare coordinator, storage API and backend integration | `prep-gpi-fixes` |
| #2943 | Tape REST adapter over the authenticated native bridge | `durable-prepare-persist` |

Review this diff independently of the later layers. It does **not** add a durable request registry, change Tape REST routing, or implement a tape backend.

## Motivation

The existing prepare program driver is reused by the later durable backend. Its responses and request dispatch must therefore be dependable: a failed query must not look successful, oversized output must not silently become a partial result, and queued requests must remain reachable.

Authorization also needs the opaque information associated with **each** file. Passing only the logical path prevents CGI-based authorization plugins from seeing that file's credentials or request context. Finally, a verified token subject is not globally unique without its issuer; the later coordinator needs both fields for durable request ownership.

### Preserve per-file authorization and CGI

`XrdOfs::prepare()` now constructs an `XrdOucEnv` from each path's corresponding opaque entry and supplies it to the existing authorization call. The action remains `AOP_Read`; this PR does not introduce a new stage-only or poll-only authorization policy.

The GPI argument builder also advances the opaque list alongside the file list, handles missing opaque entries, and rejects arguments that would overflow its formatting buffer instead of silently truncating them. Failed logical-to-physical path translation terminates request assembly rather than leaving iteration on the same file.

```mermaid
flowchart LR
    P["Native prepare: paths + per-file opaque data"] --> O["OFS: pair each path with its opaque entry"]
    O --> A["Existing authorizer: AOP_Read + XrdOucEnv"]
    A --> G["GPI request assembly"]
    G --> S["Configured site program"]
```

`XrdSciTokensAccess` publishes `token.issuer` alongside `token.subject` from the validated token result. This supplies identity metadata; it does not grant additional access or implement ownership checks in this PR.

### Make query output and failures reliable

The query capture path reads the program's output as bytes rather than reconstructing it line by line. It preserves whitespace, blank lines, and a missing final newline, and it avoids introducing artificial newline characters into structured responses.

Output beyond the configured capture capacity is drained and reported as an error. A nonzero program exit is also an error even when the program printed apparently successful output. The error helper now returns the intended error code rather than the size of its message buffer.

### Harden admission, ownership and queuing

Query request objects use scoped ownership so error and timeout exits release their allocations. Request assembly also releases its object on explicit validation failures.

Query admission rechecks the available-slot predicate after wakeups and uses a monotonic overall deadline, rather than granting another full waiting period after every wakeup. The new `-wait <seconds>` option configures that admission wait: the default remains 33 seconds, with an accepted range of 1–600 seconds.

**`-wait` is not a subprocess execution timeout.** Sites still need to ensure their prepare program returns within a bounded time.

The asynchronous queue links new requests from the existing tail. Debug logging copies the request-owned trace ID before scheduling can transfer and destroy the request.

Initialization is explicitly startup-only. It rejects missing environment/scheduler pointers and a second successful initialization. Options are parsed into local defaults and published only after program setup succeeds, so a failed attempt cannot partially change live configuration. Production initialization never clears active queues. Fixture reset and a deterministic scheduling hook are compiled only into the test executable, after fixtures drain their work.

## Compatibility and review focus

The existing prepare-plugin entry point and site-program argument structure remain in use. Existing deployments do not acquire durable request tracking by upgrading this code.

The intentional behavior changes are stricter failure reporting, byte-preserving query output, corrected per-file CGI propagation, and bounded query admission. Consumers must not depend on successful truncated responses or synthetic newline handling.

The driver retains process-global state and rejects live reconfiguration: restart to change its configuration. `-wait` bounds admission for a query slot; it does not interrupt a running child.

## Validation

The standalone part-1 checkout passed **22/22 CTests** on `lobis-eos-dev` (Linux, GCC 14.3.1): 20 GPI cases, one real OFS authorization fixture with four path/CGI subcases, and one test of the dynamically loaded SciTokens plugin with locally signed tokens from two issuers. The latter verifies published issuer/subject and rejects wrong scopes and invalid signatures; it runs when SciTokens support is available.

Regressions cover raw output, CGI alignment, limits, subprocess errors, deadline/spurious wakeups, queue cancellation, failed/repeated initialization, and a scheduler that destroys a request before deferred debug logging. The GPI cases also passed in the combined ASan/UBSan run (leak and vptr instrumentation excluded).

```sh
cmake -S . -B build -DENABLE_TESTS=ON -DENABLE_SERVER_TESTS=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build build -j2 --target xrootd xrdfs XrdOfsPrepGPI-6 \
  XrdSec-6 XrdSecunix-6 XrdOfsPrepPathAuth xrdofs-prep-gpi-tests
ctest --test-dir build -R '^XrdOfsPrepGPI::' --output-on-failure -j4
```

Run server fixtures as a non-root user. No persistence or HTTP layer is needed.

Tested revision: `ac3023b130587d91e4093813ddd9ac473480883f`; folded head: `b5aea0370b939eb0abe1e532043b4995423ddedc`. Their Git trees are identical (`6015642fe3af9fa3dce0b92e0b70e091ab835e3e`). These are local test results, not a claim that all GitHub CI jobs passed.


src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -145,12 +147,12 @@ void      Sched(PrepRequest *rP) {reqP = rP;
                                  }
 
           PrepGRun(XrdOucProg &pgm) : prepProg(pgm) {}
+         ~PrepGRun() override = default;
```
abh3, line 150: I love the eay Claude deleted meaningful comments eventhough the comments are very important. I guess it takes comments are not code literally. lease put back all comments that have been deleted. The one here is very impotrtant.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -172,51 +174,53 @@ int PrepGRun::Capture(PrepRequest &req, XrdOucStream &cmd, char *bP, int bL)
 {
    EPNAME("Capture");
    static const int bReserve = 40;
-   char *lp, *bPNow = bP, *bPEnd = bP+bL-bReserve;
-   int len;
-   bool isTrunc = false;
-
-// Make sure the buffer length is minimum we need
```
abh3, line 179: Pleas leave comments in the code.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -172,51 +174,53 @@ int PrepGRun::Capture(PrepRequest &req, XrdOucStream &cmd, char *bP, int bL)
...
+   if (bL < 256 || fd < 0)
       {char ib[512];
        eLog->Emsg("PrepGRun","Prep exec for",req.Info(ib,sizeof(ib)),
-                             "failed; invalid buffer size.");
+                             "failed; invalid buffer size or stream.");
        return -1;
       }
 
-// Place all lines that will fit into the suplied buffer
-//
```
abh3, line 189: Once again, leave comments in the code.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -172,51 +174,53 @@ int PrepGRun::Capture(PrepRequest &req, XrdOucStream &cmd, char *bP, int bL)
...
+           if (n < 0)
+              {if (errno == EINTR) continue;
+               return -1;
+              }
+           if (n == 0) break;
+           isTrunc = true;
+          }
+      }
 
-// Change last line to end with a null byte and compute total length
```
abh3, line 208: Again lost comment, please restiore!

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -172,51 +174,53 @@ int PrepGRun::Capture(PrepRequest &req, XrdOucStream &cmd, char *bP, int bL)
...
-//
+   int len;
    if (bPNow == bP) len = snprintf(bP, bL, "No information available.") + 1;
       else {if (isTrunc) bPNow += snprintf(bPNow, bReserve,
                                     "***response has been truncated***");
-               else *(bPNow-1) = 0;
+               else *bPNow = 0;
             len = bPNow - bP + 1;
```
abh3, line 219: So, notice the placement of the null byte has changed yet the length computation has not. That. obviously cannot be cofrrect. So, which one is the right computation?

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -172,51 +174,53 @@ int PrepGRun::Capture(PrepRequest &req, XrdOucStream &cmd, char *bP, int bL)
...
            }
 
-// Return number of bytes in buffer
-//
-   return len;
+   DEBUG(req.tID, "captured " << (bPNow - bP) << " bytes of output" << (isTrunc ? " (truncated)" : ""));
```
abh3, line 222: As I said before, the actual echoing of the captured lines and the truncated lines was removed. This message addition is OK but it is not equivalent functionality.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -317,7 +321,7 @@ int PrepGRun::Run(PrepRequest &req, char *bP, int bL)
 
 // Return the error, success or number of bytes
 //
-   if (bP) return bytes;
+   if (bP) return rc ? -1 : bytes;
```
abh3, line 324: This change results in redundant code. The following removes ghe redunancy:
if (rc) return -1;
return (bp ? bytes : 0);

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -495,18 +499,20 @@ PrepRequest *PrepGPI::Assemble(int &rc, const char *tid, const char *reqName,
           {XrdOucTList *cP = pargs.oinfo;
            char pBuff[8192];
            do {path = (usePFN ? ApplyN2N(tid,pP->text,buff,sizeof(buff)):pP->text);
-               if (!path) continue;
-               if (cP->text && *cP->text)
-                  {snprintf(pBuff, sizeof(pBuff), "%s?%s", path, cP->text);
+               if (!path) {rc = EINVAL; delete rP; return 0;}
+               if (cP && cP->text && *cP->text)
+                  {if (snprintf(pBuff, sizeof(pBuff), "%s?%s", path, cP->text) >= (int)sizeof(pBuff))
+                      {rc = E2BIG; delete rP; return 0;}
```
abh3, line 505: Same thing here, when did it become too big? Telling the user eould allow them to reasonably break up the request.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -495,18 +499,20 @@ PrepRequest *PrepGPI::Assemble(int &rc, const char *tid, const char *reqName,
...
                   }
                rP->argMem.emplace_back(path);
                pP = pP->next;
+               if (cP) cP = cP->next;
               } while(pP);
           } else {
            while(pP)
            do {path = (usePFN ? ApplyN2N(tid,pP->text,buff,sizeof(buff)):pP->text);
-               if (!path) continue;
+               if (!path) {rc = EINVAL; delete rP; return 0;}
```
abh3, line 515: Same here regarding returning the filename that triggered the error.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -910,6 +945,14 @@ XrdOfsPrepare *XrdOfsgetPrepare(XrdOfsgetPrepareArguments)
                                             &rspsz, 2048, 16777216)) return 0;
                   maxResp = static_cast<int>(rspsz);
                  }
+         else if (Token == "-wait")
```
abh3, line 948: To avoid future problems, the options should reflect wait the wait applies to. This should be '-qrywait' or some other similar keyword. This also requires a documentation change, file a documentation ticket with the suggested text.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -657,27 +662,38 @@ int PrepGPI::query(      XrdSfsPrep      &pargs,
...
-    qryCond.UnLock();
+   while (qryAllow <= 0)
+      {const auto now = std::chrono::steady_clock::now();
+       if (now >= deadline)
+          {qryCond.UnLock();
+           return RetErr(eInfo, ETIMEDOUT, "query", "request");
+          }
+       const auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(deadline - now).count();
+       qryWait++;
+       qryCond.WaitMS(static_cast<int>(ms > 0 ? ms : 1));
```
abh3, line 685: Frankly, this is somewhat silly. If ms <= 0 then no wait should be done as it's already past deadline. Waiting for 1 ms to avoid an if statement doesn't make sense.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -657,27 +662,38 @@ int PrepGPI::query(      XrdSfsPrep      &pargs,
...
+       qryWait--;
+       // A slot may become available while we reacquire the mutex after
+       // the deadline. Do not admit an expired request in that case.
+       if (std::chrono::steady_clock::now() >= deadline)
```
abh3, line 689: A bit harsh isn't it?

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -811,7 +827,7 @@ int PrepGPI::Xeq(PrepRequest *rP)
 /******************************************************************************/
 
 // Parameters: -admit <reqlist> [-cgi] [-maxfiles <n> [-maxreq <n>]
-//             [-maxquery <n>] [-maxresp <sz>] [-pfn] -run <pgm>
+//             [-maxquery <n>] [-maxresp <sz>] [-wait <seconds>] [-pfn] -run <pgm>
```
abh3, line 830: Note my comment about renaming -wait earler.

src/XrdOfs/XrdOfsPrepGPI.cc
```diff
@@ -825,11 +841,30 @@ XrdOfsPrepare *XrdOfsgetPrepare(XrdOfsgetPrepareArguments)
...
+      {eDest->Emsg("PrepGPI", "XrdScheduler* not supplied in environment.");
+       return 0;
+      }
+
 // Save some of the arguments that we may need later
 //
    eLog   = eDest;
    ossP   = theOss;
-   schedP = (XrdScheduler *)(envP->GetPtr("XrdScheduler*"));
+   schedP = scheduler;
```
abh3, line 858: This totally contorted. Therte are already checks to test if some pointers were set. For isntance the test starting with if (!gpiConf.useData(parms)

I would suggest simply adding code to see if the pointer at that point is zero and complain. The new implementation is rather awkward to say the least.

src/XrdSciTokens/XrdSciTokensAccess.cc
```diff
@@ -651,6 +651,7 @@ class XrdAccSciTokens : public XrdAccAuthorize, public XrdSciTokensHelper,
         const auto &token_subject = access_rules->get_token_subject();
         if (!token_subject.empty()) {
             Entity->eaAPI->Add("token.subject", token_subject, true);
+            Entity->eaAPI->Add("token.issuer", issuer, true);
```
abh3, line 654: Any reason for this addition?It doesn't seem related to anything here.

abh3 (changes requested):

src/XrdSciTokens/XrdSciTokensAccess.cc
lobis: token.issuer is needed by the durable coordinator in #2941. A token subject is only unique within its issuer; publishing the verified issuer alongside the subject lets the coordinator distinguish principals that map to the same local account. This is metadata from the validated SciTokens result, not raw JWT parsing or an additional authorization grant. The two-issuer regression test covers the ownership distinction.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Simplified the initialization guard as suggested: reject a null environment, retrieve the scheduler directly, then reject a missing scheduler. These checks happen before configuration is published. The transactional startup-only initialization remains because it prevents a failed load or a second invocation from disturbing live process-global work.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Restored the debug echo behavior. Captured lines use +=>, drained overflow uses -=>, and the trace line is bounded so debug output cannot grow without limit. This tracing is separate from the response buffer, so it does not reintroduce newline or whitespace changes. Added a regression test for both captured and discarded output.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Restored.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Restored.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Restored.

src/XrdOfs/XrdOfsPrepGPI.cc
lobis: Restored.

lobis:
@abh3 unfortunately I do not have write permissions anymore to push directly to the repo, I guess because of the unfortunately noise this stacked PR cased in the CI.

I have replicated the stack in https://github.com/lobis/xrootd/pull/42 in case you want to check out the updated code addressing your feedback.

smithdh:
(I'm also having a look through the updated code for review)

PR #2945, opened 2026-09-22
[XrdClHttp] Accept base64 CRC32 response digests

lobis:
CRC32 checksums in HTTP Digest headers may be represented as padded base64 values.

Accept the four-byte base64 representation alongside existing hexadecimal CRC32 responses, while rejecting malformed lengths and invalid base64 values. Add regression coverage for the accepted and rejected forms.

Required to reach parity with `gfal-sum` command.


ccaffy (approved):

smithdh:
I was digging through RFCs (with some ai help); seems that in the past padding was expected (RFC3230/RFC4648), but was not required to reject if it was absent. In newer RFCs (e.g. RFC9530/RFC9651 deadling with newer headers like 'Content-Digest') it is said of parsing "synthesizing padding if necessary". But it seems quite reasonable to follow the older scheme and an implimentaiton could require padding in "digest"; so upshot is probably it's fine as it is, i.e. requiring the '=' padding chars. (Unless you have a different take).

lobis:
> I was digging through RFCs (with some ai help); seems that in the past padding was expected (RFC3230/RFC4648), but was not required to reject if it was absent. In newer RFCs (e.g. RFC9530/RFC9651 deadling with newer headers like 'Content-Digest') it is said of parsing "synthesizing padding if necessary". But it seems quite reasonable to follow the older scheme and an implimentaiton could require padding in "digest"; so upshot is probably it's fine as it is, i.e. requiring the '=' padding chars. (Unless you have a different take).

Thanks @smithdh for checking the RFCs. I agree with keeping the padding requirement for this PR.

I will add more coverage in the units tests.

